PR #1, opened 2018-08-30
Fixed bug that caused memory_limit=-1 to lead to 'not enough memory' …

delassiter:
Firstly thank you for bringing iTop to github! I've been waiting for this day for a long time.

### Issue
When running setup with memory_limit = -1 (no limit), this error is presented:

> memory_limit (-1) is too small, the minimum value to run the application is 33554432. 

### Solution
Since there were multiple places where the memory was checked, I've created a new function that ignores -1 and returns a boolean (true = memory limit is fine, false = memory limit below required value).


application/utils.inc.php
```diff
@@ -559,6 +559,18 @@ public static function ConvertToBytes( $value )
...
+    /**
+     * Checks if the memory limit is at least what is required
+     *
+     * @param int $memoryLimit set limit in bytes
+     * @param int $requiredLimit required limit in bytes
+     * @return bool
+     */
+    public static function IsMemoryLimitOk($memoryLimit, $requiredLimit)
+    {
+        return $memoryLimit >= $requiredLimit || $memoryLimit == -1;
```
piRGoif: Please add parenthesis in this test...

piRGoif: I would suggest this method do : 
* the current test
* call ini_set
* if ini_set returns false, then throw an exception (maybe a new CoreMemoryException ?)
* returns true otherwise

Hipska: Good idea!

piRGoif: No, I really meant the ini_set call, because in all callers you have the same scheme : 

* test
* ini_set call
* statement if ini_set fails
* statement if ini_set succeeds

So I would factorize both the test and the ini_set... But we should call the method differently to reflect that it tries also to update the memory parameter. Maybe TestAndSetMemoryLimit ?

What are the reason why you wouldn't put ini_set in there ?

Hipska: Keeping methods small and single purpose.

Best would be to first have `IsMemoryLimitOk` and later add your `TestAndSetMemoryLimit` which then calls this method.

This to avoid a lot more refactoring (and possible bugs) in 1 step.

piRGoif: I totally agree !

delassiter: As @Hipska said, methods should have a single purpose. In this case I was concerned with the assertion. Fetching the memory limit and setting a new memory limit are different concerns.

I do agree that I could have refactored more than just the assertion, but I hardly think it has anything to do with the actual bug. If you really think this should be done as a part of this pull request, then I will also add it.

What I should have added is a unit test. I kind of shied away from the test suite, which was wrong of me. I'll also add that later today or tomorrow.

piRGoif: Hello,

As said before I totally agree to keep `IsMemoryLimitOk` as it (but with parenthesis in the test ;) ), and also I would prefer to add a second method factorizing both the `IsMemoryLimitOk` and the `ini_set` calls.  
Thanks to add it in this pull request !

And also a test would be great, many thanks for your suggestion ! 

Lack of tests is a big issue for now in iTop, we added just few of them (in /test), but we're planning to have a real big work on this in the future - can't give a date nor a version now, but this os one of the major concern in our roadmaps.

piRGoif:
Hello, thanks for the test, perfect ! And for the parenthesis too :)
I'm going to add the last factorization, fix the 2 calls that were removed by the develop merge, and then close this PR :)

application/utils.inc.php
piRGoif: Never mind, I'm closing this PR without this extra function.
A method that would check and set, and could have 3 return states (memory ok and set, memory ok but can't be set, memory nok) is not a good idea indeed. And we won't change the `ini_set` calls to another method, that would be nonsense. 
Allright I changed my mind, I'm ok with this PR for now :)

Hipska:
Whoohoo 🎉 First user PR merged! Congrats!

Molkobain:
w00t w00t !

piRGoif:
Dennis, your are now part of iTop contributors listed in the README.md :)

PR #2, opened 2018-09-06
Sort select/radio fields by translation

jbostoen:
Makes it more natural, sorts select/radio attributes by their translation.
Should we make pull requests to 'master' or 'develop'?

For instance, in English you could have this order:

- arts
- sports
- youth

while in Dutch it would be 

- jeugd (youth)
- kunst (arts)
- sport (sports)

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -1951,6 +1951,7 @@ EOF
 							$aEventsList[] ='change';
 							$sHTMLValue = "<div class=\"field_input_zone field_input_{$sDisplayStyle}\">";
 							$bVertical = ($sDisplayStyle != 'radio_horizontal');
+							asort($aAllowedValues, SORT_REGULAR);
 							$sHTMLValue .= $oPage->GetRadioButtons($aAllowedValues, $value, $iId, "attr_{$sFieldPrefix}{$sAttCode}{$sNameSuffix}", $bMandatory, $bVertical, '');
 							$sHTMLValue .= "</div>{$sValidationSpan}{$sReloadSpan}\n";
 							break;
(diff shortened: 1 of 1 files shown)


Molkobain:
_TLDR; Too soon to tell you. I'm curious about @dflaven and @v-dumas thoughts on this._

It would definitively be a property of the AttributeEnum class, but as it would also need its counterpart in the XML definition, which requires to be upgraded to a new version number (current version is 1.5, 1.6 coming with iTop 2.6) for both the compiler & the ITSM Designer to handle it.

As this would be introduced as a new feature, we can imagine that we would also like to be able to choose _on what_ it is sorted (value vs label) but also _how_ (ascending vs descending order).

IMHO, this needs to be discussed.

@dflaven & @v-dumas might have some insights from existing requests on this matter.

jbostoen:
@v-dumas , with backward compatibility in mind, is 'rank' really needed? Couldn't it be assumed that the rank is simple the one you have in the XML file? (unless you're thinking about extensions adding values, but then again, you might need to redefine the entire ranking anyway in the extension)

Hipska:
Why not order on value id then? It is currently only used to define and change/remove the values..

v-dumas:
Good idea if we where starting from scratch, but not very flexible on existing installed base. I doubt that we can change those id within the Standard Datamodel, without breaking a lot of existing customization/extensions.

Hipska:
The same counts if you go from
```xml
<value id="name">name</value>
```
to
```xml
<value id="name">
    <code>name</code>
    <rank>3</rank>
</value>
```

jbostoen:
For compatibility/minimal work, I'd suggest just using one tag (sort_on) with:
- **value** (or 'code' or ...)
- **label**  (or 'translation' or ...)
- **xml** 

By default, it could assume 'value', which is the way it works today. Fully compatible with previous work, I guess.

Or is there an issue I'm missing with this approach?

v-dumas:
Using the XML order instead of a rank tag, is again not a good idea, as it forces full reordering by an extension, which may not know what values you have, so its bad. A rank is not perfect, but still better as most extensions will assume Standard Datamodel + maybe other values brought by other extensions, and will decide to rank their value between 2 standard values assuming they have not been moved. We use ranking with decimal, so you don't need to reorder to insert a value.

v-dumas:
@Hispka, iTop compiler is already able to cope with such double syntax

dflaven:
My thoughs on the topics are simple: a PR is not the place to discuss such internal iTop design topics.

PR #4, opened 2018-09-25
AttributeExternalField: enhancement - support AttributeImage

jbostoen:
* enhances AttributeExternalField. Makes it possible to point to AttributeImage. With some changes, could be adapted to AttributeBlob as well I think.

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -5562,6 +5562,12 @@ class AttributeExternalField extends AttributeDefinition
 	{
 		if ($sPrefix == '')
 		{
+
+			// AttributeImage needs to return different SQLExpressions (_data, _mimetype, _filename)
+			if( get_class($this->GetFinalAttDef()) == "AttributeImage") { 
+				return $this->GetFinalAttDef()->GetSQLExpressions( $sPrefix );
+			}
+			
 			return array('' => $this->GetCode()); // Warning: Use GetCode() since AttributeExternalField does not have any 'sql' property
 		}
 		else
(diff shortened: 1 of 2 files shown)


core/attributedef.class.inc.php
```diff
@@ -5822,9 +5828,61 @@ public function FromSQLToValue($aCols, $sPrefix = '')
...
 	{
+		
 		$oExtAttDef = $this->GetExtAttDef();
-		return $oExtAttDef->GetAsHTML($value, null, $bLocalize);
+		 
+		if( get_class($this->GetFinalAttDef()) != "AttributeImage") { 
+			
+			// Wait, why was this 'null'?
+			// Let's change it to $oHostObject 
+			return $oExtAttDef->GetAsHTML($value, $oHostObject, $bLocalize);
```
jbostoen: Please revise this. Not sure if null was intended or not. It caused an issue earlier on when I was trying to figure out why it didn't work, but 'null' was not the cause here I guess.

Molkobain:
PR related to ticket [1651](https://sourceforge.net/p/itop/tickets/1651/) on SourceForge.

Molkobain:
Hi @jbostoen ,

Thank you for your contribution! 😊

There is indeed an issue affecting the AttributeImage when used through an external field. Actually, AttributeBlob, AttributeStopwatch & other composite attributes have the same issue; it seems that the delegation process for the SQL expression and HTML rendering (& probably CSV, XML, ...) is not working as expected.

As your PR is focused on fixing the HTML rendering of the AttributeImage, we won't be able to accept it as-is, as we rather address a more generic and global approch for all types. So I'm closing the PR but I tracked the issue as an enhancement and linked the ticket you opened on SF.

Thank you for pointing this out!

Molkobain:
It's been tracked but will most likely not be part of iTop 2.6 as we are closing on the beta version.

Hopefully it will be part of 2.7 or 2.8 as it can be useful in some corner cases, but as we had very few demands on this, it's not among the top priority fixes.

PR #12, opened 2018-11-01
Session id regeneration on login.

Nould:
This may be solution of a session fixation vulnerability.

PR #20, opened 2018-11-13
Session id regeneration on login.

Nould:
This is #12. I have renamed my branches.


bruno-ds:
Since we want to ship it with iTop 2.6, could you please rebase it to release/2.6 ?

Molkobain:
Careful, this PR causes fatal errors on webserver using PHP7+ and memcached (see [here](https://bugs.php.net/bug.php?id=71187)).
We are looking into it. 🧐

piRGoif:
A workaround has been found. To prevent any other issues, we will soon add a "regenerate_session_id_enabled" config parameter, default value true, hidden by default.

PR #25, opened 2018-12-03
REST API : new params "limit" and "page" (pagination)

delassiter:
The Webservice currently doesn't allow limiting the results to return. This causes problems when fetching huge datasets (timeout, memory exceeded).

I've added two **optional** parameters to the "core/get" operation:

- **limit** (int): Amount of results to return (default: 0 = no limit)
- **page** (int): Page number to return

The comment still returns the real number of found results.


core/restservices.class.inc.php
```diff
@@ -774,4 +776,16 @@ public function DeleteObjects($oResult, $aObjects, $bSimulate)
...
 	}
+
+    /**
+     * Returns the Offset for a given page number
+     *
+     * @param int $limit
+     * @param int $page
+     * @return int
+     */
+	private function getOffsetFromLimitAndPage($limit, $page)
```
bruno-ds: Since it is private and not called by your modifications, this seems to be a dead function .

Hipska: It gets called by the [previous hunk](https://github.com/Combodo/iTop/pull/25/files#diff-ef8dc8870bd1e00e5503344c54de04f2R442).

piRGoif:
Hello,
Thanks Dennis for the suggestion ! We need to discuss this in the team, we will get back to you.

delassiter:
Sorting is a whole other can of worms. I should've implemented this as well. If I find some time later, I'll propose something.

If not otherwise specified, data will be returned from the DB in the order it was written. So, in most use cases it will be sorted by ID ascending, but it that's more or less a coincidence.

jbostoen:
Not really looking to 'sort' (although it could be nice), but if you're paging, it should be consistent. But if it's by default ordered by ID ascending, this would never be a problem.

Hipska:
The ordering is handled by the DB, it isn’t a problem.

Hipska:
@piRGoif any feedback from @Combodo?

Molkobain:
@Hipska Not yet, the 2.6 is keeping us quite busy right now. We will get back to you when we can.

piRGoif:
Hello,

Sorry those last weeks we were pretty busy with the 2.6.0 products version packaging.

The dev team had a chat though about this PR, and we agree that this would be a nice addition ! As @jbostoen pointed out pagination cannot be used if the record ordering is not predictable... Adding an ORDER BY statement in OQL cannot be done right now, but there is already something in iTop now : see in the datamodel the classes/class/properties/order nodes. So in principle, we totally agree on this PR idea !

But of course we need to review the code, we will in the upcoming days (after christmas O:) ) and give a feedback !

Thanks !

Hipska:
The ordening might not look predictable, but it is consistent, and that’s the only thing you initially need.

piRGoif:
This seems ok by me, I'm ok to merge this in the develop branch.
Just a question, when limit is set and page is 0, you get the first page (same behavior whereas page is not present, has 0 or 1 as value). Shouldn't page not present or 0 brings an error ? Or no results ?

Those 2 new parameters needs to be documented in the official Wiki, I will do this (this won't be visible until 2.7.0-beta is out)
Can I squash & merge, or do you want to do it yourself ?

delassiter:
I agree that page 0 being the same as page 1 is not ideal. I'll have a look at throwing an error this week.

delassiter:
Sorry about the weird commit tree... rebasing did weird things.
Anyway, I think I was able to implement the change that I promised. Sorry for the delay.

piRGoif:
Hello,
Many thanks Dennis, perfect !
I wasn't able to merge, indeed your rebasing gave weird results :D
But I cherry picked from your fork and commit to fd55bdf9.
After updating our documentation (will be out with 2.7.0-beta) the PR could be closed.

piRGoif:
Done ! I'm closing this PR. Thanks again !

piRGoif:
Just to let you know, the modification was cherry-picked to the support/2.5 (e2c8237b) and so is also available on the master branch.
Also, we decided to upgrade the REST API version from 1.3 to 1.4 (d6a0a279)

PR #27, opened 2018-12-04
add visual studio

purplegrape:
add visual studio

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,5 @@ local.properties
 .scala_dependencies
 .worksheet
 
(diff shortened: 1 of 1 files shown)


piRGoif:
Hello,

There are already lots of things in our ignore file, we won't add this as there are no reason why we should prevent VS files in a PHP+MySQL app.

However you can ignore those files on your side by adding a custom ignore file in your repo (.git/info/exclude) or globally (~/git/ignore). See https://git-scm.com/docs/gitignore.

PR #39, opened 2018-12-08
fix core/expressioncache.class.inc.php multiple language support

annProg:
`Dict::GetUserLanguage()` get correct result only after `LoginWebPage::DoLogin()`, it made expressioncache always generated by default language(`EN US`), format define for `friendlyname` in other language will not work.

For example, default language is `EN US`, define
```php
// no Name define for EN US
Dict::Add('ZH CN', 'Chinese', '简体中文', array(
    'Class:ApplicationSolution/Name' => '%1$s / %2$s',
));
```
when user language is `ZH CN` , the format define above not work. It will display as default format `%1$s %2$s`

### DIFF
--- a/core/expressioncache.class.inc.php
+++ b/core/expressioncache.class.inc.php
@@ -21,13 +21,13 @@ class ExpressionCache
 {
 	static private $aCache = array();
 
-	protected function GetLanguge() {
+	protected function GetLanguage() {
 		return str_replace(" ", "", Dict::GetUserLanguage());
 	}
 
 	protected function GetCacheClass() {
-		$lang = self::GetLanguge();
-		return "ExpressionCacheData" . $lang;
+		$sLang = self::GetLanguage();
+		return "ExpressionCacheData" . $sLang;
 	}
 
 	static public function GetCachedExpression($sClass, $sAttCode)
@@ -35,7 +35,7 @@ class ExpressionCache
 		// read current cache
 		@include_once (static::GetCacheFileName());
 
-		$className = self::GetCacheClass();
+		$sClassName = self::GetCacheClass();
 
 		$oExpr = null;
 		$sKey = static::GetKey($sClass, $sAttCode);
@@ -45,11 +45,11 @@ class ExpressionCache
 		}
 		else
 		{
-			if (class_exists($className))
+			if (class_exists($sClassName))
 			{
-				if (array_key_exists($sKey, $className::$aCache))
+				if (array_key_exists($sKey, $sClassName::$aCache))
 				{
-					$sVal = $className::$aCache[$sKey];
+					$sVal = $sClassName::$aCache[$sKey];
 					$oExpr = unserialize($sVal);
 					static::$aCache[$sKey] = $oExpr;
 				}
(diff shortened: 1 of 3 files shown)


core/expressioncache.class.inc.php
```diff
@@ -21,11 +21,22 @@ class ExpressionCache
 {
 	static private $aCache = array();
 
+	protected function GetLanguge() {
```
jbostoen: Typo, also in other parts of this commit. Should be GetLanguage()

core/expressioncache.class.inc.php
```diff
@@ -21,11 +21,22 @@ class ExpressionCache
...
+		return str_replace(" ", "", Dict::GetUserLanguage());
+	}
+
+	protected function GetCacheClass() {
+		$lang = self::GetLanguge();
```
jbostoen: By Combodo standards, this should be $sLang

annProg:
@jbostoen Hi, I had fix typo and variable name. 

And I found this fix would not work with `Person` objects,  how to fix it?

- delete `data/cache-production/*`
- account with language `ZH CN` refresh  `Person` details page, friendlyname work fine
- another acount with language `EN US`  access iTop
- account with language `ZH CN`  refresh `Person` details page, friendlyname back to format defined in `EN US`

annProg:
@jbostoen Hi, the third commit had fix `Person` friendlyname format

core/expressioncache.class.inc.php
piRGoif: Yes, thanks Jeffrey ! 
The complete guidelines can be found here : https://www.itophub.io/wiki/page?id=2_5_0%3Acustomization%3Acoding_standards
We are working on a CONTRIBUTING.md file in this repo, should be out in a few weeks.

purplegrape:
I am planing to rename chinese translation(ZH CN) file from zh.dict*  to zh_cn.dict*

core/expressioncache.class.inc.php
annProg: I had rename $lang and $className to $sLang, $sClassName

piRGoif:
Hello,
Thanks for the feedback and the proposed fix !

At first sight, not sure your solution is the best one : having `ExpressionCache::Warmup()` in each UI.php call seems to be a bad idea.
Plus, the root cause is having user dependent data in a shared cache, so I would prefer having this fixed ?

We need to look at this with @eespie, stay tune !

eespie:
@annProg , @jbostoen Hi,
I had a look at your proposal, but I think I will change a little bit the way to handle the problem. It seems simpler to generate once at startup all the cache files for all the languages and then include the one corresponding to the user preferences at runtime.
I need more time to check the fix before I can integrate it in the develop branch.

eespie:
Hi,
After doing some tests, it seems that the expression cache is not used when the OQL is already cached. So in order to make this feature work, the OQL cache must also be dependent on the language.
I will investigate that.

bruno-ds:
Fixed by @eespie with these commits:
 - https://github.com/Combodo/iTop/commit/6ad27b43ab47342755a06804947de8902dd29ab2
 - https://github.com/Combodo/iTop/commit/b8b468195cbf568d41baec8246530e69607ac140

thank you @annProg, @purplegrape & @jbostoen

PR #41, opened 2018-12-12
rename chinese translation file from zh.dict*php to zh_cn.dict*php .

purplegrape:
in case guys from HongKong ,TaiWan or Singapore wana have their own translations.


annProg:
Dict filename is not the key to the problem, in fact ,you can add new language with any dict filename by format `language.dic.modulename.php`, The key is `Dict::Add` method
```
Dict::Add('ZH HK', 'Chinese HK', '繁体中文 香港', array(
 'Class:Kubernetes/Attribute:status' => '狀態',
));
```

image

purplegrape:
I knew this , 
but how to name a zh_hk translation file ? 
the best way is zh_hk.dict*php ,is'nt it?

piRGoif:
Thanks purplegrape, this is nice ! I'm merging this !

PR #43, opened 2018-12-15
improve alert with jquery-confirm

annProg:
I am not sure if this PR is necessary or whether it is implemented in an appropriate way. If not, just close this PR. But I hope that iTop can achieve a similar solution.

### preview
itop-alert1

itop-alert2


piRGoif:
Hello,
Thanks for the idea !
Since iTop 2.6 some of the errors are now thrown using a box similar to the SessionMessage (this enhancement was bring by the new uniqueness constraint, give it a try !). Mandatory fields errors are still using alert() though, but this may change in the future as the console layout will certainly evolve.
In conclusion I'm rejecting this PR, but you can still achieve the same thing with a dedicated module (see iPageUIExtension to add the js and css files in all of the console pages, plus add a specific JS to replace alert() with jQueryAlert calls).

PR #47, opened 2018-12-22
🔒 Adding more options to external authentication

jbostoen:
Extends **UserExternal** with two options. Currently only username is considered.

* ability to pass a token which will be validated (**optional** to be backward compatible)
* ability to specify an optional external URL login portal (used if the only authentication method is 'external' and user authentication within iTop fails)

---

* https://www.itophub.io/wiki/page?id=2_6_0%3Aadmin%3Auser_authentication_options 

I was a bit surprised to see an example where $_SERVER is not used, but $_COOKIE. Cookies can be easily manipulated. I'd at least recommend changing that to $_SESSION, but it may be vulnerable as well?

I can even imagine the easiest way to pass on information might still be to set a cookie in some cases, but then I think at least a second check (token) should be added. Hence the proposal to extend with an **optional** token-like variable. 

It's been tested before with the release/2.6 branch, I tried to find all things I changed and made a request now so it's open for discussion.

* Ticket: https://sourceforge.net/p/itop/tickets/1693/
* Thread: https://sourceforge.net/p/itop/discussion/1113906/thread/873dfa400c/

### DIFF
--- a/application/loginwebpage.class.inc.php
+++ b/application/loginwebpage.class.inc.php
@@ -564,12 +564,15 @@ EOF
 
 				case 'external':
 				// Web server supplied authentication
-				$bExternalAuth = false;
-				$sExtAuthVar = MetaModel::GetConfig()->GetExternalAuthenticationVariable(); // In which variable is the info passed ?
+				$sExtAuthVar = MetaModel::GetConfig()->GetExternalAuthenticationVariable(); // In which variable is the username passed?
+				$sExtAuthTokenVar = MetaModel::GetConfig()->GetExternalAuthenticationTokenVariable(); // In which variable is the token passed?
 				eval('$sAuthUser = isset('.$sExtAuthVar.') ? '.$sExtAuthVar.' : false;'); // Retrieve the value
-				if ($sAuthUser && (strlen($sAuthUser) > 0))
+				eval('$sAuthPwd = isset('.$sExtAuthTokenVar.') ? '.$sExtAuthTokenVar.' : false;'); // Retrieve the value. Named 'password' here to be in line with other authentication methods for further processing.
+
+				// Either the user variable is set; or if this is the only method (external) enabled to authenticate: make sure redirect URL can be used if set.
+				// The count is one and it is obviously an external case, so login mode is external (only). This is important later to check if redirect to a URL is needed.
+				if ( ($sAuthUser && (strlen($sAuthUser) > 0)) || count($aAllowedLoginTypes) == 1 ) 
 				{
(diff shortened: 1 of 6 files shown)


application/loginwebpage.class.inc.php
```diff
@@ -564,12 +564,15 @@ protected static function Login($iOnExit)
 
 				case 'external':
 				// Web server supplied authentication
-				$bExternalAuth = false;
```
jbostoen: This was present in the original code, but I couldn't find where it's used. Probably a leftover from a previous implementation?

jbostoen:
I noticed there's also some work being done on 2FA (Google Authenticator); perhaps this request can be considered in a similar way?

This would be a generic approach where people could easily add their own external authentication (including 2FA) of their choice as well. 

Our use case: we'd like to link this to an existing portal which offers authentication through national ID card; SMS; authentication apps (such as Google Authenticator but also a similar national app which might be even more popular since it's also used by major banks and government); ...

jbostoen:
Any further opinions on this?

I think it can't be turned into a separate extension since it's related to the authentication. Making it a bit more flexible by allowing redirect to another (web) authentication portal is definitely a nice-to-have for the purpose I have in mind. The token part could make it more secure. Backward compatibility should not be a problem. :)

bruno-ds:
after internal review, while being important, this is not a top priority for now.
When we'll move we'll probably try to make the `LoginWebPage::Login()` more extensible (example of possible code [here](https://github.com/Combodo/iTop/blob/feature/2FactorAuth/application/loginwebpage.class.inc.php#L527) 
but for the loginType handling (ie: [here](https://github.com/Combodo/iTop/blob/feature/2FactorAuth/application/loginwebpage.class.inc.php#L571) and [here](https://github.com/Combodo/iTop/blob/feature/2FactorAuth/application/loginwebpage.class.inc.php#L653))

I do not want to make promises that I can keep up with, so let's be clear: there is no work planned on this appart some spare time @eespie and I spend on.

jbostoen:
Thanks for the feedback @bruno-ds !
Can you just clarify what you mean with "when we'll move"? Do you mean move the code parts, making it more extensible or ...?

Molkobain:
I think he probably meant "When we will get to it".

bruno-ds:
Since we do not plan to include this into the 2.7 release, I'm really sorry to close this PR :( 

Since you already create the [sourceforge ticket 1693](https://sourceforge.net/p/itop/tickets/1693/), we'll keep trace of your proposal.

Thank you for your time!

PR #49, opened 2018-12-26
N°1947 add font DroidSansFallback to tcpdf

purplegrape:
raise this PR just to discuss CJK support in PDF export

since font dejavusans is not good to support CJK ,the exported pdf file is even not readable

so I add font DroidSansFallback to tcpdf to support CJK characters，it also works well with latin 

need further test for other language


piRGoif:
Hello,
This is indeed a problem now...

We're using TCPDF lib (/lib/tcpdf) that is bundled with some TTF fonts, DejaVu is part of them - DroidSansFallback is not : thanks for the integration work you've done in this PR !

The fonts are embedded in the generated PDF, so weight of the font file is a concern ! The main DroidSansFallback file weights 1.43MB, this is far too much to add it to every single PDF file generated !

We might instead use an option to choose a specific font when needed - and a font containing only one unicode script glyphs for example, instead of a very generic one ?

piRGoif:
Oh, and also, this should be based on the develop branch, as this is definitely a new feature, and has to be tested widely !

purplegrape:
font DroidSansFallback is wildly used in android smart phones 
it's small enough than others as far as I see
it's a good choice if you guys have not found  better fonts
what's more , it's good-looking and free to use

piRGoif:
My understanding was that the font was embedded in each PDF file. As I saw that the file weight 1.43MB, I was a bit worried !! :fearful: 

After testing on a contact list PDF export on my system, I'm getting a 147KB file before modification, and 254KB file after it, both files with the correct fonts included. I guess the PDF lib picks only the necessary parts of the font file :+1: 

I would like some modifications though, I'm adding a review :mag:

piRGoif (changes requested):
Please add those changes :
* base the modification on the develop branch instead of support/2.5
* add and use a new optional config parameter for the PDF font value, defaults to DroidSansFallback

Thanks !

purplegrape:
you are right，but .. I'm really not a PHPer. :worried:

piRGoif:
This might be easier with some explanations :)

Config parameter can be added in core/config.class.inc.php, \Config::$m_aSettings attribute. There are already lots of them, notice the keys 'default' (as default value, set to 'DroidSansFallback') and 'show_in_conf_sample' (to be set to false, so that the param won't be present by default).
I would suggest adding using the name "pdf_export_font".

To get the parameter, just use \MetaModel::GetConfig and \Config::Get, like this :
`$sFont = MetaModel::GetConfig()->Get('pdf_export_font');`

I would add the font name and size as PDFPage constructor parameters, with default values - might create a constant to share the value between Config and PDFPage. The \TCPDF::SetFont call in \PDFBulkExport::GetFooter could then be removed.

Are you ok to do those modifications ? If not, I could report modifications myself, but this may take me some time...

purplegrape:
I just wanna report this issue
this PR can be closed at any time if you guys have an acceptable solution
thanks !

piRGoif:
Hello,
Ok :)

I've just pushed 931bc74921148c2b1796998e8aba6e521bf51953, an implementation of the modifications. I will see with the rest of the dev team what they think about it... I will give a feedback here, and I will close this PR then :)

piRGoif:
Hello, 
The dev team is ok with the changes :)
I need to see product team now...

Can you give me some good examples of the benefits to switch fonts ? That means some strings that weren't supported before and will be after ?
You may attach a CSV I could import directly as a demo dataset ?

Thanks !

purplegrape:
1 DroidSansFallback itself is free to use ,good looking and has been widely proven on andriod smart phones to support many fonts including latin, CJK and others

2 DroidSansFallback works well via tcpdf ,the size of the output pdf file is acceptable ,not too large，it's also printable

3 iTop together via  DroidSansFallback  has been well tested for years without problem （for me since 2016）

the last thing we need attention is DroidSansFallback  licenced under apache 2.0 not GPL

see also https://en.wikipedia.org/wiki/Droid_fonts

purplegrape:
for licence issue, 
shall we add an announcement at iTop setup wizard just like tcpdf，jquery etc ？

Molkobain:
> for licence issue,
> shall we add an announcement at iTop setup wizard just like tcpdf，jquery etc ？

I think so yes.

piRGoif:
Hello,

> for licence issue,
> shall we add an announcement at iTop setup wizard just like tcpdf，jquery etc ？

You're perfectly right, I was wrong to forget this !
I've just force pushed the feature/pr49_pdf_font branch, the license was added in a1b253f26c36ad24099c4c7725e7f5407bf89383.

Can you send me some text samples to show what type of characters are now supported with this font ?

purplegrape:
> Hello,
> 
> > for licence issue,
> > shall we add an announcement at iTop setup wizard just like tcpdf，jquery etc ？
> 
> You're perfectly right, I was wrong to forget this !
> I've just force pushed the feature/pr49_pdf_font branch, the license was added in [a1b253f](https://github.com/Combodo/iTop/commit/a1b253f26c36ad24099c4c7725e7f5407bf89383).
> 
> Can you send me some text samples to show what type of characters are now supported with this font ?

please refer to google or wikipedia , I don't even have a full font list which DroidSansFallback support

piRGoif:
> > Can you send me some text samples to show what type of characters are now supported with this font ?
> 
> please refer to google or wikipedia , I don't even have a full font list which DroidSansFallback support

You said in the original description "the exported pdf file is even not readable", I supposed you had some issues with the previous version in your daily iTop use ? I am just searching for some text samples to justify the change ?

purplegrape:
> > > Can you send me some text samples to show what type of characters are now supported with this font ?
> > 
> > 
> > please refer to google or wikipedia , I don't even have a full font list which DroidSansFallback support
> 
> You said in the original description "the exported pdf file is even not readable", I supposed you had some issues with the previous version in your daily iTop use ? I am just searching for some text samples to justify the change ?

chinese sample text for your test 

Android是一种基于Linux的自由及开放源代码的操作系统，主要使用于移动设备，如智能手机和平板电脑，由Google公司和开放手机联盟领导及开发。尚未有统一中文名称，中国大陆地区较多人使用“安卓”或“安致”。Android操作系统最初由Andy Rubin开发，主要支持手机。2005年8月由Google收购注资。2007年11月，Google与84家硬件制造商、软件开发商及电信营运商组建开放手机联盟共同研发改良Android系统。随后Google以Apache开源许可证的授权方式，发布了Android的源代码。第一部Android智能手机发布于2008年10月。Android逐渐扩展到平板电脑及其他领域上，如电视、数码相机、游戏机、智能手表等。2011年第一季度，Android在全球的市场份额首次超过塞班系统，跃居全球第一。 2013年的第四季度，Android平台手机的全球市场份额已经达到78.1%。 2013年09月24日谷歌开发的操作系统Android在迎来了5岁生日，全世界采用这款系统的设备数量已经达到10亿台。

you can copy/paste it info iTop then export as pdf

piRGoif:
Hi, Product team is ok to integrate this modification :+1: , but having the existing font as default.
I'm going to do this change and merge, when finish I'll add a comment here :)

piRGoif:
Merge done on the develop branch (future 2.7.0 version)
Commits :
2f5f92ddca8b061f7615fc23699bc1a87b41ba5d
bdedd83368153ca695580f84cc2f1d9a0b32d868
5723e9a77e80ea804a9dea6d838bdabe216653cd

I'm closing this PR.

Thanks @purplegrape !

purplegrape:
cheers!

PR #50, opened 2019-01-03
:art: vlan_tag is always a number

chifu1234:


piRGoif:
Hello, can you give us some explanations ?

PR #51, opened 2019-01-03
:art: vlan_tag is always a number

chifu1234:


v-dumas:
In details, such change would impact all existing customers having text data in that field for whatever reason, It would prevent them from upgrading. We have a long experience of datamodel changes, they are very disruptive. This is why we avoid them in iTop core, especially the changes on existing object/field.
Nevertheless, thanks for your accurate feedback, and sorry if I looked a bit rude in my previous reply.

chifu1234:
@v-dumas 
In details, such change would impact all existing customers having text data in that field for whatever reason, It would prevent them from upgrading.
> If you do have text in this field it will overwrite the text with an "0".

We have a long experience of datamodel changes, they are very disruptive. This is why we avoid them in iTop core, especially the changes on existing object/field.
Nevertheless, thanks for your accurate feedback, and sorry if I looked a bit rude in my previous reply.
> I understand if you don't like to change the data model. It's just about the sorting algorithm, which with "int" data type is much better than "varchar".


I still can modify this with my own extension.
Thanks though for the quick reply!

PR #52, opened 2019-01-04
Update link to always reflect latest version

jduartedj:
This way there is no need for further updates of version number on link

### DIFF
--- a/install.txt
+++ b/install.txt
@@ -1,3 +1,3 @@
 For installation instructions, please refer to:
 
-https://wiki.openitop.org/doku.php?id=2_4_0:install:start
+https://wiki.openitop.org/doku.php?id=latest:install:start


bruno-ds:
@v-dumas: do you agree?
I think that it will lead some people to follow the "latest" instruction for "older" versions of iTop.

jduartedj:
@bruno-ds: True but in this case all the files must be maintained correctly, I updated this file from v2.4 meaning in v2.5 no one updated it.
Still if someone is doing an installation **it should be latest version always, correct?** I think this would even push people with older installs to install newer versions, it would raise awareness.

So I agree with you but I still think it would benefit iTop more to keep the link to latest.

piRGoif:
Indeed, I may have merged the PR a little too soon, as this has lead to some discussions in Combodo O:).
I will write here a feedback of the decisions made :)

piRGoif:
Hello,
We have decided that each branch should point to its corresponding version. We will fix this in the following days. We shouldn't forget this in the future as this action was added to our release checklist... but you are welcome to check the next times ;) 
Thanks !

jduartedj:
Great!
These comments also provide some kind of documentation for this issue so I trust it will not be forgotten.

Eventually It could be automated from the build (?) a simple script could populate the text file.

piRGoif:
> Eventually It could be automated from the build (?) a simple script could populate the text file.

Yes, we have an internal build tool that already does lots of things, we are currently studying its roadmap, this is one of the changes we would like to introduce.

PR #53, opened 2019-01-04
:bug: Synchro - Trigger - Before Update - dest_id could be null or equal to 0.

GurneyHallack:
A SynchroReplica dest_id is nullable but, as the default value in PHP for this field is 0, the TRIGGER BEFORE_UPDATE should check it as well.

A test for the bug is added as well. Here a manual test:
- Create a synchro, for example OSFamily (mine had id 10)
- Set the Full load interval to 30s (to avoid waiting to long for testing)
- Set to update the Delete policy and name:test_obsoleting the Update rules
- Create a new entry by SQL (`INSERT INTO synchro_data_osfamily_10(primary_key,name) VALUES ('test','test');`)
- Wait 40s (just to be sure)
- Check the SynchroReplica's status . Should be 'new' (`SELECT SynchroReplica WHERE sync_source_id = 10`)
- Launch a synchro_exec.php by cli/API (postman)
- Check the SynchroReplica's status . Should be 'obsolete'
- Update the entry by SQL (`UPDATE synchro_data_osfamily_10 SET name = 'test2345' WHERE primary_key = 'test';`)
- Check the SynchroReplica's status . Should be 'modified' with 'dest_id' equal to 0.
- Launch the synchro before the full load interval. It should raise an exception

The Synchro must but cannot find a object having id equal to 0.

> <p>#3071,Organization::0,>>> Beginning of SynchroReplica::Synchro, replica status is 'modified'.</p>
>         <p>ERROR: No result for the single row query: 'SELECT DISTINCT `OSFamily_typology`.`id` AS `OSFamilyid`, `OSFamily_typology`.`name` AS `OSFamilyname`, `OSFamily_typology`.`finalclass` AS `OSFamilyfinalclass`, CAST(CONCAT(COALESCE(`OSFamily_typology`.`name`, '')) AS CHAR) AS `OSFamilyfriendlyname` FROM `typology` AS `OSFamily_typology` WHERE ((`OSFamily_typology`.`id` = 0) AND COALESCE((`OSFamily_typology`.`finalclass` IN ('OSFamily')), 1))  '</p>

With checking if dest_id IS NULL OR dest_id = 0, the trigger will set the correct status.

### DIFF
--- a/synchro/synchrodatasource.class.inc.php
+++ b/synchro/synchrodatasource.class.inc.php
@@ -844,7 +844,7 @@ EOF
 		$sTriggerUpdate .= "   FOR EACH ROW";
 		$sTriggerUpdate .= "   BEGIN";
 		$sTriggerUpdate .= "      IF @itopuser is null THEN";
-		$sTriggerUpdate .= "         UPDATE `{$sReplicaTable}` SET status_last_seen = NOW(), `status` = IF(`status` = 'obsolete', IF(`dest_id` IS NULL, 'new', 'modified'), IF(`status` IN ('synchronized') AND ($sIsModified), 'modified', `status`)) WHERE sync_source_id = {$this->GetKey()} AND id = OLD.id;";
+		$sTriggerUpdate .= "         UPDATE `{$sReplicaTable}` SET status_last_seen = NOW(), `status` = IF(`status` = 'obsolete', IF(`dest_id` IS NULL OR `dest_id` = 0, 'new', 'modified'), IF(`status` IN ('synchronized') AND ($sIsModified), 'modified', `status`)) WHERE sync_source_id = {$this->GetKey()} AND id = OLD.id;";
 		$sTriggerUpdate .= "         SET NEW.id = OLD.id;"; // make sure this id won't change
 		$sTriggerUpdate .= "      END IF;";
 		$sTriggerUpdate .= "   END;";
--- a/test/phpunit.xml.dist
+++ b/test/phpunit.xml.dist
@@ -50,6 +50,9 @@
         <testsuite name="Application">
             <directory>application</directory>
         </testsuite>
+        <testsuite name="Synchro">
+            <directory>synchro</directory>
+        </testsuite>
     </testsuites>
 
     <!-- Code coverage white list -->
--- a/test/synchro/SynchroDataSourceTest.php
+++ b/test/synchro/SynchroDataSourceTest.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Combodo\iTop\Test\UnitTest\Synchro;
+
+use Combodo\iTop\Test\UnitTest\ItopDataTestCase;
+use CMDBSource;
+use MetaModel;
+use SynchroExecution;
+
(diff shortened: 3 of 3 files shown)


eespie:
Thank you for the fix.
It seems that the test is failing.

> Failed asserting that two strings are equal.
> Expected :'obsolete'
> Actual   :'synchronized'
> itop/test/synchro/SynchroDataSourceTest.php:58

It's the same behavior without or with the fix. May be I'm missing something.

GurneyHallack:
Hi,

Here is my test without the fix

```
sudo php vendor/bin/phpunit --testsuite Synchro
PHPUnit 6.5.8 by Sebastian Bergmann and contributors.
F                                                                   1 / 1 (100%)
----------
---------- testSynchroReplicaStatus
----------
Created Organization::3
Created Organization UnitTestOrganization
Created SynchroDataSource::2
Removing SynchroDataSource::2
Removing Organization::3

Time: 20.97 seconds, Memory: 6.00MB
There was 1 failure:
1) Combodo\iTop\Test\UnitTest\Synchro\SynchroDataSourceTest::testSynchroReplicaStatus
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'new'
+'modified'
/var/www/html/iTop/test/synchro/SynchroDataSourceTest.php:65
FAILURES!
Tests: 1, Assertions: 4, Failures: 1.
```

And with

```
gcouronne@DESKTOP-4B4BAE0:~/Documents/Works/html/iTop/test$ sudo php vendor/bin/phpunit --testsuite Synchro
PHPUnit 6.5.8 by Sebastian Bergmann and contributors.
.                                                                   1 / 1 (100%)
----------
---------- testSynchroReplicaStatus
----------
Created Organization::4
Created Organization UnitTestOrganization
Created SynchroDataSource::3
Removing SynchroDataSource::3
Removing Organization::4

Time: 21.39 seconds, Memory: 4.00MB
OK (1 test, 5 assertions)
```

It's strange that we don't have the same result...

piRGoif:
Hello, 

I was able to reproduce very easily, thanks for the detailed procedure !
I wasn't able to retrieve this bug in our database, I'm quite surprised. I'm wondering if the `dest_id IS NULL` test was working on older MySQL versions ? It is not on my dev machine (MySQL 5.7.19 on Windows), what version did you use on your side ?

Your fix seems all good to me, but doesn't fix existing triggers...
In consequence we prefer to close this PR, the bug was added in our database (reference N°2786), and our product team will review and plan it !

GurneyHallack:
Hi,

I don't remember and I won't try it again soon :s. Perhaps with a mysql docker, and checking version.

No, it doesn't fix existing triggers but when you do a setup, the setup will check updated Class and update/fix triggers if their SynchroDataSource should be updated (CheckDBConsistency).
Perhaps it could added in the db_analyzer, as a new function (eg. : check SynchroDataSource) ?

piRGoif:
Hello,
Thanks for your answer - well, it was just out of curiosity :)
Thanks also for your suggestion about existing triggers, this is indeed a good lead !

PR #55, opened 2019-01-11
add docker support

purplegrape:
1 add docker support just for quick deployment
2 mysql database and apcu module  not include
3 need further test and other improvemnt

add this file  for discussion

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM debian:stretch
+
+RUN apt-get update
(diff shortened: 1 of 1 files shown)


purplegrape:
basic usage：
docker build -t iTop .
docker run -d -p 80:80 iTop

Hipska:
I would not make a Docker image with a complete LAMP stack, as this is against the idea behind Docker and containerisation as principle.

jbostoen:
If Combodo wants to include this, I think part of the debate should also be about the versions used (PHP). Personally I'd go for highest compatible and stable version.

purplegrape:
just close it

PR #56, opened 2019-01-25
:sparkles: Add status.php for getting iTop's status

GurneyHallack:
Allow for HAProxy and monitoring to get iTop's status
:white_check_mark: Add tests for status

Add status.php to allow load balancer and monitoring to check iTop's instance status (if the MetaModel could startup the model only).
Returns a HTTP Code 200 if ok or a 500 if not.
Not remplacing extensive tests to check the integrity of the databases or iTop's model/workflows.

Signed-off-by: Guy Couronné <gcouronne:@sapiens.biz>


Hipska:
Wouldn't it be better if this was a new core operation of the REST API?

GurneyHallack:
I was thinking the status page not as only for REST but for all "clients".
I would prefer using ITOP_URL/status.php than ITOP_URL/webservices/rest.php?json_data=...{"operations":"status"} (more difficult for a admin user just checking if iTop is ok)
As for JIRA (with its /status and /rest/api/latest), the meaning is separeted).

rest.php would required a login user. That means that load balancers should use an account to check the status of iTop. Even if you use a specific account for it, i'm not fan of it.
Morever, this status page checks if iTop's configuration is setted, usable to startup iTop and the connection with the Database is correct. No more operation is done. 
With rest.php, each call to it would requires a select to the database (check if the user is logged on), and more PHP files would be loaded.
I have 6 web instance of iTop linked to the same database (with ProxySQL and cluster galera) but I don't want to overloard the databases (HAP checks it every 5 seconds).

But your idea might more than a simple Status (iTop could be started and DB connection is good) but a complete one (DB integrity is good, with results from CheckHKeys, CheckDataSources).

I'm working on an other idea : checking the satus of SynchroDataSouce by using API (last completed import date, if it was synchronized before the full load periodicity's end), this could be added in parallel.

piRGoif:
Hello,
Thanks for your contribution !
Why not simply add this page in an extension ?

GurneyHallack:
It could be added as an extension but I thought that if it's requiring a extension to simply check iTop's status, people will use custom made check instead of the possible "community" one (this status page is a start and could be optimized).

The path to check the status will be harder to remember (ITOP_URL/env-production/x-status-page/status.php) as well.

Why not add this page (or a better version of it ) directly into iTop ?

piRGoif:
> Why not add this page (or a better version of it ) directly into iTop ?

I was just asking question to get pros and cons :)

There is a concern on security with such a unauthenticated page, as there is with setup/phpinfo.php for example... How to mitigate it ?

GurneyHallack:
I returned the question to get your views about it :).

setup/phpinfo.php show sensible information and as it, it should have security about.

This status page does not use user's inputs, nor show confidential information (only returning HTTP 200/500 if iTop might be usable).
It's up to the web server to show exception or not, and this is not specific to this page.

The "login" page (ITOP_URL/pages/UI.php) is also a unauthenticated page, show the logo and load more information from iTop (check the session, load more PHP files).
I thought of this status page as a simpler and compresser way of checking iTop's status than check if ITOP_URL/pages/UI.php is up.

A example for another tool : https://jira.atlassian.com/status (i'm not authenticated but I can get JIRA's status).

status.inc.php
```diff
@@ -0,0 +1,64 @@
...
+function StatusGetAppRoot($sAppRootFilename = 'approot.inc.php') 
+{
+        $sAppRootFile = dirname(__FILE__).'/'.$sAppRootFilename;
+
+        /*
+        * Check that the approot file exists and has the appropriate access rights
+        */
+        if (!file_exists($sAppRootFile) || !is_readable($sAppRootFile)) 
+        {
+            throw new \Exception($sAppRootFile . ' is not readable');
```
Molkobain: We usually don't throw / show absolute paths as it would expose the application to a *full path disclosure* security flaw. So we would recommand to change the exceptions accordingly.

```suggestion
            throw new \Exception($sAppRootFilename . ' is not readable');
```

status.inc.php
```diff
@@ -0,0 +1,64 @@
...
+        StatusGetAppRoot();
+        
+        $sConfigFile = APPCONF.ITOP_DEFAULT_ENV.'/'.$sConfigFilename;
+
+        /**
+         * Check that the configuration file exists and has the appropriate access rights
+         */  
+        if (!file_exists($sConfigFile) || !is_readable($sConfigFile)) 
+        {
+                throw new \Exception($sConfigFile . ' is not readable');
```
Molkobain: Same as the previous comment.
```suggestion
                throw new \Exception($sConfigFilename . ' is not readable');
```

Molkobain:
Hi @GurneyHallack 

It's been a while sorry, we have been busy lately. Anyway, thank you for integrating the suggestions regarding the security concerns from a previous post!

You have reviewed your PR and are ok to integrate it in the project if you are willing to make those last changes:
- Keep the 200/5xx HTTP status code
- Make a JSON response instead of just a blank page. This would provide a more intuitive usage of the page by a human. The example you gave on Jira is great, you can do something similar.
- Surround the *StatusStartup()* with a try/catch and put the status code / error message in the JSON response as well.
- Change comments in the unit test files with your name instead of @eespie 's 😄

After that we should be all set for merger!

Molkobain:
Hi @GurneyHallack ,

Thanks for the updates! We'll talk about it in the coming days.

bruno-ds:
Since the tests wasn't compatible with our CI, we made some changes to the code, we used this occasion to add namespaces, move a file, ... :
https://github.com/Combodo/iTop/pull/79

PR #59, opened 2019-03-11
:bug: composer.json dependencies correctness

bruno-ds:
Two dependencies which should not block composer are moved into the `suggest` block.
Several wrong dependencies are removed from `require`.
The `config` block was misused.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,10 +1,6 @@
 {
   "require": {
     "php": ">=5.6.0",
-    "ext-mysql": "*",
-    "ext-ldap": "*",
-    "ext-mcrypt": "*",
-    "ext-cli": "*",
     "ext-soap": "*",
     "ext-json": "*",
     "ext-zip": "*",
@@ -12,6 +8,14 @@
     "ext-dom": "*",
     "ext-iconv": "*",
     "ext-gd": "*"
+  },
+  "suggest": {
+    "ext-libsodium": "Required to use the AttributeEncryptedString.",
+    "ext-openssl": "Can be used as a polyfill if libsodium is not installed",
+    "ext-mcrypt": "Can be used as a polyfill if either libsodium and openssl are not installed (libsodium and openssl are more secure)",
+    "ext-ldap": "Required to use LDAP as an identity provider",
+    "ext-posix": "Not required by the core, but some extensions uses it.",
+    "ext-imap": "Required by the extension \"Mail to ticket automation\""
   },
   "config": {
     "platform": {


composer.json
```diff
@@ -13,9 +9,8 @@
     "ext-iconv": "*",
     "ext-gd": "*"
   },
-  "config": {
-    "platform": {
-      "php": "5.6.0"
-    }
+  "suggest": {
+    "ext-mcrypt": "can be used as a polyfill if either libsodium and openssl are not installed (libsodium and openssl are more secure)",
+    "ext-ldap": "This extension is required to use LDAP as an identity provider"
```
Molkobain: Suggestion

```suggestion
    "ext-ldap": "Required to use LDAP as an identity provider"
```

composer.json
```diff
@@ -13,9 +9,8 @@
     "ext-iconv": "*",
     "ext-gd": "*"
   },
-  "config": {
-    "platform": {
-      "php": "5.6.0"
-    }
+  "suggest": {
+    "ext-mcrypt": "can be used as a polyfill if either libsodium and openssl are not installed (libsodium and openssl are more secure)",
```
Molkobain: Typo

```suggestion
    "ext-mcrypt": "Can be used as a polyfill if either libsodium and openssl are not installed (libsodium and openssl are more secure)",
```

composer.json
```diff
@@ -1,10 +1,6 @@
 {
   "require": {
     "php": ">=5.6.0",
-    "ext-mysql": "*",
-    "ext-ldap": "*",
-    "ext-mcrypt": "*",
-    "ext-cli": "*",
```
piRGoif: [Is documented as required](https://wiki.combodo.com/doku.php?id=latest:install:installing_itop#installing_the_prerequisites_on_debian_or_ubuntu_1604_and_above), but this package is only to have a /usr/bin/php, [as documented on Debian packages](https://packages.debian.org/fr/sid/php-cli)
So this is not a required php extension at all !

piRGoif (dismissed):

composer.json
```diff
@@ -13,9 +9,8 @@
     "ext-iconv": "*",
     "ext-gd": "*"
   },
-  "config": {
-    "platform": {
-      "php": "5.6.0"
-    }
+  "suggest": {
+    "ext-mcrypt": "Can be used as a polyfill if either libsodium and openssl are not installed (libsodium and openssl are more secure)",
+    "ext-ldap": "Required to use LDAP as an identity provider"
```
piRGoif: Only used for UserLDAP, non blocking for login (see \SetupUtils::CheckPhpAndExtensions)

piRGoif:
Discussed with Bruno : PhpStorm uses composer.json to fix PHP language level (see https://blog.jetbrains.com/phpstorm/2017/07/configuring-with-composer-in-phpstorm-2017-2/). Removing config/platform/php might not be a good idea...

bruno-ds:
I force-pushed a new version with 
```json
 "config": {
    "platform": {
      "php": "5.6.0"
    }
  }
```

As explained by @piRGoif it is required in order to 
 - force PHPstorm to syntax check the source code on the lowest supported version
 - force composer to install only dependencies compatible with the lowest supported version (instead of the local CLI version)

composer.json
```diff
@@ -13,6 +9,10 @@
     "ext-iconv": "*",
     "ext-gd": "*"
   },
+  "suggest": {
+    "ext-mcrypt": "Can be used as a polyfill if either libsodium and openssl are not installed (libsodium and openssl are more secure)",
+    "ext-ldap": "Required to use LDAP as an identity provider"
+  },
   "config": {
     "platform": {
       "php": "5.6.0"
```
bruno-ds: - force PHPstorm to syntax check the source code on the lowest supported version (see https://blog.jetbrains.com/phpstorm/2017/07/configuring-with-composer-in-phpstorm-2017-2/)
 - force composer to install only dependencies compatible with the lowest supported version (instead of the local CLI version)

piRGoif:
> I force-pushed a new version with

http://www.commitstrip.com/en/2019/03/11/the-force/

;o)

composer.json
```diff
@@ -13,6 +9,11 @@
     "ext-iconv": "*",
     "ext-gd": "*"
   },
+  "suggest": {
+    "ext-mcrypt": "Can be used as a polyfill if either libsodium and openssl are not installed (libsodium and openssl are more secure)",
+    "ext-ldap": "Required to use LDAP as an identity provider",
+    "ext-posix": "Not required by the core, but some extensions uses it."
```
bruno-ds: Required at least by itop-integrity. 
Adding this suggest entry removes PHPstorm warnings.

Hipska:
Maybe also suggest IMAP. Or is this already included? (Used by mail-to-ticket)

composer.json
Hipska: Isn't that needed to run the cron.php?

piRGoif:
> Maybe also suggest IMAP. Or is this already included? (Used by mail-to-ticket)

I created the /composer.json initially just to have an automatic switch in PHP Storm for the language level config entry. I also added some extensions to have an idea of what was mandatory to run iTop, but that was just the cherry on the cake ;).

Now Bruno really uses the file for its real purpose, and the as it was not really well written he created this PR as a place to exchange on its evolution.

Our main goal for now is to have a file :
* that won't raise an error when using Composer
* usable by PHP Storm to fix its language level
* eventually, *that would list PHP extensions required to run iTop*

As iTop requirements are listed in the official Wiki, the last point is really just an option, that we could drop. Concerning extensions requirements, we might have something in the extensions themselves (nested composer files ??), but we will see that later.

composer.json
piRGoif: Eric confirms that this is not at all an iTop requirement, but only PHP Storm is concerned.

bruno-ds:
I use composer, but only on feature branches, in order to create quick POC.

At the moment there is no plan to use composer farther than that.

piRGoif:
Comment after our team review : just add the openssl lib as suggest :)

piRGoif (changes requested):
add openssl  in suggest

composer.json
```diff
@@ -10,9 +10,12 @@
     "ext-gd": "*"
   },
   "suggest": {
+    "ext-libsodium": "Required to use the AttributeEncryptedString.",
+    "ext-opênssl": "Can be used as a polyfill if libsodium is not installed",
```
Hipska: `ê` -> `e`

bruno-ds: corrected, thank you

bruno-ds:
On my development env. I now have :
```shell
composer check-platform-reqs
ext-dom     20031129    success  
ext-gd      7.2.15      success  
ext-iconv   7.2.15      success  
ext-json    1.6.0       success  
ext-mysqli  7.2.15      success  
ext-soap    7.2.15      success  
ext-zip     1.15.4      success  
php         7.2.15      success  
```

piRGoif (approved):

PR #60, opened 2019-03-12
Update cron.cmd

Hipska:
Updated the `cron.cmd` file so that an update of iTop does not require you to update this file afterwards. The location to `php.exe` is now an argument.

Also removes references to old and unsupported php version.

### DIFF
--- a/webservices/cron.cmd
+++ b/webservices/cron.cmd
@@ -2,17 +2,17 @@
 REM
 REM To be scheduled by the following command:
 REM
-REM schtasks /create /tn "iTop Cron" /sc minute /tr "\"C:\Program Files\EasyPHP-5.3.6.0\www\iTop-trunk\webservices\cron.cmd\""
+REM schtasks /create /tn "iTop Cron" /sc minute /tr "\"C:\www\iTop\webservices\cron.cmd\""
 REM
 REM
 REM PHP_PATH must point to php.exe, adjust the path to suit your own installation
-SET PHP_PATH=C:\Program Files\EasyPHP-5.3.6.0\php\php.exe
+SET PHP_PATH=%~f1
 REM PHP_INI must contain the full path to a PHP.ini file suitable for Command Line mode execution
-SET PHP_INI=C:\Program Files\EasyPHP-5.3.6.0\php\php-cli.ini
+SET PHP_INI=%~dp1php.ini
 REM The double dash (--) separates the parameters parsed by php.exe from the script's specific parameters
 REM %~p0 expands to the path to this file (including the trailing backslash)
-SET CRON_SCRIPT=%~p0cron.php
+SET CRON_SCRIPT=%~dp0cron.php
 REM Adjust the path below if you use a param files stored in a different location
-SET PARAMS_FILE=%~p0cron.params
+SET PARAMS_FILE=%~dp0cron.params
 REM Actual PHP invocation
-"%PHP_PATH%" -c "%PHP_INI%" -f "%CRON_SCRIPT%" -- --param_file="%PARAMS_FILE%" --verbose=1 >> "%~p0log.txt"
+"%PHP_PATH%" -c "%PHP_INI%" -f "%CRON_SCRIPT%" -- --param_file="%PARAMS_FILE%" --verbose=1 >> "%~dp0..\log\cron.log"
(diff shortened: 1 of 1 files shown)


piRGoif:
Hello Thomas, hope you made a good trip back home and enjoyed yesterday event :)

As I was saying I'm not very competent in windows batch syntax (I was though but it was a long time ago and it was DOS :o) )

I found this wiki page that has lots of info :
https://en.wikibooks.org/wiki/Windows_Batch_Scripting#Paths
Do you have any other source you can share ?

webservices/cron.cmd
```diff
@@ -2,17 +2,17 @@
 REM
 REM To be scheduled by the following command:
 REM
-REM schtasks /create /tn "iTop Cron" /sc minute /tr "\"C:\Program Files\EasyPHP-5.3.6.0\www\iTop-trunk\webservices\cron.cmd\""
+REM schtasks /create /tn "iTop Cron" /sc minute /tr "\"C:\www\iTop\webservices\cron.cmd\""
```
piRGoif: I would recommend to have an example with the PHP interpreter path parameter ?

webservices/cron.cmd
```diff
@@ -2,17 +2,17 @@
...
 REM PHP_INI must contain the full path to a PHP.ini file suitable for Command Line mode execution
-SET PHP_INI=C:\Program Files\EasyPHP-5.3.6.0\php\php-cli.ini
+SET PHP_INI=%~dp1php.ini
 REM The double dash (--) separates the parameters parsed by php.exe from the script's specific parameters
 REM %~p0 expands to the path to this file (including the trailing backslash)
-SET CRON_SCRIPT=%~p0cron.php
+SET CRON_SCRIPT=%~dp0cron.php
 REM Adjust the path below if you use a param files stored in a different location
-SET PARAMS_FILE=%~p0cron.params
+SET PARAMS_FILE=%~dp0cron.params
```
piRGoif: Might not be a good idea to put this file in the itop/webservices dir ? Maybe in conf ?

piRGoif:
I didn't put this in the review as I'm not sure if it's possible but I would recommend adding variables at the top of the script with a clear name to store both paths ? The script (%0) and PHP interpreter (%1) ?
It would improve readability A LOT for people like me who forgot all about .bat files syntax O:)

webservices/cron.cmd
Hipska: It has always been like that, I didn't change the location, only added support if iTop is not on the C drive

Hipska:
They are? See line 9 (and partially in line 14)

Reference to parameter extensions: https://ss64.com/nt/syntax-args.html

webservices/cron.cmd
piRGoif: Right :) We'll see if we change that later.
I'm solving this conversation !

piRGoif:
Thanks for the pointer :)

Sorry I wasn't very comprehensive on my request :D I meant something like :

```
$ParameterValue = ...;
$CurrentScriptPath = ...;
...
SET PHP_PATH=$Parametervalue
...
SET PHP_INI=$ParameterValue php.ini
...
SET CRON_SCRIPT=$CurrentScriptPath cron.php
```

Hipska:
Okay, but that also wasn't the case before. And your suggestion will not really work as `$ParameterValue` would be `C:\php\php.exe` and then `PHP_INI` would not have a correct value. For `CRON_PATH` this might work.

Preferably, I would then even get rid of `PHP_PATH` and `CRON_SCRIPT` and use the following as last line:
```
"%~f1" -c "%PHP_INI%" -f "%CRON_PATH%cron.php" -- --param_file="%PARAMS_FILE%" --verbose=1 >> "%CRON_PATH%..\log\cron.log"
```

piRGoif:
Well, before there was paths that everyone can understand, now something like %~dp1 is a bit cryptic for lots of people I guess...

I would suggest to have variables for :
* PHP command (%~f1)
* PHP path (%~dp1)
* current script dir (%~dp0)

And build from there variables for :
* php.ini full path
* cron.php full path
* parameters file full path
* log file full path

I insist as this file will live long without modifications, and I'm pretty sure no one in Combodo (or its partners, clients, ...) will even remember what %~dp1 means in 6 months or so, just like I forgot everything I knew about batch syntax when I stopped wrote such files.

piRGoif:
Many thanks @Hipska, I'm merging the PR !

PR #61, opened 2019-03-13
Make setup backup location and name similar as other backups

Hipska:
As discussed in [#1309 Backup location during setup](https://sourceforge.net/p/itop/tickets/1309/), I have made a change so that the backup file that the setup generates is similar in naming convention as the backups created by the cron job.

I also included a check if the directory to place the backup exists.

### DIFF
--- a/setup/wizardsteps.class.inc.php
+++ b/setup/wizardsteps.class.inc.php
@@ -220,7 +220,7 @@ class WizStepInstallOrUpgrade extends WizardStep
 		$sPreviousVersionDir = '';
 		if ($sInstallMode == '')
 		{
-			$sDBBackupPath = APPROOT.'data/'.ITOP_APPLICATION.strftime('-backup-%Y-%m-%d');
+			$sDBBackupPath = APPROOT.'data/backups/manual/__DB__-%Y-%m-%d_%H_%M';
 			$bDBBackup = true;
 			$aPreviousInstance = SetupUtils::GetPreviousInstance(APPROOT);
 			if ($aPreviousInstance['found'])
--- a/setup/wizardsteps.class.inc.php
+++ b/setup/wizardsteps.class.inc.php
@@ -220,7 +220,7 @@ class WizStepInstallOrUpgrade extends WizardStep
 		$sPreviousVersionDir = '';
 		if ($sInstallMode == '')
 		{
-			$sDBBackupPath = APPROOT.'data/backups/manual/__DB__-%Y-%m-%d_%H_%M';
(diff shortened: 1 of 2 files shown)


setup/wizardsteps.class.inc.php
```diff
@@ -220,7 +220,7 @@ public function Display(WebPage $oPage)
 		$sPreviousVersionDir = '';
 		if ($sInstallMode == '')
 		{
-			$sDBBackupPath = APPROOT.'data/'.ITOP_APPLICATION.strftime('-backup-%Y-%m-%d');
+			$sDBBackupPath = APPROOT.'data/backups/manual/__DB__-%Y-%m-%d_%H_%M';
```
Hipska: I have chosen `manual` as it is then as well visible in the backups status. Maybe a separate directory `setup` eg could be used?

piRGoif: This seems like a very good choice as the backups will then appear in the console (admin tools / Scheduled backups) and the user will be able to make a restore or download it very easily !

Hipska: If we later would switch to `setup` eg, the console page could be updated to also list files in that directory together with the ones from `manual` directory? Anyway these are things for later 😉

piRGoif: Yeah I agree !
Indeed we reviewed this PR with the team this morning, we thought it wasn't worst the effort to have a dedicated dir on the disk and group in the console page.

piRGoif: I would suggest keeping the strftime() call to have the final value directly in the UI : admins installing iTop are not PHP dev and I think they would be pretty confused seeing a path containing such strange characters...

Hipska:
The warning comes from [wizardsteps.class.inc.php line 2517-2518](/Combodo/iTop/tree/develop/setup/wizardsteps.class.inc.php#L2517-L2518) where it checks for the not replaced filename.

Maybe instead of setting `__DB__` (which will be replaced by DB name) we could use `setup`, so a file named `setup-2019-03-20_12_26.tar.gz` would be created?

piRGoif:
Hello !

Thanks for the investigation !

> Maybe instead of setting `__DB__` (which will be replaced by DB name) we could use `setup`, so a file named `setup-2019-03-20_12_26.tar.gz` would be created?

I agree, it would also help to differentiate with manual backups in the "scheduled backups" iTop page !
Do that and I'll merge the PR (doing a squash & merge if you are ok with it ?)

Hipska:
Thanks, did the change and pushed. Ready to merge.

PR #63, opened 2019-03-13
:note: PhpDoc generation for the public API

bruno-ds:
The goal of this PR is to generate a phpdoc of the object manipulation API.

It will be an iterative process, with many drafts.


core/dbobject.class.php
```diff
@@ -179,19 +225,42 @@ protected function UpdateMetaAttributes($aAttCodes = null)
 	}
 
 	// Read-only <=> Written once (archive)
```
piRGoif: Remove this or move it to the PHPDoc block !

core/dbobject.class.php
```diff
@@ -179,19 +225,42 @@ protected function UpdateMetaAttributes($aAttCodes = null)
...
 	public function RegisterAsDirty()
 	{
-		// While the object may be written to the DB, it is NOT possible to reload it
-		// or at least not possible to reload it the same way
-		$this->m_bDirty = true;	
+		$this->m_bDirty = true;
 	}
 
+    /**
+     * Whether the object is already persisted in DB or not.
```
piRGoif: This is duplicating the m_bIsInDB attribute phpdoc... I would either replace all of this with a @see, or the opposite (only a @see on the attribute and the doc in the method). I would prefer the 1st solution

core/dbobject.class.php
```diff
@@ -179,19 +225,42 @@ protected function UpdateMetaAttributes($aAttCodes = null)
...
 		return (!$this->m_bIsInDB);
 	}
 
-	// Returns an Id for memory objects
+    /**
+     * Returns an Id for memory objects
+     * 
+     * @internal
+     * 
+     * @param $sClass
```
piRGoif: add the type hint string ?

core/dbobject.class.php
```diff
@@ -203,6 +272,17 @@ static protected function GetNextTempId($sClass)
 		return (- self::$m_aMemoryObjectsByClass[$sRootClass]);
 	}
 
+    /**
+     * String representation of the object
+     *
+     * Only a few meaningful information will be returned.
+     * This representation is subject to change.
```
piRGoif: are @api and "subject to change" compatible ? O:)

core/dbobject.class.php
```diff
@@ -214,12 +294,31 @@ public function __toString()
...
 	public function DBRevert()
 	{
 		$this->Reload();
 	}
 
+    /**
+     * Is the current instance is fully or partially loaded.
+     *
+     * This method compute the state in realtime.
+     * In almost every case it is preferable to use {@see DBObject::m_bFullyLoaded}.
```
piRGoif: I would have written @link instead of @see, but maybe I'm wrong ?

core/dbobject.class.php
```diff
@@ -234,10 +333,15 @@ protected function IsFullyLoaded()
...
 
 	/**
-	 * @param bool $bAllowAllData DEPRECATED: the reload must never fail!
+     * Fetch the object from the DB.
+     *
+     * This will erase any pending changes.
+     *
+     * @api
+     *
+	 * @param bool $bAllowAllData @deprecated the reload must never fail
```
piRGoif: This phrase is not very clear, I don't understand it... The attribute is not used in the method anymore, what was it used for ?

core/dbobject.class.php
```diff
@@ -381,13 +500,42 @@ protected function FromRow($aRow, $sClassAlias = '', $aAttToLoad = null, $aExten
...
+    /**
+     * Protected raw Setter
+     *
+     * This method is an internal plumbing : it set the value without doing any of the required processes.
+     * The exposed API Setter is {@see DBObject::>Set()}
+     *
+     * @internal
+     * 
+     * @param $sAttCode
+     * @param $value
```
piRGoif: don't forget to add type hinting...

core/dbobject.class.php
```diff
@@ -504,12 +673,39 @@ public function SetTrim($sAttCode, $sValue)
 		$this->Set($sAttCode, $sValue);
 	}
 
+    /**
+     * Get the {@see AttributeDefinition}'s label.
```
piRGoif: Definitively, here it should be a @link instead of a @see

core/dbobject.class.php
```diff
@@ -504,12 +673,39 @@ public function SetTrim($sAttCode, $sValue)
...
+     * 
+     * Shortcut to the field's {@see AttributeDefinition->GetLabel()}
+     *
+     * @api
+     * 
+     * @param string $sAttCode
+     *
+     * @return string
+     *
```
piRGoif: there are lots of spaces on this line, beware of configuring your code formatter properly

core/dbobject.class.php
```diff
@@ -1162,7 +1621,10 @@ public function GetStateDescription()
 	}
 
 	/**
-	 * Overridable - Define attributes read-only from the end-user perspective
+	 * Define attributes read-only from the end-user perspective
+     * 
+     * @api 
+     * @overwritable-hook This method is an extension point meant to be extended in order to define the read-only attributes
 	 * 	 
 	 * @return array List of attcodes
```
piRGoif: Sure ? The code does : `return null;`

core/dbobject.class.php
```diff
@@ -1232,12 +1699,17 @@ public function GetAttributeFlags($sAttCode, &$aReasons = array(), $sTargetState
...
-	 */
+    /**
+     * Whether the attribute is read-only
+     *
+     * @internal
+     *
+     * @param string $sAttCode
+     * @param array  $aReasons To store the reasons why the attribute is read-only (info about the synchro replicas)
+     *
+     * @return int
```
piRGoif: You should add some @link to the corresponding constants

core/dbobject.class.php
```diff
@@ -1476,17 +1959,21 @@ public function CheckValue($sAttCode, $value = null)
 	/**
 	 * check attributes together
 	 *
-	 * @return bool
-	 * @api
+     * @overwritable-hook You can extend this method in order to provide your own logic.
+     * 
+	 * @return bool
```
piRGoif: it would be interesting to know here what happens if the method returns false (or why should you override this ?)

core/dbobject.class.php
```diff
@@ -4142,6 +4963,15 @@ public function IsArchived($sKeyAttCode = null)
 		return $bRet;
 	}
 
+    /**
+     * is the instance obsolete
+     *
+     * @param null $sKeyAttCode
```
piRGoif: string instead of null

core/dbobject.class.php
```diff
@@ -4131,6 +4941,17 @@ public function ExecAction($sVerb, $aParams, $aSourceObjects)
 		}
 	}
 
+    /**
+     * Is the instance archived
+     *
+     * @api
+     *
+     * @param null $sKeyAttCode
```
piRGoif: string instead of null

core/dbobjectsearch.class.php
```diff
@@ -17,9 +17,14 @@
 //   along with iTop. If not, see <http://www.gnu.org/licenses/>
 //
 
-// Dev hack for disabling the some query build optimizations (Folding/Merging)
+/** @internal Dev hack for disabling the some query build optimizations (Folding/Merging) */
```
piRGoif: "for disabling some query ..." (remove "the")

piRGoif (changes requested):
Some generic comments :
* always add type hinting when adding @param
* beware of the useless { and [
* sometimes @link is a better fit than @see
* not sure "@todo: better document this method" are a good idea... I would better create a ticket referencing all of the methods that needs to have better doc ?
* some lines with useless spaces, beware of your code formatter

piRGoif (commented):
Is the  ast.dump file should be versionned ?

core/dbobject.class.php
bruno-ds: on the resulting phpdoc, the navigation is not really easy. I'd rather attach meaningful information the the api endpoint.

core/dbobject.class.php
bruno-ds: mmh, the difference is so subtle that I don't understand it:
> https://docs.phpdoc.org/guides/docblocks.html#list-of-tags
> link: indicates a relation between the associated element and a page of a website.
> see: indicates a reference from the associated element to a website or other elements.

for now I decided to use @see for internal relations and @link for link to external websites.
Do you disagree?

core/dbobject.class.php
bruno-ds: AllowAllData has a strong sematic: if true, then the user rigths overs the classes are not check. it's like a [ring 0](https://en.wikipedia.org/wiki/Protection_ring)

core/dbobject.class.php
bruno-ds: I tried to respect a schema: 
/**
* Title
* [space]
* long description, possibly multiline
* [space]
* @api and associates
* [space]
* @param
* [space]
* @return
* [space]
* @throws
*/

What format would you prefer?

Hipska: No space between @ items.

PS, use ` to comment code.

core/dbobject.class.php
bruno-ds: By reading the use of the method, the signature should be 
```
@return bool|string true mean the consistency is correct, returning a string mean the consistency is not correct, the string should be a human readable description of the problem
```

core/dbobjectsearch.class.php
bruno-ds: corrected.
But since it has the @internal tag, it is now excluded, soit should be out of scope.

core/dbobject.class.php
piRGoif: I meant, on the line 686 you had *<space><space>... (trailing spaces)

core/dbobject.class.php
piRGoif: That's better :) But if you return a string that means persisting is cancelled ?

bruno-ds: if the !== true then DoCheckToWrite() append a $this->m_aCheckIssues[], then CheckToWrite return fale, then,  DBObject trigger a  CoreCannotSaveObjectException while cmdbAbstractObject simply refuse to write the object.

=> the final behaviour vary, but in general, yes, "persisting is cancelled".

@rquetiez will probably finalize the comments, I'll let him coose how far we go in the phpdoc.

core/dbobject.class.php
bruno-ds: finaly, the attributes are excluded from the doc, they generated too much noise.

piRGoif: Ok :) It would be nice to have only a @see on the attribute so...

core/dbobject.class.php
bruno-ds: We'll debate of this internaly

core/dbobject.class.php
```diff
@@ -308,7 +306,7 @@ public function DBRevert()
 	}
 
     /**
-     * Is the current instance is fully or partially loaded.
+     * Is the current instance fully or partially loaded.
```
piRGoif: I prefer this to be on the @return line, what do you think ?

core/dbobject.class.php
```diff
@@ -153,7 +153,8 @@ abstract class DBObject implements iDisplay
      * You should preferably use MetaModel::NewObject() instead of this constructor.
      * The whole collection of parameters is [*optional*] please refer to DBObjectSet::FromRow()
      *
-     * @api The availability of this method is not guaranteed in the long term, you should preferably use MetaModel::NewObject().
+     * @deprecated The availability of this method is not guaranteed in the long term, you should preferably use MetaModel::NewObject().
+     * @see MetaModel::NewObject()
```
bruno-ds: beware: 
Since `MetaModel` is not yet included in the scope of the generated doc, 
this `@see MetaModel::NewObject()` will result in a dead link on the generated API page.

core/dbobject.class.php
```diff
@@ -153,7 +153,7 @@ abstract class DBObject implements iDisplay
      * You should preferably use MetaModel::NewObject() instead of this constructor.
      * The whole collection of parameters is [*optional*] please refer to DBObjectSet::FromRow()
      *
-     * @deprecated The availability of this method is not guaranteed in the long term, you should preferably use MetaModel::NewObject().
+     * @internal The availability of this method is not guaranteed in the long term, you should preferably use MetaModel::NewObject().
```
bruno-ds: Putting an @internal will prevent this method to be present in the generated doc.
It is probably by intent, so feel free to close this without comment  ;)

core/dbobject.class.php
```diff
@@ -287,7 +286,7 @@ public function __toString()
         $sClass = get_class($this);
         $sRootClass = MetaModel::GetRootClass($sClass);
         $iPKey = $this->GetKey();
-        $sFriendlyname = $this->Get('friendlyname');
+        $sFriendlyname = $this->GetAsHTML('friendlyname');
```
bruno-ds: For the sake of clarity of those commits, you should avoid changes in the code ;)

core/dbobject.class.php
bruno-ds: I'm ok with you, but a title as to be present otherwise the next bloc of text will become the title, it would be awful in the generated doc.

core/dbobject.class.php
```diff
@@ -697,13 +690,14 @@ public function GetLabel($sAttCode)
 	}
 
     /**
-     * Getter
+     * Getter : get a value from the current object of from a related object
```
bruno-ds: typo `of from`

core/dbobject.class.php
```diff
@@ -1488,9 +1496,11 @@ public function GetIcon($bImgTag = true)
 	}
 
 	/**
-	 * Get the name as defined in the dictionary
+	 * Get the label of a class
+	 *
+	 * Returns the label as defined in the dictionary for the language of the current user
```
bruno-ds: shouldn't this be in the @return tag?

core/dbobject.class.php
```diff
@@ -2851,8 +2851,11 @@ public function MakeInsertStatements($aAuthorizedExtKeys, &$aStatements)
 		}
 	}
 
-	/**
+	/*
+	* Persist an object to the DB, for the first time
+	*
      * @api
+     * @see DBWrite
```
bruno-ds: @see must have a very special signature as stated in the known limitations here : 
https://github.com/Combodo/iTop/blob/feature/doc-object-manipulation/.doc/README.md

core/dbobject.class.php
```diff
@@ -3286,13 +3290,14 @@ protected function DBDeleteSingleObject()
     /**
      * Delete an object
      *
-     * And guarantee data integrity
+     * Pior to the delete the object, it checks if this is feasible to guarantee the DB integrity.
```
bruno-ds: multiple typos : 
> *Pior* to *the* delete the object
Prior to delete the object

core/dbobject.class.php
```diff
@@ -3579,11 +3585,10 @@ public function ApplyStimulus($sStimulusCode, $bDoNotWrite = false)
 	}
 
 	/**
-	 * Lifecycle action: Recover the default value (aka when an object is being created)
+	 * Helper to recover the default value (aka when an object is being created)
+     * Suitable for use as a lifecycle action
```
bruno-ds: maybe we could create a @lifecycle  in order to put an emphasis on this key principle?
(I could render it as IO did for the @api-advanced)

piRGoif:
Rebased manually branch with @bruno-ds to avoid future conflicts (there was one on /composer.json)

piRGoif (approved):

PR #64, opened 2019-03-15
Update Person list view

Hipska:
In the default data model, a list of persons has the columns friendly name, followed by first name. You can sort on both columns, but as friendly name also starts with the first name, it is not possible to sort by last name. Sorting by last name is  more logical in this context of persons, so I replaced the column `first_name` by `name` (the last name). By this, sorting on last name is possible.


piRGoif:
I feared this could cause problem, as any change in the default datamodel could have impacts on all of the customatizations made by all iTop users (we are working on adding this in our contribution guide by the way)

But I checked with our product team (especially @v-dumas) and they agreed to integrate this contribution !

PR #65, opened 2019-03-15
:zap: Optimize Column load when using api core/get

GurneyHallack:
Getting in SQL only desired columns to show
Used when output_fields != '*' || '*+'

When using to get 500 Server's name by a complex OQL, we reduce the delay to 
0.9s to 0.8s on the first query
Then 0.9s to 0.2s for the following (SQL cache)

Signed-off-by: Guy Couronné <gcouronne@sapiens.biz>

### DIFF
--- a/core/restservices.class.inc.php
+++ b/core/restservices.class.inc.php
@@ -467,6 +467,12 @@ class CoreServices implements iRestServiceProvider
             }
 			else
 			{
+                                if (!$bExtendedOutput && RestUtils::GetOptionalParam($aParams, 'output_fields', '*') != '*') 
+                                {
+                                        $aAttToLoad = array($oObjectSet->GetClassAlias() => $aShowFields[$sClass]);
+                                        $oObjectSet->OptimizeColumnLoad($aAttToLoad);
+                                }
+                                
 				while ($oObject = $oObjectSet->Fetch())
 				{
 					$oResult->AddObject(0, '', $oObject, $aShowFields, $bExtendedOutput);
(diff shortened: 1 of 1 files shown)


Hipska (approved):

GurneyHallack:
I updated the files as it was generating a error when using 'id' as output_fields.

Hipska (approved):

core/restservices.class.inc.php
```diff
@@ -467,6 +467,18 @@ public function ExecOperation($sVersion, $sVerb, $aParams)
             }
 			else
 			{
+                                if (!$bExtendedOutput && RestUtils::GetOptionalParam($aParams, 'output_fields', '*') != '*') 
+                                {
+                                        $aFields = $aShowFields[$sClass];
+                                        //Id is not a valid attribute to optimize
+                                        if (in_array('id', $aFields)) 
+                                        {
+                                            unset($aFields[array_search('id', $aFields)]);
```
jbostoen: Just wondering: why not simply unset($aFields['id']) ?

bruno-ds: 'id' is a value within the array, not a key :)

jbostoen: Of course, so obvious!

Molkobain:
Hello @GurneyHallack , sorry this took so long to be processed. Now that the LTS version is almost behind us, we can consider integrating contributions of this kind.

It was submitted in Combodo DB as N°2766, candidate for iTop 2.8 if:
- Performance gain
- Doesn't bring regression

Hipska (approved):

eespie (approved):
tested and approved for 2.8

PR #66, opened 2019-03-15
Updated Demo Organization parent/child relation

Hipska:
This updates the demo data so that "IT Department" is now a child of "Demo" organization

### DIFF
--- a/datamodels/2.x/itop-config-mgmt/data.sample.organizations.xml
+++ b/datamodels/2.x/itop-config-mgmt/data.sample.organizations.xml
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Set>
-<Organization alias="Organization" id="3">
-<name>IT Department</name>
-<code>IT</code>
-<status>active</status>
-<parent_id>0</parent_id>
-<deliverymodel_id></deliverymodel_id>
-</Organization>
-<Organization alias="Organization" id="2">
-<name>Demo</name>
-<code></code>
-<status>active</status>
-<parent_id>0</parent_id>
-<deliverymodel_id></deliverymodel_id>
-</Organization>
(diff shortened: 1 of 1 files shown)


Molkobain:
Hi @Hipska ,

What do you have in mind with this? *IT Dept* is not necessarely part of the *Demo* organization.

Hipska:
As I have seen in most of Combodo demo's it is and it is also easier to explain new customers/users if that structure is already reflected in default example data.

Molkobain:
As this is rather a _product owner_ decision, I'll forward it to @v-dumas .

v-dumas:
We could do it, but should we lose time on such insignificant change? I don't think so.

Hipska:
Well, I did it already for you, so you don't have to loose time 😉

v-dumas:
All our tests rely on the demo data, changing them may break them, just thinking and checking if it does or not is a lost of time

Hipska:
Okay, could you point me to that specific test that verifies the organization parent? Then I will update that one as well.. 👍

zipkid:
> All our tests rely on the demo data, changing them may break them, just thinking and checking if it does or not is a lost of time

This kind of reaction makes me doubt if iTop is a product to adopt.... :-(

Molkobain:
Hi guys,

What Vincent meant is that sample data are not subject to PRs yet, especially as they are part of our functional tests (part automated through Behat, part manual).

As a friendly reminder, for now we only consider PRs on *bug fixes* and *translations* (you can take a look at the [contributing guide](https://github.com/Combodo/iTop/blob/develop/CONTRIBUTING.md) for more detailled information).

Regarding this particular demand, the hierarchy was not intended but maybe we send mixed signals. That being said, you can easily bring your own samples through an extension if necessary 😊

PR #67, opened 2019-03-15
:chart_with_upwards_trend: Add KPI on API Rest

GurneyHallack:


Hipska:
Hi Guy, could you elaborate a little?

GurneyHallack:
Hi Thomas,

When searching to optimize a API query (a complexe OQL query with Server and 6 other classes and returning only the name of the server) which takes 0.9 s for 437 elements, I found that they weren't any KPI available, to identify what it's causing the delay.

After implementing some, I found that the SQL query was taking 0.67 s, because it was getting all the fields of Server (#65 created as a result).

I think it would good to be able to analyse the performance of the API. Thus, I added some KPI but feel free to optimize them.

webservices/rest.php
```diff
@@ -118,9 +118,16 @@ function json_last_error_msg() {
 try
 {
 	utils::UseParamFile();
-
+        
+	$oKPI = new ExecutionKPI();
+	$oKPI->ComputeAndReport('Data model loaded');
```
piRGoif: Are there any use for this log ?

GurneyHallack: Hi,

I based this commit, from https://github.com/Combodo/iTop/blob/develop/pages/UI.php
Line 354 for "Data model loaded"
Line 353 & 356 for recreating the mode.

I could update my code to use only one instance (so creating before the Try/Catch)

piRGoif: Ok ! The lines with ExecutionKPI in UI.php are quite old and not the best way to do :)

You could keep the 'Data model loaded' trace, but we would prefer only one ExecutionKPI instance to be used. Commit this and I'll merge !

Thanks !

piRGoif:
Hello,
Thanks for the modifications, ok to merge !

PR #71, opened 2019-03-26
make demo_mode effect more expressive

bruno-ds:
Since an empty json could be a bug, 
Make demo_mode effect expressive.

### DIFF
--- a/datamodels/2.x/itop-hub-connector/launch.php
+++ b/datamodels/2.x/itop-hub-connector/launch.php
@@ -234,7 +234,7 @@ function MakeDataToPost($sTargetRoute)
 	if (MetaModel::GetConfig()->Get('demo_mode'))
 	{
 		// Don't expose such information in demo mode
-		$aDataToPost = array();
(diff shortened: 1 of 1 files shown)


Molkobain:
Why don't we send anything when the instance is in demo mode? It could be someone using the iTop as a showcase but still getting extensions from the Hub from time to time. So I think we could use having the instance information on our side.

bruno-ds:
with @dflaven we tough it could possibly lead to security issue, so we preferred to not divulgate anything.

rquetiez (approved):

piRGoif:
The team did validate this in PR review this afternoon !

@bruno-ds, a last though, shouldn't this be merged to master or support/2.5 ? If you agree, can you do a rebase of your branch ? (there's a mean to rebase a PR in GitHub but it seems to give very strange results...)

piRGoif:
Was cherry picked because rebase wasn't working :/ => closing this PR !

PR #72, opened 2019-03-26
[POC] reduce metamodel serialized size

bruno-ds:
# [POC] reduce metamodel serialized size ...
... by storing class name instead of class instance (in the APC cache). 

This could be extended to more properties.

for now the metamodel is a 2.3Mo serialized string, this is huge.
The unserialization of this takes 16% of the homepage rendering on my laptop:
image


Hipska:
I don't see any difference as you initiate them anyway, which now makes it even bigger..

Hipska (changes requested):
Indentation is wrong/different compared to other code in this file.

bruno-ds:
The data stored in APC cache was an array of instances, it is now an array of class names.
The in memory footprint is a little bigger, but the serialized footprint (stored in APC) is way smaller.

For the indent: let's consider it's just a POC I'll work on it it the poc is accepted.

Hipska:
Oh okay, you are talking about data you put in APC. I thought you were talking about memory size of `MetaModel` class. It wasn’t clear, but now it is, and it’s fine for me.

core/metamodel.class.php
```diff
@@ -6335,7 +6331,7 @@ public static function LoadConfig($oConfiguration, $bAllowCache = false)
 				$oKPI = new ExecutionKPI();
 
 				$aCache = array();
-				$aCache['m_aExtensionClasses'] = self::$m_aExtensionClasses;
+				$aCache['m_aExtensionClassNames'] = self::$m_aExtensionClassNames;
```
bruno-ds: This is where the magic happens: storing  `array of string` instead of `array of instances`

bruno-ds:
To move forward, some benchmark would be needed in order to determine if there is a real gain

test/core/MetaModelTest.php
```diff
@@ -83,4 +83,134 @@ public function ApplyParamsProvider()
...
+			];
+		return $aInterfaces;
+	}
+
+	/**
+	 * @dataProvider getPluginsProvider
+	 *
+	 * @param $expectedInstanciationCalls
+	 * @param $expectedResults
+	 * @param $m_aExtensionClassNames
```
Molkobain: We don't have to start properties name with "m_" anymore. It was an old convention in the begining of iTop.

test/core/MetaModelTest.php
```diff
@@ -83,4 +83,134 @@ public function ApplyParamsProvider()
...
+	 * @param $m_aExtensionClasses
+	 * @param $interface
+	 * @param $className
```
Molkobain: But we do start properties name (and variables) with it type (eg. $aArray, $iInteger, $oObject, $sString, ...)

test/core/MetaModelTest.php
Molkobain: We don't start properties name with "m_" anymore, it was an old convention from the begining of iTop.

test/core/MetaModelTest.php
Molkobain: But we do start properties and variables name with their types (eg. $aArray, $iInteger, $sString, $oObject)

rquetiez:
I think the benefit in performance is negligible: the time spent unserializing MetaModel data is 99% related to the classes and their attributes.

Anyhow, I see a great opportunity in this proposal: it is a **first step** toward slicing up MetaModel into smaller pieces, moving forward to the dependancy injection model.

As it is a first step, let's make it a template for next steps! I would therefore suggest a few improvements:

1. Implement new classes into separate files
2. How about namespaces/autoloading?

odain-cbd:
branch merged

PR #73, opened 2019-03-27
Making default visible items in case log attributes configurable

larhip:
Hi guys,
what do you thinking about making the default number of visible items in case log attributes configurable?
Could this one a way to do so?
Best
Lars

### DIFF
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -693,6 +693,15 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		),
+		'case_log_visible_items' => array(
+			'type' => 'integer',
+			'description' => 'Number of items which are visible by default in CaseLogAttributes',
+			// examples... not used (nor 'description')
+			'default' => 2,
+			'value' => 2,
+			'source_of_value' => '',
+			'show_in_conf_sample' => false,
+		),
 		'forgot_password' => array(
 			'type' => 'bool',
 			'description' => 'Enable the "Forgot password" feature',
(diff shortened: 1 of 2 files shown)


piRGoif:
Hello, I've forwarded your suggestion to our product team, I'm waiting for their answer...
In the meantime have a look at [itop-caselogs-toggler](https://github.com/Molkobain/itop-caselogs-toggler) O:)

piRGoif:
Hi Lars,

I'm really sorry but I will close this PR : indeed there are lots of thinking on iTop ergonomics, caselog attributes included. All of this aims for the upcoming 2.8 version, so in the very near future !
At least we would make opened entries configurable per class (so as a datamodel parameter), but maybe per user... Or go way beyond this : plans are not definitely fixed.
We prefer not to accept this PR that would introduce a new config that will change in the next version.

Thanks anyway for this work !
To be sure we won't forget this, I've create [a ticket on Sourceforge](https://sourceforge.net/p/itop/tickets/1738/) with a link to this PR.

Next time, for new features, you may create a ticket first to validate it with us : an idea is not always suitable for all of our customers, and we may have some plans for the future that could be incompatible with a PR... That's why we added this in the contributing guide.

Thanks, and keep the good work with iTop !

larhip:
Hi Pierre,
thanks for the feedback.
Closing this PR is fine for me, it was just an idea with a concrete suggestion how to do to so.
Because it was not really a lot of to implement it, it was just easier for me to share the the idea with directly in a pr than describe in a ticket what I want to do.

So I will follow the ticket on sourceforge and looking forward to the 2.8 release :-)

Best
Lars

piRGoif:
Hello,

Yes I understand, sometimes an idea can be described far more easier with some code than with some text :)

We are suggesting to write ticket instead of code just to avoid the feeling of a waste of time for our contributors when their PR is directly rejected... I'm glad to hear you don't feel that way :)
There are no strict rules to follow of course, every case is specific !

Anyway thanks for your contributions and ideas, they are more than welcome !

PR #74, opened 2019-04-03
🐛 Fix table sort (editing/linking objects)

jbostoen:
In editing mode when linking objects (for example Network Devices), sort doesn't work. 
This is an attempt at fixing it and relies on the value of input, select elements.

Further testing is advised to see if it has undesired consequences.

I did not change the minified version, I noticed the console doesn't use it?

( for some reason, some previous Combodo commits seem to be re-included here...  the only file which matters is js/jquery.tablesorter.js )

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ We would like to give a special thank you to the people from the community who c
  - Gumble, David
  - Hippler, Lars
  - Khamit, Shamil
+ - Kincel, Martin
  - Konečný, Kamil
  - Kunin, Vladimir
  - Lassiter, Dennis


bruno-ds:
Your PR code is hard to read: it cover 54 files.

I think you made a pull before the push and got "polluted" by 3 commits.
maybe you should cherry pick your modifications in a new branch?

piRGoif:
Just a little recommendation also : we do love emoji :heart_eyes:, but be sure to use their Github codes (like `:bug:`) and not directly the character (🐛) as Unicode support is quite poor for now in git clients... 
I've just pushed a CONTRIBUTING file update on this topic.

piRGoif:
Replaced by #75

PR #75, opened 2019-04-03
:bug: Sort tables when modifying links

jbostoen:
Fixes sort issue when modifying links (sort is not working when you edit links between objects, such as NetworkDevice to NetworkDevice)

This pull request takes into account the value which is set for input (text) and select elements, rather than textContent.

I only adjusted the "full" version of the JavaScript library, which seems to be used in the console. No adjustments to the minified version.

### DIFF
--- a/js/jquery.tablesorter.js
+++ b/js/jquery.tablesorter.js
@@ -299,14 +299,29 @@
 
                 if (!node) return "";
 
-                if (!config.supportsTextContent) config.supportsTextContent = node.textContent || false;
+                if (!config.supportsTextContent) config.supportsTextContent = ( node.textContent == node.innerHTML ? node.textContent : false );
 
                 if (config.textExtraction == "simple") {
                     if (config.supportsTextContent) {
                         text = node.textContent;
                     } else {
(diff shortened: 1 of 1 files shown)


piRGoif:
Hello Jeffrey,
We prefer to reject this contribution sorry :/

To explain :

- The TableSorter JQuery plugin was not made by us but has already some minor modifications. We prefer not to add again some modifications in an external script... 
Having an external lib / plugin / etc with custom modifications is not a very suitable situation. We are trying to limit this today, either by proposing our modifications to the author with a PR or at least by creating a custom fork in the combodo-itop-libs GitHub organization to ease updates and have a better readable history. But this was integrated a long ago and none of this was done... 
Having a better handling of our dependencies is a current work in progress.

- We understand the benefits of the modification for you but this wasn't asked by our clients... and this may introduce regressions, that we think are hard to find.
In other words, the risks / benefits ratio is bad in our opinion...

- A big rework will be done both on console UI and relations in iTop 2.8. It is almost certain that the UI to handle n-n relations will be totally different in this very near future !

Thanks anyway !

PR #76, opened 2019-04-10
Add support to optionally mention username in password reset mail

Hipska:
Currently the password reset mail only adds the password reset url (which also includes the username).

If someone wants to modify the `UI:ResetPwd-EmailBody` dictionary, and also include the username into the text, it is not possible without change of iTop code. It is even not possible to change by extension.

So I have added extra parameter `%2$s` which will be replaced by the user login.
(Maybe also add the Person friendly name as option?)

This change does not impact current dictionary or reset mail.


piRGoif:
Hi,
Seems fine !
May this also be added to UI:ResetPwd-EmailSubject ?

piRGoif:
Thanks ! Merging now !

PR #77, opened 2019-04-10
Use operational_status where possible to simplify OQL queries

Hipska:
This improves some dashboard and menu links. (Sometimes rejected status was forgotten and also included in list of open requests)


Molkobain:
Hi @Hipska 

Thanks for the PR, this modification makes sense for the standard datamodel, it should have been done when the meta-enum was introduced in the first place. 

But as always when it comes to standard datamodel modifications, the impacts are greater than they seem to be. Changing those parts could have a significant impact on our clients during a future migration (operational_status groupings being different) so we won't accept it sorry, especially as it can be done through an extension when necessary.

PR #78, opened 2019-04-17
:recycle: N°2222 Make ticket reference generation working with new sub-classes

Molkobain:
When extending the default datamodel, sub-classes of UserRequest or Change for example will not have a corresponding reference (eg. R-12345 or C-12345) but the one from the default class (T-12345). This fixes this by checking is the current object is derivated from a class instead of checking the current class itself.

### DIFF
--- a/datamodels/2.x/itop-change-mgmt-itil/datamodel.itop-change-mgmt-itil.xml
+++ b/datamodels/2.x/itop-change-mgmt-itil/datamodel.itop-change-mgmt-itil.xml
@@ -902,6 +902,17 @@
         </states>
       </lifecycle>
       <methods>
+				<method id="GetTicketRefFormat">
+					<static>true</static>
+					<access>public</access>
+					<type>Overload-DBObject</type>
+					<code><![CDATA[
+        public static function GetTicketRefFormat()
+        {
+                return 'C-%06d';
+        }
+        ]]></code>
+				</method>
         <method id="SetClosureDate">
           <comment><![CDATA[/**
 	 * To be deprecated: use SetCurrentDate() instead
--- a/datamodels/2.x/itop-change-mgmt/datamodel.itop-change-mgmt.xml
+++ b/datamodels/2.x/itop-change-mgmt/datamodel.itop-change-mgmt.xml
@@ -452,6 +452,17 @@
                 </states>
             </lifecycle>
             <methods>
+				<method id="GetTicketRefFormat">
+					<static>true</static>
+					<access>public</access>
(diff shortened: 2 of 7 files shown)


Hipska:
Why not create a MakeTicketRef method in each class? (And extend when needed)

Molkobain:
I discussed this possibility as well with @steffunky as it would be a lot nicer to have polymorphism, we agreed on it but it could introduce from migration issues with people who have customized their datamodel.

For example, many of our clients have overload the MakeTicketRef in child classes, so doing so would will introduce a conflit on migration (if there is no migration notes).

Another way would be to have a _GetFormat()_ function in each classes which would return the format only, used by _MakeTicketRef()_ natively. That way, you have the benefits of polymorphism (by extending it in child classes) without breaking anything on existing environments.

Molkobain:
_Note: I forgot to mention that this PR is just there to start a discussion with the team during the next technical point. While answering to [this question](https://sourceforge.net/p/itop/discussion/1113906/thread/1035ad746b/) on SourceForge I stumbled onto the fact that classes where hardcoded there._

Hipska:
> I discussed this possibility as well with @steffunky as it would be a lot nicer to have polymorphism, we agreed on it but it could introduce from migration issues with people who have customized their datamodel.
>
> For example, many of our clients have overload the MakeTicketRef in child classes, so doing so would will introduce a conflit on migration (if there is no migration notes).

Correct, didn't think about that.

> Another way would be to have a _GetFormat()_ function in each classes which would return the format only, used by _MakeTicketRef()_ natively. That way, you have the benefits of polymorphism (by extending it in child classes) without breaking anything on existing environments.

Great solution! Maybe make it `GetRefFormat()` (and static) then?

Molkobain:
Absolutely! It will be discussed futhermore on friday :)

Hipska:
Note that #82 might be relevant to this as well..

Molkobain:
Refactoring done. Tested in full ITIL mode.
GetTicketFormatRef() function is public static instead of protected so we can show the format anywhere needed.

Hipska (approved):

Hipska:
Only thing I'm wondering is if in _itop-request-mgmt_ it should be **I-000000** when `request_type` is incident?

Molkobain:
I was wondering about it too, but as it would change the behaviour of existing installation, we decided to keep it that way. UserRequest with `request_type`as incident are marked `R-XXXXXX` in current iTop versions.

bruno-ds (approved):

PR #80, opened 2019-04-18
Fix cron crash when MySQL connection lost

Hipska:
Currently, cron.php crashes with "PHP Fatal error:  Uncaught Exception" when for example MySQL stops working during `CronExec`.

The first thrown exception is correctly catched, but after all code, the `iTopMutex` class does execute it's `__destruct` method which is doing a `RELEASE_LOCK` on the DB, which then results in an exception being throwed.

When you unset the mutex object in a try catch, this is cleaner.

### DIFF
--- a/webservices/cron.php
+++ b/webservices/cron.php
@@ -591,6 +591,22 @@ catch (Exception $e)
 		$oP->p($e->getTraceAsString());
 	}
 }
+finally
+{
+	try
+	{
+		unset($oMutex);
+	}
+	catch (Exception $e)
+	{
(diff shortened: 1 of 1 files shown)


piRGoif:
Thanks for the idea Thomas !

I would better move the `$oMutex->Unlock();` in a finally block, included itself in a try/catch block : this would make the code easier to read and understand, and also more logic (unlock should be called whatever happens !)

What do you think of this alternative ?

Hipska:
You are absolutely right, will update asap.

PR #81, opened 2019-04-24
🐛 N°2182 Fix default usage of iTopMutex when TLS is enabled

Hipska:
When MySQL is configured to only allow SSL connections, the _itop-backup status.php_ page throws an error. I could track it down to the _iTopMutex_ constructor defaulting `$bDBTlsEnabled` to `false`.

There might be other places where the same problem occurs as it also seems to be a problem on the setup page.

Impacted: iTop 2.5 & 2.6

### DIFF
--- a/core/mutex.class.inc.php
+++ b/core/mutex.class.inc.php
@@ -45,7 +45,7 @@ class iTopMutex
 	static protected $aAcquiredLocks = array(); // Number of instances of the Mutex, having the lock, in this page
 
 	public function __construct(
-		$sName, $sDBHost = null, $sDBUser = null, $sDBPwd = null, $bDBTlsEnabled = false, $sDBTlsCA = null
+		$sName, $sDBHost = null, $sDBUser = null, $sDBPwd = null, $bDBTlsEnabled = null, $sDBTlsCA = null
 	)
 	{
 		// Compute the name of a lock for mysql


piRGoif:
Wow thanks, this is quite a bug !
The constructor parameters were added in 2.5.0 with f07bbfa1, and indeed having false for $bDBTlsEnabled was a bad idea ! If not provided, we should always look for the value in the config.

Changing the base branch of the PR gave a very unexpected result (more than 700 commits in the PR).
Can you rebase yourself your branch on the support/2.5 branch ?

Hipska:
Can't you merge the PR with `develop` and then cherry pick to also have it on 2.5 and 2.6 branches?

piRGoif:
We can but... we try to limit as much as possible cherry-picks as having two commits with the same thing is confusing : better have just one commit and immediately see in which branches / tags it is present than searching for a modification, find one commit, assume it was integrated there and no, you missed another one :/.
We would really appreciate if you can rebase on support/2.5, and then we'll do the branches merge immediately after.
Another solution would be that we commit ourselves to support/2.5 with a reference to this PR and you as author, and close this PR ?

Hipska:
I think its correct right now..

piRGoif:
Many thanks ! Merged !

Hipska:
> we try to limit as much as possible cherry-picks as having two commits with the same thing is confusing

Well, now you also have three commits: 2858d13fd, 77f757995 and 15362df69. But okay, I understand what you are trying to say.

piRGoif:
> Well, now you also have three commits: [2858d13](https://github.com/Combodo/iTop/commit/2858d13fd5fcf486c157e77fbe908f29d39a81a5), [77f7579](https://github.com/Combodo/iTop/commit/77f757995e77c5d70a4c891ce970e22a8eceae99) and [15362df](https://github.com/Combodo/iTop/commit/15362df69ac584f31626de73db0b5ac534a6956c)

Not really : when opening the file on any of the corresponding branch (support/2.5, master, develop) and doing a blame (for example https://github.com/Combodo/iTop/blame/develop/core/mutex.class.inc.php) you will always get 2858d13.

When opening this commit in a good git client, you have every tag and branch really having this commit.

If you have another commit cherry picked, you'll have to search for it to find it...

Hipska:
Correct again 😉

PR #82, opened 2019-04-29
Only set Ticket ref if not yet present via import or synchro

Hipska:
As requested by [#1475 Only generate ticket ref when empty](https://sourceforge.net/p/itop/tickets/1475/).

### DIFF
--- a/datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
+++ b/datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
@@ -209,7 +209,7 @@
           $oMutex->Lock();
           $iNextId = MetaModel::GetNextKey(get_class($this));
           $sRef = $this->MakeTicketRef($iNextId);
-          $this->Set('ref', $sRef);
+          $this->SetIfNull('ref', $sRef);
           $iKey = parent::DBInsertNoReload();
           $oMutex->Unlock();
           return $iKey;
--- a/datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
+++ b/datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
@@ -53,6 +53,14 @@
 				   </attributes>
 				 </index>
 				</indexes>
+				<uniqueness_rules>
+					<rule if="ref">
+						<attributes>
+							<attribute id="ref"/>
+						</attributes>
+						<is_blocking>false</is_blocking>
+					</rule>
+				</uniqueness_rules>
 			</properties>
 			<fields>
 				<field id="operational_status" xsi:type="AttributeMetaEnum">


Hipska:
Hi, this PR has been requested by @v-dumas, but no response from Combodo yet..

piRGoif:
Hello Thomas, 
We are currently working hard on 2 new versions, we will certainly do a review of the PR on friday, so hopefully there should be some news next week :)

piRGoif:
Hello !
Indeed your modification makes perfect sense.
But it introduces a major change as before ref would always depends on object id. This may leads to some uniqueness problem... Could you add a uniqueness constraint on the ref ?

Hipska:
Hi @piRGoif, you are free to do it. Maintainers can do edits.

piRGoif:
>> Could you add a uniqueness constraint on the ref ?
>
> Hi @piRGoif, you are free to do it. 

I ask you to add it :) : this modification as it is now can lead to integrity issues, so it can't be merged without modification.

> Maintainers can do edits.

Didn't understand what you meant ?

Hipska:
I meant this Github feature:

- [x] Allow edits from maintainers.
> If checked, **users with write access to Combodo/iTop can add new commits** to your `feature/ticket_ref_overwrite` branch. You can always change this setting later.

piRGoif:
Hello Thomas,

Oh ok I didn't knew this option, it's cool !-) There is a whole documentation here : https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

The use case that could cause problems is importing tickets in a brand new iTop, for example refs 10 to 15, then creating new tickets (beginning from ref 1) : you'll get problems when creating the 10th ticket. The solution would be to change the base for the auto increment id field, an operation that needs to be performed by the db admin, so not really easy to do...
Having a blocking uniqueness rule would be painful for users, so I'll add a non blocking rule.

piRGoif:
> The use case that could cause problems is importing tickets in a brand new iTop, for example refs 10 to 15, then creating new tickets (beginning from ref 1) : you'll get problems when creating the 10th ticket.

As @bruno-ds pointed out in a private message, this is wrong !

See https://dev.mysql.com/doc/refman/5.7/en/example-auto-increment.html :

> When you insert any other value into an AUTO_INCREMENT column, the column is set to that value and the sequence is reset so that the next automatically generated value follows sequentially from the largest column value.

But hey, we will keep anyway the uniqueness rule non blocking, to be extra careful :)

Hipska:
That would indeed be no problem if that user also imported ticket id. But I guess they would not do that most of the time. So a uniqueness rule would still be useful.

Molkobain:
Yeah, +1 with @Hipska . We said that because most of the time IDs are not exported, so we can't be sure they will be imported with it.

piRGoif:
Yes, you're perfectly right, I forgot to mention this would be the case if id were imported also. But you did understand anyway :)

piRGoif:
I've just pushed the uniqueness rule on your fork, then squash&merge !
Thanks

Hipska:
I was just thinking, even if you are also importing ticket ID, even then the ref would be calculated on the next available ID, the method does not take into account a current ID. Which is also the reason this breaks on MySQL 8.

piRGoif:
Hello,
Yes you're right : in \CMDBSource::GetNextInsertId we do SHOW TABLE STATUS and get the auto_increment value, so the max(id) + 1.

piRGoif:
Oh, by the way, we aim to fix MySQL8 incompatibilities in 2.7.0 O:)

datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
```diff
@@ -53,6 +53,14 @@
 				   </attributes>
 				 </index>
 				</indexes>
+				<uniqueness_rules>
+					<rule if="ref">
```
Hipska: Still needs to be changed to `id`

Hipska: Nevermind, just saw this is already merged and fixed..

piRGoif: Thanks for your comment Thomas - indeed as you saw Jeffrey already sent the same feedback. Sorry for this silly mistake :/

PR #83, opened 2019-05-07
Add html5 time tag when displaying attributes with date information

Hipska:
This introduces useful machine readable date/time information, to be used by browser or other extensions.

[W3](https://www.w3.org/TR/2011/WD-html5-author-20110705/the-time-element.html)
[Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time)

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -5534,7 +5534,7 @@ class AttributeDateTime extends AttributeDBField
 
 	public function GetAsHTML($value, $oHostObject = null, $bLocalize = true)
 	{
-		return Str::pure2html(static::GetFormat()->format($value));
+		return sprintf('<time datetime="%s">%s</time>', $value, Str::pure2html(static::GetFormat()->format($value)));
 	}
 
 	public function GetAsXML($value, $oHostObject = null, $bLocalize = true)
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -7968,7 +7968,7 @@ class AttributeStopWatch extends AttributeDefinition
 		switch ($sItemCode)
 		{
 			case 'timespent':
-				$sHtml = Str::pure2html(AttributeDuration::FormatDuration($value));
+				$sHtml = sprintf('<time datetime="PT%dS">%s</time>', $value, Str::pure2html(AttributeDuration::FormatDuration($value)));
 				break;
 			case 'started':
 			case 'laststart':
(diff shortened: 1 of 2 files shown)


Molkobain:
Hello @Hipska , could you give some use cases for this?

Hipska:
For example when used in notifications:

- Mail client can suggest to add calendar item for that date. (MacOS Mail can do this already, but needs to guess what part of the text is a data, now it can just get from the html tag)
- With https://github.com/itomig-de/itomig-webhook-integration/issues/3 It could be parsed/formatted so that Slack also knows that it is a date.

Many other use cases could exist, but in my opinion no real or actual use case is needed as it is backwards compatible and all html viewers/parsers not knowing the `time` tag will just ignore it. So it does not break stuff.

Molkobain:
Ok, thank you for the use cases, they are great examples of what could indeed be done through this! We'll talk about this soon.

Hipska:
Additional from the referenced [W3 spec](https://www.w3.org/TR/2011/WD-html5-author-20110705/the-time-element.html):
> This element is intended as a way to encode modern dates and times in a machine-readable way so that, for example, user agents can offer to add birthday reminders or scheduled events to the user's calendar.

From the referenced [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time):
> The HTML **`<time>` element** represents a specific period in time. It may include the `datetime` attribute to translate dates into machine-readable format, allowing for better search engine results or custom features such as reminders.

> This element is for presenting dates and times in a machine readable format. For example, this can help a user agent offer to add an event to a user's calendar.

Hipska:
Hi @Molkobain, is there anything else that blocks this?

Molkobain:
Hello @Hipska 

We discussed about this today and decided not to go with it for now for several reasons:
- GetAsHTML is not always used properly throughout the application/extensions and could display the markup as raw text.
- Datetime formats introduced in iTop 2.3 has brought a lot a bugs and we are just seeing the end of it, so we have cold feet going through that again.
- Seing that you commited changes 6 weeks after the initial post shows that this is typically a sensitive change and we might forget stuff.

So the added-value does not seem that important regarding the possible regressions it could introduced. But we are keeping this for the future global rework.

But most of all, thanks for helping modernizing iTop Thomas! :)

Hipska:
I added these recent commits because the PR was still open (otherwise I would have created another one.) and I just stumbled across some extra undocumented (!) usages of DateTime field types.

Molkobain:
> ... I just stumbled across some extra undocumented (!) usages of DateTime field types.

Unfortunately that's exactly my point.

Hipska:
Why closing the request if you say you want to keep it for future global rework?

Molkobain:
We keep the idea itself, but the code will probably be different as this won't be until iTop 2.8. also, GetAsHTML and other APIs are most likely to evolve too. So the PR won't be integrated as is, the merge won't be possible. I took the description you made about the markup and put it in our bug tracker.

PR #84, opened 2019-05-31
add ctype as required php extension

chifu1234:
Hi Guys

ctype is not in every php build integrated (for example like alpine).

cheers, Kevä

### DIFF
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -105,6 +105,7 @@ class SetupUtils
 			'zlib',
 			'zip',
 			'gd', // test image type (always returns false if not installed), image resizing, PDF export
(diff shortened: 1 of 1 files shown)


piRGoif:
Hello,
I don't think this is necessary anymore as since we change Silex to Symfony we now have the polyfill-ctype package included.
Do you confirm @chifu1234 ?

piRGoif:
The ctype polyfill is present in the Composer autoloader (see lib/composer/autoload_static.php), and reading https://github.com/symfony/polyfill we understand that nothing more should be done.
It seems quite hard to test as ctype seems to be embedded in lots of PHP dustributions...
So I'm closing this one, @chifu1234 if you think we're wrong don't hesitate to come here explain why !

PR #87, opened 2019-06-18
Add controls for treeview

scheibling:
Add controls for expanding or collapsing treeview.
Set standard to collapsed tree for user to decide state closed or open.
Add css to properly style controls.

### DIFF
--- a/application/ui.extkeywidget.class.inc.php
+++ b/application/ui.extkeywidget.class.inc.php
@@ -615,13 +615,13 @@ EOF
 		$oPage->add('</td></tr></table>');
 		$oPage->add('</div>');
 		
-		$oPage->add("<div class=\"treecontrol\" id=\"treecontrolid\"><a href=\"?#\">Collapse All</a> | <a href=\"?#\">Expand All</a></div>");
-		
+		$oPage->add('<div class="treecontrol" id="treecontrolid"><a href="?#">'.Dict::S("UI:Button:CollapseAll").'</a> | <a href="?#">'.Dict::S("UI:Button:ExpandAll").'</a></div>');
+
 		$oPage->add("<input type=\"button\" id=\"btn_cancel_{$this->iId}\" value=\"".Dict::S('UI:Button:Cancel')."\" onClick=\"$('#dlg_tree_{$this->iId}').dialog('close');\">&nbsp;&nbsp;");
 		$oPage->add("<input type=\"button\" id=\"btn_ok_{$this->iId}\" value=\"".Dict::S('UI:Button:Ok')."\"  onClick=\"oACWidget_{$this->iId}.DoHKOk();\">");
-
+			
 		$oPage->add('</div></div>');
-		$oPage->add_ready_script("\$('#tree_$this->iId ul').treeview(control: #treecontrolid);\n");
+		$oPage->add_ready_script("\$('#tree_$this->iId ul').treeview({ control: '#treecontrolid',	persist: 'false'});\n");
 		$oPage->add_ready_script("\$('#dlg_tree_$this->iId').dialog({ width: 'auto', height: 'auto', autoOpen: true, modal: true, title: '$sDialogTitle', resizeStop: oACWidget_{$this->iId}.OnHKResize, close: oACWidget_{$this->iId}.OnHKClose });\n");
 	}
 
--- a/application/ui.extkeywidget.class.inc.php
+++ b/application/ui.extkeywidget.class.inc.php
@@ -614,11 +614,14 @@ EOF
 
 		$oPage->add('</td></tr></table>');
 		$oPage->add('</div>');
+		
+		$oPage->add("<div class=\"treecontrol\" id=\"treecontrolid\"><a href=\"?#\">Collapse All</a> | <a href=\"?#\">Expand All</a></div>");
+		
 		$oPage->add("<input type=\"button\" id=\"btn_cancel_{$this->iId}\" value=\"".Dict::S('UI:Button:Cancel')."\" onClick=\"$('#dlg_tree_{$this->iId}').dialog('close');\">&nbsp;&nbsp;");
 		$oPage->add("<input type=\"button\" id=\"btn_ok_{$this->iId}\" value=\"".Dict::S('UI:Button:Ok')."\"  onClick=\"oACWidget_{$this->iId}.DoHKOk();\">");
 
 		$oPage->add('</div></div>');
-		$oPage->add_ready_script("\$('#tree_$this->iId ul').treeview();\n");
+		$oPage->add_ready_script("\$('#tree_$this->iId ul').treeview(control: #treecontrolid);\n");
 		$oPage->add_ready_script("\$('#dlg_tree_$this->iId').dialog({ width: 'auto', height: 'auto', autoOpen: true, modal: true, title: '$sDialogTitle', resizeStop: oACWidget_{$this->iId}.OnHKResize, close: oACWidget_{$this->iId}.OnHKClose });\n");
 	}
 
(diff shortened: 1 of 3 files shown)


piRGoif:
For information, this is a suggested fix for https://sourceforge.net/p/itop/tickets/1761/
Thanks Lars, we will review this !

application/ui.extkeywidget.class.inc.php
```diff
@@ -614,11 +614,14 @@ public function DisplayHierarchy(WebPage $oPage, $sFilter, $currValue, $oObj)
 
 		$oPage->add('</td></tr></table>');
 		$oPage->add('</div>');
+		
+		$oPage->add("<div class=\"treecontrol\" id=\"treecontrolid\"><a href=\"?#\">Collapse All</a> | <a href=\"?#\">Expand All</a></div>");
```
jbostoen: Needs to use dictionary entries I think (Dict::S), rather than hardcoded strings?

Also, if you use Combodo's convention of writing PHP, they usually use single quotes - which makes it unnecessary to escape the double quotes.

application/ui.extkeywidget.class.inc.php
```diff
@@ -614,11 +614,14 @@ public function DisplayHierarchy(WebPage $oPage, $sFilter, $currValue, $oObj)
...
+		
 		$oPage->add("<input type=\"button\" id=\"btn_cancel_{$this->iId}\" value=\"".Dict::S('UI:Button:Cancel')."\" onClick=\"$('#dlg_tree_{$this->iId}').dialog('close');\">&nbsp;&nbsp;");
 		$oPage->add("<input type=\"button\" id=\"btn_ok_{$this->iId}\" value=\"".Dict::S('UI:Button:Ok')."\"  onClick=\"oACWidget_{$this->iId}.DoHKOk();\">");
 
 		$oPage->add('</div></div>');
-		$oPage->add_ready_script("\$('#tree_$this->iId ul').treeview();\n");
+		$oPage->add_ready_script("\$('#tree_$this->iId ul').treeview(control: #treecontrolid);\n");
```
jbostoen: I've done some work in JavaScript, but is this correct? 

I'd be expecting something like .treeview({ control: '#treecontrolid' }) or something? (not entirely sure about which lib the treeview comes from, but it seems like you're trying to define a property)

jbostoen (changes requested):

application/ui.extkeywidget.class.inc.php
piRGoif: You're perfectly right, thanks Jeffrey !
You can refer to the translation guide : https://www.itophub.io/wiki/page?id=2_6_0%3Acustomization%3Atranslation

scheibling: Right, I'll add the dictionary entries! Regarding the quotes, I basically copied the convention used in the rest of the file. I can change that as well.

application/ui.extkeywidget.class.inc.php
scheibling: Not quite an expert on JS either, the limited documentation I could still find on that instance of treeview.js suggested this style. Since it's been deprecated in favor of jqTree, it's been removed from the jquery.com-website. I'll try your suggestion as well, I agree, if that works it seems like the right way to go.

Hipska (approved):

application/ui.extkeywidget.class.inc.php
piRGoif: Our coding standards are documented here : https://www.itophub.io/wiki/page?id=latest%3Acustomization%3Acoding_standards
Indeed it does not cover this particular point, we'll discuss with the team if it should be added.

application/ui.extkeywidget.class.inc.php
jbostoen: It does look like an old library; perhaps this is something which will change in 2.8.

application/ui.extkeywidget.class.inc.php
```diff
@@ -615,13 +615,13 @@ public function DisplayHierarchy(WebPage $oPage, $sFilter, $currValue, $oObj)
...
-		
+		$oPage->add('<div class="treecontrol" id="treecontrolid"><a href="?#">'.Dict::S("UI:Button:CollapseAll").'</a> | <a href="?#">'.Dict::S("UI:Button:ExpandAll").'</a></div>');
+
 		$oPage->add("<input type=\"button\" id=\"btn_cancel_{$this->iId}\" value=\"".Dict::S('UI:Button:Cancel')."\" onClick=\"$('#dlg_tree_{$this->iId}').dialog('close');\">&nbsp;&nbsp;");
 		$oPage->add("<input type=\"button\" id=\"btn_ok_{$this->iId}\" value=\"".Dict::S('UI:Button:Ok')."\"  onClick=\"oACWidget_{$this->iId}.DoHKOk();\">");
-
+			
 		$oPage->add('</div></div>');
-		$oPage->add_ready_script("\$('#tree_$this->iId ul').treeview(control: #treecontrolid);\n");
+		$oPage->add_ready_script("\$('#tree_$this->iId ul').treeview({ control: '#treecontrolid',	persist: 'false'});\n");
```
jbostoen: Couldn't find the proper documentation; but it seems persist expects string values of either 'location' or 'cookie'? (I looked it up because for boolean values I'd simply expect persist: false)

( I actually took a look at https://github.com/Combodo/iTop/blob/96334d859aaa695aa128fe562256dad1cd6603cf/js/jquery.treeview.js#L191 ).

application/ui.extkeywidget.class.inc.php
piRGoif: > It does look like an old library; perhaps this is something which will change in 2.8.

We're not sure yet : we began a big work on all of our dependencies (inventory, create forks when possible, ...), this is still in progress. Sure the JQuery plugins will be updated, but we don't have a clear roadmap for now, especially for this one.

The first step was to upgrade JQuery itself : we moved to 3.3.1 for 2.6.0 and change deprecated code as much as possible. We will move to 3.4 and change some plugins for 2.7.0 (we're aiming for the autocomplete for example but not sure we will have enough resources).

Step by step ;)

Hipska:
Do you also have an example image?

application/ui.extkeywidget.class.inc.php
scheibling: That's absolutely correct.

scheibling (commented):
Correction to the treeview class

scheibling:
Yes, sorry for the delay. 

image

piRGoif:
This was submitted in Combodo ticket DB as N°2764, candidate for iTop 2.8.

piRGoif:
@scheibling sorry it took us so much time to accept this PR... We were (and still are in a way) quite busy with the 2.7.0... We are aiming to integrate this in 2.8.0.
As your branch as conflicts, can you rebase on Combodo/iTop develop branch ? You may also allow us to write on your branch (see https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

scheibling:
$piRGoif No worries, I'll look at it during the week. Thanks for the heads up!
I've also opened it to you in case you see something I've missed.

piRGoif:
Hello,

As we are now ready to integrate PR for 2.8.0, I tried to test your changes. As there are still conflicts, I reported your changes myself. While testing it came to me that displaying those buttons if no child leaf is present is king of weird, so I added a test. 
See the 3 commits in the  feature/pr87_treeview_expand-collapse branch.

I'm ready to integrate this from this branch, don't hesitate to give your feedback in the coming days !

piRGoif:
I'm closing this as no answer for the original submitter, and because we already discussed on #132 with the rest of the team.

PR #89, opened 2019-08-01
Abstract implementation for iScheduledProcess

piRGoif:
In iTop, BackupExec is a iScheduledProcess implementation that contains some code (getting config parameters and next occurrence datetime) that could be made generic and reusable by extensions.

This PR moves this code to a new AbstractScheduledProcess class that can be used by other modules.

### DIFF
--- a/core/backgroundprocess.inc.php
+++ b/core/backgroundprocess.inc.php
@@ -56,7 +56,7 @@ interface iBackgroundProcess extends iProcess
  * interface iScheduledProcess
  * A variant of process that must be called at specific times
  *
- * @see \AbstractScheduledProcess for a bootstrap implementation
+ * @see \AbstractWeeklyScheduledProcess for a bootstrap implementation
  * @license     http://opensource.org/licenses/AGPL-3.0
  * @copyright   Copyright (C) 2013 Combodo SARL
  */
@@ -84,7 +84,7 @@ interface iScheduledProcess extends iProcess
  *
  * @since 2.7.0
  */
-abstract class AbstractScheduledProcess implements iScheduledProcess
+abstract class AbstractWeeklyScheduledProcess implements iScheduledProcess
 {
 	// param have default names/values but can be overriden
 	const MODULE_SETTING_ENABLED = 'enabled';
--- a/core/backgroundprocess.inc.php
+++ b/core/backgroundprocess.inc.php
@@ -56,8 +56,9 @@ interface iBackgroundProcess extends iProcess
  * interface iScheduledProcess
  * A variant of process that must be called at specific times
  *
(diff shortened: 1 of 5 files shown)


datamodels/2.x/itop-backup/main.itop-backup.php
```diff
@@ -85,16 +85,26 @@ public function ListFiles($sBackupDir)
 	}
 }
 
-class BackupExec implements iScheduledProcess
+class BackupExecProcess extends AbstractScheduledProcess
```
bruno-ds: Why did you rename it?
Could it cause BC issues?

core/backgroundprocess.inc.php
```diff
@@ -67,14 +68,180 @@ interface iScheduledProcess extends iProcess
 	public function GetNextOccurrence();
 }
 
+
 /**
- * Class ProcessException
- * Exception for iProcess implementations.<br>
+ * Implementation of {@link iScheduledProcess}, using config parameters for module
```
bruno-ds: we loose the git history, couldn't it have been possible to preserve it?

datamodels/2.x/itop-backup/main.itop-backup.php
piRGoif: There will at least be an issue with existing records in the BackgroundTask table... (no bugs but existing records kept)
But as my goal was to discuss with the rest of the team, I made this proposal also :)

core/backgroundprocess.inc.php
piRGoif: Discussed this with Bruno : he was worried about the blame that we are loosing on the 2 methods that were moved from BackupExec to the new AbstractScheduledProcess class.
But indeed those 2 methods are quite old, both are dated from 2 months in 2014, and only 1 line is more recent, but has a comment identical to the commit message.
Bruno and I agree that loosing this history won't be a problem.

datamodels/2.x/itop-backup/main.itop-backup.php
piRGoif: The dev team did a review today.
Decisions :
* rollback the BackupExec class renaming (some risk introduced with a very slight benefit)
* important to have this in 2.7.0 as it will allow to change corresponding extensions sooner
* apart of the renaming rollback ok to merge

I'm closing this conversation, many thanks again Bruno for the feedback !

Molkobain (dismissed):

Molkobain (approved):

bruno-ds (approved):
I had troubble understanding the mean of $aWEEKDAYTON 
aka : WEEKDAY TO N 

but seems all good to me :)

PR #90, opened 2019-08-16
Handle nested transactions

piRGoif:
Ticket N°679 was aiming at adding more reliability on object insertion by adding a transaction around INSERT statements for an object (there is one table per class in the object hierarchy).

But we already have transactions in iTop, especially in datasynchro : when simulating, a transaction is started and rollbacked at the end.

MySQL doesn't support nested transactions, and indeed every START TRANSACTION will automatically COMMIT the previous started transaction ! (see [official doc](https://dev.mysql.com/doc/refman/5.6/en/commit.html))

I started implementing a solution using savepoints, that would rollback only the nested transaction if rollback is called.
But after a chat with Denis and Bruno, we decided to have a simpler solution : just one transaction, and :

* if calling COMMIT then sends to db if we're at the root level, otherwise do nothing
* if calling ROLLBACK : sends to db if we're at the root level, else throw a new MySQLChildTransactionRollbackException

### DIFF
--- a/core/cmdbsource.class.inc.php
+++ b/core/cmdbsource.class.inc.php
@@ -99,6 +99,15 @@ class MySQLHasGoneAwayException extends MySQLException
 	}
 }
 
+/**
+ * @see \CMDBSource::Rollback()
+ * @since 2.7.0 N°679
+ */
+class MySQLChildTransactionRollbackException extends MySQLException
+{
+
+}
+
 
 /**
  * CMDBSource
@@ -131,7 +140,7 @@ class CMDBSource
 	protected static $m_oMysqli;
 
 	/**
-	 * @var string[] list of nested transactions saved as savepoints
+	 * @var string[] unique identifier for each nested transaction level (array[0] => 1st level, array[1] => sub transaction, etc)
 	 * @since 2.7.0 N°679
 	 */
 	protected static $m_aTransacSavepoints = array();
@@ -579,6 +588,7 @@ class CMDBSource
 	 * @return \mysqli_result|null
 	 * @throws \MySQLException
 	 * @throws \MySQLHasGoneAwayException
+	 * @throws \CoreException
 	 *
 	 * @since 2.7.0 N°679 handles nested transactions using {@link m_aTransacSavepoints}
 	 */
@@ -608,7 +618,9 @@ class CMDBSource
 	}
 
 	/**
-	 * Do nothing before sending query to the DB
+	 * Send the query directly to the DB. **Be extra cautious with this !**
+	 *
+	 * Use {@link Query} if you're not sure.
 	 *
 	 * @param string $sSql
 	 *
@@ -616,7 +628,7 @@ class CMDBSource
 	 * @throws \MySQLException
 	 * @throws \MySQLHasGoneAwayException
 	 */
-	private static function DBQuery($sSql)
+	public static function DBQuery($sSql)
 	{
 		$oKPI = new ExecutionKPI();
 		try
@@ -646,7 +658,11 @@ class CMDBSource
 	}
 
 	/**
-	 * If nested transaction, then only create a savepoint that will be used by {@link Rollback}
+	 * If nested transaction, we are not starting a new one : only one global transaction will exist.
+	 *
+	 * Indeed [the official documentation](https://dev.mysql.com/doc/refman/5.6/en/commit.html) states :
+	 *
+	 * > Beginning a transaction causes any pending transaction to be committed
 	 *
 	 * @see m_aTransacSavepoints
 	 * @since 2.7.0 N°679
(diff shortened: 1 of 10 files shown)


core/cmdbsource.class.inc.php
```diff
@@ -99,6 +99,15 @@ public function __construct($sIssue, $aContext)
 	}
 }
 
+/**
+ * @see \CMDBSource::Rollback()
+ * @since 2.7.0 N°679
+ */
+class MySQLChildTransactionRollbackException extends MySQLException
```
bruno-ds: You know I'm paranoid:
Shouldn't we extend \Exception in order to prevent some possible catch of \MySQLException?
(At least a check on the codebase would be reassuring)

core/cmdbsource.class.inc.php
```diff
@@ -130,6 +139,12 @@ class CMDBSource
 	/** @var mysqli $m_oMysqli */
 	protected static $m_oMysqli;
 
+	/**
+	 * @var string[] unique identifier for each nested transaction level (array[0] => 1st level, array[1] => sub transaction, etc)
+	 * @since 2.7.0 N°679
+	 */
+	protected static $m_aTransacSavepoints = array();
```
bruno-ds: To be removed?

core/cmdbsource.class.inc.php
```diff
@@ -570,25 +585,64 @@ public static function Quote($value, $bAlways = false, $cQuoteStyle = "'")
...
 	 * @throws \MySQLHasGoneAwayException
+	 * @throws \CoreException
+	 *
+	 * @since 2.7.0 N°679 handles nested transactions using {@link m_aTransacSavepoints}
 	 */
 	public static function Query($sSQLQuery)
+	{
+		if (preg_match('/^START TRANSACTION;?$/i', $sSQLQuery))
+		{
+			self::StartTransaction();
```
bruno-ds: Maybe we could @trigger_error E_USER_DEPRECATED (I'm not sure about the constant real name)

core/cmdbsource.class.inc.php
```diff
@@ -570,25 +585,64 @@ public static function Quote($value, $bAlways = false, $cQuoteStyle = "'")
...
+	 *
+	 * Use {@link Query} if you're not sure.
+	 *
+	 * @param string $sSql
+	 *
+	 * @return bool|\mysqli_result
+	 * @throws \MySQLException
+	 * @throws \MySQLHasGoneAwayException
+	 */
+	public static function DBQuery($sSql)
```
bruno-ds: Maybe it could be protected/private?

core/cmdbsource.class.inc.php
```diff
@@ -599,10 +653,140 @@ public static function Query($sSQLQuery)
...
+	/**
+	 * @see m_aTransacSavepoints
+	 *
+	 * @param bool $bHasExistingTransactions
+	 *
+	 * @return string name of the new savepoint
+	 *
+	 * @since 2.7.0 N°679
+	 */
+	private static function AddTransactionLevel($bHasExistingTransactions)
```
bruno-ds: If the savepoints are removed, then this method should be simplified

bruno-ds: In fact using a counter rather than an array seems simpler

bruno-ds (commented):
Maybe the three *TransactionLevel()
Should be renamed if we move from the array to a more simple counter.

bruno-ds:
The "force" option for the rollback make me wonder if it is just smart or yagni

core/cmdbsource.class.inc.php
piRGoif: Yeah, I asked myself the same question... But I didn't find in corresponding code some catch of MySQLException so I go on implementing a child MySQLException class...

core/cmdbsource.class.inc.php
piRGoif: It was private in the first implementation but I changed it to public after our discussion with Denis...

piRGoif:
> The "force" option for the rollback make me wonder if it is just smart or yagni

I added the force parameter as before the DBQuery method was private. Now it is public, so you can still query directly the DB and send a Rollback. But I prefer to have this clear way of doing a rollback anyway, I think it is simpler for the developer ?

core/cmdbsource.class.inc.php
piRGoif: I didn't removed all of the savepoint implementation, I thought it might be useful but you're right, should do one thing at a time. Will change this to a simple counter as we discussed.

core/cmdbsource.class.inc.php
```diff
@@ -599,10 +653,140 @@ public static function Query($sSQLQuery)
...
+	 *
+	 * @see m_aTransacSavepoints
+	 *
+	 * @param bool $bForce if true then do a rollback even if we're not at the transaction root level
+	 *
+	 * @throws \MySQLException
+	 * @throws \MySQLHasGoneAwayException
+	 * @since 2.7.0 N°679
+	 */
+	public static function Rollback($bForce = false)
```
piRGoif: Note for monday : should add an Exception parameter, as Rollback is called in a catch block. Passing the original exception would allow to keep it in the throw stack.

core/cmdbsource.class.inc.php
piRGoif: I'm closing this as the same subject is listed below :)

core/cmdbsource.class.inc.php
piRGoif: It's now a counter with 71ed546f

core/cmdbsource.class.inc.php
piRGoif: Done with 18387e00

piRGoif:
> Maybe the three *TransactionLevel()
> Should be renamed if we move from the array to a more simple counter.

I like to have some indirection so that each consumer doesn't have to worry about the implementation. Indeed, the "transaction level" mechanism should be an object with an interface... But this might be overkill :o)

core/cmdbsource.class.inc.php
piRGoif: See comment below, after a meeting with Denis & Eric

core/cmdbsource.class.inc.php
piRGoif: See comment below, after a meeting with Denis & Eric

piRGoif:
Here are the conclusion after a chat with @dflaven and @eespie :

* decided to change visibility of \CMDBSource::DBQuery to private (was here to provide a workaround if developer would be blocked, but we are confident in our code)
* all of the new transaction dedicated methods will become private too (don't need to change the consumer, don't need to call them directly)
* we fear that adding transactions would bring some problems, maybe on very specific infrastructure (cluster for example) : we decided to add a new config flag that would allow to disable the new transactions in \DBObject::DBInsertNoReload in case such problem would become blocker

piRGoif:
Modifications have been pushed a few minutes ago. Time to merge !

PR #93, opened 2019-10-13
restservice return pagination but not need where

chileeb:
When visit rest service of  rest.php,need to return all records or pagination data.
I want to find all data,but the key must be required field。
So that,i can set key is * to finish this problem.

### DIFF
--- a/application/applicationextension.inc.php
+++ b/application/applicationextension.inc.php
@@ -1389,8 +1389,13 @@ class RestUtils
 		}
 		elseif (is_string($key))
 		{
-			// OQL
-			$oSearch = DBObjectSearch::FromOQL($key);
+			// return all records when the key is "*"
+			if($key == "*"){
+				$oSearch = new DBObjectSearch($sClass);
+			}else{
+				// OQL
+				$oSearch = DBObjectSearch::FromOQL($key);
+			}
 		}
 		else
 		{


chileeb:
i want to find all data,but the key must be required field。

jbostoen:
Why not just use for example `SELECT UserRequest`? 
That already works today?

chileeb:
> Why not just use for example `SELECT UserRequest`?
> That already works today?

yeah，i try it successful.but the final SQL whether is a In syntax?

piRGoif:
Hello,
Sorry we did not understand. 
Can you give some uses cases where you have the problem, and explain how your code would fix it ?

jbostoen:
I think @chileeb assumes you must specify a key (which is correct), but I think he's unaware you can provide an OQL query without condition rather than a number (integer/ID) to retrieve everything?

Seems to be a shortcut to an already existing feature, but yes, now you do need to write `SELECT ClassName` rather than "*"

piRGoif:
Thanks @jbostoen I understand now :)
Personnaly I would agree with @bruno-ds... But we will give an official feedback after the whole team has reviewed the PR.

jbostoen:
I'd agree too. It can easily be implemented in the application which uses the API.
In my PHP implementation, I also removed the need to specify 'class', so...

piRGoif:
Hello,
I just did a review with our product team, they also think using an OQL query answer the need and they don't want to change the current verb : so I'm closing this PR as rejected.
Anyway many thanks for your time and effort @chileeb !
As a side note, there is a iRestServiceProvider interface you can implement to add your own verbs.

PR #96, opened 2019-10-24
Prevent an endless loop when using tagsets for contacts

larhip:


piRGoif:
Hello Lars, can you explain what this is meant to fix ? So, at least a way to reproduce, and ideally a test script ?

larhip:
HI Pierre,
unfortunately I am not able to provide a simple test script for you, but I will try to explain the issue:
I tried to add a new attribute to the class Contact (which should have the type AttributeTagSet) via XML (something like: `<field id="contact_tags" xsi:type="AttributeTagSet" _delta="define"> ...</field>`
After a setup I added some tags to a person which was linked with an user. If this user now tried to login there was an php error regarding an endless loop. 
So i tried to find out why:
After the login iTop tries to get the current contact id, within this function the TagSet will be initiated as well and the method IsValidTag will called during this.
This method calls the method GetAllowedValues which is using a DBObjectSearch. The ObjectSearch tries to add the context parameters (especially the current contact id), so it will try to get the current contact id which tries to initiate the TagSet and now we are in an endless loop.
Hope I was able to explain myself..
Anyhow. You are able to reproduce this issue as soon as you add an AttributeTagSet to the class Contact, add some Tags to a Person which is linked to an user and try to login with this user.

Are you able to follow my explanation?

piRGoif:
Thanks for the explanation ! I will try to reproduce this. Good catch :)

piRGoif:
I reproduced with the XML in this gist : https://gist.github.com/piRGoif/2a246d1d55166a505bd824f10dc086cd

Indeed there is no bug anymore with your fix :)
Will now review this with the rest of the team !

piRGoif:
Hi Lars !
The Combodo dev team reviewed this PR this morning. We prefer not to fix it this way : we want to change the way the Person object is loaded during authentication, so that only needed fields are loaded (OptimizeColumnLoad).

So I'm closing this PR, but I've created a UR on the Combodo support site (with you as caller), linked to a corresponding bug. The product team will review it and add it to the roadmap.

PR #97, opened 2019-10-31
1627 - Ticket ref sometimes duplicate

bruno-ds:
This code is a drop in replacement based on  an abstract counter based on an abstract keyValue store.
The counter can be
 - class based (in this case the counter is initialized on max(id) + 1
 - key based (in this case the counter starts at 0)

Important: on both cases the counter is no more kept aligned with the primary key.

This lead to a MySQL8 compatible implementation.


datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
```diff
@@ -203,12 +203,10 @@
     public function DBInsertNoReload()
     {
           $oMutex = new iTopMutex('ticket_insert');
```
Hipska: Unused variable?

core/cmdbsource.class.inc.php
```diff
@@ -793,6 +793,9 @@ private static function RemoveAllTransactionLevels()
 	}
 
 	/**
+	 *
+	 * @deprecated 2.7.0 N°1627 use ItopCounter insted
```
Hipska: Instead

bruno-ds:
Issue reviewed by @Hipska are resolved in https://github.com/Combodo/iTop/commit/d9ad3f5e988902f993ba7637f2035e087a859a96
thank you!

PR #98, opened 2019-11-01
:bug: iTop stuck in setup (realpath issue) (SF 1817)

jbostoen:
* fixes (Windows) issue where realpath returns lowercase path, causing str_replace to fail

---

SF https://sourceforge.net/p/itop/tickets/1817/

---

There's an issue with **realpath**.

For development purposes, I installed XAMPP x64 (C:\xampp64) recently.
I created a symbolic link so C:\xampp64\htdocs actually is C:\xampp\htdocs

Now, with realpath on Windows 10 and using this XAMPP version ( Windows NT ASUSJ 10.0 build 18362 (Windows 10) AMD64 , PHP Version 7.2.23 ), an issue occurs when running the setup (upgrade).

It happens after the step showing "Information about the upgrade from version 2.6.1.4463 to 2.6.1.4463".

The cause is in utils::GetDefaultUrlAppRoot(), and more specifically in the use of realpath().

I did some digging:
`$_SERVER['SCRIPT_FILENAME'] : C:/xampp64/htdocs/itop/web/setup/index.php`
But running realpath() on that, I get:
`c:\xampp\htdocs\itop\web\setup\index.php`

The difference is very small (it correctly identifies the real file), but the consequence is the setup crashes. Because `C: has become c:`

* https://www.php.net/manual/en/function.realpath.php

> Note: For case-insensitive filesystems realpath() may or may not normalize the character case.

Now, it works when changing str_replace to **str_ireplace** for the assignment of $sCurrentRelativePath (2 times)

**There might be more similar issues in iTop.**

Also, I did not work this out yet, but I'm wondering if this check + fallback method can't be simplified to one case with some **preg_match**

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -840,7 +840,7 @@ class utils
 
 		$sCurrentScript = realpath($_SERVER['SCRIPT_FILENAME']);
 		$sCurrentScript = str_replace('\\', '/', $sCurrentScript); // canonical path
-		$sAppRoot = str_replace('\\', '/', APPROOT); // canonical path
+		$sAppRoot = str_replace('\\', '/', realpath(APPROOT)); // canonical path
 		$sCurrentRelativePath = str_replace($sAppRoot, '', $sCurrentScript);
 	
 		$sAppRootPos = strpos($sAbsoluteUrl, $sCurrentRelativePath);


bruno-ds:
applying  `realpath()` to approot should work too and should also be more restrictive for case sensitive systems.

Hipska:
Please keep your changes from another PR separated, only add commits that have relation to this PR.

jbostoen:
> 
> 
> Please keep your changes from another PR separated, only add commits that have relation to this PR.

Yeah I know; I'm struggling a bit with the rebase, should be fixed in a couple of minutes.

piRGoif:
This code seems good to me :) Indeed simply changing the str_replace was dangerous, using real_path is more "atomic".

@jbostoen can you please give a method to reproduce the problem ? I'm on Windows too but can't manage to make it happen...

jbostoen:
The way I stumbled upon it:

* C:\xampp\htdocs directory (x86 PHP)

For some other project, it turned out I needed a x64 PHP version before it recognized some shell commands. So I installed x64 XAMPP

Now rather than copying everything, I decided to leave both PHP x86 and x64 on the system, and link the directory (similar to `ln` on Linux)

`mklink /D "C:\xampp64\htdocs" "C:\xampp\htdocs"`

piRGoif:
Ok thanks :) The new fix with the real_path call solves the problem too right ?

jbostoen:
It does seem to work too, yes.

piRGoif:
Ok thanks :)
I'll try to reproduce it myself and we should be good to merge !

piRGoif:
Didn't manage to reproduce, but ok to merge as the risk is null :)

piRGoif:
I've just closed the corresponding [SF#1817 ticket](https://sourceforge.net/p/itop/tickets/1817/) :)

bruno-ds:
@jbostoen I just push this commit : https://github.com/Combodo/iTop/commit/45536cf957ca1a62b9ee7969aa900cf64a0de683

It fix this regression: the setup has a 'app_root_url' with the final '/' trimmed.

jbostoen:
> 
> 
> @jbostoen I just push this commit : [45536cf](https://github.com/Combodo/iTop/commit/45536cf957ca1a62b9ee7969aa900cf64a0de683)
> 
> It fix this regression: the setup has a 'app_root_url' with the final '/' trimmed. :)

Actually in beta 2.7.0 it does NOT fix the issue because it's again case sensitive.


In `GetDefaultUrlAppRoot::GetDefaultUrlAppRoot();` my initial values for 
`$sAppRootPos = strpos($sAbsoluteUrl, $sCurrentRelativePath);`
are at that point  in the code (around line 853):

> absolute URL = http://127.0.0.1:8181/itop/web/setup/index.php
> app root = C:/xampp/htdocs/itop/web/ 
> current script = c:/xampp/htdocs/itop/web/setup/index.php
> 
> current relative path = c:/xampp/htdocs/itop/web/setup/index.php

Using str_ireplace right above that line (and one line below it) fixes the entire issue.

bruno-ds:
Sorry @jbostoen 😱
This is the third regression:
We definitely should write a test.

sensor-dream:
-- Please reply above this line --

Сообщение получено

-----------------------------------------------------------
Support powered by FreeScout — Free open source help desk & shared mailbox

sensor-dream:
-- Please reply above this line --

Сообщение получено

-----------------------------------------------------------
Поддержка работает на FreeScout — бесплатный хелпдек с открытым кодом и общий почтовый ящик

piRGoif:
Hello @jbostoen ,
You're getting drive letters in different cases although both values are made using a realpath() call ?!!?? This is very odd...
I've just pushed 13d31ac2, can you confirm this is solving your issue ?

jbostoen:
@piRGoif It's indeed very odd. I think it has to do with the way PHP handles it on Windows systems; whlie one is a URL and the other a local file path. The fix should work, looks the same as the changes I did on my local file last night. Thanks!

piRGoif:
Nice this is solved :)
I didn't manage to reproduce this on my Windows system... Can you send your exact PHP version ?

jbostoen:
PHP Version 7.2.23
System 	Windows NT ASUSJ 10.0 build 18363 (Windows 10) AMD64
Build Date 	Sep 25 2019 08:47:09
Compiler 	MSVC15 (Visual C++ 2017)
Architecture 	x64 

But I think being able to reproduce this also has to do with the symbolic link ( C:\xampp64\htdocs -> C:\xampp\htdocs )

piRGoif:
Thanks for the info :) I trie with PHP 7.2.24 and had no luck... Didn't see anything [in the changelog](https://www.php.net/ChangeLog-7.php#7.2.24) about this...

> But I think being able to reproduce this also has to do with the symbolic link 

Right ! Are you using the Windows mklink command, or another way ?

jbostoen:
`
mklink /D "C:\xampp64\htdocs" "C:\xampp\htdocs"`  is what I used; both directories are also lowercase in Explorer; not just in the command

piRGoif:
Ok thanks. Did the same, still no error :D
Well, we have a fix so let's keep it :)

PR #101, opened 2019-12-09
Feature/faf log modularity

bruno-ds:
introduce the concept of channels,
enable log filtering per channel.

Also add a new log level "Debug" wich is not written with the default conf.

This leverage the possibility to 
 - write debug log for only a given channel
 - mute logs bellow a given level for specifics channels

### DIFF
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -99,11 +99,11 @@ class Config
 	 * @since 2.7.0 export_pdf_font param
 	 */
 	protected $m_aSettings = array(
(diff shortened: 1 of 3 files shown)


bruno-ds:
example of possible conf: 
```
	'min_log_level' => array(
		'' => 'Error',
		'IssueLog' => true,
		'FooBar' => 'Debug',
	),
```

piRGoif (dismissed):

odain-cbd:
review ok. added testcases.

PR #102, opened 2019-12-23
:bug: Fix session start check

dlepenven:
:bug: Add a simple test on session start and trigger an exception when it fails

### DIFF
--- a/application/startup.inc.php
+++ b/application/startup.inc.php
@@ -54,7 +54,12 @@ register_shutdown_function(function()
 });
 
 session_name('itop-'.md5(APPROOT));
-session_start();
+$bSessionStart = session_start();
+// Trigger an exception if the session cannot be started
+if ($bSessionStart !== false)
+{
+        throw new Exception("Session cannot start");
+}
 $sSwitchEnv = utils::ReadParam('switch_env', null);
 $bAllowCache = true;
 if (($sSwitchEnv != null) && (file_exists(APPCONF.$sSwitchEnv.'/'.ITOP_CONFIG_FILE)) && isset($_SESSION['itop_env']) && ($_SESSION['itop_env'] !== $sSwitchEnv))
--- a/application/startup.inc.php
+++ b/application/startup.inc.php
@@ -56,7 +56,7 @@ register_shutdown_function(function()
 session_name('itop-'.md5(APPROOT));
 $bSessionStart = session_start();
 // Trigger an exception if the session cannot be started
-if ($bSessionStart !== false)
+if ($bSessionStart === false)
 {
         throw new Exception("Session cannot start");
 }


piRGoif:
Thanks, 
Relates to https://sourceforge.net/p/itop/tickets/1829/

dlepenven:
I made a mistake here, should be === false not !== false :-(

dlepenven:
In my case, I use a memcached server to handle PHP sessions.

During "huge" requests, that can take severals seconds, if I open a new tab or any itop page, I have already a request handled by the server, so the session file is locked.
The memcached system make some retries until the number configured.
in this case session_start() returns a false.

So the idea is to handle this case with a simple test.

https://www.php.net/manual/fr/function.session-start.php :
Return Values ¶
This function returns TRUE if a session was successfully started, otherwise FALSE.

piRGoif:
> During "huge" requests, that can take severals seconds

Indeed in iTop we do not do any session_write_close()... Do you have some examples of actions that could block your iTop ?

Having a chat about the solution you proposed with the rest of the R&D team, we think that there should always be an error thrown in such case, by the call to \LoginWebPage::DoLoginEx ? 
Do you have any use case were it's not ?

dlepenven:
> > During "huge" requests, that can take severals seconds
> 
> Indeed in iTop we do not do any session_write_close()... Do you have some examples of actions that could block your iTop ?
> 
> Having a chat about the solution you proposed with the rest of the R&D team, we think that there should always be an error thrown in such case, by the call to \LoginWebPage::DoLoginEx ?
> Do you have any use case were it's not ?



> Indeed in iTop we do not do any session_write_close()... Do you have some examples of actions that could block your iTop ?
> 
> Having a chat about the solution you proposed with the rest of the R&D team, we think that there should always be an error thrown in such case, by the call to \LoginWebPage::DoLoginEx ?
> Do you have any use case were it's not ?

The real problem is not really the freeze. It is the fact that the session_start could fail, and not having any informations about this. The solution I propose seems to cover my problem, but maybe it should be extended to others functions I don't know. 
An other case of session fails for exemple is when I configure a memcache server in PHP but doesn't start it. In this case, I always return to the login page in iTop.

piRGoif:
> The real problem is not really the freeze. It is the fact that the session_start could fail, and not having any informations about this.

We were thinking that because of the DoLoginEx() call, if session fails to load, being redirected to the login screen would give enough information ? And most importantly prevent the code from being executed...

Hipska:
Would the better solution then be, that user still gets redirected to the login page and a message of failed `session_start` be put into the log file? This way the user don't get useless details and the admin can find the root cause.

dlepenven:
> > The real problem is not really the freeze. It is the fact that the session_start could fail, and not having any informations about this.
> 
> We were thinking that because of the DoLoginEx() call, if session fails to load, being redirected to the login screen would give enough information ? And most importantly prevent the code from being executed...

Yes indeed! My solution is a little.... brutal...

> Would the better solution then be, that user still gets redirected to the login page and a message of failed `session_start` be put into the log file? This way the user don't get useless details and the admin can find the root cause.

 And if it is possible to show a message to the user it would be even better :D

dlepenven:
I close my request since the edits I made are not okay for the moment.

piRGoif:
Thanks @dlepenven !
I like the idea proposed by Hipska... That could be something to do in the future.

PR #105, opened 2020-01-09
:globe_with_meridians: Added ES CR translations

lazki:
As requested by @Molkobain here are some new translations :-)


Molkobain:
Wow, that was fast! 😁

Thanks @lazki

PR #106, opened 2020-01-09
:globe_with_meridians: Typos and new translations in ES CR

lazki:
@Molkobain some more translations


dictionaries/es_cr.dictionary.itop.core.php
```diff
@@ -1037,3 +1037,4 @@
 	'Class:AsyncTask/Attribute:finalclass' => 'Clase',
 	'Class:AsyncTask/Attribute:finalclass+' => 'Clase',
 ));
+:globe_with_meridians:
```
Molkobain: You have a line that should not be here @lazki

lazki: Oops!

Fixed.

Molkobain:
Great thanks! Will be part of the 2.7.0 as well ✌

PR #111, opened 2020-01-29
N°2293 DBUpdate : save changed fields and corresponding previous values

piRGoif:
In 2.7.0-beta we added DBObject::$m_aChanges
Later on we added the DBObject::ListChangesUpdated as a getter for use in \iApplicationObjectExtension::OnDBUpdate

But that was not enough as we didn't have access to orig values.

We removed this and add in DBObject :
$m_aPreviousValuesForUpdatedAttributes attribute, set at the very beginning of DBUpdate
ListPreviousValuesForUpdatedAttributes() getter

### DIFF
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -83,9 +83,9 @@ abstract class DBObject implements iDisplay
 	/** @var bool true IF the object is mapped to a DB record */
 	protected $m_bIsInDB = false;
 	protected $m_iKey = null;
-	/** @var array attcode => value : corresponding current value (the new value passed to {@link Set}). Reset during {@link DBUpdate} */
+	/** @var array attcode => value : corresponding current value (the new value passed to {@see DBObject::Set()}). Reset during {@see DBObject::DBUpdate()} */
 	private $m_aCurrValues = array();
-	/** @var array attcode => value : previous values before the {@link Set} call. Array is reset at the end of {@link DBUpdate} */
+	/** @var array attcode => value : previous values before the {@see DBObject::Set()} call. Array is reset at the end of {@see DBObject::DBUpdate()} */
 	protected $m_aOrigValues = array();
 
 	protected $m_aExtendedData = null;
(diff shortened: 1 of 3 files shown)


core/dbobject.class.php
```diff
@@ -3305,6 +3308,30 @@ public function DBUpdate()
...
+	private function InitPreviousValuesForUpdatedAttributes()
+	{
+		$aChanges= $this->ListChanges();
+		if (empty($aChanges))
+		{
+			$this->m_aPreviousValuesForUpdatedAttributes = array();
+			return;
+		}
+
+		$aPreviousValuesForUpdatedAttributes = array_intersect_key($this->m_aOrigValues, $aChanges);
```
Molkobain: Except from the result of that line that needs to be tested with a real use case, sound all good to me! 👌

Molkobain (dismissed):

core/dbobject.class.php
```diff
@@ -134,10 +136,11 @@ abstract class DBObject implements iDisplay
 	 */
 	protected $m_aModifiedAtt = array();
 	/**
-	 * @var array attname => currentvalue Persists changes for {@link DBUpdate}
+	 * @var array attname => value : value before the last {@link Set} call. Set at the beginning of {@link DBUpdate}.
+	 * @see ListPreviousValuesForUpdatedAttributes getter for this attribute
```
bruno-ds: I think you should type it : 
ListPreviousValuesForUpdatedAttributes::ListPreviousValuesForUpdatedAttributes()

cf : https://github.com/Combodo/iTop/tree/develop/.doc#known-limitations

core/dbobject.class.php
```diff
@@ -852,17 +855,13 @@ public function GetStrict($sAttCode)
 	}
 
     /**
-     * Get the value as it was before change with Set
```
bruno-ds: I don't get why you removed the explanation

bruno-ds (dismissed):
Beside the test recommended by Molkobain (and maybe a unit test if possible), 
It sound all good to me.

core/dbobject.class.php
piRGoif: I moved it to the @return tag.
Ok for you ?

core/dbobject.class.php
piRGoif: Wooops thanks, I forgot !
I fixed that for the added @see... and also I changed the @link I did put everywhere to @see, as @link should be used only for URI.
Tell me if those modifications are ok now ?

piRGoif: Saw with Bruno : the rendering is good enough for now. Just a little prb on the way @uses are printed, but we will make a review on the whole doc later on.

core/dbobject.class.php
piRGoif: Christophe did some test, this is validated, I'm closing the conversation :)

PR #114, opened 2020-02-20
Add blocks to enable customization

v-dumas:


eespie (approved):

v-dumas:
This will allow to reuse this template form in case of password expiration, without asking for the old password again as we detect it in the LoginFSM OnUserOk() and the old password was already checked.

bruno-ds:
Since there is only one known use case, and the required changes is rather simple, I would rather add conditions into the template than enable extensibility. 

This is simpler, and we would be less afraid to perform BC changes in the future.

PR #115, opened 2020-02-21
N°2889 - Add counter & triggers on file attributes / attachments downloads

Molkobain:
~Not for iTop 2.7 obsviously, but maybe iTop 2.8? 😁~
Aiming at 3.1 🤞

## What it does
* Add a downloads counter in the ormDocument object (file attributes, attachments)
  * Counter starts at 0 after migration, even for existing documents that have already been downloaded in the past.
  * Display counter in attachments table (in object details).
  * Display counter on AttributeBlob, next to the size.
  
* Add a trigger on file attributes (eg. `file` on the `DocumentFile` class)
  * There is no "target fields" attribute on the trigger because this mechanism only works with fields that are updated by the user, which is not the case
  * Placeholders
    * `$this->xxx$` for the object hosting the attribute
    * `$file->mime_type$`, `$file->file_name$`, `$file->downloads_count$`, `$file->data$`, `$file->data_as_base64$` (use the last two with caution as it can be huge) for the corresponding properties of the `ormDocument`.
      * `$file->downloads_count$` is the count BEFORE the current download. Should it rather be after the current download?
      
* Add a trigger on attachments downloads
  * Placeholders
    * Same as for the previous trigger but `$attachment->` instead of `$file->$`
    * `$attachment->xxx$` for the attachment itself. Not so relevant for user-to-user notifications (eg. Slack, email) but can be handy for app integrations via webhooks.

### Attachment download trigger
  * [x] Should it have a mandatory "target class" attribute like regular `TriggerOnObjectXXX` or can it be for attachments accross the whole datamodel?
  
### Document attribute trigger
  * [X] ~Should it have an optional "target fields" attribute like regular `TriggerOnObjectUpdate`? Will require some refactoring to avoid duplication, but can be done.~
    => Cannot be done yet as the field doesn't appear as "changed" and is therefore not pass to the trigger

### Misc.
* [X] Update PHPDoc to tell iTop version it was introduced
* [X] Use the new event system
* [X] How to avoid coupling with the Attachment module and avoid this being triggered for an Attachment object as it is already handled in the attachment module? (see `sources/Core/EventListener/AttributeDocumentEventListener.php`)
* [X] Discuss impact of adding a new column to the attachments table (and on attribute blob columns but less impact)
  As discussed with Erwan and Stephane, since MySQL 5.7 and MariaDB 10.3 there is no impact anymore (with the default DB settings)
* [x] Discuss if the notification should include the download count BEFORE or AFTER the current download.
  As discussed with Eric, it's not that simple to add a "After document download"
  * At the end of the `ormDocument::DownloadDocument()` the page isn't output yet
  * In `DownloadPage` we don't have context about the host object / attribute, so don't have much data to pass to the event

## Limitations
* When file / attachment is retrieved directly from browser's cache, counter is not increased and trigger not triggered. We could imagine adding a cache buster in certain conditions to the `"download_document"` operation of `ajax.document.php`, but that might be out of scope.


piRGoif:
You may use the new KeyValueStore class to store the count ? It might be easier, and more flexible ?

bruno-ds:
Since there is one counter per row iTop counter doesn't seems appropriate to me.
(For the ref, there is only one counter per root class)

Molkobain:
Tracked in Combodo's system under N°2889

piRGoif:
I've just removed the milestone as 3.0.0 was delivered some time ago !
@Molkobain is this PR really in the pending Combodo state ?

Molkobain:
Yes, waiting for the events service to come out.

Molkobain:
Rebased on develop and forced push.
Starting rework to use the new event system.

Molkobain:
Ready for review!

v-dumas:
> Add a trigger on file attributes (eg. file on the DocumentFile class) "_**download**_"

Without the last word, I am lost, what does it mean exactly, when is it triggered, on file download?
If yes, can it be filtered on some AttributeFile(s) like a TriggerOnObjectUpdate?

If I have on a CustomerContract class, defined two AttributeFile with code "**contract**" and "**appendix**", my placeholder for the first one would be: _$this->contract_downloads_count$_ or _$this->document_downloads_count$_
Which depending on the trigger may be guessable or not.

For attachment, there is no doubt as there aren't any field code.

Molkobain:
Discussed with Vincent:

> > Add a trigger on file attributes (eg. file on the DocumentFile class) "_**download**_"
> 
> Without the last word, I am lost, what does it mean exactly, when is it triggered, on file download? If yes, can it be filtered on some AttributeFile(s) like a TriggerOnObjectUpdate?

It is trigger on file download, which means when the user clicks the hyperlink.
It cannot be filtered yet in the trigger, that one of the pending questions. We decided to add the filtering like on other triggers.
 
> If I have on a CustomerContract class, defined two AttributeFile with code "**contract**" and "**appendix**", my placeholder for the first one would be: _$this->contract_downloads_count$_ or _$this->document_downloads_count$_ Which depending on the trigger may be guessable or not.
> 
> For attachment, there is no doubt as there aren't any field code.

The action is always executed for 1 AttributeFile (when the clicks on the hyperlink), so I'll add placeholders for both attribute code / label so we can display then in the Action.

Molkobain:

Off this PR, a discussion should take place with Eric and Pierre to discuss about where the triggers should be (within the listener?) and how to handle their errors.

Molkobain:

  * [x] Erwan and Guillaume will check impact of adding a empty column to a table with a BLOB column.
  * [x] Add downloads count in the AttributeBlob rendering dict entry (with "Open / Download")
  * [x] Add downloads count column in the portal attachments table
  * [x] Change placeholders in the actions so
    * `$document_xxx$` become `$attachment->xxx` for the attachment trigger
    * `$document_xxx$` become `$file->xxx` for the AttributeBlob trigger

Molkobain:
> * [x]  Add downloads count in the AttributeBlob rendering dict entry (with "Open / Download")

I did several mockups, none were really great, this is what I came up with.
image

Molkobain:

Regarding the interrogation on adding a new column in tables with a lot of blobs: Since MySQL 5.7 and MariaDB 10.3.7 the algorythm changed and there no temporary table created. It was tested on a very big DB (10GB+), the alteration (during the setup) took 62ms.

Nevertheless, a warning will be added in the migration notes to warn about this column addition.
Also that the default parameter `ALGORYTHM=INPLACE` in MariaDB can be overload in the MariaDB conf (`ALGORYTHM=COPY`)`, in that case the migration will take longer.

piRGoif:
> Off this PR, a discussion should take place with Eric and Pierre to discuss about where the triggers should be (within the listener?) and how to handle their errors.

We didn't took time to have this conversation :) 
I have an idea of a TriggerOnEvent, I'll try to make a PR !

PR #116, opened 2020-02-24
[WIP] POC of a dependency injection

bruno-ds:
This is a POC of integrating a dependency injection container into iTop.
As of now it does nothing but 
 - build the container into a test of `ContainerBuilderBridgeTest.php`.
 - replace `iApplicationObjectExtension::OnDBUpdate()` by an event

It only role is to demonstrate that plugging Symfony's Service Container is not as hard as it may seems.


bruno-ds:
for the new hook API, we will use event, but not symfony as rapidly poc in this PR.
see : https://github.com/Combodo/iTop/tree/feature/faf_event_service

PR #117, opened 2020-02-27
N°2793 Log rotation

piRGoif:
With N°2518 I implemented a simple system where the log file name was calculated each time we want to open it. This was not cool as usual error.log or setup.log were no longer present, the current log file name was not deterministic anymore (have to guess which policy, and which name based on current date)

Those modifications get back to the old system : we are still writing to usual file names.
But a check is done  before writing to see if we need to rotate the file.

### DIFF
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -161,7 +161,7 @@ abstract class RotatingLogFileNameBuilder implements iLogFileNameBuilder
 	 */
 	protected function GetRotatedFileName()
 	{
-		$sFileSuffix = $this->GetFileSuffix();
+		$sFileSuffix = $this->GetFileSuffix(static::$oLogFileLastModified);
 		return $this->sFilePath.DIRECTORY_SEPARATOR
 			.$this->sFileBaseName
 			.'.'.$sFileSuffix
@@ -177,10 +177,11 @@ abstract class RotatingLogFileNameBuilder implements iLogFileNameBuilder
 	abstract public function ShouldRotate($oLogDateLastModified, $oNow);
 
 	/**
+	 * @param DateTime $oDate log file last modification date
+	 *
 	 * @return string suffix for the rotated log file
-	 * @uses oLogDateLastModified
 	 */
-	abstract protected function GetFileSuffix();
+	abstract protected function GetFileSuffix($oDate);
 
 	/**
 	 * @see \LogFileRotationProcess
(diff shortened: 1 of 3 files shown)


piRGoif:
@odain-cbd  & @bruno-ds feedbacks were pushed in e68656fd

Keeping this PR opened to get feedbacks also from @rquetiez and @eespie !

piRGoif:
Did a review with @eespie, feedbacks pushed in 14a9af19
Main point : try/catch around the iTopMutex

=> ok to go, merging !

core/log.class.inc.php
```diff
@@ -158,24 +158,34 @@ public function CheckAndRotateLogFile()
...
+		{
+			$oLock = new iTopMutex('log_rotation_'.$this->sLogFileFullPath);
+			$oLock->Lock();
+			$this->ResetLastModifiedDateForFile();
+			$sNewLogFileName = $this->GetRotatedFileName($oLogFileLastModified);
+			rename($this->sLogFileFullPath, $sNewLogFileName);
+		}
+		catch (Exception $e)
+		{
+			// nothing to do, cannot log... file will be renamed on the next call O:)
```
bruno-ds: Shouldn't you log this?
just in case if it happen too often?

(honestly it's really a question: I don't know if you should)

core/log.class.inc.php
```diff
@@ -158,24 +158,34 @@ public function CheckAndRotateLogFile()
...
-		$this->ResetLastModifiedDateForFile();
-		$sNewLogFileName = $this->GetRotatedFileName($oLogFileLastModified);
-		rename($this->sLogFileFullPath, $sNewLogFileName);
-		$oLock->Unlock();
+		$oLock = null;
+		try
+		{
+			$oLock = new iTopMutex('log_rotation_'.$this->sLogFileFullPath);
+			$oLock->Lock();
+			$this->ResetLastModifiedDateForFile();
```
bruno-ds: After acquiring the lock, maybe you should verify if the work have been done.
(it may occur if several concurrent script try to rename it)

Personnaly I'd skip if `is_file($sNewLogFileName)`

piRGoif: Modification viewed with @bruno-ds, and pushed in 33f3f281

core/log.class.inc.php
piRGoif: Hard to log for code that is inside the log API...
Discussed with @bruno-ds  : we don't change this.

PR #121, opened 2020-03-06
FAF iTop objects extensibility API v2

piRGoif:
# Existing API and its pro/cons

Currently there are 2 ways to change iTop objects behavior :

* override DBObject methods (one that is tagged @overwritable-hook)
  - 👌 member method, so anything is possible !
  - 😭 only one override possible
  - 🤔 ambiguous naming (Is OnUpdate called before updating the DB?)
* create an iApplicationObjectExtension implementation
  - 👌 multiple implementations possible
  - 🤔 no rank
  - 🤔 called by every objects
  - 🤯 no exception handling by the caller in DBObject
  - 🕵️ hard to debug, each module can bring its own implementation, no centralized list of what will be called in which order
  - 🤔 ambiguous naming (Is OnDBUpdate called before updating the DB?)
  - 🤔 the interface contract is difficult to extend

Moreover, the naming is inconsistent between the two options:

  - DBObject::OnUpdate is called before the update
  - iApplicationObjectExtension::OnDBUpdate is called after the update

This id documented in this wiki page : [CRUD methods sequentiallity [iTop Documentation]](https://www.itophub.io/wiki/page?id=3_0_0%3Acustomization%3Asequence_crud)

# Proposed solution

## XML declaration
Each objects can define actions to launch. Each one of them will point to an existing method of this object.

Here is a sample XML, containing method and action definitions :
⚠ The XML structure was changed after the dev team review, see [comment below](https://github.com/Combodo/iTop/pull/121#issuecomment-598848026)

```xml
	<classes>
		<class id="FunctionalCI" _delta="must_exist">
			<methods>
				<method id="poc_update_method" _delta="define">
					<comment/>
					<static>false</static>
					<access>protected</access>
					<code><![CDATA[
protected function PocUpdateMethod()
{
    IssueLog::Info('PocUpdateMethod called !');
	throw new CoreException('testing exception in PocUpdateMethod');
}
]]></code>
				</method>
			</methods>
			<actions>
				<action id="poc_update_action" _delta="define">
					<type>update</type>
					<method>PocUpdateMethod</method>
					<rank/>
				</action>
			</actions>
		</class>
	</classes>
</itop_design>
```

## Pro/cons

* ➕ XML definition : standard way of doing extensibility in iTop using the `_delta` attribute !
* ➕ uses method definition : compatible with the ITSM Designer
* ➕ rank !
* ❓  Shouldn't we pass something to those action methods ? Event (can be resolved by indirections), ... ?
* ➖ called code is a member method, this could be quite dangerous (code segregation)
* ➖ XML so does not work for PHP classes
* ➖ cannot define something for all classes (for example at the cmdbAbstractObject class level)
* ➖ testability

# TODO

* [X] XML syntax v1 POC
* compiler v1 POC
  - [X] helper to get module defining the node => \Combodo\iTop\DesignDocument::GetItopNodeSourceModule
  - [X] generate the method to launch
  - [X] called method exception handling
* Reviews 
  - [X] Dev team (13/03/2020)
  - [x] Christophe (19/03/2020)
  - [x] Product team (07/04/2020)
* compiler v2 POC
  - [ ] event/action structure
  - [ ] parent call
* Specs / cost evaluation
  - [ ] list of different events we need (CRUD + apply_stimulus + ??)
  - [ ] mapping between existing interfaces and new events
  - [ ] existing modules audit (which ones will need to be migrated ? How long will it take to implement the new extensibility in those ?)
  - [ ] Designer part (events node edition, ) : cost evaluation
* [ ] XML syntax final
* compiler final
  - [ ] `action` node should point to `method` node id instead of directly method name
  - [ ] inconsistencies handling
  - [ ] handle class hierarchy
  - [ ] handle action rank
  - [ ] log each action object changes diff (diff of ListChanges() before action call and after)
  - [ ] helper in MetaModel to get list of actions (MetaModel::EnumActionsForClass() ?)
* [ ] where to plug our events in the current code

# Other related topics

* [ ] use the new \Combodo\iTop\DesignDocument::GetItopNodeSourceModule on compiler errors (tody we only have the node XPath and not the module where this node is declared)


rquetiez:
"cannot define something for all classes (for example at the cmdbAbstractObject class level)"

I think that this is feasible (inheritance could apply to the PHP class tree, not only the business class trees), but that is an option when the implementation gets finalized, no ?

rquetiez:
"called code is a member method, this could be quite dangerous (code segregation)"

Given a white list of callable verbs (or considering "public" methods only), static PHP code analysis could cope with that (check at compile time). That list should be part of the meta tag, allowing for any extension to declare wether a new verb should be "public" or "private".

bruno-ds:
> "called code is a member method, this could be quite dangerous (code segregation)"
> 
> Given a white list of callable verbs (or considering "public" methods only), static PHP code analysis could cope with that (check at compile time). That list should be part of the meta tag, allowing for any extension to declare wether a new verb should be "public" or "private".

Isn't it un-intuitive to re-implement a scope on Metamodel level rather than on PHP language level (by using public/private/protected) ?

piRGoif:
Some notes after a meeting with some of the dev team members this afternoon.


# Sum-up

Seems everyone agree with this proposition. We need to identify all of the extensions point now... They are numerous, not only object CRUD but other events too like adding a linked object, etc. Indeed everything that calls a stack containing DBObject methods is a candidate !


# Parent call
We discussed about the hierarchy handling : a nice solution came up ! First group all events in a event node with some action nodes as children. Each action can have a rank indicating whether it needs to be called before of after the parent call : negative rank to be called before, positive rank to be called after.
A specific node can be specified under the event node to specify if we will generate the parent call or not.

Here is a sample XML :

```xml
<events>
  <event id="onBeforeUpdate">
    <generateParentCall>true</generateParentCall>
    <action id="beforeParent">
      <method>MyMethod1</method>
      <rank>-1</rank>
    </action>
    <action id="afterParent">
      <method>MyMethod2</method>
      <rank>1</rank>
    </action>
  </event>
</events>
```

# iTop compatibility
When this new API will be published, we need a compatibility break : extensions should implement both old api (either DBObject or iApplicationObjectExtension) and new one (event/action node in XML).
Actually we don't have a way in extensions to tell "use this API if iTop version is N and use this other one if iTop version is N+1". So having only the new API supported will allow to have extensions compatible with older and current iTop versions.

# Testability
For now it is impossible to mock called methods to test if the stack calls what it is intended to call (for example, method1 and method2 called when doing an update).
We have some solutions (indirections using an array containing methods names, call to methods in other objects, ...) but nothing very elegant.

piRGoif:
I'm closing this PR : another work by @eespie was made in branch feature/faf_event_service, and will certainly be the solution kept.
Merge with the work done here was already done.

PR #129, opened 2020-03-20
[WIP] boot a symfony kernel during iTop own startup

bruno-ds:
This POC launch a symfony kernel during MetaModel::Startup().

It is **not** an official combodo project, I do it (mostly) on my free time because I'm really curious about how far we can go.
Since we really enjoy being the only master of our framework, it may not go beyond this POC.
But is the liberty we have worth the price of every single feature?

---------------------------
**WHY** (meta version):
 - Our product is the Extendable model put on top of the framework. And we are a niche framework
 - Symfony's framework is their product.  And it is one of the most used in the world

So, while Symfony has to produce a framework as good as possible, we only have to produce the framework as good as required considering the requirements of our product. 
So it is quite normal not to have the same investment put on not core iTop feature.

------------------------------
**WHY** (business version):
non-core functionalities should be externalized while core functionalities should be given the focus on. Once this done, a better effort can be done over what brings value.
On one hand, some part of the code are at the very core of the attractiveness of iTop, and have no reason to be delegated. 
On the other hand, other parts are just required functionalities, the effort invested in them is proportional. For them, we could benefit from using external components.

----------------------
**Further reading:**
I find this blog post a really good POV about how re-using instead of coding everything helps: https://www.freecodecamp.org/news/we-fired-our-top-talent-best-decision-we-ever-made-4c0a99728fde/

--------------------------------------
**symfony components leverage:** 
 - the wonderfull profiler
 - the autodiscovery of the conf based on customizable conventions
   - even for the extensions
 - the router / controller / twig
   - even for the extensions
 - the console
   - and its debug tooling
   - and it easy as hell new command develpment
      - even for the extensions
 - the service container autowired
 - the event system
 - any o f the high quality symfony components
   - cache
   - https://symfony.com/doc/current/components/messenger.html
   - ...
 - PSR compatible log class (who said monologue??)
 - so many more components from the symfony ecosystem
 - an easier learning curve since
    - a lot of PHP developer already know symfony 
    - symfony  is easy to learn
 - a better DX (developper experience): symfony  encompass the developer experience as a key feature
 - stability / cost reduction
    - since symfony is well tested, we only have to worry about our implementation on top of it
 - testability
    - symfony help leveraging best practice easing unit/functionnal tests 

-----------------------------------
**A note about the PR's code**:
The current work has not a production grade quality.
For instance, you should see how in integrated UI.php inside a controller... beurk!

-----------------------------------
**psr4 legacy layer proposal: **
This code containes also a psr4 legacy layer proposal.
It was the first developped feature, but I did not pushed it as far as I should, since in my mind, the next step was tu use the PSR-4 compatible componenents within the dependency injection.

Exaplanation of the proposal:  `.make/namespaces-generator/runner.php` contains a script that generate classe alias located in psr4 compatible path, leveraging an auto-discovery/auto-wiring inside the service container.

----------------------
**To test this POC:**
 - checkout : feature/faf_embrace_symfony
 - at the root of the project, create a .env.local containing `APP_ENV=dev` 
 - log in
 - "et voila"!

------------------
**Screenshots**
image

image

image

image

image

image


bruno-ds:
The namespaces are based on a mapping : 
```php
$aScannedDirs = array(
	'core' => array(
		'exclude' => array('oql'),
		'namespaces' => array(
			'apc-.*' => '\\Combodo\\iTop\\Core\\Cache\\Apc',
			'attributedef.class.inc.php' => '\\Combodo\\iTop\\Core\\AttributeDefinition',
			'db.*' => '\\Combodo\\iTop\\Core\\Orm\\DbObject',
			'oql.*' => '\\Combodo\\iTop\\Core\\Orm\\Oql',
		),
	),
);
```

bruno-ds:
This was never meant to be integrated as it, I close the PR (but not my ❤️, <3 Symfony).

PR #99, opened 2019-11-04
Support array for json_data posted in rest/json service

Ousret:
When using x-www-form-urlencoded we could leverage the possibility to send array instead of raw json as PHP support it out of the box.

This change/bugfix is not BC-Break.

I do not know if this is a bug or not, when reading [current documentation](https://www.itophub.io/wiki/page?id=latest%3Aadvancedtopics%3Arest_json), it is not clearly stated that `json_data` should be a string.

```
auth_user=my_user&auth_pwd=my_password&json_data%5Boperation%5D=core%2Fget&json_data%5Bclass%5D=ServiceChange&json_data%5Bkey%5D=SELECT+ServiceChange+WHERE+id%3D1&json_data%5Boutput_fields%5D=request_state
```
**instead of**
```
auth_user=my_user&auth_pwd=my_password&json_data=%7B%22operation%22%3A+%22core%2Fget%22%2C+%22class%22%3A+%22ServiceChange%22%2C+%22key%22%3A+%22SELECT+ServiceChange+WHERE+id%3D1%22%2C+%22output_fields%22%3A+%22request_state%22%7D
```

It is more convenient to post json_data[operation]=core/get instead of json_data={"operation": "core/get"..

PR #130, opened 2020-03-27
REST API : new array syntax for json_data parameter

piRGoif:
Before iTop 2.8.0, REST API json_data param needed a JSON value. So you would call the API like this :

```php
CURLOPT_POSTFIELDS => array(
	'auth_user' => 'admin',
	'auth_pwd' => 'admin',
	'json_data' => '{
	   "operation": "core/get",
	   "class": "Person",
	   "key": "SELECT Person", "limit": "10", "page": "1"
	}'
)
```

With #99, it became possible to use this syntax :

```php
CURLOPT_POSTFIELDS => array(
	'auth_user' => 'admin',
	'auth_pwd' => 'admin',
	"json_data[operation]" => "core/get",
	"json_data[class]" => "Person",
	"json_data[key]" => "SELECT Person",
	"json_data[limit]" => 10,
	"json_data[page]" => 1
)
```

Now with this modification we can use :

```php
CURLOPT_POSTFIELDS => array(
	'auth_user' => 'admin',
	'auth_pwd' => 'admin',
	"operation" => "core/get",
	"class" => "Person",
	"key" => "SELECT Person",
	"limit" => 10,
	"page" => 1
)
```

### DIFF
--- a/webservices/rest.php
+++ b/webservices/rest.php
@@ -70,6 +70,8 @@ $oCtx = new ContextTag(ContextTag::TAG_REST);
 
 $sVersion = utils::ReadParam('version', null, false, 'raw_data');
 $sOperation = utils::ReadParam('operation', null);
+
+// Can be a either a string or an array
 $sJsonString = utils::ReadParam('json_data', null, false, 'raw_data');
 $sProvider = '';
 
--- a/webservices/rest.php
+++ b/webservices/rest.php
@@ -128,7 +128,17 @@ try
 		throw new Exception("Missing parameter 'version' (e.g. '1.0')", RestResult::MISSING_VERSION);
 	}
 	
-	if ($sJsonString == null)
+	if ($sJsonString === null)
+	{
+		$aNonJsonDataKeysList = array('auth_user', 'auth_pwd');
+		$sJsonString = $_REQUEST;
+		foreach ($aNonJsonDataKeysList as $sNonJsonDataKey)
+		{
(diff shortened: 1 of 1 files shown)


webservices/rest.php
```diff
@@ -126,7 +128,17 @@ function json_last_error_msg() {
 		throw new Exception("Missing parameter 'version' (e.g. '1.0')", RestResult::MISSING_VERSION);
 	}
 	
-	if ($sJsonString == null)
+	if ($sJsonString === null)
+	{
+		$aNonJsonDataKeysList = array('auth_user', 'auth_pwd');
```
Hipska: I think `version` and `callback` need to be added as well.

piRGoif: You're perfectly right, thanks !
Fixed !

Hipska (approved):

piRGoif:
Thanks @Hipska for the review !
I'll present this to the dev team on friday, and we'll see from there...

piRGoif:
Combodo bug reference : N°2767

piRGoif:
Hello,

The dev team reviewed this, and there were a majority of votes against this implementation for the following reasons : 

* having the json_data parameters at the same level as the API is quite dangerous (collisions in the future), 
* also considering that any verb can be added though a `iRestServiceProvider` implementation... there are a lot of parameters we can't have the knowledge of

So to speak, the dev team doesn't want to support this third way of transmitting parameters to the REST API.

I'm closing this PR.

PR #131, opened 2020-03-31
N°2901 - Add inline images trace logs

bruno-ds:
This PR propose new log entries (cf N°2901 / R-024611).

they are written only if
```php
	'log_level_min' => array(
		'InlineImage' => LogAPI::LEVEL_TRACE,
		'UserRequest' => LogAPI::LEVEL_TRACE,
	),
```
You can then monitor the logs using something like : `tail -f log/error.log | grep "| InlineImage\|| UserRequest" -A 20`

One drawback of the log I've written are that the UserRequest logs are only written for some operations of UI.php, maybe I should add less detailed information within DBObject, What do you think?

For iTop <2.7, the code must be altered in order to remove the channel & move the extra info within the message string. (thus, it is not filterable, and must be used as a separate temporary build activating more functionalities 😭

### DIFF
--- a/core/inlineimage.class.inc.php
+++ b/core/inlineimage.class.inc.php
@@ -191,7 +191,7 @@ class InlineImage extends DBObject
 				'$sTempId' => $sTempId,
 				'$aInlineImagesId' => $aInlineImagesId,
 				'$sUser' => UserRights::GetUser(),
-				'HTTP_REFERER' => $_SERVER['HTTP_REFERER'],
+				'HTTP_REFERER' => @$_SERVER['HTTP_REFERER'],
 			));
 		}
 		else
@@ -200,7 +200,7 @@ class InlineImage extends DBObject
 				'$sObjectClass' => get_class($oObject),
 				'$sTransactionId' => $iTransactionId,
 				'$sUser' => UserRights::GetUser(),
-				'HTTP_REFERER' => $_SERVER['HTTP_REFERER'],
+				'HTTP_REFERER' => @$_SERVER['HTTP_REFERER'],
 			));
 		}
 
@@ -242,7 +242,7 @@ class InlineImage extends DBObject
 			'$sTempId' => $sTempId,
 			'$aInlineImagesId' => $aInlineImagesId,
 			'$sUser' => UserRights::GetUser(),
-			'HTTP_REFERER' => $_SERVER['HTTP_REFERER'],
+			'HTTP_REFERER' => @$_SERVER['HTTP_REFERER'],
 		));
 	}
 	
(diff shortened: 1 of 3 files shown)


pages/ajax.render.php
```diff
@@ -2650,6 +2660,16 @@ function(data){
...
+
+						IssueLog::Trace('InlineImage created', 'InlineImage', array(
+							'$operation' => $operation,
+							'$aResult' => $aResult,
+							'secret' => $oAttachment->Get('secret'),
+							'temp_id' => $sTempId,
+							'item_class' => $sObjClass,
+							'user' => $sUser = UserRights::GetUser(),
+							'HTTP_REFERER' => @$_SERVER['HTTP_REFERER'],
+						));
```
piRGoif, line 2672: Shouldn't this be put in the InlineImage class constructor ? Here we're only handling this single creation case...

core/inlineimage.class.inc.php
```diff
@@ -176,13 +176,35 @@ public static function FinalizeInlineImages(DBObject $oObject)
...
+			$aInlineImagesId = array();
 			while($oInlineImage = $oSet->Fetch())
 			{
+				$aInlineImagesId[] = $oInlineImage->GetKey();
 				$oInlineImage->SetItem($oObject);
 				$oInlineImage->Set('temp_id', '');
 				$oInlineImage->DBUpdate();
 			}
+
+			IssueLog::Trace('FinalizeInlineImages (see $aInlineImagesDebugTrace for a list of updated id)', 'InlineImage', array(
```
piRGoif, line 188: I was confused by the log message I found pretty vague... But the channel is printed also (see \FileLog::Write) so should not be a problem

pages/ajax.render.php
```diff
@@ -2690,6 +2710,15 @@ function(data){
 					$oAttachment->Set('contents', $oDoc);
 					$oAttachment->Set('secret', sprintf('%06x', mt_rand(0, 0xFFFFFF))); // something not easy to guess
 					$iAttId = $oAttachment->DBInsert();
+
+					IssueLog::Trace('InlineImage created', 'InlineImage', array(
```
piRGoif, line 2714: This will be logged whenever we're adding a new image in CKE, so we would have in the log almost all the cases we want. Cool !

piRGoif (approved):
A little hesitation on the InlineImage log messages.
But this in a whole is good, in my opinion !

piRGoif:
You might want to rebase on support/2.7 for this to be on the 2.7.1 version ?

pages/ajax.render.php
bruno-ds: I'm on my phone, so I can't verify, but if I remember well, the inline image constructor does not have access to all of the data passed to the context array

piRGoif:
The log were pushed for the next 2.7.1 version with 839bbc42

PR #134, opened 2020-04-08
Enable external classes to use bypass function as well

larhip:
Hey there,
with this PR I would like to add the possibility that implementations iApplicationObjectExtension interface could ask the given object whether to bypass user right checks or not. This offers the possibility to bypass some additional checks in an extension as well.
Hope it was okay, that I did not create a ticket before, like you wrote in CONTRIBUTING.md.
From my point of view this change was such small, that it was easier to touch it in the code than explaining it in a ticket.
Best
Lars

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -4091,6 +4091,16 @@ EOF
 	}
 
 	/**
+	 * Whether to bypass the checks of user rights when writing this object, could be used in {@link \iApplicationObjectExtension::OnCheckToWrite()}
+	 *
+	 * @return bool
+	 */
+	public function GetAllowWrite()
+	{
+		return $this->bAllowWrite;
+	}
+
+	/**
 	 * Bypass the check of the user rights when deleting this object
 	 *
 	 * @param bool $bAllow True to bypass the checks, false to restore the default behavior


Molkobain:
Hello @larhip ,

Could you give "real world" examples of its usage?

Thanks,
Guillaume

larhip:
HI @Molkobain ,

my "real world" goal is quite complex, but I will do my best to explain the usage as easy as possible:

The class Ticket is linked via a many2many relation with contacts (lnkContactToTicket). 
An implementation of iApplicationObjectExtension checks that not every user is allowed to link new contacts to the ticket (modify this list). Only a user linked to a contact, with special attributes, is allowed to do so.
But in some cases I want to add automatically contacts to this list (i.e. UpdateImpactedItems) in such a case I thought about using "AllowWrite" to bypass my own checks.
But the implementation of iApplicationObjectExtension is not able to ask for this attribute.

I real my use case is a lot more complex, but I hope this mini explanation gives you the possibility to follow my thoughts. I imagined that there could be other user cases as well.

But honestly speaking: After thinking about it a little more and having really trouble to find at least one simple "real world" example of its usage, I could also imagine that my needs are really to special to integrate it in the core of iTop.
Which means, if you are not able to follow my thoughts I am totally fine, if you just close this PR :-)

Best
Lars

bruno-ds:
Hello,
Reading from your explanation, I have the feeling that you want to add a setter in order to be able to force "AllowWrite", though you proposed a getter (and so you won't be able to change the boolean value).

Hipska:
No I really think he means a getter, as his code that uses the value is inside an iApplicationObjectExtension.

bruno-ds:
Ok, you want to be able to propagate the allowWrite call over the classes plugged to it through iApplicationObjectExtension and allow them to vary in there implementation regarding the value of allowWrite.

So basically, the DBObject would store a data that you set in a part of your code and read in one other part.
I'm thinking out loud: it may make sense if it is really closely related to the behavior of the DBObject (if the bypass of the security is performed both into the DBObject and your extension).

For me it make sense,  and I'm inclined to accept the (minor) change I'll let my colleague speak also.

bruno-ds:
@larhip Could you please respond to my previous comment, we want to be sure that we understood well your willing.

larhip:
Hi @bruno-ds ,
you described my willing in your previous comment  100 percent correctly, otherwise I would have already answered :-)
Sorry that I did not confirm your description earlier, anyway.
Best
Lars

bruno-ds (approved):

bruno-ds:
We just reviewed your PR, thank you for the contribution & sorry for the delay.

PR #135, opened 2020-04-16
Check changes attributes in TriggerOnPortalUpdate

larhip:
This PR adds a feature in TriggerOnPortalUpdate to check for changes attributes (similar to TriggerOnObjectUpdate) using the new method ListPreviousValuesForUpdatedAttributes

Previously I implemented my own Trigger with this feature via an extension, but I though it might be interesting to integrate this in the core and make it available to all iTop users.

Best
Lars

### DIFF
--- a/core/trigger.class.inc.php
+++ b/core/trigger.class.inc.php
@@ -229,6 +229,7 @@ abstract class TriggerOnObject extends Trigger
 	 *
 	 * @throws \ArchivedObjectException
 	 * @throws \CoreException
+	 * @throws \Exception
 	 */
 	public function DoActivate($aContextArgs)
 	{
@@ -237,7 +238,7 @@ abstract class TriggerOnObject extends Trigger
 		{
 			/** @var \DBObject $oObject */
 			$oObject = $aContextArgs['this->object()'];
-			$bGo = $this->IsTargetObject($oObject->GetKey(), $oObject->ListChanges());
+			$bGo = $this->IsTargetObject($oObject->GetKey(), $this->GetChanges($oObject));
 		}
 		if ($bGo)
 		{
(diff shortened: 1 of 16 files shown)


eespie:
Hi Lars,
If you want to have _TriggerOnObjectUpdate_ working only in the context of the portal you can use a new attribute named... _context_ to filter in which context you want your trigger to work.
If this answer if ok for you, just close the PR.

larhip:
Hi @eespie ,
I already mentioned the context attributes, but in fact it is sligthly different to this one.
Using _TriggerOnObjectUpdate_ I am not able to check the attributes **after** the object modification. _TriggerOnObjectUpdate_ is the only trigger checking the OQL filter before the object modification, which makes the handling a little bit different.
I.e. if some one is adding a contact to the contacts_list via portal and then writes a public log entry. I am not able to send this log entry to the newly added contact using _TriggerOnObjectUpdate_
Using the implementation in this PR I am able to do so.

Best
Lars

bruno-ds:
Hi Lars,
we won't be able to work on your PR for some months.

It's not that it's good (or bad), it's just that we have some hard deadlines to meet.

We are really sorry for the delay, 💔

Molkobain:
Discussed during technical review today: We don't feel comfortable with relying on the assumption that the `TriggerOnPortalUpdate` is later in the processing and therefore gives you access to changed attributes. This positioning in the CRUD could be moved.

On the other hand, in the portal you can specify which triggers you want to be activated, maybe this could save you as you could do an OQL like "SELECT TriggerOnPortalUpdate UNION SELECT TriggerOnObjectUpdate WHERE context MATCHES portal" (it's pseudo OQL). Can you check if that would do the job for you?

Check [this page](https://www.itophub.io/wiki/page?id=3_0_0:customization:portal_xml) for more info.
image

larhip:
Hey @Molkobain 

This PR is pretty old (about 3 years). As far as I am able to remember, that time the motivation for this PR was, that 

- There was no Context-Filter for TriggerOnUpdate, and
- TriggerOnUpdate was activated _before_ object was written

Both issues are not valid anymore., which means the need for this PR might be obsolete. I forgot to update this PR / close this PR. 

Cheers
Lars

Molkobain:
Thanks for the feedback!

PR #136, opened 2020-04-22
Faster compilation of themes

dflaven:
The SCSS compilation of a theme takes quite a long time during the setup (or when running the toolkit). This pull request introduces a new tag <precompiled_stylesheet> under the "theme" node, to specify the path (relative to the module's root dir) to a precompiled CSS file.

To be effective, CSS precompiled files must contain a signature (generated during the compilation of the theme). The compiler is then able to check that the precompiled theme is still up-to-date based on the signature which covers the imports and stylesheets files as well as the variables.

On my system the compilation of the two standard themes used to take about 14.5 s, whereas the signature check takes a little bit less than 10ms (compared to about 1.5 ms when just checking the timestamps of the files).


test/application/ThemeHandlerTest.php
```diff
@@ -367,48 +376,22 @@ public function testCompileThemes($ThemeParametersJson, $CompileCSSFromSASSCount
...
-			2 => array("pipe", "w"),  // stderr
-		);
-		$process = proc_open($cmd, $descriptorspec, $pipes, $workdir, null);
-
-		$stdout = stream_get_contents($pipes[1]);
-		fclose($pipes[1]);
-
-		$stderr = stream_get_contents($pipes[2]);
-		fclose($pipes[2]);
+	public function testToto()
```
bruno-ds: Great name 🤪

test/application/ThemeHandlerTest.php
```diff
@@ -0,0 +1,580 @@
...
+				else {
+					copy($src . '/' . $file,$dst . '/' . $file);
+				}
+			}
+		}
+		closedir($dir);
+	}
+
+	/**
+	 * Test used to be notified by CI when precompiled styles are not up to date anymore in code repository.
```
bruno-ds, line 95: Great idea!

PR #137, opened 2020-04-26
Add Russian translations for 2.7.0-1

vbkunin:


Molkobain:
Hello @vbkunin ,

Thanks for thé update on the translations! Can you change the target branch to support/2.7 instead of develop? This way is will be part of the next 2.7.x, otherwise it will be for the 2.8.

Thanks
Guillaume

PR #147, opened 2020-06-23
Add "software requirements" section to README

Molkobain:
At first I added a complete section with a table and all currently supported versions of iTop, then @bruno-ds pointed out that it would only be redundant with the wiki page and just another source of information to maintain (which is totally true, thanks!). So instead it's just a simple link in the "resources" section.

Below is the original table, just because I learned how to make them in GitHub 😁

## Software requirements
| iTop version | PHP versions | MySQL versions | MariaDB versions | Chrome support | Firefox support | Edge support | IE support | Limitations |
|:------------:|:-------------|:---------------|:-----------------|:--------------:|:---------------:|:------------:|:----------:|:---|
| 2.8.x | **7.1-7.4** | **5.7**-8.0 | **10.3+** | ✔ | ✔ | ✔ | ❌       | - |
| 2.7.x | 5.6-7.3     | 5.6-**8.0** | 10.1+     | ✔ | ✔ | ✔ | **IE10+** | Add MySQL 8 support, but not recommended for large database |
| 2.6.1 | 5.6-**7.3** | 5.6-5.7     | 10.1+     | ✔ | ✔ | ✔ | IE9+      | Add PHP 7.3 support |
| 2.6.0 | 5.6-7.2     | 5.6-5.7     | 10.1+     | ✔ | ✔ | ✔ | IE9+      | - |

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,16 +26,18 @@ iTop also offers mass import tools and web services to integrate with your IT
  - [iTop Forums][1]: community support
  - [iTop Tickets][2]: for feature requests and bug reports
  - [Releases download][3]
- - [Documentation][4] covering both iTop and its official extensions
- - [iTop Hub][5] : discover and install extensions !
+ - [Software requirements][4]
+ - [Documentation][5] covering both iTop and its official extensions
+ - [iTop Hub][6] : discover and install extensions !
 
 
 
 [1]: https://sourceforge.net/p/itop/discussion/
 [2]: https://sourceforge.net/p/itop/tickets/
 [3]: https://sourceforge.net/projects/itop/files/itop/
-[4]: https://www.itophub.io/wiki
-[5]: https://store.itophub.io/en_US/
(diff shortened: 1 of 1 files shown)


piRGoif (approved):

Molkobain:
3 thumbs up and 1 discussion with Bruno, I'm merging it.

Hipska:
The table looked nicer in the email notification than it does here on the web.

Oh, and the contents of it are also very nice. I don't think the wiki has such a good comparison?

Molkobain:
@Hipska the rendering was also slightly better in the README than in the PR as well.

In the wiki, I think we could do some enhancements, what did you liked?
- Browser columns
- Min-Max versions for softwares

I can talk about it with @v-dumas

Hipska:
The fact everything is visible in one blink of the eye. For me, the software are more important than the browsers, but both are interesting..

PR #148, opened 2020-07-04
fix: DB schema check, MySQL 8 int and tinyint ignore length

annProg:
If use MySQL 8，toolkit db schema check will report:

- `found int DEFAULT 0 while expecting INT(11) DEFAULT 0`
- `found tinyint DEFAULT 0 while expecting TINYINT(1) DEFAULT 0`
- `varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT while expecting VARCHAR(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT '\''`

this pr try to fix it.

### DIFF
--- a/core/cmdbsource.class.inc.php
+++ b/core/cmdbsource.class.inc.php
@@ -605,7 +605,7 @@ class CMDBSource
 	 */
 	private static function RemoveSurroundingQuotes($sValue)
 	{
-		if (utils::StartsWith($sValue, '\'') && utils::EndsWith($sValue, '\''))
+		if ($sValue != "'" && utils::StartsWith($sValue, '\'') && utils::EndsWith($sValue, '\''))
 		{
 			$sValue = substr($sValue, 1, -1);
 		}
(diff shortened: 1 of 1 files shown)


piRGoif:
Many thanks for the feedback !

I can see that the int & tinyint size not present anymore was added in MySQL 8.0.17 (as stated in the [8.0.17 released notes](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-17.html#mysqld-8-0-17-deprecation-removal))

In the MySQL 8 ref there is a chapter dedicated to the changes, a sort of cumulative changelog said otherwise : [What Is New in MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html#mysql-nutshell-deprecations)
I see there are lots of changes and we might not handle them all... Something on the float and double might generate useless ALTER TABLE queries as well...

We won't fix this as you did (we might better find a way to generate a SQL for each DB we have to deal with ?), so I'm closing this PR. But I've added two tickets in the (private) Combodo system :
N°3168 to fix int/tinyint/double/float issues
N°3167 to study if we have missed any other MySQL 8 change

Thanks again !

PR #149, opened 2020-07-04
fix: Typology translation for zh_cn

annProg:
**Typology** means **类型**
**Topology** means **拓扑**

The previous translator may have confused these two words


Molkobain:
Thanks @annProg !

Molkobain (approved):

PR #150, opened 2020-07-08
Generate lifecycle as SVG

Hipska:
Here a request to change the Graphviz generation of a lifecycle to output SVG instead of PNG.

Pro:
- Smaller image size (Roughly 10x smaller files)
- Easier readable when there are a lot of transitions
- Reduced php memory consumption by using `readfile` instead of `echo file_get_contents`.

Con:
- Might not be supported by old browsers

### DIFF
--- a/pages/graphviz.php
+++ b/pages/graphviz.php
@@ -143,6 +143,7 @@ if (file_exists($sDotExecutable))
 	else
 	{
 		header('Content-type: image/svg+xml');
+		header('Content-Disposition: inline; filename="'.$sClass.'.svg"');
 		readfile($sImageFilePath);
 	}
 	@unlink($sDotFilePath);
@@ -150,6 +151,7 @@ if (file_exists($sDotExecutable))
 else
 {
 	header('Content-type: image/png');
+	header('Content-Disposition: inline; filename="'.$sClass.'.png"');
 	readfile($sImageFilePath);
 }
 
--- a/pages/graphviz.php
+++ b/pages/graphviz.php
@@ -143,14 +143,14 @@ if (file_exists($sDotExecutable))
 	else
 	{
 		header('Content-type: image/png');
-		echo file_get_contents($sImageFilePath);
+		readfile($sImageFilePath);
 	}
 	@unlink($sDotFilePath);
 }
 else
 {
 	header('Content-type: image/png');
-	echo file_get_contents($sImageFilePath);
+	readfile($sImageFilePath);
 }
 
 ?>
--- a/pages/graphviz.php
+++ b/pages/graphviz.php
@@ -112,7 +112,7 @@ $sDotExecutable = MetaModel::GetConfig()->Get('graphviz_path');
 if (file_exists($sDotExecutable))
 {
 	// create the file with Graphviz
-	$sImageFilePath = APPROOT."data/lifecycle/".$sClass.".png";
+	$sImageFilePath = APPROOT."data/lifecycle/".$sClass.".svg";
 	if (!is_dir(APPROOT."data"))
 	{
 		@mkdir(APPROOT."data");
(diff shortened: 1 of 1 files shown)


pages/graphviz.php
```diff
@@ -142,15 +142,15 @@ function GraphvizLifecycle($sClass)
 	}
 	else
 	{
-		header('Content-type: image/png');
-		echo file_get_contents($sImageFilePath);
+		header('Content-type: image/svg+xml');
+		readfile($sImageFilePath);
 	}
 	@unlink($sDotFilePath);
```
Hipska, line 149: PS, why is the dot file being removed and the generated image isn't?

Shouldn't both be removed? Or keep them both?

Molkobain: Hello Thomas,

Out of the box there are already some pre-generated lifecycle images for the existing classes (UserRequest / Incident). They are used as a fallback if you don't have GraphViz installed and a standard DM. That's why the file isn't deleted I think, otherwise it would only work the first time.

That being said, it would be clearer if the fallback image was in a dedicated variable. Also if the image isn't deleted afterwards, we should check before generating it again and reuse it.

Molkobain:
> Might not be supported by old browsers

As of iTop 2.8, IE will no longer be supported. Only Chrome, Firefox, Edge, so I think it will be ok 👍

pages/graphviz.php
Hipska: I have seen the fallback image, but that one is on a different location (in the module) as well. I'm clearly only talking about the generated image only.

Hipska:
I also looked at replacing Graphviz at all, other tools provide other functionality: 
[Simple UserRequest livecycle](https://www.planttext.com/?text=ZLAxRiCm3Dpr5Vm1_u278a7c14NQJa46GM8SePPKY7Jw-wKjF-Ga2BfHlECTAN93h1DtlbKjNXWuGdBDbS6wX8PT34gf0_v2LRt1biWr0ItKOvYHazLqHNiMnX4fjZUqGkWuUitEGCevDLVtKkyOh0kDA8sHcZ9L8IFmNgTlq3Ifs8Ntn1QU-PR2ZnEikdhbkn3_QxoIBfoFgBFtNFr0gE8Ftk6xDjA03e6QrMluHEsBvxR1m6eQU_vWQKjJlyvbHqFKzKi64oqMoCUOBrEORiw8SMwaXY_212GJqqNdIraX9de_DB8HVxclFm00)

Molkobain:
Did you tested on Chrome / Firefox / Edge? Besides this I don't see any reason not to accept the PR.

Hipska:
Tested on Chrome, Firefox and Safari.

Molkobain:
Thanks, it'll be reviewed technically by the dev. team tomorrow, if everything is fine we'll explain the benefits to the product team in order to get their approval.

pages/graphviz.php
Hipska: @Molkobain ?

Molkobain: Hello @Hipska , are you replying about the thread or the status of the PR?

Hipska: This actual thread. I'm still looking for an answer to my initial question from 8 days ago.

Molkobain: I think it was for some caching purpose but it's clearly not using it well as the image is re-generated every time. As Denis is on vacation I can't ask him what the intention was.

Molkobain (approved):

Molkobain:
@Hipska R&D and Product teams gave a GO as well, can we merge the PR or do you still have some adjustments to do?

Hipska:
Thanks, on my side it is ready!

Molkobain:
Done, thanks !

pages/graphviz.php
Hipska: Can you try again now?

Molkobain: @dflaven any input on this?

Hipska: Can you contact @dflaven directly internally?

Hipska: Any response from @dflaven ?

Molkobain: Hello @Hipska , just talked with him IRL. It seems to be a mistake from the past, you can fix it 😁

Hipska: Keep both? Or remove both?

Molkobain: Remove both

Hipska: This seems not to be done yet, could you quickly do it in a commit? Or do I need to create a new branch and PR for that?

```suggestion
	@unlink($sDotFilePath);
	@unlink($sImageFilePath);
```

Molkobain: I thought you would make a PR, but sure just did it in 7613ac7 😊

PR #151, opened 2020-07-10
N°3171 - Friendly name and obsolescence flag not refreshed

rquetiez:
- Compute any type of expression on server side
- Recompute friendly name and obsolescence flag on server side (DBOBject)
- Bonus : compute dependency for external keys

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -6659,7 +6659,13 @@ class AttributeExternalKey extends AttributeDBFieldVoid
 		$oExpression = DBSearch::FromOQL($this->GetValuesDef()->GetFilterExpression())->GetCriteria();
 		foreach ($oExpression->GetParameters('this') as $sAttCode)
 		{
-			$aAttributes[] = $sAttCode;
+			// Skip the id as it cannot change anyway
+			if ($sAttCode =='id') continue;
+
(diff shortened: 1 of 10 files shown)


Molkobain (approved):
Blind approval

bruno-ds:
I don't see any doc referring to this new feature (neither in the changelog nor in the functional doc.).

Don't forget it ;)

Hipska:
There even isn’t any description what this PR is for…

But looking at the title it looks like this is a bugfix instead of a new feature?

piRGoif:
Hello,
The doc is in the merged commit... I've updated the description with this text.

Hipska:
Sorry, it is still not clear what the goal of this PR is. Is it a new feature? Is it a bugfix? It has something to do with the calculation of friendlyname, but apart from that not much info..

bruno-ds:
The description lack some words: 

This is a new feature :
> Compute any type of expression on server side

This feature leverage for example the possibility to 
> Recompute friendly name and obsolescence flag on server side (DBOBject)

(before this PR, a refresh from the databse was required)


as a bonus, this PR also add the possibility to automatically
> compute dependency for external keys




In other words : before this PR, the expression had only one way to be evaluated : be converted to an SQL statement sent to the MySQL server which evaluated it and returned the result. Now, PHP itself can evaluate the expression 

is it better?

PR #152, opened 2020-07-16
Fix AttributeDecimal to be empty in synchro

Hipska:
### Problem
When having fields of type AttributeDecimal (or AttributeInteger, ..) and you want to do a data synchronisation with them, it is not possible to synchronise when there is no value. MySQL will complain with:
> Incorrect decimal value: '' for column 'decimal_field' at row 1
### Solution
Fix the query building process to not generate `''` as value for these fields, but use `NULL` instead. There was an `if` clause intending to do that, but never got used.

### DIFF
--- a/synchro/synchro_import.php
+++ b/synchro/synchro_import.php
@@ -534,7 +534,7 @@ try
 				$aValues = array(); // Used to build the insert query
 				foreach ($aRow as $iCol => $value)
 				{
-					if ($value === null)
+					if (!strlen($value))
 					{
 						$aValues[] = 'NULL';
 					}
@@ -604,7 +604,11 @@ try
 					}
 
 					$sCol = $aInputColumns[$iCol];
-					if ($aIsDateToTransform[$iCol] !== false)
+					if(!strlen($value))
+					{
+						$aValuePairs[] = "`$sCol` = NULL";
+					}
+					elseif ($aIsDateToTransform[$iCol] !== false)
 					{
 						$bDateOnly = false;
 						$sFormat = $sDateTimeFormat;


piRGoif:
Hi Thomas,
Can you give some step by step examples to reproduce this problem with the default datamodel ?

Also, this code part is not currently covered by automatic testing, adding such tests would be I think mandatory before modifying this code... But it doesn't seem to be very testable yet :/

Hipska:
This came from discussion with @etaloc. He knows more about it.

1. Pick any class with a decimal or integer attribute.
2. Create a data synchro for it
3. Upload csv with empty values for the decimal or integer field to this synchro
4. See errors on tis field

piRGoif:
Ok thanks I'll have a chat with him !
This is still not ready for R&D review in my opinion.

piRGoif:
Here is a sample XML to create an object with an AttributeDecimal attribute :

```xml
	<classes>
		<class id="Pr152" _delta="define">
			<parent>cmdbAbstractObject</parent>
			<properties>
				<is_link>0</is_link>
				<category>bizmodel</category>
				<abstract>false</abstract>
				<key_type>autoincrement</key_type>
				<db_table>pr152</db_table>
				<db_key_field>id</db_key_field>
				<db_final_class_field/>
				<display_template/>
				<icon/>
				<naming>
					<attributes>
						<attribute id="title"/>
					</attributes>
				</naming>
				<reconciliation>
					<attributes>
						<attribute id="title"/>
					</attributes>
				</reconciliation>
			</properties>
			<fields>
				<field id="title" xsi:type="AttributeString">
					<sql>title</sql>
					<default_value></default_value>
					<is_null_allowed>false</is_null_allowed>
				</field>
				<field id="mydecimal" xsi:type="AttributeDecimal">
					<sql>mydecimal</sql>
					<default_value>0</default_value>
					<digits>4</digits>
					<decimals>2</decimals>
				</field>
			</fields>
			<methods/>
			<presentation>
				<details>
					<items>
						<item id="title">
							<rank>10</rank>
						</item>
						<item id="mydecimal">
							<rank>20</rank>
						</item>
					</items>
				</details>
				<search>
					<items>
						<item id="title">
							<rank>10</rank>
						</item>
						<item id="mydecimal">
							<rank>20</rank>
						</item>
					</items>
				</search>
				<list>
					<items>
						<item id="title">
							<rank>10</rank>
						</item>
						<item id="mydecimal">
							<rank>20</rank>
						</item>
					</items>
				</list>
			</presentation>
		</class>
	</classes>
</itop_design>
```

And a CSV to reproduce the problem :

```csv
"primary_key";"title";"mydecimal"
"1";"NoValue";
```

Before patch, executing synchro/synchro_import.php throw this error :

```
#MySQLException: Failed to issue SQL query: query = INSERT INTO `synchro_data_pr152_1` (`primary_key`, `title`, `mydecimal`) VALUES ('1', 'NoValue', ''), mysql_errno = 1366, mysql_error = Incorrect decimal value: '' for column 'mydecimal' at row 1 in C:\Dev\wamp64\www\itop-dev\core\cmdbsource.class.inc.php:689
Stack trace:
#0 C:\Dev\wamp64\www\itop-dev\core\cmdbsource.class.inc.php(648): CMDBSource::DBQuery('INSERT INTO `sy...')
#1 C:\Dev\wamp64\www\itop-dev\synchro\synchro_import.php(573): CMDBSource::Query('INSERT INTO `sy...')
#2 {main}
```

After applying patch, no error anymore. But I get a new object with the "0.00" value instead of blank (an the same value in the replica). Saving the object with no value doesn't do the same : after saving you have a blank value in the field and not "0.00".

I'm adding this to the PR to review by the R&D team list, we'll keep you posted.

Hipska:
> After applying patch, no error anymore. But I get a new object with the "0.00" value instead of blank (an the same value in the replica). Saving the object with no value doesn't do the same : after saving you have a blank value in the field and not "0.00".

So you are saying that there's something else that needs to be fixed as well?

piRGoif:
Hello,

> > After applying patch, no error anymore. But I get a new object with the "0.00" value instead of blank
> 
> So you are saying that there's something else that needs to be fixed as well?

Yes, on my configuration I didn't get a blank value as expected...

Hipska:
Strange, in my setup(s) this didn't happen and the field is empty as expected. I Just double checked..

synchro/synchro_import.php
```diff
@@ -534,7 +534,7 @@ public function add_comment($sText)
 				$aValues = array(); // Used to build the insert query
 				foreach ($aRow as $iCol => $value)
 				{
-					if ($value === null)
+					if (!strlen($value))
 					{
 						$aValues[] = 'NULL';
```
Hipska, line 539: This line never got reached because data from a csv file field will never be null.

piRGoif:
All of my test conditions are explained in my previous comment (https://github.com/Combodo/iTop/pull/152#issuecomment-661738955)
Can you try with those et see if this happens on your side ?

Hipska:
Ah, but in there you put `0` as default value, then it is normal to have `0.00` as result when putting nothing (`null`) into the field.

piRGoif:
> Ah, but in there you put `0` as default value, then it is normal to have `0.00` as result when putting nothing (`null`) into the field.

Oh good catch, I feel shamy, sorry for this obvious mistake !
I've done the test once again with no default value (that means `<default_value/>`in the XML for the mydecimal field), and no problem, the object is created as it should, with a blank value in the decimal field.
Sorry again.

The dev team will review this !

synchro/synchro_import.php
rquetiez: Good catch: the code could not be reached, but...
Unfortunately, an empty string in the CSV means "I want to reset the attribute value" whereas null in the data table means "I want to leave this attribute unchanged".
Yes, the documentation (see the tip at the end of [this chapter](https://www.itophub.io/wiki/page?id=2_7_0%3Aadvancedtopics%3Adata_synchronization#description)) is unclear about that... and only date attributes can be reset as of now (Change made in iTop 2.4, see https://github.com/Combodo/iTop/commit/e41a883).
The point has already been submitted and is **not** yet planned for the next major release.
The solution should adress Integer, Decimal and Enum types of attributes.

Anyhow, I would suggest a quick and dirty workaround : tune MySQL STRIC_MODE might help getting rid of the error... does it reset the value ?

And the dead code could be removed!
Sorry for having left so many traps in there!

Hipska: Does this mean my PR will not be accepted? Please discuss in the main thread then..

rquetiez: I was wrong about the dead code.
The initial code could be reached: `"<NULL>"` in the CSV input will be interpreted as `null` by the CSV Parser.
Anyhow, the initial code would set `"NULL"` (string literal) in the data table. And unfortunately it will fail for columns that are not in the VARCHAR format.

Decision: We will interpret the literal `"<NULL>"` as "_I want to ignore this column for that row_".

rquetiez: Done in a1f5d80

rquetiez:
1. A blank value should be intepreted as a wish to reset the attribute value (this is intuitively the case for string attributes).
2. <NULL> will be interpreted as a wish to leave the attribute unchanged

1 requires more work. It is the continuation of e41a883 and is planned for iTop 2.9
2 has just been implemented in a1f5d80.

Hipska:
These assumptions do not compute very well since you specify in the datasource what will be done to the attribute (overwrite, ignore or set if empty)

synchro/synchro_import.php
Hipska: > Anyhow, the initial code would set "NULL" (string literal) in the data table. And unfortunately it will fail for columns that are not in the VARCHAR format.

That is not correct. The resulting SQL would be ```INSERT INTO `synchro` (field) VALUES(NULL)```.

Hipska:
Okay, I think this will be the solution for this problem then: https://github.com/Combodo/iTop/commit/a1f5d8041e64e64c3ea68a46963b725f7278c0c7#commitcomment-41808635

Only sad this could not be discussed within this PR.

PR #155, opened 2020-08-14
N°3253 Fix setup crashing on incompatible PHP versions

piRGoif:
## Context

A couple of time after adding some PHP code in iTop core, we saw that the setup cannot be launched anymore when using old PHP versions. 

Some past examples :

* adding a finally block in \MetaModel::GetObjectWithArchive (commit 0adadeb2 in iTop 2.6.0) made iTop setup incompatible with PHP < 5.5.0
* adding constant visibility in `ormLinkSet` (commit b58a084d in iTop 2.8.0 dev version) broke setup for PHP < 7.1.0

## Setup homepage error

Sample error when opening the setup with an incompatible PHP version :

parse error on setup homepage

🤯

We should instead display a clear message stating that the current PHP version is incompatible with iTop !

## Risks

* This is something we don't test for now (we are aiming to do so but this is complicated as docker images with such old PHP version are hard to find/maintain !)
* Next major iTop version will update min PHP version from 5.6.0 to 7.1.3, and as this is directly configured in dev team PHPStorm's we will introduce more and more incompatible code in iTop core !

## Proposition

This PR propose a lightweight setup index that will first check PHP version. This new index page includes the bare minimum files to avoid PHP parse errors.

If the test succeeds the setup is launched, without changing the browser's URL.
If not an error message is displayed.

## Refactorings

Two new files were created, they are the only dependencies for the new setup/index.php : 

* setup/setuputilslight.class.php
* setup/setupconst.class.php

The below constants/methods were moved/introduced :

* SetupUtils const for PHP / MySQL versions
* SetupUtils::CheckPhpVersion method and CheckResult object
* WebPage::PAGES_CHARSET
* \WebPageLight::EmbedSetupPageContent
* ITOP_* const that were defined in core/config.class.inc.php


application/xmlpage.class.inc.php
```diff
@@ -42,7 +42,7 @@ function __construct($s_title, $bPassThrough = false)
 		parent::__construct($s_title);
 		$this->m_bPassThrough = $bPassThrough;
 		$this->m_bHeaderSent = false;
-		$this->add_header("Content-type: text/xml; charset=".self::PAGES_CHARSET);
```
dflaven, line 45: Je mélangerai pas le charset des pages web et celui des pages XML... le XML est quasiment toujours en UTF-8, ça n'a rien à voir avec les pages web je pense.

core/ormlinkset.class.inc.php
```diff
@@ -30,8 +30,8 @@
 
 class ormLinkSet implements iDBObjectSetIterator, Iterator, SeekableIterator
 {
-	const LINK_ALIAS = 'Link';
-	const REMOTE_ALIAS = 'Remote';
+	public const LINK_ALIAS = 'Link';
+	public const REMOTE_ALIAS = 'Remote';
```
dflaven, line 34: C'est quoi le rapport de cette modif avec le setup ??

dflaven (commented):
J'aime bien l'idée de protéger le setup contre un problème de PHP, mais est-ce qu'on peut pas faire un truc plus simple et qui permettrait aussi de se protéger contre une installation où PHP ne fonctionne même pas? Je pensais à une page index.php qui soit essentiellement du HTML avec juste la détection de version de PHP qui produirait un markup qu'on vérifie en JS. Si c'est bon, on passe à la page du setup proprement-dit, sinon on affiche (toujours en JS) un message indiquant que PHP n'est pas configuré ou n'est pas la bonne version...

core/ormlinkset.class.inc.php
piRGoif: This is the latest modification that broke the setup (again) : this is the one I'm talking about in the second point in the "context" chapter of this PR description.
The first commit in this PR restores this so we can reproduce the problem.

application/xmlpage.class.inc.php
piRGoif: The same constant was already used in a lots of place, including XMLPage. As I've moved this constant to another file, XMLPage is affected too.
I've made atomic commits in the branch, the one containing this modification is 0f9bc5d0c18f5845ab8147921484482a0aced1c3

piRGoif:
Thanks Denis for the feedback, your solution is simplier indeed : a simple index that generated a string in PHP, and controlling what was generated server side by a JS code.

Implementing this we shouldn't forget to check also PHP version (maybe generate a `php -v` server side ?) as the root of the problem was parse errors generated on the setup homepage.
I like that during the setup the URL displayed by the browser don't change, it isn't mandatory but I think we should keep it this way... Might be hard to implement in your proposed solution ?

bruno-ds:
Php expose it's version through this collection of constants:
https://www.php.net/manual/fr/reserved.constants.php

Ie: PHP_VERSION

piRGoif:
As discussed with @rquetiez, we decided to go for a different solution !
I'm closing this and opening another PR : #177

PR #158, opened 2020-08-21
add boilerplate function and use it to intercept/enrich trigger exception loops…

odain-cbd:
we had difficulties to identify why a notification on timeout event was not sent. we focused on the wrong trigger. another one trigger a PHP exception and lost us. it took time to find the root cause of such a ticket. 

I propose to intercept the exception and enrich it with the ticket classname and its id. 

boilerplate function can be used at other places of the code....


application/utils.inc.php
```diff
@@ -2316,4 +2316,27 @@ public static function ToCamelCase($sInput)
...
+			throw $oException;
+		}
+
+		$sMessage = $oException->getMessage() . " (" . get_class($oCmdbAbstract) . "::" . $oCmdbAbstract->Get('id') . ")";
+		$reflectedObject = new \ReflectionClass(get_class($oException));
+		$property = $reflectedObject->getProperty('message');
+		$property->setAccessible(true);
+		$property->setValue($oException, $sMessage);
+		$property->setAccessible(false);
+		throw $oException;
```
piRGoif: This seems quite weird
Why not encapsulate exception as this is commonly done ? 
Having for example a TriggerException and doing something like :
`return new TriggerException($oTrigger, $oException);`

odain-cbd: i agree with you. that's what i initially intended to do (encapsulating the original exception). 

but i'd prefer to keep the original exception to avoid to generate any regression in iTop core. somewhere we may rely on a specific kind of exception to do specific additional treatment...

last, we agreed with quality team to keep that PR in next 2.7.2. only modifying exception message attribute is safer.

odain-cbd: as discussed with Eric we will keep this change as is for the below reasons.

eespie (dismissed):

application/utils.inc.php
piRGoif: Depends on what catch we are talking about O:)
Triggers are executed by iTop Core, so we should identify which catch are triggered. I bet they are `catch(Exception $e)` ? If so, throwing a typed expression won't affect iTop, and it will be far better for future uses (a specific catch when launching triggers).

We might need to discuss this directly ?

application/utils.inc.php
```diff
@@ -2316,4 +2316,27 @@ public static function ToCamelCase($sInput)
...
+	/**
+	 * @param \cmdbAbstractObject $oCmdbAbstract
+	 * @param \Exception $oException
+	 *
+	 * @throws \Exception
+	 */
+	public static function EnrichRaisedException($oCmdbAbstract, $oException)
+	{
+		if (is_null($oCmdbAbstract) ||
+			! is_a($oCmdbAbstract, 'cmdbAbstractObject'))
```
Hipska:
```suggestion
			! is_a($oCmdbAbstract, \cmdbAbstractObject::class))
```

application/utils.inc.php
Hipska: And what about making `Trigger::DoActivate` to not throw exceptions? But return a true/false on success? 

It seems it is always used inside a loop.

application/utils.inc.php
piRGoif: Thanks Hipska, this is indeed better for IDEs !
And this is compatible PHP 5.5 : https://www.php.net/manual/fr/language.oop5.basic.php#language.oop5.basic.class.class
And was already used in iTop

application/utils.inc.php
piRGoif: As trigger are written in modules, they can throw any exception, we don't have any control on this :/
This PR aims to add some traces that helps when doing support, but without changing any behavior so that we can commit to a support branch (warning @odain-cbd you're based on develop by the way !!). We will certainly improves trigger exception catching in the next release (develop branch), but that will be done apart.

piRGoif: Here is the catch I found !

Using the [Crud stack documentation](https://www.itophub.io/wiki/page?id=latest%3Acustomization%3Asequence_crud), we can see that triggers are launched by DBObject::DBInsertNoReload and DBObject::DBUpdate.

Here are the call stack we doing an insert or an update from the console UI :

```
UI.php operation=apply_new
  catch(CoreException $e)
  catch(Exception $e)
    catch (CoreCannotSaveObjectException $e)
      cmdbAbstractObject::DBInsert
        cmdbAbstractObject::DBInsertNoReload
          DBObject::DBInsertNoReload

UI.php operation=apply_modify
  catch(CoreException $e)
  catch(Exception $e)
    catch (CoreCannotSaveObjectException $e)
    catch (DeleteException $e)
      cmdbAbstractObject::DBUpdate
        CMDBObject::DBUpdate
          DBObject::DBUpdate
```

The portal does it using this stack :

```
\Combodo\iTop\Portal\Form\ObjectFormManager::OnSubmit
  \DBObject::DBWrite
    \DBObject::DBInsert // if m_bIsInDB
    \DBObject::DBUpdate // otherwise
```

There are no catch here at all.

So as a conclusion there are no place where we have very specific catch, and we can use exception encapsulation.

Hipska: Looks much better now!

core/coreexception.class.inc.php
```diff
@@ -28,7 +28,7 @@
 
 class CoreException extends Exception
 {
-	public function __construct($sIssue, $aContextData = null, $sImpact = '')
+	public function __construct($sIssue, $aContextData = null, $sImpact = '', Throwable $previous = null)
```
Hipska:
```suggestion
	public function __construct($sIssue, $aContextData = null, $sImpact = '', Throwable $oPrevious = null)
```

core/coreexception.class.inc.php
```diff
@@ -66,7 +66,7 @@ public function __construct($sIssue, $aContextData = null, $sImpact = '')
 			}
 			$sMessage .= implode(', ', $aContextItems);
 		}
-		parent::__construct($sMessage, 0);
+		parent::__construct($sMessage, 0, $previous);
```
Hipska:
```suggestion
		parent::__construct($sMessage, 0, $oPrevious);
```

application/utils.inc.php
odain-cbd: thanks. yes code seems nicer. but stack trace seems worst to me now. as we loose the root cause of the original exception raised. this is not final code.

application/utils.inc.php
```diff
@@ -2316,4 +2316,24 @@ public static function ToCamelCase($sInput)
...
+	 *
+	 * @throws \Exception
+	 */
+	public static function EnrichRaisedException($oCmdbAbstract, $oException)
+	{
+		if (is_null($oCmdbAbstract) ||
+			! is_a($oCmdbAbstract, \cmdbAbstractObject::class))
+		{
+			throw $oException;
+		}
```
Hipska, line 2333: Why not like this?
```suggestion
	public static function EnrichRaisedException(\cmdbAbstractObject $oCmdbAbstract, Throwable $oException){
		if (is_null($oCmdbAbstract)){
			throw $oException;
		}
```

application/utils.inc.php
```diff
@@ -2316,4 +2316,24 @@ public static function ToCamelCase($sInput)
...
+
+		$sMessage = $oException->getMessage() . " (" . get_class($oCmdbAbstract) . "::" . $oCmdbAbstract->Get('id') . ")";
```
Hipska: Why not use predefined `__toString` method?

> HTML String representation of the object
> 
> Only a few meaningful information will be returned.
> This representation is for debugging purposes, and is subject to change.
> The returned string is raw HTML

```suggestion
		$sMessage = $oCmdbAbstract . " " . $oException->getMessage();
```

application/utils.inc.php
Hipska: It's not lost, it's only not represented. I think it's just a matter of calling `$this->getPrevious()->GetTrace()` inside `CoreException::getTraceAsHtml()` to be added into `$aBackTrace`.

odain-cbd: thanks i'll try

core/coreexception.class.inc.php
odain-cbd: yes i'll change

application/utils.inc.php
odain-cbd: good idea for code refactoring. but __toString() currently returns html content. isn't it weard in logs?

application/utils.inc.php
odain-cbd: chaining the exception is nicer. as you proposed (Hipska), I can adapt CoreException->getTraceAsHtml() to keep previous stack trace. but i confirm that stacktrace is lost in getTraceAsString(). Reflection manipulation remains better to me.

https://stackoverflow.com/questions/31562724/php-chaining-error-handling

https://stackoverflow.com/questions/34923644/does-re-throwing-an-exception-in-php-destroy-the-stack-trace

Hipska: These are 2 irrelevant links, see this example to prove the stack trace is not lost: https://3v4l.org/NnPbt
It's really is only matter of correctly retrieving the trace.

application/utils.inc.php
Hipska: True, but that is a weirdness of the `__toString` method.

Note that most of these messages do get displayed on the UI (Hence the existence of `CoreException::getTraceAsHtml()`)

application/utils.inc.php
odain-cbd: trace is not lost when using __tostring($e) of your chained exception. on the other hand we dont have root cause when using getTrace() or getTraceAsString(). this is really annnoying as we cannot override them (final public methods).

as discussed with Pierre, we will keep your proposal (chaining the exception) and add a new method getFullStackTraceAsString().

application/utils.inc.php
odain-cbd: i will used __tostring() as recommanded.

application/utils.inc.php
odain-cbd: Type hinting is a good idea. but it is not authorized until latest develop branch. ie 2.8.

as we would like to benefit from this work in 2.7.x (long term support) I will not apply your proposal. even if it is better to me.

Hipska: So, starting from which version is type hinting authorised? I think there are lots of other places that already do type hinting..

core/coreexception.class.inc.php
Hipska: Although this conversation was marked as resolved, it is still not resolved.

pages/UI.php
```diff
@@ -1997,7 +1997,7 @@ function DisplayNavigatorGroupTab($oP)
 				$oLog->Set('userinfo', '');
 				$oLog->Set('issue', $e->GetIssue());
 				$oLog->Set('impact', 'Page could not be displayed');
-				$oLog->Set('callstack', $e->getTrace());
+				$oLog->Set('callstack', $e->getFullStackTraceAsString());
```
Hipska, line 2007: Previous method returned an array and now a string is returned. This is a huge difference.

pages/UI.php
```diff
@@ -2007,7 +2007,7 @@ function DisplayNavigatorGroupTab($oP)
 			}
 		}
 
-		IssueLog::Error('UI.php operation='.$operation.', error='.$e->getMessage()."\n".$e->getTraceAsString());
+		IssueLog::Error('UI.php operation='.$operation.', error='.$e->getMessage()."\n".$e->getFullStackTraceAsString());
```
Hipska, line 2017: Aren't you over-complicating thing?
```suggestion
		IssueLog::Error('UI.php operation='.$operation.', error='.$e->getMessage()."\n".$e);
```
Or this as the toString does give the message as well:
```suggestion
		IssueLog::Error('UI.php operation='.$operation.', error='.$e);
```

Hipska (changes requested):

application/utils.inc.php
piRGoif: We decided to add type hinting only starting from 2.8.0 (develop branch), and only for new methods : a very conservative approach, but we wouldn't like to add bugs because of this...

piRGoif: Note that the [coding standard wiki page](https://www.itophub.io/wiki/page?id=latest%3Acustomization%3Acoding_standards) wasn't updated yet, as this decision was made only few weeks ago, and we want to test drive this internally first...

Hipska: Well, correct me if I'm wrong, but isn't this a new method?

core/coreexception.class.inc.php
```diff
@@ -28,7 +28,7 @@
 
 class CoreException extends Exception
 {
-	public function __construct($sIssue, $aContextData = null, $sImpact = '')
```
piRGoif, line 31: There are some CoreException children that redefines the constructor, but we chose not to update them yet, as this modification aims to be merged to support/2.7 branch.
The work will continue later on, actually handling exceptions throwned by non iTop core code (modules, datamodel methods overrides, ...) is a vast subject O:)

piRGoif (dismissed):

pages/UI.php
piRGoif: You're right ! But the value is written to a string field anyway.

pages/UI.php
piRGoif: Adding a new method allows us to have a clear PHPDoc explaining why we're doing this. And we're also not totally confident that Exception::toString will suit our needs in the future.

pages/UI.php
Hipska: Correct, but the structure of it will be completely different.

pages/UI.php
Hipska: Okay, can be solved for me.

application/utils.inc.php
piRGoif: There are 2 conditions to add type hinting :
* code starting on the 2.8.0
* new method

We are adding a new method but on a previous version.

For now this PR is based on develop but this was a mistake and will be fixed very soon : it will be merged to the support/2.7 branch.

pages/UI.php
piRGoif: Right again, but this is debug traces and we though changing format won't matter.

application/utils.inc.php
Hipska: Oh, okay. That was not clear. 👍

core/coreexception.class.inc.php
piRGoif: Conversation reopened, thanks for your notice ! I'll tell @odain-cbd

application/utils.inc.php
Hipska: I didn't recommend it. I was just asking if this wouldn't be the better way.

core/coreexception.class.inc.php
Hipska: Same here @odain-cbd @piRGoif . Oh, and it has type-hinting ;)

core/coreexception.class.inc.php
```diff
@@ -81,6 +81,16 @@ public function getHtmlDesc($sHighlightHtmlBegin = '<b>', $sHighlightHtmlEnd = '
 		return $this->getMessage();
 	}
 
+	/**
+	 * getTraceAsString() cannot be overrided and it is limited as only current exception stack is returned.
+	 * we need stack of all previous exceptions
```
piRGoif, line 86: Having a PSR-3 compatible logger would allow to get rid of this... In such logger the exception is passed directly to the log method, whereas iTop logger only accept a string so logger consumer needs to call itself the getTrace() method.

We might need to add this to LogAPI, will discuss this with the team...

core/coreexception.class.inc.php
piRGoif: And also there is type hinting... And Throwable seems to have been added in PHP 7 only (https://www.php.net/manual/fr/class.throwable.php)

odain-cbd: type hinting removed

core/coreexception.class.inc.php
odain-cbd: code cleanup done

Hipska (dismissed):

Hipska (dismissed):

piRGoif (dismissed):

odain-cbd:
PR cherry-picked and commited in support/2.7 (4e0eed6e1).

Thanks for the active collaboration of all reviewers. we went deeper than I would have imagined!

PR #160, opened 2020-09-02
Fix filename headers when downloading

Hipska:
The [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) is either `inline` or `attachment` with optional filename.

The wrong combination resulted in (some?) browser not to use the correct name for the downloaded file.

Also, the `Content-Description` header is not a HTTP header, but a MIME header, so I removed that as well.

### DIFF
--- a/sources/application/TwigBase/Controller/Controller.php
+++ b/sources/application/TwigBase/Controller/Controller.php
@@ -437,7 +437,7 @@ abstract class Controller
 
 		if ($bFileTransfer)
 		{
-			header('Content-Disposition: attachment; filename="'.$sDownloadArchiveName);
+			header('Content-Disposition: attachment; filename="'.$sDownloadArchiveName.'"');
 		}
 
 		header('Expires: 0');
--- a/sources/application/TwigBase/Controller/Controller.php
+++ b/sources/application/TwigBase/Controller/Controller.php
@@ -437,8 +437,7 @@ abstract class Controller
 
 		if ($bFileTransfer)
 		{
-			header('Content-Description: File Transfer');
-			header('Content-Disposition: inline; filename="'.$sDownloadArchiveName);
+			header('Content-Disposition: attachment; filename="'.$sDownloadArchiveName);
 		}
 
 		header('Expires: 0');


Molkobain:
Hello @Hipska 

Can you provide an example on how to reproduce the incorrect case? For example what kind of file you used and how you downloaded it.

Thanks,
Guillaume

Hipska:
In iTop Professional, there is the System Information module, where you can download a report. The name of the downloaded file is (with Safari) `exec.php` instead of something like `itop-system-information-report-2020-09-02-09-33-27.zip`.

Molkobain:
Thanks, I'll try to reproduce it and we'll give you a feedback on friday.

Molkobain (dismissed):
Even though I don't have a Mac to reproduce, the code makes sense and follows what we do in other parts of the app.

sources/application/TwigBase/Controller/Controller.php
```diff
@@ -437,8 +437,7 @@ public function DownloadZippedPage($aParams = array(), $sTemplateName = null)
 
 		if ($bFileTransfer)
 		{
-			header('Content-Description: File Transfer');
-			header('Content-Disposition: inline; filename="'.$sDownloadArchiveName);
+			header('Content-Disposition: attachment; filename="'.$sDownloadArchiveName);
```
eespie: There is a missing '"' at the end (bug from previous version). Without it I have the same error on Safari (file named exec.php).

```suggestion
			header('Content-Disposition: attachment; filename="'.$sDownloadArchiveName.'"');
```

Tested on Safari & Firefox

Hipska: Good find. Didn't notice this as on my Safari it worked 🙄

Molkobain (approved):
Good catch from @eespie , thanks for testing on Mac. I'll push it to the review on friday. No reason why it will not be accepted.

piRGoif (approved):

Molkobain:
Merged, thanks @Hipska @eespie

PR #161, opened 2020-09-02
Correctly detect if connection to client is secure

Hipska:
The current code breaks the setup page when the iTop server is behind a proxy that handles [SSL termination](https://en.wikipedia.org/wiki/TLS_termination_proxy).

I have applied a better way to detect if iTop has been reached over HTTPS based on code already present in the project:
https://github.com/Combodo/iTop/blob/97e58c2d79355c71b91e44071d24796fe6449e67/datamodels/2.x/authent-cas/vendor/apereo/phpcas/source/CAS/Client.php#L3591-L3610

### DIFF
--- a/setup/setuppage.class.inc.php
+++ b/setup/setuppage.class.inc.php
@@ -144,7 +144,7 @@ class SetupPage extends NiceWebPage
 
 	public function output()
 	{
-		$sLogo = utils::GetAbsoluteUrlAppRoot().'/images/itop-logo.png';
+		$sLogo = utils::GetAbsoluteUrlAppRoot(true).'/images/itop-logo.png';
 		$this->s_content = "<div id=\"header\"><h1><a href=\"http://www.combodo.com/itop\" target=\"_blank\"><img title=\"iTop by Combodo\" alt=\" \" src=\"{$sLogo}?t=".utils::GetCacheBusterTimestamp()."\"></a>&nbsp;".htmlentities($this->s_title,
 				ENT_QUOTES, self::PAGES_CHARSET)."</h1>\n</div><div id=\"setup\">{$this->s_content}\n</div>\n";
 
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -925,10 +925,19 @@ class utils
 	{
 		$bSecured = false;
 
-		if (!empty($_SERVER['HTTPS']) && (strtolower($_SERVER['HTTPS']) != 'off'))
+		if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO']))
 		{
-			$bSecured = true;
+			$bSecured = ($_SERVER['HTTP_X_FORWARDED_PROTO'] === 'https');
 		}
+		elseif (!empty($_SERVER['HTTP_X_FORWARDED_PROTOCOL']))
+		{
+			$bSecured = ($_SERVER['HTTP_X_FORWARDED_PROTOCOL'] === 'https');
+		}
+		elseif (isset($_SERVER['HTTPS']) && !empty($_SERVER['HTTPS']))
+		{
+			$bSecured = (strcasecmp($_SERVER['HTTPS'], 'off') !== 0);
+		}
+
 		return $bSecured;
 	}
 
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -834,7 +834,7 @@ class utils
 	{
 		// Build an absolute URL to this page on this server/port
 		$sServerName = isset($_SERVER['SERVER_NAME']) ? $_SERVER['SERVER_NAME'] : '';
-		$sProtocol = self::IsConnectionSecure() ? 'https' : 'http';
+		$sProtocol = self::IsConnectionSecure(true) ? 'https' : 'http';
 		$iPort = isset($_SERVER['SERVER_PORT']) ? $_SERVER['SERVER_PORT'] : 80;
 		if ($sProtocol == 'http')
 		{
@@ -920,16 +920,20 @@ class utils
 	 * Though the official specs says 'a non empty string', some servers like IIS do set it to 'off' !
 	 * nginx set it to an empty string
 	 * Others might leave it unset (no array entry)	 
+	 *
+	 * @param bool $bTrustProxy
+	 *
+	 * @return bool
 	 */	 	
-	public static function IsConnectionSecure()
+	public static function IsConnectionSecure($bTrustProxy=false)
 	{
 		$bSecured = false;
 
-		if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO']))
+		if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO']) && $bTrustProxy)
 		{
 			$bSecured = ($_SERVER['HTTP_X_FORWARDED_PROTO'] === 'https');
 		}
-		elseif (!empty($_SERVER['HTTP_X_FORWARDED_PROTOCOL']))
+		elseif (!empty($_SERVER['HTTP_X_FORWARDED_PROTOCOL']) && $bTrustProxy)
 		{
 			$bSecured = ($_SERVER['HTTP_X_FORWARDED_PROTOCOL'] === 'https');
 		}
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -790,11 +790,13 @@ class utils
     /**
      * Returns the absolute URL to the application root path
      *
+     * @param bool $bTrustProxy
+     *
      * @return string The absolute URL to the application root, without the first slash
      *
      * @throws \Exception
      */
-	public static function GetAbsoluteUrlAppRoot()
+	public static function GetAbsoluteUrlAppRoot($bTrustProxy=false)
 	{
 		static $sUrl = null;
 		if ($sUrl === null)
@@ -802,7 +804,7 @@ class utils
 			$sUrl = self::GetConfig()->Get('app_root_url');
 			if ($sUrl == '')
 			{
-				$sUrl = self::GetDefaultUrlAppRoot();
+				$sUrl = self::GetDefaultUrlAppRoot($bTrustProxy);
 			}
 			elseif (strpos($sUrl, SERVER_NAME_PLACEHOLDER) > -1)
 			{
@@ -826,15 +828,17 @@ class utils
 	 * For most usages, when an root url is needed, use utils::GetAbsoluteUrlAppRoot() instead as uses this only as a fallback when the
 	 * app_root_url conf parameter is not defined.
 	 *
+	 * @param bool $bTrustProxy
+	 * 
 	 * @return string
 	 *
 	 * @throws \Exception
      */
-    public static function GetDefaultUrlAppRoot()
+    public static function GetDefaultUrlAppRoot($bTrustProxy=false)
 	{
 		// Build an absolute URL to this page on this server/port
 		$sServerName = isset($_SERVER['SERVER_NAME']) ? $_SERVER['SERVER_NAME'] : '';
-		$sProtocol = self::IsConnectionSecure(true) ? 'https' : 'http';
+		$sProtocol = self::IsConnectionSecure($bTrustProxy) ? 'https' : 'http';
 		$iPort = isset($_SERVER['SERVER_PORT']) ? $_SERVER['SERVER_PORT'] : 80;
 		if ($sProtocol == 'http')
 		{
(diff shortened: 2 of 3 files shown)


piRGoif (dismissed):
Seems OK to me :)
After integrating this PR we should change the CAS code to call the utils method...

bruno-ds:
http headers should be taken into account only for trusted remote.
If this is not done then, for instance, a man in the middle could trick iTop into assuming the connection is secure.

So, I think you should add a way to perform this configuration.


for instance, symfony name this conf "trusted proxies":
https://symfony.com/doc/current/deployment/proxies.html#solution-settrustedproxies

bruno-ds:
With a little more time to think about your code, I think that your assumption is harmless for your use case. But it indeed introduce  a misleading API, maybe conditioning the check of the headers to an optional second argument with a phpdoc mentioning the trust issue would be a good compromise since it would avoid the difficulty of setting a conf before the initial setup

bruno-ds:
I just found that there is a documented solution that seems simple to use: 
https://www.itophub.io/wiki/page?id=latest:admin:settings&s[]=RequestHeader#accessing_itop_from_different_urls

Doesn't it work for your use case?

Hipska:
1. The instance is not accessed from different urls.
2. It's on Azure, I can't modify proxy headers of the SSL terminator.
3. During initial setup there is no config file yet, so the method `utils::GetDefaultUrlAppRoot()` gets executed.

bruno-ds:
> 2. It's on Azure, I can't modify proxy headers of the SSL terminator.

I don't get it, you can't add this on the iTop's webserver configuration?
```apache
RequestHeader set HTTPS On
```
This should be feasible either on the vhost's configuration or in a .htaccess file.

Hipska:
Ah, yes that should be possible indeed.

But I see and understand your concerns, I see 2 uses of `utils::IsConnectionSecure()`:
https://github.com/Combodo/iTop/blob/0542a8e4f8c46f70c640c12f79e7480a899d94e9/application/loginwebpage.class.inc.php#L438-L442
https://github.com/Combodo/iTop/blob/0542a8e4f8c46f70c640c12f79e7480a899d94e9/application/utils.inc.php#L833-L837

Correct me if I'm wrong, but I think the security concern is only for the first one. So what if we add a `$bTrustProxy` parameter and set it `true` by default. And set it to `false` when use the method on the login page?

bruno-ds:
> Correct me if I'm wrong, but I think the security concern is only for the first one. So what if we add a `$bTrustProxy` parameter and set it `true` by default. And set it to `false` when use the method on the login page?

This is was I proposed you but since there is a solution in the documentation, I'd rather close the PR if you're ok with that.
If you're not, I'll trigger an internal debate over your proposal.

Please note that while it has the drawback to require the admin to find the wiki entry, I still think the wiki's solution is better because: 
 - it do not add complexity in the code
 - it put the administrator in charge of a security concern

Hipska:
The documentation has a workaround, not a 'solution'.

I don't agree to close the PR as I'm having the impression that PR's are too quickly/easily closed without any good reason or feedback. So please discuss internally first and come back with a good reason. I don't bite 😂

The workaround in the documentation also is not a good fit if iTop is reached via this SSL proxy for external clients, but if internal systems at the same side of the proxy are reaching over HTTP.

bruno-ds:
> I don't agree to close the PR as I'm having the impression that PR's are too quickly/easily closed without any good reason or feedback. So please discuss internally first and come back with a good reason. I don't bite 😂

I think that I gave you both reason and feedback, bu you have the right to think they are not good enough, so ok, we'll discuss it internally 😄 


> The workaround in the documentation also is not a good fit if iTop is reached via this SSL proxy for external clients, but if internal systems at the same side of the proxy are reaching over HTTP.

I don't think so, you'll just have to use two virtual host (one for the ssl, one for the http) and handle this conf. on the virtual host side.

Hipska:
> I think that I gave you both reason and feedback, bu you have the right to think they are not good enough, so ok, we'll discuss it internally 😄

I'm not only talking about this PR, you came with clear arguments indeed. Only sad that it is a workaround, while this PR (unintentionally) aims to fix the need of a workaround.

> I don't think so, you'll just have to use two virtual host (one for the ssl, one for the http) and handle this conf. on the virtual host side.

I could have thought that one would come up with this "workaround on a workaround" solution. I don't get why you as a vendor would prefer this over a fix in the code, so all these workarounds are not needed.

Discuss it internally, and let me know if it is not wanted, I just tried to help to fix stuff as I know there are a lot of bugs/features to be handled and you don't have time to work on these small ones.

Hipska:
Hi @bruno-ds, do you have an update?

bruno-ds:
> Hi @bruno-ds, do you have an update?

Sorry for the delay (my family was hit by the covid), I just restarted to work this monday, we will soon work on your PR.

Hipska:
Okay, no problem. I hope everything is ok.

odain-cbd:
Hi Hipska,

we just reviewed current PR with Pierre and Bruno. it seems nice to us if you agree to finalize it. 

here are following remaining things to be completed:
- add optional parameter and implement discussed logic. instead of `$bTrustProxy`,  `$bTrustUnsecureHeaders` seems better to us.
- enhance PHP doc to reflect the security impact of the parameter
- (optional) add unit test to cover/document this enhancement. 
if you don't feel confident with testing, no pb. we will do this part for you.

BR
the team

Hipska:
The optional parameter should default to `true` or `false`?

odain-cbd:
"So what if we add a $bTrustProxy parameter and set it true by default. And set it to false when use the method on the login page?"

let's stick to your initial proposal. defaut is true.

piRGoif:
I was surprised by @odain-cbd answer, so I've just had a chat with him... And indeed we both agreed that `$bTrustProxy=false` is preferable : this way previous calls or modified calls with default parameter value will lead to the exact same behavior as before.

If a developer wants to had the extra check using the other http headers, then he'll have to specify the parameter, and it's a value that isn't the default one so we can assume the developer has read the phpdoc :o)

Thanks !

bruno-ds:
There are two reasons to set `$bTrustProxy=false` as default :
 - We do not want to introduce unnecessary breaking changes
 - Security must take precedence, and the default signature mustn't use insecure headers

Hipska:
Sorry for the late response, but here are my changes ;-)

piRGoif (changes requested):
For now the $bTrustProxy is passed directly in \utils::GetDefaultUrlAppRoot, but it's not this method responsibility to know this.
Indeed this parameter value should be set by the final consumer : calls in WizStep* should pass true, all others nothing or false.

Also note that \utils::GetDefaultUrlAppRoot is called from another util method : \utils::GetAbsoluteUrlAppRoot. This one too should propagate this parameter with the same default value !
It is called multiple times, but never seems to be called directly from the setup : so we should just add the parameter and don't change the consumer so that default value would be used.

Hipska:
Done

setup/setuppage.class.inc.php
```diff
@@ -144,7 +144,7 @@ public function collapsible($sId, $sTitle, $aItems, $bOpen = true)
 
 	public function output()
 	{
-		$sLogo = utils::GetAbsoluteUrlAppRoot().'/images/itop-logo.png';
+		$sLogo = utils::GetAbsoluteUrlAppRoot(true).'/images/itop-logo.png';
```
piRGoif, line 147: Good catch !

piRGoif (changes requested):
Many thanks !

As you found the call to \utils::GetAbsoluteUrlAppRoot in \SetupPage::output, I did a search of GetAbsoluteUrlAppRoot in the setup dir, and found another one in \MFCompiler::PathToPHP. Could you add true here as well ?

Hipska:
Hmm, this gets reflected in compiled code, so outside of the setup. So I think it should actually be false in that case.

piRGoif:
Do you have any call stack in mind ? I supposed calls to the compilator is done only in safe places like setup, toolkit, Designer or Hub connector ?

Hipska:
The call is indeed done in the setup, but the code you are referring to is this:
https://github.com/Combodo/iTop/blob/76fbd3790ff4a48e2de3d9730a67bb1107f08f97/setup/compiler.class.inc.php#L728-L767

That doesn't do the actual call to `utils::GetAbsoluteUrlAppRoot` but will place that in the generated datamodel php classes. And there, that piece of code will be called at normal iTop runtime, not inside the setup.

See https://github.com/Combodo/iTop/blob/76fbd3790ff4a48e2de3d9730a67bb1107f08f97/setup/compiler.class.inc.php#L2114-L2123

So I guess this is all that is needed for this PR and it can be merged?

piRGoif:
You are perfectly right ! My mistake, sorry :/
I'm changing my review !

piRGoif (approved):

piRGoif:
Awaiting a second Combodo review to merge, I've just pinged Bruno :)

bruno-ds (approved):
Alas some minor coding convention (`bTrustProxy = false` instead of `bTrustProxy=false`) and an explanation of the security implication of trusting insecure proxies inside the PhpDoc, everything seems fine.

I think that the merge can be performed as it (we would then fix the above-mentioned minor issues)

bruno-ds:
FYI, this commit: https://github.com/Combodo/iTop/commit/7b093a6bba505810d00f7937754e26e626f829b3 is built upon your PR.
It enhance it by providing getters for the most important parts of the url (the uri part seems to not be really used, we didn't even find sufficient information about how it behave).

A few commit will add a few extra api endpoints (or: GetCurrentAbdoluteUrl), but the core logic is here.
Thanks @Hipska for pointing us in the right direction :)

piRGoif:
Also 7b093a6 was committed to the support/2.7 branch and will be available in iTop 2.7.4. This PR was merged to develop so next iTop 3.0.0.

bruno-ds:
> Also [7b093a6](https://github.com/Combodo/iTop/commit/7b093a6bba505810d00f7937754e26e626f829b3) was committed to the support/2.7 branch and will be available in iTop 2.7.4. This PR was merged to develop so next iTop 3.0.0.

Yes, and since it will then be merged to develop and overwrite part of Hipka's work. I thought he deserved an explanation (and a thank you).

Hipska:
Hey, thanks for pinging back. I don't know if Cherry-picking would have helped to keep my name in the `git blame` records.. That's maybe something to consider in the future.

I still want to express I am very happy that this PR got merged even it had some resistance and comments in the beginning, but in the end I could convince you all and it got merged! 👍

Hipska:
Hey all, just a heads-up. I just tried to install iTop Professional 2.7.4 in Azure and the setup is still broken.

It tries to load all resources as https://NAME.azurewebsites.net:80/... (I was under the impression it would have been solved in that version?) It seems the protocol is now trusted, but the port isn't?

Molkobain:
Hello Thomas,

Thanks for the feedback, I've just forwarded your message to the QA team.

bruno-ds:
Did you properly set `behind_reverse_proxy`, and could you please check if the requirements set in the link below are met?
https://www.itophub.io/wiki/page?id=latest%3Ainstall%3Areverse_proxy

Thanks

Hipska:
The link is private, but no, it’s an initial install, so settings cannot be set yet.

piRGoif:
>> Did you properly set behind_reverse_proxy, and could you please check if the requirements set in the link below are met?
>> https://www.itophub.io/wiki/page?id=latest%3Ainstall%3Areverse_proxy
> 
> The link is private

Here is the public link : 
https://www.itophub.io/wiki/page?id=3_0_0%3Ainstall%3Areverse_proxy

bruno-ds:
> The link is private, but no, it’s an initial install, so settings cannot be set yet.

When I said that I wanted you to check against [this page](https://www.itophub.io/wiki/page?id=latest%3Ainstall%3Areverse_proxy), I had this in mind:
Could you please check/indicate which ` X-Forwarded-*`  HTTP headers are present?

I think that either the port or the protocol is missing.

Hipska:
It seems only `X-Forwarded-Proto` is available:
image

The problem seems to be at https://github.com/Combodo/iTop/blob/410948fb4e261f1bf4063525288873dd97508ff8/application/utils.inc.php#L1074-L1078

I think there should be assumed the default port if the proxy does not specify the `X-Forwarded-Port` parameter?

bruno-ds:
your capture seems to lack some informations, like:
 - on what port the webserver is contacted by the reverse proxy?
 - same for the protocol?

By inferring this from your symptoms, I would state that the reverse proxy forward the requests to the webserver using the port 80 in place of the 443 but does not send the header `X-Forwarded-port`.

Could you please check this, and add the `X-Forwarded-port` if I'm right?

PR #163, opened 2020-09-04
Cleanup references to Flash

Hipska:
While Flash support is deprecated, the codebase still has a lot of references to Flash.

This PR removes all the swf files and an unused swf library that get's loaded on every page. It deletes a php page with abandoned code. Since it never got referenced, it is better/safer to completely remove it. It also renames the action to load the relation graph as it still got a swf reference.


pages/UI.php
```diff
@@ -1853,7 +1853,7 @@ function DisplayNavigatorGroupTab($oP)
 
 		///////////////////////////////////////////////////////////////////////////////////////////
 		
-		case 'swf_navigator': // Graphical display of the relations "impact" / "depends on"
+		case 'navigate': // Graphical display of the relations "impact" / "depends on"
```
piRGoif: I would recommend keeping the old operation, as there might be some modules out there depending on this. This will allow us to add the depreciation in our migration notes, and remove this operation on the next major version.

```suggestion
		case 'swf_navigate': // SWF was removed in iTop 
		case 'navigate': // Graphical display of the relations "impact" / "depends on"
```

piRGoif:
Many thanks, we should have done this long ago !

pages/UI.php
Hipska: Good suggestion!

Hipska: Maybe adding an `@depricated` doc as well?

piRGoif: I did hesitate to add it : it suits very well in a function / class PHPDoc, but for just one line, I'm not sure ? But indeed it's a keyword we're looking before releasing :) Maybe add it in the inline comment like this :
```suggestion
		case 'swf_navigate': // @deprecated SWF was removed in iTop 
		case 'navigate': // Graphical display of the relations "impact" / "depends on"
```

Hipska: I'm not sure this works with phpdoc. Can you verify first?

It would at least needed to be 
```suggestion
		case 'swf_navigate': /** @deprecated SWF was removed in iTop */
		case 'navigate': // Graphical display of the relations "impact" / "depends on"
```

piRGoif: I don't think that the deprecated tag is handled by IDEs for anything else than a class or a function... I didn't find any reference though :/ I was remembering that deprecated usages are strike through since a few versions in PHPStorm but I didn't find any release note about this (I didn't take time to do an exhaustive search...)

You version is ok for me :)

Hipska: Done: e0079a4

piRGoif (dismissed):
Ok for me, but we will need to do a team review... will keep you posted !

piRGoif:
Will be reviewed by Romain & Denis

rquetiez (changes requested):
I would rather have 'navigate_relations' because 'navigate' is too common and grep is the solely option to connect the dots.

Hipska:
Or just 'relations', like you have 'details' and 'stimulus' for example?

rquetiez (dismissed):
Yes, 'relations' is more meaningful than 'navigate', but I would prefer a more discriminant id like 'view-relations' because this will be helpfull on the long run when searching where this operation is used (and therefore perform a change or cleanup.

Hipska:
Done

piRGoif (dismissed):

application/displayblock.class.inc.php
```diff
@@ -1677,11 +1677,11 @@ public function GetRenderContent(WebPage $oPage, $aExtraParams = array(), $sId)
 						{
 							if (array_key_exists('down', $aRelationInfo))
 							{
-								$aActions[$sRelationCode.'_down'] = array ('label' => $aRelationInfo['down'], 'url' => "{$sRootUrl}pages/$sUIPage?operation=swf_navigator&relation=$sRelationCode&direction=down&class=$sClass&id=$id{$sContext}") + $aActionParams;
+								$aActions[$sRelationCode.'_down'] = array ('label' => $aRelationInfo['down'], 'url' => "{$sRootUrl}pages/$sUIPage?operation=view-relations&relation=$sRelationCode&direction=down&class=$sClass&id=$id{$sContext}") + $aActionParams;
```
Molkobain: I know @rquetiez told you to put a dash but it was a typo, we rather use underscores for operations like in the rest of the PHP endpoints.

```suggestion
								$aActions[$sRelationCode.'_down'] = array ('label' => $aRelationInfo['down'], 'url' => "{$sRootUrl}pages/$sUIPage?operation=view_relations&relation=$sRelationCode&direction=down&class=$sClass&id=$id{$sContext}") + $aActionParams;
```

application/displayblock.class.inc.php
```diff
@@ -1677,11 +1677,11 @@ public function GetRenderContent(WebPage $oPage, $aExtraParams = array(), $sId)
...
 							}
 							if (array_key_exists('up', $aRelationInfo))
 							{
-								$aActions[$sRelationCode.'_up'] = array ('label' => $aRelationInfo['up'], 'url' => "{$sRootUrl}pages/$sUIPage?operation=swf_navigator&relation=$sRelationCode&direction=up&class=$sClass&id=$id{$sContext}") + $aActionParams;
+								$aActions[$sRelationCode.'_up'] = array ('label' => $aRelationInfo['up'], 'url' => "{$sRootUrl}pages/$sUIPage?operation=view-relations&relation=$sRelationCode&direction=up&class=$sClass&id=$id{$sContext}") + $aActionParams;
```
Molkobain:
```suggestion
								$aActions[$sRelationCode.'_up'] = array ('label' => $aRelationInfo['up'], 'url' => "{$sRootUrl}pages/$sUIPage?operation=view_relations&relation=$sRelationCode&direction=up&class=$sClass&id=$id{$sContext}") + $aActionParams;
```

pages/UI.php
```diff
@@ -1852,8 +1852,9 @@ function DisplayNavigatorGroupTab($oP)
 		break;
 
 		///////////////////////////////////////////////////////////////////////////////////////////
-		
-		case 'swf_navigator': // Graphical display of the relations "impact" / "depends on"
+
+		case 'swf_navigate': /** @deprecated SWF was removed in iTop */
```
Molkobain: Did you mean "swf_navigat**or**"?

```suggestion
		case 'swf_navigator': /** @deprecated SWF was removed in iTop */
```

pages/UI.php
```diff
@@ -1852,8 +1852,9 @@ function DisplayNavigatorGroupTab($oP)
...
+		case 'view-relations': // Graphical display of the relations "impact" / "depends on"
```
Molkobain:
```suggestion
		case 'view_relations': // Graphical display of the relations "impact" / "depends on"
```

js/swfobject.js
```diff
@@ -1,5 +0,0 @@
-/* SWFObject v2.1 <http://code.google.com/p/swfobject/>
```
Molkobain, line 1: Careful this file is included in the "approval-base" and "customer-survey" modules.

If we continue as is, there will be a 404 error for this file in the browser and the webserver logs until we make a new version of the extension (strategy TBD to keep BC).

We also need to check if these 2 modules completely abandoned usage of Flash (I can't anwser this without having a look at the code)

Molkobain (changes requested):

Molkobain:
Just made a few comment, once the one on the complementary modules is discussed and the small changes on the operation name fixed, it will be all good for me.

Thanks for the cleanup @Hipska 👍

pages/UI.php
piRGoif: Ouch, nice catch Guillaume !

application/displayblock.class.inc.php
Hipska: Thanks, will combine them in 1 commit if no problem..

pages/UI.php
Hipska: Whoops indeed. Came from e0079a41e5df4bd106441ca518a107bdd31330da

application/displayblock.class.inc.php
Molkobain: Yeah sure 😊

js/swfobject.js
Hipska: Since there isn't any browser that supports Flash anymore (out of the box), I highly doubt these modules still actively use the framework referenced. But it's indeed better to check these modules and remove the include line as done  in `itopwebpage.class.inc.php` before this gets merged into the development branch.

Molkobain: Just checked, these 2 modules are included in the iTop commercial packages, so they will be worked on before the 2.8.0 release. Corresponding tickets have been created N°3344 and N°3345. We can proceed with this!

Molkobain (approved):

Molkobain:
All good, thanks @Hipska ! ✌

Molkobain:
We got a "This branch cannot be rebased safely: Rebasing the commits of this branch on top of the base branch cannot be performed automatically as this would create a different result than a regular merge."

Can you try rebasing your branch on develop? Or is it because your branch have diverged from develop?

Hipska:
That's why I don't like rebase's.. I checked and don't see any merge conflicts. Is a regular merge not possible instead?

piRGoif:
Hi @Hipska, Guillaume had an error message but I didn't. I've just done the merge !
Usually we are doing squash & merge, unless we absolutely want to keep the PR branch history - I've done squash here.

Molkobain:
Thanks @piRGoif !

Hipska:
Hi, could you please also add the `hacktoberfest` tag to this repo? digitalocean/hacktoberfest#596

piRGoif:
This was done here too

PR #165, opened 2020-09-06
fix error: Invalid uniqueness rule declaration : class=FunctionalCI, …

annProg:
follow the wiki：https://www.itophub.io/wiki/page?id=2_7_0%3Acustomization%3Auniqueness-rules

define  **Unique name within sub-classes**

```xml
        <uniqueness_rules>
          <rule id="functionalci_name" _delta="define">
            <!-- field or combination of "FunctionalCI" fields which must be unique -->
            <attributes>
              <attribute id="name"/>
              <attribute id="finalclass"/>
            </attributes>
          <!-- ... -->
```

will report error:

> Invalid uniqueness rule declaration : class=FunctionalCI, rule=functionalci_name, reason=Uniqueness rule : non existing field 'finalclass'

This pr try to fix the error. This may not be the right fix, I just want to point out the issue.

### DIFF
--- a/core/metamodel.class.php
+++ b/core/metamodel.class.php
@@ -2889,6 +2889,7 @@ abstract class MetaModel
 					if (!empty($aCurrentClassUniquenessRules))
 					{
 						$aClassFields = self::GetAttributesList($sPHPClass);
+						$aClassFields[] = "finalclass";
 						foreach ($aCurrentClassUniquenessRules as $sUniquenessRuleId => $aUniquenessRuleProperties)
 						{
 							$bIsRuleOverride = self::HasSameUniquenessRuleInParent($sPHPClass, $sUniquenessRuleId);


piRGoif:
Hello,
Thanks for this, I'll have a look ! This is an interesting use case we didn't think about !

piRGoif:
I confirm I reproduce this on all of those code bases : support/2.7 and support/2.6 branches, 2.6.0 tag.

core/metamodel.class.php
```diff
@@ -2889,6 +2889,7 @@ public static function InitClasses($sTablePrefix)
 					if (!empty($aCurrentClassUniquenessRules))
 					{
 						$aClassFields = self::GetAttributesList($sPHPClass);
+						$aClassFields[] = "finalclass";
```
piRGoif, line 2892: `finalclass` field name can be customized (`classes/class/properties/db_final_class_field` node in the XML)

I suggest using this instead of a hardcoded string :

```suggestion
						$aClassFields[] = self::DBGetClassField($sPHPClass);
```

But anyway the uniqueness rule check is done in the wrong place as magix fields are not already added to classes...

piRGoif:
There are more fields to add that only finalclass... Actually there are lots of such "magic" attributes, for example : friendly_name, obsolescence_*, ...
The check is done in \MetaModel::InitClasses, but before all the magic fields are added...
Indeed this is a wider issue :/ 

So we'll certainly implement our own fix... But we need to discuss this with the product team, it should take some weeks. In the meantime I keep this open, and I'll get back to you here !

piRGoif (commented):
In addition to PR comment, a little info about finalclass field name !

annProg:
> There are more fields to add that only finalclass... Actually there are lots of such "magic" attributes, for example : friendly_name, obsolescence_*, ...
> The check is done in \MetaModel::InitClasses, but before all the magic fields are added...
> Indeed this is a wider issue :/
> 
> So we'll certainly implement our own fix... But we need to discuss this with the product team, it should take some weeks. In the meantime I keep this open, and I'll get back to you here !

Yes, wait for your fix

piRGoif:
Hello,
I've created a bug in our (private) DB : N°3371. I hope that it will be planned in iTop 2.8.0...

The correct fix would be to extract the \MetaModel::CheckUniquenessRuleValidity and move it to a new loop self::GetClasses() at the very end of \MetaModel::InitClasses.

I'm closing this PR, many thanks again for the bug report !

piRGoif:
Hello,
The fix was postponned by the product team... But the wiki page was updated : https://www.itophub.io/wiki/page?id=2_7_0%3Acustomization%3Auniqueness-rules#unique_server_name

Hipska:
@piRGoif could you have a second look at that wiki section? It has multiple inconsistencies or wrong statements..

piRGoif:
Hello Thomas, could you create a SF ticket pointing out the different errors, and eventually proposed fixes ? 
Thanks !

Hipska:
It appears to be only 2 small errors:
image

piRGoif:
Thanks !
This is now fixed in both pages : 
https://www.itophub.io/wiki/page?id=2_7_0%3Acustomization%3Auniqueness-rules#unique_server_name
https://www.itophub.io/wiki/page?id=3_0_0%3Acustomization%3Auniqueness-rules#unique_server_name

PR #166, opened 2020-09-09
Fix SQL error introduced in a1f5d80

Hipska:
As discussed in the comments of a1f5d80, this introduced SQL errors. This PR solves the errors.

### DIFF
--- a/synchro/synchro_import.php
+++ b/synchro/synchro_import.php
@@ -534,11 +534,7 @@ try
 				$aValues = array(); // Used to build the insert query
 				foreach ($aRow as $iCol => $value)
 				{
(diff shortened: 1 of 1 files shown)


rquetiez:
Yes, you are right, this simplifies the code.

As a coincidence, I am working today on a fix for that. It took time because I take the opportunity to port automated tests on the data synchro from my own odd framework into the PHPUnit framework.

Also, I did a quick refactoring to make it clear that there is no security breach and there is not trap anymore:  CMDBSource::Quote is invoked **once for all the values injected** into the INSERT statement.

And also, I've included your simplification too.

See commit f5a3bb2.

Hipska:
A merge of this PR would have been nice for the credits though …

PR #167, opened 2020-09-17
Advanced search: only list AttributeExternalKey items which is used by the objects

annProg:
Update 2020/10/26
Latest commit use a new method, only list AttributeExternalKey items which is used by the objects. No need to define `<filter>` any more. 

--------
Assuming such a scenario, we manage a variety of devices, they have different brand models, there may be dozens of them. At this time, when the user views the server or network device list, many items can be seen in the brand search box. 
image

The system administrator may not want to see the brand of the router, and the network administrator does not want to see the brand of the server. Therefore, I hope to reduce the number of brands listed through filter. Through the following OQL statement: `SELECT Brand AS b JOIN Model AS m ON m.brand_id=b.id WHERE m.type=:this->finalclass`. 

```xml
<class id="PhysicalDevice">
	<fields>
		<field id="brand_id" xsi:type="AttributeExternalKey"  _delta="redefine">
			<sql>brand_id</sql>
			<target_class>Brand</target_class>
			<on_target_delete>DEL_MANUAL</on_target_delete>
			<filter>SELECT Brand AS b JOIN Model AS m ON m.brand_id=b.id WHERE m.type=:this->finalclass</filter>
		</field>
	</fields>
</class>
```
It works fine when edit an object, but has no effect in the list interface. Looking at the code, I found that the search form does not support OQL statements with variables. This is normal, after all, lists is objectset, and variables for a single object, like `:this->org_id`, `:this->location_id`,  are almost useless. But `:this->finalclass` is not for a single object, I think it is worth supporting, to improve the user experience in this scenario. .
image

This pr try to resolve the issue. This may not be the right fix, I just want to point out the issue.

### DIFF
--- a/sources/application/search/searchform.class.inc.php
+++ b/sources/application/search/searchform.class.inc.php
@@ -45,11 +45,10 @@ use TrueExpression;
 use utils;
 use WebPage;
 use UserRights;
+use ValueSetObjects;
 
 class SearchForm
 {
-	private $sClass;
-
 	/**
 	 * @param \WebPage $oPage
 	 * @param \CMDBObjectSet $oSet
@@ -65,7 +64,6 @@ class SearchForm
 		$sHtml = '';
 		$oAppContext = new ApplicationContext();
 		$sClassName = $oSet->GetFilter()->GetClass();
-		$this->sClass = $sClassName;
 		$aListParams = array();
 
 		foreach($aExtraParams as $key => $value)
@@ -440,7 +438,7 @@ class SearchForm
      * @throws \MySQLException
      * @throws \MySQLHasGoneAwayException
      */
-	public static function GetFieldAllowedValues($oAttrDef, $sClass = '')
+	public static function GetFieldAllowedValues($oAttrDef)
 	{
 		$iMaxComboLength = MetaModel::GetConfig()->Get('max_combo_length');
 		if ($oAttrDef->IsExternalKey(EXTKEY_ABSOLUTE))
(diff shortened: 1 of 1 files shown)


Hipska:
Wouldn't it be better that the list only contains items that are being used by the objects? With an extra "Show more/all ▾" link to show all other options as well?

annProg:
> Wouldn't it be better that the list only contains items that are being used by the objects? With an extra "Show more/all ▾" link to show all other options as well?

Yes, I agree

Hipska:
While I do like your suggestion for the filter of `brand_id`, it might cause problems when a brand exists, but no models for the type of CI is already created. Then the user would be tempted to create the Brand (duplicate) and the Model instead of only the Model if the Brand would not have been filtered out.

annProg:
@Hipska yes，this is a problem. May add uniqueness rule for Brand.  and change OQL to
```
SELECT Brand AS b JOIN Model AS m ON m.brand_id=b.id WHERE m.type=:this->finalclass UNION SELECT Brand WHERE id NOT IN (SELECT Brand AS b JOIN Model AS m ON m.brand_id=b.id)
```

Hipska:
That query still does not show Brands that have other types of models, so I think filtering of brand does not add much when create/modify an object.

Hipska (dismissed):

annProg:
> Wouldn't it be better that the list only contains items that are being used by the objects? With an extra "Show more/all ▾" link to show all other options as well?

Latest commit can list only contains items that are being used by the objects.

bruno-ds:
This proposal is quite interesting, but would also require some time to fully embrace this behavior.
We just requested an evaluation of our product team because we (the developers) can't decide to accept/refuse such an impacting feature on our own (private ticket n°3410).

Cheers for the idea and 🤞 !


PS : I'm always uncomfortable when I have this kind of answer, in part because you already have invested so much time for the code. Our contributing guideline try to prevent this by asking you to submit the idea for an approval before coding, I don't know if you read it, I'm just suggesting :)

piRGoif:
Hello,
I'm sorry to tell that we are rejecting this PR as this seems quite some work to integrate fully and test (search is a generic mechanism, and we must check that all iTop objects works well in all scenario), and no clients expressed this need yet.
Thanks anyway for your proposal !

PR #168, opened 2020-09-25
🎨 Fix the filter icon on Data model page

Hipska:
For some unknown reason, the icon used to run a filter query on the datamodel page was [JUNO](https://unicode-table.com/en/26B5/). Sadly this icon does not work on macOS and an empty square is shown, so I replaced it with the 🔎 emoji.

If you find any better icon, please let me know.


Molkobain:
@steffunky what was the original intention?

steffunky:
I tried to find an icon that looked like a "filter" icon, but I guess it's really a random Unicode char.
The magnifying glass emoji is okay for me, but maybe a FontAwesome icon would be better as we could style it to look like an external link (as the "Filter" text next to it isn't clickable and it may not be obvious that the emoji is)?

Molkobain:
+1 on the FontAwesome icon, it allows us to have a consistent rendering on all platforms and it can be stylized as Stephen said (_Note: Please use a dedicated span tag within the anchor._)

Hipska:
I agree to have FA icon as well..

Molkobain:
Once done, I don't see any reason no to merge this 😁

Hipska:
Do you expect me to craft an FA icon or will someone provide it?

Molkobain:
You can use this one : https://fontawesome.com/icons/search?style=solid
(Spoiler : `<i class="fas fa-search"></i>`)

Hipska:
Done!

Molkobain:
Great thanks!

Molkobain (approved):

PR #169, opened 2020-09-26
Use filter of external key on csv import

Hipska:
This allows an extra check to see if the provided values are compliant with the specified filter. Example `model_id` on `PhysicalDevice`.

It also reduces the need of adding `Model->Brand name` field in csv next to `Model->Name` when importing a device. If you have duplicate model names (for different brands or different device types), the import tool is unable to pick the correct one without the addition of brand name. In some cases we are importing devices with a brand name, but without model, then this is also not working because the import tool searches for a Model with empty name and a brand name.

I didn't test this with any object that has a hierarchical key.

### DIFF
--- a/core/bulkchange.class.inc.php
+++ b/core/bulkchange.class.inc.php
@@ -378,7 +378,7 @@ class BulkChange
 			}
 			else
 			{
-				$oReconFilter = new DBObjectSearch($oExtKey->GetTargetClass());
+				$oReconFilter = $oExtKey->GetAllowedValuesAsFilter();
 
 				$aCacheKeys = array();
 				foreach ($aKeyConfig as $sForeignAttCode => $iCol)
@@ -417,7 +417,7 @@ class BulkChange
 				else
 				{
 					// Cache miss, let's initialize it
-					$oExtObjects = new CMDBObjectSet($oReconFilter);
+					$oExtObjects = new CMDBObjectSet($oReconFilter, array(), $oTargetObj->ToArgs());
 					$iCount = $oExtObjects->Count();
 					if ($iCount == 1)
 					{


Hipska:
Hi, can there be any feedback please?

piRGoif:
Hello Thomas,
It's not that easy to review this PR... I guess adding some use cases with some data will help us ?

piRGoif:
I've just sent an email to the products team to get their opinion.

Hipska:
Hi @piRGoif, what is the feedback of the products team?

Molkobain:
Hello @Hipska ,

@piRGoif is out fo the office until monday, @v-dumas wasn't sure to fully understand this, they will get back to you soon.

Hipska:
Let's hope so, I'm waiting for any relevant feedback since september.

piRGoif:
Hello, I need to take some time with our product team to talk about this... Will let you know !

piRGoif:
Ok, sorry to say this but you'll have to wait a few more weeks : there were not much time slots to talk about this face to face, but hourray this is planned now.

piRGoif:
Hello !
I've had a chat with @v-dumas and @rquetiez, and well it was worth it because there was some lack of knowledge on my side :o)

First, congratulations, you did find the right place to modify, which impacts both GUI and CLI imports !

To sum up the discussion we had :

External key filter were designed to add very few constraints in the application : still as of today they are meant to be only helpers, and not guaranty data integrity. If strict integrity control is needed, then the solution is the `\DBObject::CheckToWrite` method (uniqueness rules were implemented this way).
Extkey filters are only used as filters when filling objects data in the GUI, they never were in the import.

Combodo wants the import to be straightforward, of course with data migration in mind! When an administrator processes multiple CSV imports, if an error occurs he will be very confused and will need help to find the proper way to fix. If extkey filters were handled in such a scenario, importing lots of data from another tool can become quickly a nightmare, and you'll need to involve multiple persons to finally push your legacy data into iTop.
But that is not the sole reason: using import to populate/modify as easily as possible the DB is important to us!

Also, extkey filters aren't meant to be the exact truth of an object. During the object lifecycle, you can imagine that the constraint would be more or less valid... Besides legacy data import I talked about earlier, for example sometimes the fields involved can be readonly, and sometimes only in the console not in the portal, etc.

So for all these reasons, we want import to accept as much data as the user need. And so we prefer to reject this PR.

⟳ Possible workarounds as you say is to have 2 columns (both Model name and brand). We can also suggest to change Model friendlyname to include both model name and brand name.
📝 Note that @v-dumas is going to change the import wiki page to explain that extkey filters are ignored and why.
📅 And also, as we of course do understand that having constraints in object data is important, we are aiming to dedicate a release (maybe 3.3 ?) on the (vast !) subject of data integrity. Part of this will be improving the audits. For extkey filters, users would be able to select which one to monitor using an audit, and also customize them (by filtering the objects by state for example).

Thanks again !

Hipska:
Okay thanks for the detailed reasoning.

Imagine we are importing devices, some have a model name and some don't. So for the ones that don't, the import searches for a model named '' (empty string), which of course doesn't exist. If I remember correctly, this PR also solves the issue that this introduces. So, the only way to work around this is to split the csv into one with model name and one without? Please correct me if I'm wrong.

Is there also an easier way to import/sync the Brand name together with the Model name? In my view, I need to add 2 columns that with the same content for this to work: `Brand`, `Model->Brand name` and the actual `Model->Name`.

Important detail in this situation is that there are model names that appears for multiple brands.

piRGoif:
Hello,

> Imagine we are importing devices, some have a model name and some don't. So for the ones that don't, the import searches for a model named '' (empty string), which of course doesn't exist. If I remember correctly, this PR also solves the issue that this introduces. So, the only way to work around this is to split the csv into one with model name and one without?

I'll have to check with the products team... will keep you posted !

piRGoif:
The products team will get back to you directly by commenting here :)

Hipska:
@piRGoif sadly, the products team didn't put any comment yet. Maybe reopen until all questions are resolved? As it might turn out this PR is still valid and could be merged?

piRGoif:
Sad to see this :/ I've send them a notice

Hipska:
@piRGoif any update?

piRGoif:
I can't give you any : I'm not part of the products team.

Hipska:
My question is more like, "Are they still going to get in contact with me?".

And please reopen the PR as there are still unresolved questions.

piRGoif:
To my knowledge this is still the products team that is handling this PR.

Hipska:
Yes, but it's closed, so I think they won't do much..

v-dumas:
Hello Hiska,
We have provided a detailed answered for why we are not going to do this. I don't see what are the open points which do not have an answer yet. Yes, you have to provide **1 extra column** to specify the 'Model->Brand name' on top of 'Brand->Name' and the actual 'Model->Name', that's true but that's not a big deal and remains generic. 
In addition, if you are providing an empty string for 'Model->Brand name' and 'Model->Name' as long as Model is not mandatory, the import will work smoothly. 
You are writing that "importing devices with a brand name, but without model, is not working", I am sorry but this is wrong! Of course the columns 'Model->Name' and 'Model->Brand name' must be **booth** empty, and the column 'Brand->Name' can be documented.
This allows you to import stupid data that's true, but it's a deliberate choice to leave it open and we are not going to change that choice.

Hipska:
> Of course the columns 'Model->Name' and 'Model->Brand name' must be **booth** empty, and the column 'Brand->Name' can be documented.

Okay thanks, that was indeed a part that I missed. Seems that this indeed solves the problems I had:
image
Many thanks!

PR #170, opened 2020-09-28
Advanced search: Fix string criterion contains '0' returning all results

annProg:
criterion `contains 0` will not work as expect
image

The redevelopped query expression should be 
```
SELECT `Server` FROM Server AS `Server` WHERE (`Server`.`friendlyname` LIKE '%0%')
```

The problem is caused by `empty()` function considered "0"(0 as a string) to be empty.

### DIFF
--- a/sources/application/search/criterionconversion/criteriontooql.class.inc.php
+++ b/sources/application/search/criterionconversion/criteriontooql.class.inc.php
@@ -114,12 +114,21 @@ class CriterionToOQL extends CriterionConversionAbstract
 		return addslashes($aValues[$iIndex]['value']);
 	}
 
+	private static function IsEmpty($sValue)
+	{
+		if ($sValue === "0")
+		{
+			return false;
+		}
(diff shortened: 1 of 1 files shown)


Hipska (changes requested):
The correct way to do it would be to use `strlen`.

annProg:
> The correct way to do it would be to use `strlen`.

new commit use `strlen`

Molkobain:
Hello @annProg ,

We reviewed this among the dev team and agreed that it should be merged. We are just checking if a unit test could be added to cover this, then we'll check with the product team if it's okay to put this in the 3.0.

Thanks and sorry for the delay!

Take care,
Guillaume

Molkobain:
Go for integration in 3.0, will be merged just before the beta

Hipska (approved):

Molkobain (approved):

Molkobain:
Corresponding unit test done in 43290c0f

PR #171, opened 2020-09-30
Add curl as optional php extension

Hipska:
I'm open to suggestions for the description text.

### DIFF
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -2154,7 +2154,7 @@ JS
 			],
 			'ldap' => 'LDAP authentication will be disabled.',
 			'mbstring' => 'For CryptEngine implementations, trace in Mail to ticket automation', // N°2891
-			'curl' => 'cURL for data collectors and API connectors',
+			'curl' => 'cURL might be needed for some extensions.',
 		];
 
 		if (utils::IsDevelopmentEnvironment()) {
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -2154,6 +2154,7 @@ JS
 			],
 			'ldap' => 'LDAP authentication will be disabled.',
 			'mbstring' => 'For CryptEngine implementations, trace in Mail to ticket automation', // N°2891
+			'curl' => 'cURL for data collectors and API connectors',
 		];
 
 		if (utils::IsDevelopmentEnvironment()) {


Hipska:
Please, also have a look on https://github.com/Combodo/iTop/commit/7d1a61a2f9655bb53b6447dbe20e3069c2130ecd#commitcomment-42845823 and https://sourceforge.net/p/itop/tickets/1286/

setup/setuputils.class.inc.php
```diff
@@ -2154,6 +2154,7 @@ public static function GetPHPOptionalExtensions() {
 			],
 			'ldap' => 'LDAP authentication will be disabled.',
 			'mbstring' => 'For CryptEngine implementations, trace in Mail to ticket automation', // N°2891
+			'curl' => 'cURL for data collectors and API connectors',
```
Molkobain: IMHO data collectors should not be mentioned as they are not an iTop module and can/should be in separated folder. As for the "API connectors" I don't have a better formulation. Maybe mention that it can be required by some extensions?

Molkobain (dismissed):

Molkobain:
Will be presented to the product team, thanks.

Hipska:
Thanks, can you let them also have a look on https://sourceforge.net/p/itop/tickets/1286/ and give feedback there or here?

Molkobain:
Sure thing.

Molkobain:
Hello Thomas,

We just discussed this within the R&D and Product teams. As for the moment there are only a very few extensions using the cURL lib (Combodo's case exchange, Combodo's webhooks -still in beta-, ITOMIG's webhook, ... ?) we are not confortable with prompting a warning message at each setup to the administrators as 95% of them won't have the need for it.

We would rather have a system in iTop allowing extensions to give their prerequisites, which would be displayed during the setup. I know your PR is quick and cost almost nothing, but we rather keep the setup clean for most administrators and make an extra effort in the extensions documentation to explain that cURL is indeed needed.

As for you post on SourceForge about the APC extension, we are okay to put it in the list of the optional PHP extensions only if the message clearly stands what your are missing by not having it in your system. Do you have a proposal of sentence?

Cheers,
Guillaume

Hipska:
Okay, I agree. Would love to see ability of extensions to define on which php extensions they rely on.

About APC:
No, sorry i'm not that good in crafting good sentences, as you might have seen in this PR. 😂 I will try to do my best.
Ps do you know the correct name of the extension that is needed? Is it only `apcu` or `apcu_bc` as well?

Molkobain:
I'll check it out

Molkobain:
@Hipska seems to be "apcu", @Rorcha07 told me that we had plenty of issues with "apcu_bc" in the past and that we remove it systematically from the webservers now.

Hipska:
I know, a little late, but there it is: #201

PR #172, opened 2020-10-02
Cleanup utils::GetSessionLog()

Hipska:
This eliminates the use of output buffering just for printing an array.

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -946,16 +946,11 @@ class utils
 
 	/**
 	 * Get the _SESSION variable for logging purpose
-	 * @return false|string
+	 * @return string
 	 */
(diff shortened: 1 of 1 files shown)


bruno-ds (approved):

piRGoif:
When I need to do that, I usually call var_export() with the second parameter to true. Are there any reasons to choose between print_r() and var_export() ?

Hipska:
The output of var_export is expected to be used as php source code to load the var as it was.

Print_r is indeed what you want when logging the value of a var.

piRGoif (approved):

Molkobain (approved):

bruno-ds:
good catch 👍

PR #173, opened 2020-10-13
configurable default favorite_orgs

annProg:
If iTop manage lots of Organizations, It will be inconvenient when switch organization at the left upper corner, this pr add default favorite_orgs through config file.

### DIFF
--- a/application/itopwebpage.class.inc.php
+++ b/application/itopwebpage.class.inc.php
@@ -784,7 +784,7 @@ JS
 		if (MetaModel::IsValidClass('Organization'))
 		{
 			// Display the list of *favorite* organizations... but keeping in mind what is the real number of organizations
-			$aFavoriteOrgs = appUserPreferences::GetPref('favorite_orgs', null);
+			$aFavoriteOrgs = appUserPreferences::GetPref('favorite_orgs', MetaModel::GetConfig()->Get('favorite_orgs'));
 			$oSearchFilter = new DBObjectSearch('Organization');
 			$oSearchFilter->SetModifierProperty('UserRightsGetSelectFilter', 'bSearchMode', true);
 			$oSet = new CMDBObjectSet($oSearchFilter);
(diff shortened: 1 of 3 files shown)


Molkobain:
Hello @annProg sorry for the very long delay...
This will be discussed next week during the functional review to see if the feature itself will be accepted.

Regarding the technical aspect, we will most likely rather have an OQL query in the configuration than IDs, but we'll come back to you next week.

Cheers,
Guillaume

Molkobain:
Hello annProg,

We discussed this during today's review and would like several changes for this to included in the upcoming iTop version:
  * Rename the conf. param. to `user_preferences.default_favorite_organizations` as we might tend to have a configurable default param for each user preference
  * Use an OQL instead of an array of IDs, this way we can configure default org to the current user customers for example, it will be very flexible.
  * Finally, with the current implementation a user won't be able to overide the default conf and show all orgs. To fix that, we would like you to add in the preferences page a switch button or radio button so the user can specify whether it wants to use the default conf (OQL should be displayed) or its own selection.

Thanks!
Guillaume

Molkobain:
Hello @annProg , any update?

annProg:
> Hello @annProg , any update?

Sorry for not updating for so long.
My work and life have changed a lot recently. I can't go back to Beijing because of covid-19, and I don't have a PC for iTop development. It is expected that life will return to normal after I move out of Beijing at the end of this year.
So, maybe close this PR, or wait until next year?

Hipska:
@annProg you can mark it as draft if you still plan on working on it..

annProg:
> @annProg you can mark it as draft if you still plan on working on it..

How to mark as draft？

Hipska:
In the right column: 
image

annProg:
@Hipska done, thanks

Molkobain:
No problem @annProg , hope you are well :)
Take care!

PR #174, opened 2020-10-13
support custom favicon like branding logo

annProg:
Same way as logo
```xml
	<classes>
	</classes>
	<branding>
		<favicon _delta="define">opsitop-main/images/favicon.png</favicon>
	</branding>
	<menus>
	</menus>
	<snippets>
	</snippets>
</itop_design>
```


Molkobain:
Hello @annProg ,

Thank you very much for the PR but this is actually being worked on already in the release to come! 😁 It's tracked under bug N°3363 in our private tracker.

For your information, when working on the datamodel like this, we also have to update the iTopDesignFormat::From17To18() and iTopDesignFormat::From18To17() methods to prepare / cleanup the XML when it is being migrated from one XML version to another.

Keep posted for a commit message named "N°3363 - XXX" to know that it is part of the release.

Cheers,
Guillaume

PR #175, opened 2020-10-19
Remove dataprovider in iTopModulesPhpVersionIntegrationTest::testITopModulesPhpVersion

piRGoif:
We had problems with this dataProvider (see 2221e054), but actually we can remove it !

What do you think @odain-cbd and @bruno-ds ?

### DIFF
--- a/test/integration/iTopModulesPhpVersionChecklistTest.php
+++ b/test/integration/iTopModulesPhpVersionChecklistTest.php
@@ -54,7 +54,12 @@ class iTopModulesPhpVersionIntegrationTest extends ItopTestCase {
 		return [['2.8.0', '2.8'], ['3.0.0', '3.0'], ['3.', null], ['3', null]];
 	}
 
-	private function CheckItopModulePhpVersion($sExpectedVersion, $sPhpFile) {
+	/**
+	 * @param string $sPhpFile iTop module file
+	 *
+	 * @return string module version
+	 */
+	private function GetItopModuleVersion(string $sPhpFile): ?string {
 		$sModulePath = realpath($sPhpFile);
 		$sModuleFileName = basename($sModulePath);
 		$sModuleName = preg_replace('/[^.]+\.([^.]+)\.php/', '$1', $sModuleFileName);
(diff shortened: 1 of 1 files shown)


test/integration/iTopModulesPhpVersionChecklistTest.php
```diff
@@ -106,21 +90,10 @@ public function iTopModulesPhpVersionProvider() {
...
-		}
+		$sExpectedVersion = \utils::GetItopMinorVersion().'\.\d+';// ie: 2.7\.\d+   (and yes, the 1st dot should be escaped, but, hey, it is good enough as it, ans less complex to read)
 
-		$aTestCases = array();
 		foreach ($aPhpFiles as $sPhpFile) {
-			$aTestCases[$sPhpFile] = array(
-				'sExpectedVersion' => $sExpectedVersion,
-				'sPhpFile' => $sPhpFile,
-			);
+			$this->CheckItopModulePhpVersion($sExpectedVersion, $sPhpFile);
```
odain-cbd: if one datamodel file test fails immediately without checking all the others after.

it could be interesting to collect all the failures and then report the phpunit test failed with failed file list to correct.

odain-cbd: removing the dataprovider will speed up the test execution. this is a good side effect

piRGoif: Indeed, this is a good idea, remove the \PHPUnit\Framework\Assert::assertRegExp call and return or throw something... I'll try this !

piRGoif: I've just pushed a commit that implements that, see comment below. 
I'm solving this as it is outdated now.

piRGoif:
The test now does only one assert at the end, so that every modules are tested on each test function run (instead of failing on the first error).

All results are stored in an array, and at the end we call an assert to check that the result array is empty.

If there is a difference, the assert message is :

> Some modules have wrong versions ! They should match /^3\.0\.\d+$/

And the actual array contains for example :

```
Array (
    'C:\Dev\wamp64\www\itop-dev/datamodels/2.x/itop-backup/module.itop-backup.php' => '3.9.0'
    'C:\Dev\wamp64\www\itop-dev/datamodels/2.x/itop-portal-base/module.itop-portal-base.php' => '2.8.0'
)
```

odain-cbd (approved):

PR #176, opened 2020-11-06
:sparkles: iTop mutliple url apc_cache

sg-gcouronne:
Hi,

Sorry i tried to think of tests, but it's kind of a difficult case with my time available for the moment. 

It's something that I already implemented in a client, but feel free to improve it !

The case was that multiple urls were allowed in a iTop (using $_SERVER['HTTP_HOST']), but if iTop was setup with http://itop-1.com, then all icons, display_template and especially default images were hardcoded to this url in the apc_cache.
Then if http://itop-1.com was not working anymore (eg.: proxy not working ) and i asked people to use http://itop-2.com, the default_image were still with itop-1, thus generating pages with content not loaded (as itop-1 url was still down).
It will also allow for a public and a private url, each not accessible by the other group of persons (internal having no DNS to public, and public no access to private).

In this PR, I removed the hardcoded link from the compiler (XML) and also from PHP (because there are still some extensions that are not XMLed for the datamodel...), and appending the current absolute url when generating the pages.

Have a great weekend and keep you healty

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -7966,7 +7966,35 @@ class AttributeImage extends AttributeBlob
 	 */
 	public function __construct($sCode, $aParams)
 	{
+		//Remove AbsoluteUrlModulesRoot hardcoded to the default_image
+		if (!empty($aParams) && array_key_exists('default_image', $aParams) && !empty($aParams['default_image'])) 
+		{
+			$sAbsoluteUrlModulesRoot = utils::GetAbsoluteUrlModulesRoot();
+			if (strpos($aParams['default_image'], $sAbsoluteUrlModulesRoot) !== false) 
+			{
+				$aParams['default_image'] = str_replace($sAbsoluteUrlModulesRoot, '', $aParams['default_image']);
+			}
+		}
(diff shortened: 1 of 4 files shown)


Molkobain:
_Note: This has been refered to as Bug N°3677 in our bug tracker_

Molkobain:
@piRGoif will check if the issue is still present as it might have been fixed in 3.0.0

piRGoif:
Yes, I did reproduce the issue : 

1. create a new lmhosts entry
2. run setup using an URL with this entry
3. remove the entry (comment lmhosts declaration + run `nbtstat -R`)
4. access the previously installed iTop instance with locahost

I'll try to report the fix on my install to see if it fixes it !

piRGoif:
Pushed 2 branches with equivalent commits : 
* issue/3677_url_31 based on support/3.1
* issue/3677_url30 base on support/30

Seems like solely changing app_root_url (not running the setup again) is OK now !

Closing this PR and reopening a new one !

piRGoif:
Note that : 
* display template were removed in 3.0.0 with N°3233
* the icon compilation is done now in MFCompiler::GenerateStyleDataFromNode, which calls an ormStyle

PR #177, opened 2020-11-16
N°3253 Fix setup crashing on incompatible PHP versions - JS check

piRGoif:
# What this is

This replaces #155.

This is a lighter solution to the same problem : be able to display an error message in the setup even if using an incompatible PHP version. 
Today we're getting PHP parse error, even for code in deep iTop core (because of requires hell).

# How does it work
The idea is to have a specific setup landing page, that does only this :
* JS : adds an error message print on DOM ready
* PHP : require the real setup index page and capture any PHP error thrown
* JS : cancel the dom ready code and redirects to the setup, otherwise go on with the dom ready code

A POC was made in https://github.com/piRGoif/poc_detect_parse_error

# Behavior

## PHP < 7.0
For example with PHP 5.6 you get :

image

This is thrown by bootstrap.inc.php (added in 3.0.0 with 9e9b192b3c285322e945f0c2558dd5ef7dcf9490)

## PHP >= 7.0.0 && < 7.1.3
A parse error should be thrown by the setup (I restored one of the problematic commit with 47832a90)
With the new setup landing page you get :

image

## PHP >= 7.1.3
The setup is launched normally, but with a different URL
image


piRGoif:
Wooops wrong base branch my bad !
Closing this and creating again..............

piRGoif:
Done in https://github.com/Combodo/iTop/pull/178

PR #178, opened 2020-11-16
N°3253 Fix setup crashing on incompatible PHP versions - JS check (based on fullmoon branch)

piRGoif:
# What this is

This replaces #155 (and it is the same as #177 but with the correct base branch !).

This is a lighter solution to the same problem : be able to display an error message in the setup even if using an incompatible PHP version. 
Today we're getting PHP parse error, even for code in deep iTop core (because of requires hell).

# How does it work
The idea is to have a specific setup landing page, that does only this :
* JS : adds an error message print on DOM ready
* PHP : require the real setup index page and capture any PHP error thrown
* JS : cancel the dom ready code and redirects to the setup, otherwise go on with the dom ready code

A POC was made in https://github.com/piRGoif/poc_detect_parse_error

# Behavior

## PHP < 7.0
For example with PHP 5.6 you get :

image

This is thrown by bootstrap.inc.php (added in 3.0.0 with 9e9b192b3c285322e945f0c2558dd5ef7dcf9490)

## PHP >= 7.0.0 && < 7.1.3
A parse error should be thrown by the setup (I restored one of the problematic commit with 47832a90)
With the new setup landing page you get :

image

## PHP >= 7.1.3
The setup is launched normally, but with a different URL
image


piRGoif:
Force pushed the branch as the setup is fixed (07cc8bf508ae892529e5828c2b49940545604dfa)

Rorcha07:
Omg @piRGoif what is this setup screen ??? my eyes are bleeding !!! 😁

piRGoif:
It's a critical error, meaning the setup (and certainly iTop) cannot execute at all on the system, so... I played a little with CSS :)

Hipska:
I don't like the fact that you have a new and different URL when test succeeds. Can't there be another solution for this? Otherwise users will likely just go to that new URL to start setup.

piRGoif:
The setup does a PHP version check as it always did. So if you call directly `index.setup.php` and you have an incompatible PHP version, the first setup wizard step will be blocked on a error message :

image

_note: normal mini PHP version for iTop 2.7.* is 5.6.0 but I modified it to be able to produce this screenshot_

That is... If you are lucky, and don't have such an old version that any modern code will trigger a parse error... Indeed this setup wizard page is loading lot of PHP code, that could cause parse error even before the PHP version check has the ability to run !
This is the goal of this new page : just check that index.setup.php won't throw a PHP parse error, and if so display a clear error message (in quite a funky way as Stephane pointed out).

Having 2 URL was considered as not an issue when discussing this with Romain...

We could still use the same mean to launch setup that was used in the previous PR (#155) : `include 'index.setup.php';`. This would also require to capture everything that is printed by the script, and flush it if there is an error, cancel if no error. 
But anyway, we would still have the index.setup.php file that could be called manually... So I'm not sure this would be really necessary ?

bruno-ds:
maybe instead of redirecting, you could try to load the `index.setup.php` page inside a "fullscreen" iframe ?
or you could just try an `History.replaceState(null, , '', '/setup')` inside a try/catch block for the compatible browsers?

or you could also juste let the code as it is since its already quite an improvement :)

piRGoif:
Thanks for the hints Bruno (I didn't knew about History.replaceState !)
But anyway the question for me is the same : what would be the benefit ? We would still have 2 files, each could be called directly.

bruno-ds:
> what would be the benefit ? We would still have 2 files, each could be called directly.

the benefit would be to "hide" the second url so users can't easily skip the first one (cf @Hipska _"Otherwise users will likely just go to that new URL to start setup."_).

But as I said, this is not necessary since your code is already quite an improvement \o/

piRGoif:
As you said, "hide" with the double quotes :o) So really I'm not sure this is worse the extra dev effort ?
We will review this PR in the next team meeting, let's have this discussion face to face all together !

piRGoif:
We discussed this with the dev team and we decided to do the following modifications :

* rename index.setup.php to wizard.php
* use the solution Bruno suggested to replace browser history : `History.replaceState` (beware surround this call with a try/catch or do a if before to avoid crashing on browsers not implementing this)
* is using an anonymous function in register_shutdown_function is a good idea ? It may cause problems on old PHP versions...

Molkobain:
We skipped the UI part during the discussion. Shouldn't we stick to something more in the iTop's graphical charter? More over the critical error page is (IMHO) very aggressive.

piRGoif:
Yes I'll see next week with Stephen.
We should have no pain to reuse the iTop CSS and produce a markup with the correct css classes.

piRGoif:
Seems like anonymous function were introduced in PHP 5.3 so indeed it's a good idea not to use them in this case !

piRGoif:
I've pushed all decided modifications but the ability to redirect without changing the URL, as I didn't managed to implement it. @bruno-ds if you can provide any help with that that'l be much appreciated ! O:)

About the UI, I've used standard setup style and now the error message looks like this :

image

piRGoif:
> use the solution Bruno suggested to replace browser history : History.replaceState

We tried yesterday with Bruno, but this can't work as the wizard submits to itself... So adding replaceState on the first wizard page, and you'll submit to /setup/, use the redirection and get... the first wizard step.

We might still use an iframe sized to the viewport full width / height, but we didn't go this way as this would again use resources. The dev team will discuss this on friday again !

piRGoif:
After another team review, the error message was changed to this :

image

The link points to https://www.itophub.io/wiki/page?id=latest%3Ainstall%3Ainstalling_itop#software_requirements

I'm going to merge the PR

PR #180, opened 2020-12-04
ormPassword: allow use of different password hash algo

samsk:
Hi,
  this pull request introduces possibility to change password_hash algorithm via config file, ie. to use new, more advanced one, or to to use older algorithm if PASSWORD_DEFAULT changes.

example config:
```'password_hash_algo' => PASSWORD_ARGON2ID```

Along with this ormPassword::CheckPassword was fixed, because with different hash algo password_get_info() will return string and not integer as algo identificator, what results in code assuming that it is checking a legacy password.

Thank you

### DIFF
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -1450,6 +1450,11 @@ class Config
 	protected $m_aCharsets;
 
 	/**
+	 * @var array Password hash algorithm to use.
+	 */
+	protected $m_iPasswordHashAlgo;
+
+	/**
 	 * Config constructor.
 	 *
 	 * @param string|null $sConfigFile
@@ -1492,6 +1497,7 @@ class Config
 		$this->m_sExtAuthVariable = DEFAULT_EXT_AUTH_VARIABLE;
 		$this->m_aCharsets = array();
 		$this->m_bQueryCacheEnabled = DEFAULT_QUERY_CACHE_ENABLED;
+		$this->m_iPasswordHashAlgo = PASSWORD_DEFAULT;
 
 		//define default encryption params according to php install
 		$aEncryptParams = SimpleCrypt::GetNewDefaultParams();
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: core/config.class.inc.php [hunk 1]
```diff
@@ -1449,6 +1449,11 @@ public function Get($sPropCode)
 	 */
 	protected $m_aCharsets;
 
+	/**
+	 * @var array Password hash algorithm to use.
```
Hipska:
```suggestion
	 * @var integer Password hash algorithm to use.
```

core/ormpassword.class.inc.php
```diff
@@ -96,18 +97,18 @@ public function CheckPassword($sClearTextPassword)
...
-				}
-				break;
-			default:
-				$bResult = password_verify($sClearTextPassword, $this->m_sHashed);
+			//unknown, assume it's a legacy password
+			$sHashedPwd = $this->ComputeHash($sClearTextPassword);
+			if ($this->m_sHashed == $sHashedPwd)
+			{
+				$bResult = true;
+			}
```
Hipska: This suggestion is actually a question towards Combodo if this is wanted or not.
```suggestion
			$bResult = ($this->m_sHashed == $sHashedPwd);
```

Hipska (changes requested):

core/ormpassword.class.inc.php
Molkobain: Yep, 💯

Molkobain:
Sounds like a nice feature improvement!

That being said I'm not an expert on that area of the code, so I can't approve the PR. But looks like you took care of the previously stored passwords and the backward compatibility right?

samsk:
Yes, backward compatibility is preserved. This only changes how password hashes are created, so hashes generated with different algo will still work.

Hipska (dismissed):

steffunky:
Hello,

These changes seem good to me, I'll wait for the R&D team to review it but it would be a nice addition to iTop and it's perfectly integrated with the way we intended to handle password with 2.6.0 👍
We may need to add a unit test as ormPassword is becoming configurable and may have to handle legacy passwords and maybe passwords with different hash algorithms.

steffunky:
Hi,
We just reviewed this feature and we think it would be safer to write a test along with this change.
I can find time in the next couple of weeks to write it and the team will review the test and merge this PR.

core/ormpassword.class.inc.php
Hipska: Okay, @samsk, can you apply the suggestion?

piRGoif:
Test was proposed with https://github.com/promitilus/iTop/pull/1
Awaiting this to be merged... @samsk please poke here when done !

Hipska (dismissed):

samsk:
Sorry that you've to wait, I missed somehow that pull notification.

Hipska (approved):

Molkobain (approved):

Molkobain:
Seems all good to me. Don't forget to create/update a/the corresponding ticket in the internal bug tracker.

steffunky (approved):

steffunky:
Merged, thanks for your contribution!
Internal bug ref: N°3662

PR #181, opened 2020-12-21
🌐 Dutch translations: distinguish between "approval" and "acceptance"

jbostoen:
* distinguish between approval (goedkeuring) <=> acceptance (acceptatie). Both were translated as "datum goedkeuring" till now.

@Hipska agree? Or different translation suggestion?
@piRGoif not sure if I should do the pull request for "develop" or perhaps "support/2.7" or both?


piRGoif:
Hello,
Many thanks @jbostoen !
I'm adding the question label, awaiting for Thomas confirmation.

About the base branch, you can use support/2.7 so that this will be integrated in both future 2.7.4 and 3.0.0 versions. 
The develop branch will only be used by the 3.0.0 version.

Hipska:
When is this acceptance comment requested? Does that happen while approving?

Edit: it seems to be used in the validate step?

jbostoen:
> 
> 
> When is this acceptance comment requested? Does that happen while approving?
> 
> Edit: it seems to be used in the validate step?

In an ITIL normal change.
Now it just lists "datum goedkeuring" twice in the details view.

jbostoen:
Can't change branch here. Will make a new pull request.

PR #185, opened 2021-01-08
DBSearch::Init did not respect default obsolete value

Hipska:
By using this a (core/extension) developer always needs to call `DBSearch::SetShowObsoleteData` in order to respect the setting. With this change, that is done automatically.

If this is not accepted, at least the following should be put in place:
```php
$this->m_bShowObsoleteData = MetaModel::GetConfig()->Get('obsolescence.show_obsolete_data');
```

### DIFF
--- a/core/dbsearch.class.php
+++ b/core/dbsearch.class.php
@@ -91,7 +91,7 @@ abstract class DBSearch
 	protected function Init()
 	{
 		$this->m_bArchiveMode = utils::IsArchiveMode();
-		$this->m_bShowObsoleteData = true;
+		$this->m_bShowObsoleteData = utils::ShowObsoleteData();
 	}
 
 	/**


Molkobain:
Personally I'm 💯 with this PR as it should have been in the ORM from the beginning.
But it might introduce "regressions" as it might change the results of some OQLs.

Poke @v-dumas @rquetiez @etaloc

Hipska:
The same regression happened when objects started to get archived, that was accepted as well, so that should be no problem then.

v-dumas:
Doesn't it reverse the paradigme? Before: every search aimed to be displayed to a user, had to take this parameter into account (with a manual call DBSearch::SetShowObsoleteData) and all other searches made for internal purposes, had nothing to do to include obsolete data.

I am wondering what will happen to a deletion plan, which because of this change may ignore obsolete objects, thus generate errors or database incoherence. Maybe "AllowAllData()" already address this and is always called on deletion plans.

Another risk, is all existing extensions/customization which on purpose did not call DBSearch::SetShowObsoleteData(), would have to be changed, is that correct?

Molkobain:
> Doesn't it reverse the paradigme? Before: every search aimed to be displayed to a user, had to take this parameter into account (with a manual call DBSearch::SetShowObsoleteData) and all other searches made for internal purposes, had nothing to do to include obsolete data.

Very good point, all internal processing (core or extensions) would be affected if their code is not migrated.

Hipska:
Correct indeed. As I said, this is similar as to the introduction of the AllData methods; If you didn't use that, you would have less results than before. So one would need to update the code to take the change into account.

bruno-ds:
From what I understood, it was a design choice made by @rquetiez .
IMHO, while it is quite seducing to have a cleaner API, changing this precise behavior would have tremendous impacts on all the pre-existing code!

Hipska:
Nobody is disagreeing on that. So indeed, I understand this will not be merged/adapted right away, but will take a long time or a project to take care of the impacts of it.

piRGoif:
We need @rquetiez opinion on this... I'll ask him !

Molkobain (dismissed):

Molkobain:
I personally approve this PR, it should be done and work just like the AllowAllData() flag/method. Meaning that:
- We will have to find and fix all "internal" queries that should behave without restrictions
- We will no longer have bugs like "Hey that part is showing obsolete data when it shouldn't", much more robust in the future

But I don't think now is the time to merge this, we have so much on going with the upcoming 3.0 that we can't spare time looking for this queries.

So IMHO
- we should plan this on a future release, maybe 3.1 or 3.2
- @Hipska , adding to the PR fixes on the concerned queries would ease pushing it forward toward acceptance

Hipska:
Wow, that puts some pressure on me 😛

How could I identify the queries that need this?

Molkobain:
Well, that's a good question 😅

Maybe looking for queries using the AllowAllData flag would be a good start?

piRGoif:
Hello,
I've just had a chat with @rquetiez :)

He told me this : [obsolescence](https://www.itophub.io/wiki/page?id=latest%3Afeature%3Aobsolescence) was added in iTop 2.4.0, and it came also with [archiving](https://www.itophub.io/wiki/page?id=latest%3Afeature%3Aarchiving). Those functionalities address different needs but aims to work together.

We can sum-up by saying : 

* Purge : remove objects to decrease DB size and improve performances. This is a radical operation !
* Archiving : we could also call this a "soft delete" as the objects are non longer visible anywhere in the application by default, but are still present in the DB. Helps on performance issues, and allows admin to rollback if needed.
* Obsolescence : just a flag to be used in the console, to mark or hide objects in certain screens. Objects are still visible in the rest of the application. Allow vast variety of users to be aware that an object has no activities and might need to be archived/purged.

Obsolescence was designed as this from the ground, having archiving and purge in mind.

Do you still think this principle should be changed ? If so, could you list some uses cases where this was a problem to you ? Any arguments to change our minds ?

Hipska:
The principle should not be changed. The only problem is that this option is often forgotten by devs who implement new console features (either in core iTop or in extensions). See for example the mentioned issue in itop-time-tracking module.

piRGoif:
We should have a chat with the rest of the dev team to clear this up :)
Will keep you informed !

piRGoif:
Will be reviewed b ythe team in a dedicated meeting in 2 weeks.

piRGoif:
Hello,
We did discussed about this !

We are rejecting this PR right now : we still want to keep obsolescence as "opt-in", that means consumers need to specify if they want to enable this filter. Indeed there are few consumers which need this filter enabled as explained above (some screens in the console and that's it).
In opposition archiving is "opt-out" : filter is on by default, consumer need to call `SetArchiveMode` to disable it. Same for user rights of course (with the `AllowAllData` method to disable filter).

But @eespie suggested an interesting idea : deal with obsolescence filter user preference the same way we are filtering organization silo, by adding it to the ApplicationContext mechanism.
This would be a nice compromise, as we would still be "opt-in", but all the screens where we want this filter applied would be impacted without modification.

We referenced this as N°3759. Not planned yet, but won't be in 3.0 for sure.

Thanks again @Hipska  !

PR #186, opened 2021-01-22
Feature/delegated user management

dflaven:
- Hide "administrator" users and the "Administrator" profile from non-admin users
- Fix for N°2841 - Let non admin users (with the appropriate rights, i.e. "User Manager" profile) delete users.


dflaven:
Not good, the js file has nothing to do in this PR !!!

PR #188, opened 2021-01-26
Integrate the current values present the column into the column defin…

dflaven:
Make sure that update of the DBSchema does not break when an ENUM column contains values which are not (or no longer) part of the field definition.

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -5212,7 +5212,7 @@ class AttributeEnum extends AttributeString
 				$sHostClass = $this->GetHostClass();
 				$sSQLTableName = MetaModel::DBGetTable($sHostClass, $this->GetCode());
 			}
-			$aValues = array_unique(array_merge($aValues, $this->GetActualValuesInDB()));
+			$aValues = array_unique(array_merge($aValues, $this->GetActualValuesInDB($sSQLTableName)));
 		}
 
 		if (count($aValues) > 0)
@@ -5266,7 +5266,6 @@ class AttributeEnum extends AttributeString
 					$aValues[] = $aRow['value'];
 				}
 			}
-			IssueLog::Debug('Actual values for '.$sHostClass.'::'.$this->GetCode().': '.print_r($aValues, true));
 		}
 		catch(MySQLException $e)
 		{
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -5179,6 +5179,20 @@ class AttributeEnum extends AttributeString
 
 	protected function GetSQLCol($bFullSpec = false)
 	{
+		// Get the definition of the column, including the actual values present in the table
+		return $this->GetSQLColHelper($bFullSpec, true);
+	}
+
+	/**
+	 * A more versatile version of GetSQLCol
+	 * @since 3.0.0
+	 * @param bool $bFullSpec
+	 * @param bool $bIncludeActualValues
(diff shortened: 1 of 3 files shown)


bruno-ds:
So smart!

Is there any easy way to produce a flash message alerting the admin that the DB has such rogue values?

I think you can easily collect them using this before the array_merge:
```php
$aRogueValues = array_diff($this->GetActualValuesInDB(), $aValues);
```

core/attributedef.class.inc.php
```diff
@@ -5206,6 +5210,28 @@ protected function GetSQLCol($bFullSpec = false)
 		}
 	}
 
+	/**
+	 * Get the list of the actual 'enum' values present in the database
+	 * @return string[]
```
Molkobain, line 5253: Please add a `@since 3.0.0` so we know it's only available from there (if not accepted in the 3.0.0 we can always change it later before merging the PR)

core/attributedef.class.inc.php
```diff
@@ -5188,6 +5188,10 @@ protected function GetSQLCol($bFullSpec = false)
 		{
 			$aValues = array();
 		}
+
+		// Preserve the values already present in the database to ease migrations
+		$aValues = array_unique(array_merge($aValues, $this->GetActualValuesInDB()));
```
Molkobain: I'm no expert on this part of the app. but looks like the `AttributeDefinition::GetSQLCol()` gets called in `AttributeDefinition::GetSQLColumns()` which is called in the "Datamodel viewer" page and when synchros are executed in "SynchroExecution::PrepareProcessing()" (among other less sensitive places).

Wouldn't this cause significant slowdowns on high volume DBs? Or will it use the query cache? Either way, this should be tested on big DBs, I find it a bit dangerous without an explicit cache of this somewhere at compile time.

Molkobain:
That could indeed be a big step forward as it is a real pain in the 🍍 right now, for both the community and those using the ITSM Designer!

Just concerned about the performance consequences though.

> ...
> Is there any easy way to produce a flash message alerting the admin that the DB has such rogue values?
> ...

A flash message for the admin to see right after setup/MTP would be a good idea indeed.
It should also appear in the DB tools IMHO, as it is something to fix.

Molkobain:
_(Self-assigned because it has be to be assigned to someone so it can be discussed in the team meeting)_

jbostoen:
This would indeed be a big step forward! I've seen it several times. A proper warning would be really helpful.

core/attributedef.class.inc.php
dflaven: Done.

core/attributedef.class.inc.php
dflaven: I replaced the call in the Datamodel viewer with the static method IsBasedOnDBColumns() which seem sufficient for what it was doing.

For the data synchro, getting the values dynamically may be useful/needed anyway...

odain-cbd:
please wait before merging in develop. I reported an issue when migration from 2.7 to this current branch.

Molkobain:
Hello @odain-cbd , can you tell us more about the issue? Do you have a bug number?

Molkobain:
Discussed with @odain-cbd , the issue is fixed, we can move forward with this PR. I'll contact Denis ASAP.

Molkobain:
Denis is currently working on it to improve the data synchro part.

Molkobain:
Discussed with Denis, all good!

PR #190, opened 2021-02-02
N°4092 setup generates symlinks when run on dev env

piRGoif:
## Introduction

When developing, we often have to launch the setup instead of the toolkit : to select a new module to install for example. Doing so will cause symlinks to be removed, and so we need to run toolkit again afterwards.

## New functionalities added
With this PR : 

* The compiler will now generate low level (in `\MFCompiler::DoCompile`) a flag when symlinks are set (`data/.compilation-symlinks` file)
* In the setup, new "symlink" checkbox (in the misc params step, the one containing approot & graphviz path)
  * displayed if and only if : 
    * running on a dev env (`\utils::IsDevelopmentEnvironment`)
    * the symlink PHP function exists
    * symlink flag is set
  * unchecking the checkbox will remove the flag using ajax (so we don't have to save this option value all along the wizard steps)
  * if symlink flag is present, then compile with the symlinks option
* The flag is used (same conditions as the setup) in all connectors that are not setting the UseSymlinks option : Core Update, Designer, Hub
* For consumers that are setting the option (toolkit, custom scripts, ...), behavior is unchanged and the flag is set/unset according to the option value

Setup misc step screenshot with the new option : 
image

## What changed in the code
Compilation is launched in \MFCompiler::Compile.
The default $bUseSymbolicLinks parameter value is changed from false to null so that we know if the consumer has set this option, and this default value is changed also in the callee.

At the top of the method a test is added to set the option based on the flag.

We have 2 main consumers of this method : 
* `ApplicationInstaller` for the setup (ExecuteStep then Compile methods)
* `RuntimeEnvironment` for anything else : toolkit, connectors, ...

Note that for the setup in 5c7604f9 a param was used in `\ApplicationInstaller::ExecuteStep` to set the `$bUseSymbolicLinks` boolean value... but I couldn't find a documentation on how to use this parameter ? This code is left unchanged.

### DIFF
--- a/setup/applicationinstaller.class.inc.php
+++ b/setup/applicationinstaller.class.inc.php
@@ -276,17 +276,16 @@ class ApplicationInstaller
 					$sTargetDir = $this->GetTargetDir();
 					$bUseSymbolicLinks = false;
 					$aMiscOptions = $this->oParams->Get('options', array());
-					if (isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks'])
-					{
-						if (function_exists('symlink'))
-						{
+					if (isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks']) {
+						if (function_exists('symlink')) {
 							$bUseSymbolicLinks = true;
 							SetupLog::Info("Using symbolic links instead of copying data model files (for developers only!)");
-						}
-						else
-						{
+						} else {
 							SetupLog::Info("Symbolic links (function symlinks) does not seem to be supported on this platform (OS/PHP version).");
 						}
+					}
+					if (utils::IsDevelopmentEnvironment()) {
+						$bUseSymbolicLinks = true;
 					}
 
 					self::DoCompile($aSelectedModules, $sSourceDir, $sExtensionDir, $sTargetDir, $sTargetEnvironment,
--- a/setup/applicationinstaller.class.inc.php
+++ b/setup/applicationinstaller.class.inc.php
@@ -284,7 +284,8 @@ class ApplicationInstaller
 							SetupLog::Info("Symbolic links (function symlinks) does not seem to be supported on this platform (OS/PHP version).");
 						}
 					}
-					if (utils::IsDevelopmentEnvironment()) {
+
+					if (MFCompiler::HasUseSymlinksFile()) {
 						$bUseSymbolicLinks = true;
 					}
 
--- a/setup/applicationinstaller.class.inc.php
+++ b/setup/applicationinstaller.class.inc.php
@@ -285,7 +285,7 @@ class ApplicationInstaller
 						}
 					}
 
-					if (MFCompiler::HasUseSymlinksFile()) {
+					if (MFCompiler::IsUseSymbolicLinksFlagPresent()) {
 						$bUseSymbolicLinks = true;
 					}
 
--- a/setup/applicationinstaller.class.inc.php
+++ b/setup/applicationinstaller.class.inc.php
@@ -276,7 +276,7 @@ class ApplicationInstaller
 					$sTargetDir = $this->GetTargetDir();
 					$bUseSymbolicLinks = false;
 					$aMiscOptions = $this->oParams->Get('options', array());
-					if ((isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks']) || MFCompiler::IsUseSymbolicLinksFlagPresent()) {
+					if ((isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks'])) {
 						if (function_exists('symlink')) {
 							$bUseSymbolicLinks = true;
 							SetupLog::Info("Using symbolic links instead of copying data model files (for developers only!)");
--- a/setup/applicationinstaller.class.inc.php
+++ b/setup/applicationinstaller.class.inc.php
@@ -276,17 +276,13 @@ class ApplicationInstaller
 					$sTargetDir = $this->GetTargetDir();
 					$bUseSymbolicLinks = false;
 					$aMiscOptions = $this->oParams->Get('options', array());
-					if (isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks']) {
+					if ((isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks']) || MFCompiler::IsUseSymbolicLinksFlagPresent()) {
 						if (function_exists('symlink')) {
 							$bUseSymbolicLinks = true;
 							SetupLog::Info("Using symbolic links instead of copying data model files (for developers only!)");
 						} else {
 							SetupLog::Info("Symbolic links (function symlinks) does not seem to be supported on this platform (OS/PHP version).");
 						}
-					}
-
-					if (MFCompiler::IsUseSymbolicLinksFlagPresent()) {
-						$bUseSymbolicLinks = true;
 					}
 
 					self::DoCompile($aSelectedModules, $sSourceDir, $sExtensionDir, $sTargetDir, $sTargetEnvironment,
--- a/setup/applicationinstaller.class.inc.php
+++ b/setup/applicationinstaller.class.inc.php
@@ -274,8 +274,9 @@ class ApplicationInstaller
 					$sExtensionDir = $this->oParams->Get('extensions_dir', 'extensions');
 					$sTargetEnvironment = $this->GetTargetEnv();
 					$sTargetDir = $this->GetTargetDir();
-					$bUseSymbolicLinks = false;
 					$aMiscOptions = $this->oParams->Get('options', array());
+
+					$bUseSymbolicLinks = null;
 					if ((isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks'])) {
 						if (function_exists('symlink')) {
 							$bUseSymbolicLinks = true;
@@ -500,8 +501,7 @@ class ApplicationInstaller
 	{
 		$oBackup = new SetupDBBackup($oConfig);
 		$sTargetFile = $oBackup->MakeName($sBackupFileFormat);
-		if (!empty($sMySQLBinDir))
-		{
+		if (!empty($sMySQLBinDir)) {
 			$oBackup->SetMySQLBinDir($sMySQLBinDir);
 		}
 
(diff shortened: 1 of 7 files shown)


Molkobain:
IMHO this could be a good idea but we must stick to some principles:

**1** - There must be a checkbox in the setup screen so we can choose whether or not to preserve the symlinks, this for 2 reasons:
- We often run the setup to clean everything up, including removing symlinks.
- When working with a particular deployment configuration / set-up, includes/requires don't work well with symlinks so we need to be able to work with a "standard" file system (hence the previous point).

**2** - As you stated, this option should only be proposed on a dev environment, that part is covered by the PR 😊

piRGoif:
I understand the need to be extra careful and make this enabled only when needed ! 👍

But I'm not very found of the solution to add a checkbox in the setup, mainly as this will gives too much visibility to this functionality. And also, I can't decide where to put this option... 😕

Here is my suggestion 😣 : 
* change the toolkit : with the symlink option, create a `data/.use_symlinks` file, otherwise remove this file. It could be created manually also of course !
* during the setup, check if the files exists, if so compile to symlinks

This will allow to set the file manually when needed.
And if running the toolkit with the symlink option, and the you need to run the setup again, you won't remove the symlinks anymore.

I hope we'll be able to discuss this during the next dev team 🙏

Hipska:
Sounds like a good solution, this way the checkbox on toolkit could also be checked if refreshing the page and you used symlinks the last time.

piRGoif:
Team agreed to : 
* change the setup test to check for a file (for example `/data/.compilation-symlinks`)
* add a checkbox in the setup, and to avoid problems with coding information transport between wizard steps, simply create/remove file on click
* add the file creation in the toolkit

Now just need to review with @Molkobain if he's ok with this :)

piRGoif:
After discussing with @Molkobain and @bruno-ds, some more details : 
* the file will be created low level in iTop core (`\SetupUtils::copydir` for example ?) so that any consumer will be impacted
* the best place to add the checkbox would be in the "additional parameters" step, in a dedicated fieldset "dev parameters", displayed only if `\utils::IsDevelopmentEnvironment` is true or if the file is already present

Molkobain:
Bruno just modified the compiler (f328e58c025f093f68f0329969da05f5d4e208cb) to restore the toolkit behavior of deleting the env-production content before generating it again. You might want to test that it has no impact on your changes.

bruno-ds:
> Bruno just modified the compiler ([f328e58](https://github.com/Combodo/iTop/commit/f328e58c025f093f68f0329969da05f5d4e208cb)) to restore the toolkit behavior of deleting the env-production content before generating it again. You might want to test that it has no impact on your changes.

This is a little more subtle: the `env-*` was indeed sometimes purged by the old toolkit even if you selected the symlink method, but in fact it was only because of a compilation without symlink happening only during the rendering of the index page and making it so unbearable.
(So consecutive compilations with the symlink option would not have emptied the `env-*` either)

piRGoif:
I've just implemented the modifications decided with the team. 
I removed the draft flag, PR is ready for review !

piRGoif:
About the place where to write the file, here is the call stack : 

```
\ApplicationInstaller::ExecuteStep
  \ApplicationInstaller::DoCompile
    \MFCompiler::Compile
      \MFCompiler::DoCompile
        \SetupUtils::copydir
```

copydir is called for each module... by a loop in DoCompile ! Therefore I chose to add the usesymlinks file generation / removal in the later.

setup/applicationinstaller.class.inc.php
```diff
@@ -276,19 +276,19 @@ public function ExecuteStep($sStep = '', $sInstallComment = null)
...
-						}
-						else
-						{
+						} else {
 							SetupLog::Info("Symbolic links (function symlinks) does not seem to be supported on this platform (OS/PHP version).");
 						}
 					}
 
+					if (MFCompiler::HasUseSymlinksFile()) {
+						$bUseSymbolicLinks = true;
```
bruno-ds: shouldn't you reuse the above condition?
`if (function_exists('symlink')) {`

as an alternative, you could plug the `MFCompiler::HasUseSymlinksFile()` as a OR condition to `if (isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks']) {`

ie: 
```
if ((isset($aMiscOptions['symlinks']) && $aMiscOptions['symlinks']) || (MFCompiler::HasUseSymlinksFile())) {
```

setup/compiler.class.inc.php
```diff
@@ -90,54 +98,81 @@ protected function Log($sText)
...
 		return $this->aLog;
 	}
-	
+
+	/**
+	 * @return bool
+	 * @uses USE_SYMBOLIC_LINKS_FILE_PATH
+	 * @since 3.0.0
+	 */
+	public static function HasUseSymlinksFile(): bool
```
bruno-ds: This is not an important matter, but I would have used something like `IsSymlinksForced`.
For sure, the behavior's trigger is less understandable this way, but the behavior itself is more understandable, and I thinks this is what count for an API.

setup/compiler.class.inc.php
```diff
@@ -90,54 +98,81 @@ protected function Log($sText)
...
+	{
+		return (file_exists(static::USE_SYMBOLIC_LINKS_FILE_PATH));
+	}
+
+	/**
+	 * @uses USE_SYMBOLIC_LINKS_FILE_PATH
+	 * @since 3.0.0 method creation
+	 */
+	public static function ToggleUseSymlinksFile(bool $bUseSymbolicLinks): void
```
bruno-ds: I don't understand: why do you use the word `Toggle` and not `Set`?

setup/wizardsteps.class.inc.php
```diff
@@ -1161,6 +1161,30 @@ public function Display(WebPage $oPage)
...
+			$oPage->add('<fieldset>');
+			$oPage->add('<legend>Dev parameters</legend>');
+			$oPage->p('<input id="use-symbolic-links" type="checkbox" checked><label for="use-symbolic-links">&nbsp;Create symbolic links instead of creating a copy in env-production (useful for debugging extensions)');
+			$oPage->add('</fieldset>');
+			$oPage->add_ready_script(<<<'JS'
+$("#use-symbolic-links").on("click", function() {
+	var $this = $(this),
+		bUseSymbolicLinks = $this.prop("checked");
+	if (!bUseSymbolicLinks){
+		if (!window.confirm("This will disable symbolic links generation permanently, are you sure ?")) {
```
bruno-ds: Since the nominal case is to not have symboloink links, I propose:
>  "This will disable symbolic links generation (you till can re-enable them using the toolkit), are you sure ?"

bruno-ds (commented):
As a side note: having code cleanup mixed with real modification is making the review harder.

setup/compiler.class.inc.php
piRGoif: I suggest IsUseSymbolicLinksFlagPresent

setup/compiler.class.inc.php
piRGoif: commited in f9d8e91b

setup/compiler.class.inc.php
piRGoif: commited in f9d8e91b

setup/applicationinstaller.class.inc.php
piRGoif: Not sure to understand, we'll have to talk about this

setup/wizardsteps.class.inc.php
piRGoif: Here is my proposition : 

image

piRGoif: Commited with 7d96a761

setup/compiler.class.inc.php
piRGoif: Reviewed with Bruno : he's ok :)

setup/compiler.class.inc.php
piRGoif: Reviewed with Bruno : he's ok :)

setup/wizardsteps.class.inc.php
piRGoif: Reviewed with Bruno : he's ok :)

setup/applicationinstaller.class.inc.php
piRGoif: Commited in 7722d266 
Thanks Bruno !

piRGoif:
This PR has been reviewed by the dev team for the second time.
Conclusions : 
* we should use the flag ONLY when \utils::IsDevelopmentEnvironment
* need to test a deployement with the Designer connector (I did test setup, toolkit and hub connector, and even having a module in the data/production-modules dir, but not really the designer connector :/)

piRGoif:
Rebased the branch as there were LOTS of commit on the base branch

piRGoif:
About the team review : I modified \MFCompiler::IsUseSymbolicLinksFlagPresent to always return false if not in a dev env ✔.

Concerning Hub connector and Designer connector : I had a chat with Denis. He doesn't think there could be problems using symlinks in those contexts. He just warned me to check if the call stack will trigger my code in all situations ! Cause... connectors and toolkit uses \RunTimeEnvironment base class or overrides, and some of them overwrites whole methods. I did check, and actually I thought I had checked the Hub Connector but I didn't :/ The hub connector always resets symlinks, because the flag I added was in \ApplicationInstaller::ExecuteStep. Toolkit doesn't go though there :/

Here is a list of \MFCompiler::Compile callers by consumers :

* Setup : \ApplicationInstaller::ExecuteStep
* Toolkit : \RunTimeEnvironment::CompileFrom
* Hub Connector : \HubRunTimeEnvironment::CompileFrom()
* Designer Connector : \RunTimeEnvironmentDesignerConnector::CompileFrom()
* Core Updater : \RunTimeEnvironmentCoreUpdater::CompileFrom()

Maybe one solution would be to have a dedicated RuntimeEnvironment impl for the toolkit, and for all others use the flag ?
I have to think more about that 🤔

piRGoif:
Rebased branch, push some old modifications that wasn't yet pushed, and updated PR description !

piRGoif:
Rebased branch O:)

piRGoif:
Since commit b3876b78 the flag is used in all connectors impl.
I've modified this PR's description.
It is now ready for dev team review !

piRGoif:
Pushed some fix to numerous bugs :/
Also modified the description to allow an easier review !

piRGoif:
Reviewed with the team : 
* OK to merge 🎉
* But : 
  * add the `data/.compilation-symlinks` file in the repo so that any git clone will work with symlinks (beware not to include it in the packaging !)
  * create a new PR to add a configuration parameter to trigger dev env (`\utils::IsDevelopmentEnvironment`)

piRGoif:
Combodo internal ref : N°4092

PR #191, opened 2021-02-04
iTop 3.0.0 badge dashlet : make whole badge clickable

piRGoif:
**Before :** only the text link was clickable (object name, "create" link)

**Now :** the whole badge is clickable.
* When clicking on the "create" label (the corresponding A tag actually) the create form opens.
* Otherwise the list appears.


Molkobain (changes requested):
This would bring a way better UX to the badges indeed! ✌

Just a few remark regarding the current implementation to better match the UIBlock concepts and guidelines:
1. All the JS should be in a dedicated jQuery UI widget instantiated in the `layout.js.twig` of the block, not in a separated file, otherwise badges won't work on others pages (Friendly reminder: Blocks must be atomic and independant, meaning that they must work in any page)
2. Use the `data-role` attributes for JS selectors to keep segregation between style and interaction (you can check examples on other blocks)
3. Remove the HTML tags from the selectors (eg. `div`), again for the sake of the segregation as the markup could evolve.
4. Note: The jQuery UI widget provides an easier way to interact with elements within the block, making selectors simpler/clearer. (see points 2. and 3.)

I admit that I'm not fond of using all this JS just to work around the limitation of having an anchor within an anchor; but in this particular case I think it's worth it (and we don't have another choice 😁)

For the record, another (lighter?) implementation could be to have to "main" (list) anchor around the whole badge, than a `<span href="xxx" onclick="Someting to redirect here;">` instead of the anchor but:
- It will not be great for accessibility as the span could be focused
- It would be a shady hack not as easy to maintain as something following the blocks guidelines

application/dashboard.class.inc.php
```diff
@@ -567,6 +567,21 @@ public function Render($oPage, $bEditMode = false, $aExtraParams = array(), $bCa
...
+			$oPage->add_ready_script(<<<'JS'
+$("div.ibo-dashboard").on("click", "div.ibo-dashlet-badge", function(event) {
+	let $eventTarget = $(event.target);
+
+  if ($eventTarget.is("div.ibo-dashlet-badge--actions > a")) {
+    return;
+  }
+
+	let $listLink = $eventTarget
+    .closest("div.ibo-dashlet-badge")
```
piRGoif: Using `$(this)`should allow to remove this DOM browsing !
TODO tomorrow !

piRGoif:
We'll need to have a chat about your propositions so I get a clear picture :)

piRGoif:
I forgot to say : currently there is just one click handler set in the dashboard code. I would prefer to keep it that way, as having one handler per page is far better than an handler per badge dashlet !

But of course this code is of the DashletBadge component responsibility... So we might add a method in DashletContainer to be implemented in each dashlet components, and having the dashboard code to call this method on each dashlet impl to get the JS code to return ?

piRGoif:
Discussed with @Molkobain : having only one handler on the dashboard level is nice, but we need to keep code responsability which is badge dashlet component.

So we agreed to : 
* create a dashlet badge JQuery widget
* instantiate this widget in all badge instance using a JS twig file

We will later on (maybe 3.1.0 ?) improve the API by adding a mean for components to add "static" JS code (added only once event if multiple instances provided). Combodo internal ref N°3710

piRGoif:
I pushed the last version that use : 
* a JQuery widget, instanciate in each component (JS twig)
* data-role attributes as selectors

Note that a badge dashlet contains 2 containers : one to contain the whole dashlet, and one again created inside the component (?!??)
So to get to the highest container I had to do weird stuffs :/ (see 209b91bd)

js/components/dashlet-badge.js
```diff
@@ -0,0 +1,69 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ */
+
+;
+$(function () {
+	$.widget('itop.dashletBadge',
```
Molkobain: Convention is (see other blocks)
```suggestion
	$.widget('itop.dashlet_badge',
```

js/components/dashlet-badge.js
```diff
@@ -0,0 +1,69 @@
...
+		{
+			options: {},
+			js_selectors:
+				{
+					dashletContainer: '[data-role="ibo-dashlet"]',
+					dashletActions: '[data-role="ibo-dashlet-badge--actions"]',
+					dashletActionList: '[data-role="ibo-dashlet-badge--action-list"]'
+				},
+			dashletContainer: null,
```
Molkobain: No camel case properties, only snake case, see other blocks

sources/application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -12,6 +12,10 @@ class DashletBadge extends DashletContainer
 {
 	public const BLOCK_CODE = 'ibo-dashlet-badge';
 	public const DEFAULT_HTML_TEMPLATE_REL_PATH = 'base/components/dashlet/dashletbadge';
+	public const DEFAULT_JS_TEMPLATE_REL_PATH = 'base/components/dashlet/dashletbadge';
```
Molkobain: Template should be `dashlet-badge` (even though I understand you followed the naming from the HTML template, which should be fixed as well; if you could fix it while you are at it, it would be great 👍)

sources/application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -12,6 +12,10 @@ class DashletBadge extends DashletContainer
...
+	public const DEFAULT_JS_FILES_REL_PATH = [
+		'js/components/dashlet-badge.js',
```
Molkobain: Could you put this file in a `dashlet` sub-folder just like for the templates / stylesheets?

sources/application/UI/Base/Component/Dashlet/DashletContainer.php
```diff
@@ -14,4 +14,11 @@ class DashletContainer extends UIContentBlock
 {
 	public const BLOCK_CODE = 'ibo-dashlet';
 	public const DEFAULT_HTML_TEMPLATE_REL_PATH = 'base/layouts/content-block/layout';
+
+	public function __construct(string $sId = null, array $aContainerClasses = [])
+	{
+		parent::__construct($sId, $aContainerClasses);
+
+		$this->AddDataAttribute('role', 'ibo-dashlet');
```
Molkobain:
```suggestion
		$this->AddDataAttribute('role', static::BLOCK_CODE);
```

sources/application/UI/Base/Component/Dashlet/DashletContainer.php
```diff
@@ -14,4 +14,11 @@ class DashletContainer extends UIContentBlock
 {
 	public const BLOCK_CODE = 'ibo-dashlet';
 	public const DEFAULT_HTML_TEMPLATE_REL_PATH = 'base/layouts/content-block/layout';
+
+	public function __construct(string $sId = null, array $aContainerClasses = [])
```
Molkobain, line 18: I'm not fond of this at it's clear that there is clearly a design problem in the dashlets block, this should be in a parent template, but for now I might it's acceptable. We might have to refactor the dashlets a bit later 😥

Molkobain (changes requested):

application/dashboard.class.inc.php
piRGoif: Outdated comment, closing conversation !

sources/application/UI/Base/Component/Dashlet/DashletContainer.php
piRGoif: Comment well noted I agree as said in our IM !
Out of scope for now, so resolving this conversation

js/components/dashlet-badge.js
piRGoif: Will be solved by future commit

js/components/dashlet-badge.js
piRGoif: Will be solved by future commit

sources/application/UI/Base/Component/Dashlet/DashletContainer.php
piRGoif: Will be solved by future commit

sources/application/UI/Base/Component/Dashlet/DashletBadge.php
piRGoif: Will be solved by future commit

sources/application/UI/Base/Component/Dashlet/DashletBadge.php
piRGoif: You'll have to explain me directly, not sure to understand ?

piRGoif:
Changing state : not a draft anymore

sources/application/UI/Base/Component/Dashlet/DashletBadge.php
piRGoif: Discussed with Guillaume, changed in 59cd940b

piRGoif:
@Molkobain is OK to merge, so let's go !

PR #192, opened 2021-02-09
N°3717 Fix create CMDBChange without any corresponding CMDBChangeOp

piRGoif:
## Introduction

In iTop, a history functionnality exists so that user can have a clear vision of the modifications made to an object. All of this is handled in `CMDBObject` methods, and persisted in `CMDBChange` and `CMDBChangeOp` objects.

Initially the history needed to be handled manually : each developper needed to create and persist the `CMDBChange` and `CMDBChangeOp` objects.

Then in iTop 2.0 (commit 7dbbb1c2990702347780cd38130600f39f088d0c) was added an automatic generation and persistence done by the ORM itself. 
Possibility for the developper to specify his own CMDBObject was kept though (`\CMDBObject::SetCurrentChange`, and also `SetTrackInfo` and `SetTrackOrigin` methods), the ORM creating and attaching the corresponding CMDBChangeOp objects automatically.
Also were kept all the previous object CRUD methods in DBObject where you need to specify your CMDBChange having already its CMDBChangeOp attached : DBInsertTracked / DBInsertTrackedNoReload / DBUpdateTracked / DBDeleteTracked.

## Bugs in iTop 2.7.x versions

🐛 In the 2.7.* iTop versions, we had quite a hard time dealing with this object history API.
Here are the related tickets :
* 2.7.0 N°2361 deprecate all DB*Tracked and adapt code that was calling those methods
  => everything was working by chance before, and using the API as intended (instead of calling `DB*Tracked`, use the regular API but with a call to `\CMDBObject::SetCurrentChange`) caused all the previously hidden problems to appear 👻
* 2.7.0-2 N°2952 Auth API automatic provisionning not working due to wrong CMDBChange.origin
* 2.7.2 N°3218 in the "Mail to Ticket automation" extension :
  - wrong value for CMDBChange.origin
  - only one CMDBChange is generated for the whole BackgroundProcess execution
* 2.7.2 N°3219 cron.php : only one CMDBChange for the whole cron.php execution 😱
* 2.7.4 N°3716 Mail To Ticket orphan CMDBChange : since iTop products 2.7.2 or extension v3.2.0 each message read in the mailbox by the extension persists a new `CMDBChange` object. This leads to enormous volume in the corresponding table, possibily slowing down the whole application :(
  Mail to Ticket automation 3.2.1 now only compatible with iTop 2.7.2+ (in order to be able to call `CMDBObject::SetCurrentChange(null);`)
* 2.7.4 N°3464 the itop-fence module sets a CMDBChange that is not reset after
* 2.7.6 N°4486 errors due to combodo-cmdbchange-cleaner removing CMDBChange objects of ongoing modifications

## Proposition

This PR aims to improve current history API, to improve mandatory calls when working in a context when we need to customize API behavior like in `cron.php` (see commit 07bd6b853931b68a8d23d72bda90a7845b20d1ca for N°3219 in iTop 2.7.2)

* 💾 Currently we need to persist CMDBChange passed to `\CMDBObject::SetCurrentChange` : this is no longer the case. Persistance is now done in `\CMDBChangeOp::OnInsert` when needed, as this is indeed CMDBChangeOp that adds this constraint. Also, persistence isn't done anymore in `\CMDBObject::CreateChange`
* 👍 When we needed to reset current change (for example what we did for N°3219) there were multiple lines to add. For convenience we are adding a new `\CMDBObject::SetCurrentChangeFromParams` helper method : now there is only 1 line to write (`cron.php` modified as an example)

## Remaining issues

❗ Note that there is still something to solve : when setting yourself `CMDBChangeOp.change` field with your own `CMDBChange` not persisted instance, in `\CMDBChangeOp::OnInsert` we will replace silently your `CMDBChange` instance by the API's current change !
More generally, developers should be warned that history (that is `CMDBChange` and `CMDBChangeOp` objects) are the API concern, and they shouldn't deal with it directly !
This may be improved by a future PR...

❗ Also, `\CMDBObject::SetTrack*` methods are confusing, as they are doing nothing if the current change was already created... As the point before, this is something that we should improve later on.

## TODO

- [ ] Wait for 3.0.1 release
- [ ] Merge for 2.7.7 / 3.0.2 / 3.1.0
- [ ] Starting in support/3.0 add the CMDBChange.user_id field support (N°3279)

### DIFF
--- a/core/cmdbchangeop.class.inc.php
+++ b/core/cmdbchangeop.class.inc.php
@@ -79,13 +79,11 @@ class CMDBChangeOp extends DBObject
 	protected function OnInsert()
 	{
 		if ($this->Get('change') <= 0) {
-			$this->Set('change', CMDBObject::GetCurrentChange());
-		}
-
-		/** @var \CMDBChange $oChange */
-		$oChange = $this->Get('change');
-		if ($oChange->IsNew()) {
-			$oChange->DBWrite();
+			$oChange = CMDBObject::GetCurrentChange();
+			if ($oChange->IsNew()) {
+				$oChange->DBWrite();
+			}
+			$this->Set('change', $oChange);
 		}
 
 		parent::OnInsert();
--- a/core/cmdbchangeop.class.inc.php
+++ b/core/cmdbchangeop.class.inc.php
@@ -70,15 +70,24 @@ class CMDBChangeOp extends DBObject
 	}
 
 	/**
-	 * Safety net: in case the change is not given, let's guarantee that it will
-	 * be set to the current ongoing change (or create a new one)	
-	 */	
+	 * Safety net:
+	 * * if CMDBChange isn't persisted yet, do it !
+	 * * in case the change is not given, let's guarantee that it will be set to the current ongoing change (or create a new one)
+	 *
+	 * @since 2.7.4 do persist the CMDBChange if needed
+	 */
 	protected function OnInsert()
 	{
-		if ($this->Get('change') <= 0)
-		{
+		if ($this->Get('change') <= 0) {
 			$this->Set('change', CMDBObject::GetCurrentChange());
 		}
+
+		/** @var \CMDBChange $oChange */
+		$oChange = $this->Get('change');
+		if ($oChange->IsNew()) {
+			$oChange->DBWrite();
+		}
+
 		parent::OnInsert();
 	}
 }
--- a/core/cmdbchangeop.class.inc.php
+++ b/core/cmdbchangeop.class.inc.php
@@ -71,10 +71,10 @@ class CMDBChangeOp extends DBObject
 
 	/**
 	 * Safety net:
-	 * * if CMDBChange isn't persisted yet, do it !
+	 * * if change isn't persisted yet, use the current change and persist it if needed
 	 * * in case the change is not given, let's guarantee that it will be set to the current ongoing change (or create a new one)
 	 *
-	 * @since 2.7.4 do persist the CMDBChange if needed
+	 * @since 2.7.4 do persist the current change if needed
 	 */
 	protected function OnInsert()
 	{
--- a/core/cmdbchangeop.class.inc.php
+++ b/core/cmdbchangeop.class.inc.php
@@ -85,11 +85,29 @@ class CMDBChangeOp extends DBObject
 			return parent::Set($sAttCode, $value);
 		}
 
-		if ($value->IsNew()) {
-			throw new CoreUnexpectedValue("Cannot set CMDBChangeOp.change with a non persisted CMDBChange !");
+		if (is_int($value)) {
+			if ($value <= 0) {
+				$this->ThrowInvalidCMDBChangeException();
+			}
+
+			return parent::Set($sAttCode, $value);
 		}
 
-		return parent::Set($sAttCode, $value);
+		if (is_object($value)) {
+			if ($value->IsNew()) {
+				$this->ThrowInvalidCMDBChangeException();
+			}
+
+			return parent::Set($sAttCode, $value);
+		}
+
+		// we should never get here O:)
+		throw new CoreUnexpectedValue("Invalid value set to CMDBChangeOp.change");
+	}
+
+	protected function ThrowInvalidCMDBChangeException()
+	{
+		throw new CoreUnexpectedValue("Cannot set CMDBChangeOp.change with a non persisted CMDBChange !");
 	}
 
 	/**
--- a/core/cmdbchangeop.class.inc.php
+++ b/core/cmdbchangeop.class.inc.php
@@ -74,7 +74,7 @@ class CMDBChangeOp extends DBObject
 	 * * if change isn't persisted yet, use the current change and persist it if needed
 	 * * in case the change is not given, let's guarantee that it will be set to the current ongoing change (or create a new one)
 	 *
-	 * @since 2.7.4 do persist the current change if needed
+	 * @since 2.7.7 3.0.2 3.1.0 N°3717 do persist the current change if needed
 	 */
 	protected function OnInsert()
 	{
--- a/core/cmdbchangeop.class.inc.php
+++ b/core/cmdbchangeop.class.inc.php
@@ -70,38 +70,6 @@ class CMDBChangeOp extends DBObject
 	}
 
 	/**
-	 * @param string $sAttCode
-	 * @param mixed $value
-	 *
-	 * @return bool
-	 * @throws \CoreException
-	 * @throws \CoreUnexpectedValue
-	 *
-	 * @since 2.7.4 reject non persisted CMDBChange
-	 */
-	public function Set($sAttCode, $value)
-	{
-		if ($sAttCode !== 'change') {
-			return parent::Set($sAttCode, $value);
-		}
-
-		if ((is_int($value)) && ($value <= 0)) {
-			$this->ThrowInvalidCMDBChangeException();
-		}
-
-		if ((is_object($value)) && ($value->IsNew())) {
-			$this->ThrowInvalidCMDBChangeException();
-		}
-
-		return parent::Set($sAttCode, $value);
-	}
-
-	protected function ThrowInvalidCMDBChangeException()
-	{
-		throw new CoreUnexpectedValue("Cannot set CMDBChangeOp.change with a non persisted CMDBChange !");
-	}
-
-	/**
 	 * Safety net:
 	 * * if CMDBChange isn't persisted yet, do it !
 	 * * in case the change is not given, let's guarantee that it will be set to the current ongoing change (or create a new one)
--- a/core/cmdbchangeop.class.inc.php
+++ b/core/cmdbchangeop.class.inc.php
@@ -85,24 +85,15 @@ class CMDBChangeOp extends DBObject
 			return parent::Set($sAttCode, $value);
 		}
 
-		if (is_int($value)) {
-			if ($value <= 0) {
-				$this->ThrowInvalidCMDBChangeException();
-			}
-
-			return parent::Set($sAttCode, $value);
+		if ((is_int($value)) && ($value <= 0)) {
+			$this->ThrowInvalidCMDBChangeException();
 		}
 
-		if (is_object($value)) {
-			if ($value->IsNew()) {
-				$this->ThrowInvalidCMDBChangeException();
-			}
-
-			return parent::Set($sAttCode, $value);
+		if ((is_object($value)) && ($value->IsNew())) {
+			$this->ThrowInvalidCMDBChangeException();
 		}
 
-		// we should never get here O:)
-		throw new CoreUnexpectedValue("Invalid value set to CMDBChangeOp.change");
+		return parent::Set($sAttCode, $value);
 	}
 
 	protected function ThrowInvalidCMDBChangeException()
--- a/core/cmdbchangeop.class.inc.php
+++ b/core/cmdbchangeop.class.inc.php
@@ -63,10 +63,33 @@ class CMDBChangeOp extends DBObject
 
 	/**
 	 * Describe (as a text string) the modifications corresponding to this change
-	 */	 
+	 */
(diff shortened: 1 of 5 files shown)


piRGoif:
Latest commits to : 
* fix "Call to a member function IsNew() on int" bug
* add a constraint on CMDBChangeOp::Set when called on the change attribute

There could be some unit test associated, will do this later

bruno-ds:
As seen with @piRGoif : 

 - The throw in the CMDBChangeOp::Set() seem dangerous since it change the behavior of iTop on a minor version. (@piRGoif has checked against itop and all our extensions but still, there is room for private ext. and customization)
 - the type returned by Get seems to always be an integer (I didn't know this part of the API) : we should check it against @dflaven or @rquetiez or @Molkobain 
 

there is a complex auto lock behavior : since the change is persisted very late (on insert), and the the api does not return the (not yet persisted) object but it's id, we simply can't persist it anymore.
This is why @piRGoif has made the setter enforce a persisted instance.

so, the behavior change are that now on 
 - the API let set a "current change" (on CMDBObject)  not yet persisted
 - the API forbid the set of a "change" (on CMDBChangeOp)  not yet persisted and/or as integer

I think that setting a change (on CMDBChangeOp)  with an id > 0 should be authorized. 
I think that setting a negative integer or a not persisted instance (on CMDBChangeOp) , maybe trigger an IssueLog and fall back on the OnInsert code that fetch the current change.
Doing so would not impact the behavior of a minor version. If we want to migrate to the exception, maybe we could deprecate it with just an issue log, and throw on the 3.0?

piRGoif:
Thanks @bruno-ds ! I will do a review with @rquetiez next week and let you know !
In the meantime I'll have to fix a bug in \CMDBChangeOp::Set : for now it is crashing if setting using an int !

core/cmdbchangeop.class.inc.php
```diff
@@ -63,22 +63,53 @@ public static function Init()
...
+	 */
+	public function Set($sAttCode, $value)
+	{
+		if ($sAttCode !== 'change') {
+			return parent::Set($sAttCode, $value);
+		}
+
+		if ($value->IsNew()) {
+			throw new CoreUnexpectedValue("Cannot set CMDBChangeOp.change with a non persisted CMDBChange !");
+		}
```
Hipska: I would revert this logic so that the exception is thrown first and otherwise continue to to existing `return parent::Set()`
```suggestion
		if ($sAttCode == 'change' && $value->IsNew()) {
			throw new CoreUnexpectedValue("Cannot set CMDBChangeOp.change with a non persisted CMDBChange !");
		}
```

That would improve the readability of the code.

Hipska (changes requested):

core/cmdbchangeop.class.inc.php
```diff
@@ -63,22 +63,71 @@ public static function Init()
...
+		if (is_object($value)) {
+			if ($value->IsNew()) {
+				$this->ThrowInvalidCMDBChangeException();
+			}
+
+			return parent::Set($sAttCode, $value);
+		}
+
+		// we should never get here O:)
+		throw new CoreUnexpectedValue("Invalid value set to CMDBChangeOp.change");
```
Hipska: Again, now you made it worse. There are now 3 places where you call the exact same return.

To improve readability, I would suggest to revert the logic to:
```suggestion
		if ($sAttCode == 'change') {
			if (is_int($value) && $value <= 0) {
				$this->ThrowInvalidCMDBChangeException();
			}
			elseif (is_object($value) && $value->IsNew()) {
				$this->ThrowInvalidCMDBChangeException();
			}
		}

		return parent::Set($sAttCode, $value);
```

Hipska (changes requested):

core/cmdbchangeop.class.inc.php
piRGoif: I like your first proposal : exit condition first. Will change this !

piRGoif: Sorry I misread on friday : main exit condition is already on top. After reading again and more carefully your proposition, I agree grouping the error conditions is better. 
I've just pushed 38acbc87

piRGoif:
Rewrote the description according to the latest changes.

piRGoif:
We did a review with @rquetiez ! Here is a summary of what we said : 

* Romain agreed on the minimum modification, that is : 
  - removing DBInsert call in `\CMDBObject::CreateChange`
  - in `\CMDBChangeOp::OnInsert` persist the current change if it needs to be saved

* We agreed that the existing test in `\CMDBChangeOp::OnInsert` that set the current change without any information is a big risk for the developper. 
But the `\CMDBChangeOp::Set` overriding I did is a big risk, for a small gain...
We should warn clearly the developer that history should be handled by the API. Maybe add a warning when calling directly `\CMDBChangeOp::Set('change', ...)` or `CMDBChange::DBInsert` (only if dev env, call trigger_error ?)
There are only 2 extensions, but some core calls also...
This has to be discussed with the rest of the team, and we would upgrade the code in a separate PR.

* About `\CMDBObject::SetTrack*` methods : what I modified was not enough, as the current change could be already persisted, and simply modifying it would have no effect whatsoever !
We could change this so that if the current change is persisted, then we would create a new change.
This should be done in a dedicated PR !

core/cmdbchangeop.class.inc.php
Hipska: Please see my suggestion again, the call to parent should not be inside an `if` for readability of the code..

piRGoif: After a review with Romain (see last comment below), I will drop this to lower this PR's scope.
Thanks for your feedback !
I'm closing this conversation.

core/cmdbobject.class.inc.php
```diff
@@ -114,6 +114,33 @@ public static function SetCurrentChange($oChange)
...
+	 * @param string $sOrigin
+	 * @param $oDate
+	 *
+	 * @throws \CoreException
+	 *
+	 * @since 2.7.4 3.0.0 N°3717 new method to reset current change
+	 */
+	public static function SetCurrentChangeFromParams($sUserInfo, $sOrigin = null, $oDate = null)
+	{
+		$oChange = MetaModel::NewObject('CMDBChange');
```
Hipska:
```suggestion
		self::$m_oCurrChange = MetaModel::NewObject('CMDBChange');
```

core/cmdbobject.class.inc.php
```diff
@@ -114,6 +114,33 @@ public static function SetCurrentChange($oChange)
...
+
+		self::$m_oCurrChange->Set("userinfo", $sUserInfo);
```
Hipska: Why is `self` used instead of `static` like in the SetTrack* methods?

core/cmdbobject.class.inc.php
```diff
@@ -114,6 +114,33 @@ public static function SetCurrentChange($oChange)
...
+
+		if (!is_null($sOrigin)) {
+			self::$m_oCurrChange->Set("origin", $sOrigin);
+		}
+
+		if (is_null($oDate)) {
+			$oDate = time();
+		}
```
Hipska: Is this needed?

core/cmdbobject.class.inc.php
```diff
@@ -114,6 +114,33 @@ public static function SetCurrentChange($oChange)
...
+	 * @since 2.7.4 3.0.0 N°3717 new method to reset current change
+	 */
+	public static function SetCurrentChangeFromParams($sUserInfo, $sOrigin = null, $oDate = null)
+	{
+		$oChange = MetaModel::NewObject('CMDBChange');
+
+		self::$m_oCurrChange->Set("userinfo", $sUserInfo);
+
+		if (!is_null($sOrigin)) {
+			self::$m_oCurrChange->Set("origin", $sOrigin);
```
Hipska: Why not
```suggestion
			static::SetTrackOrigin($sOrigin);
```

Hipska (changes requested):

core/cmdbobject.class.inc.php
piRGoif: Wooopsy
Nice catch !
Will be fixed in the next commits

core/cmdbobject.class.inc.php
piRGoif: Will be fixed in next commits

core/cmdbobject.class.inc.php
piRGoif: There are no default value for this field, as we need to call a PHP function to get it
No overrides will set it
So this is a workaround to get a default value O:)

core/cmdbobject.class.inc.php
piRGoif: This method update will be removed in next commits, closing this conversation

piRGoif:
I've just pushed modifications discussed with @rquetiez
Removing draft flag, and setting this PR to be reviewed by the dev team

core/cmdbobject.class.inc.php
```diff
@@ -114,6 +114,31 @@ public static function SetCurrentChange($oChange)
...
+		static::$m_oCurrChange->Set("userinfo", $sUserInfo);
+
+		if (!is_null($sOrigin)) {
+			static::$m_oCurrChange->Set("origin", $sOrigin);
+		}
+
+		if (is_null($oDate)) {
+			$oDate = time();
+		}
+		static::$m_oCurrChange->Set("date", $oDate);
```
Hipska: What about
```suggestion
		static::SetTrackInfo($sUserInfo);
		static::SetTrackOrigin($sOrigin);
		static::CreateChange();

		if (!is_null($oDate)) {
			static::$m_oCurrChange->Set("date", $oDate);
		}
```

As it seems you are doing double code right now and it might be better to just reuse already existing code.

piRGoif: I like this ! I'm committing this suggestion. Thanks !

Hipska (approved):

piRGoif:
Dev team is OK, but as I'm proposing this for the 2.7.5 version we will review it with the products team.

test/core/CMDBObjectTest.php
```diff
@@ -0,0 +1,75 @@
...
+		$oInitialCurrentChange = CMDBObject::GetCurrentChange();
+		$sInitialTrackInfo = CMDBObject::GetTrackInfo();
+		// reset current change
+		CMDBObject::SetCurrentChange(null);
+
+		//-- new object with only track info
+		$sTrackInfo = 'PHPUnit test';
+		CMDBObject::SetTrackInfo($sTrackInfo);
+		/** @var \DocumentNote $oTestObject */
+		$oTestObject = MetaModel::NewObject('DocumentWeb');
```
Hipska: `DocumentNote` vs `DocumentWeb`

piRGoif: Wooops, good catch thanks !
Fixed in a2e3940f

piRGoif:
Presentation was done to the products team. Waiting for their answer (maybe after 2.7.4 ?)

piRGoif:
Presentations to the product team was done 6 months ago, and no feedback for now :(
Version 2.7.6 will be out soon and contains another fix as a workaround for this API...

I updated this PR description to explain better the whole story, and will ask again the product team.

piRGoif:
Presentations to the product team was done 6 months ago, and no feedback for now :(
Version 2.7.6 will be out soon and contains another fix as a workaround for this API...

I updated this PR description to explain better the whole story, and will ask again the product team.

piRGoif:
Meeting planned for the end of next week

piRGoif:
We agreed to have a dedicated meeting on this : it is planned for next week.

piRGoif:
The meeting happened, I have some actions to do before going on !
Stay tune :)

piRGoif:
Actions done, next meeting planned in 2 weeks.

piRGoif:
The PR was accepted after review with the product team 🥳

But as we are aiming to deliver 3.0.1 very soon and don't want those modifications to be part of this release, we will merge after 3.0.1 delivery.

PR #193, opened 2021-03-05
N°3731 log deprecated calls

piRGoif:
@Hipska suggested that we trigger warnings when calling deprecated files : https://github.com/Combodo/iTop/commit/b85b4d00674884491938a6ed02a2389838106727#r46416569

We created a deprecate/remove policy last year, but were eagger to add warnings as this could break ajax calls the way they are done in iTop for now :/
After Hipska suggestion, we decided we should try : 
* adding a dedicated LogAPI implementation, that would write to a specific file
* this logger should also trigger_error() if \utils::IsDevelopmentEnvironment
* we should call this logger in every deprecated code and files

This is a first try to implement this, as a base to discuss this with the rest of the dev team !

### DIFF
--- a/bootstrap.inc.php
+++ b/bootstrap.inc.php
@@ -50,18 +50,19 @@ if (function_exists('microtime')) {
 	$fItopStarted = 1000 * time();
 }
 
-if (! isset($GLOBALS['bBypassAutoload']) || $GLOBALS['bBypassAutoload'] == false)
-{
+if (!isset($GLOBALS['bBypassAutoload']) || $GLOBALS['bBypassAutoload'] == false) {
 	require_once APPROOT.'/lib/autoload.php';
 }
+
+require_once APPROOT.'core/log.class.inc.php';
+DeprecatedCallsLog::Enable();
 
 //
 // Maintenance mode
 //
 
 // Use 'maintenance' parameter to bypass maintenance mode
-if (!isset($bBypassMaintenance))
-{
+if (!isset($bBypassMaintenance)) {
 	$bBypassMaintenance = isset($_REQUEST['maintenance']) ? boolval($_REQUEST['maintenance']) : false;
 }
 
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -688,11 +688,7 @@ abstract class LogAPI
 		}
 
 		if (isset($sLogLevelMin[static::CHANNEL_DEFAULT])) {
-			return $sLogLevelMin[static::CHANNEL_DEFAULT];
-		}
-
-		if (isset($sLogLevelMin[''])) {
-			return $sLogLevelMin[''];
+			return $sLogLevelMin[$sChannel];
 		}
 
 		return static::GetLevelDefault();
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -606,50 +606,64 @@ abstract class LogAPI
 		static::Log(self::LEVEL_TRACE, $sMessage, $sChannel, $aContext);
 	}
 
+	/**
+	 * @throws \ConfigException if log wrongly configured
+	 */
 	public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = array())
 	{
-		if (! static::$m_oFileLog)
-		{
+		if (!static::$m_oFileLog) {
 			return;
 		}
 
-		if (! isset(self::$aLevelsPriority[$sLevel]))
-		{
+		if (!isset(self::$aLevelsPriority[$sLevel])) {
 			IssueLog::Error("invalid log level '{$sLevel}'");
+
 			return;
 		}
 
-		if (is_null($sChannel))
-		{
+		if (is_null($sChannel)) {
 			$sChannel = static::CHANNEL_DEFAULT;
 		}
 
-		$sMinLogLevel = self::GetMinLogLevel($sChannel);
-
-		if ($sMinLogLevel === false || $sMinLogLevel === 'false')
-		{
+		if (!static::IsLogLevelEnabled($sLevel, $sChannel)) {
 			return;
-		}
-		if (is_string($sMinLogLevel))
-		{
-			if (! isset(self::$aLevelsPriority[$sMinLogLevel]))
-			{
-				throw new Exception("invalid configuration for log_level '{$sMinLogLevel}' is not within the list: ".implode(',', array_keys(self::$aLevelsPriority)));
-			}
-			elseif (self::$aLevelsPriority[$sLevel] < self::$aLevelsPriority[$sMinLogLevel])
-			{
-				//priority too low regarding the conf, do not log this
-				return;
-			}
 		}
 
 		static::$m_oFileLog->$sLevel($sMessage, $sChannel, $aContext);
 	}
 
 	/**
-	 * @param $sChannel
+	 * @throws \ConfigException if log wrongly configured
+	 * @uses GetMinLogLevel
+	 */
+	final public static function IsLogLevelEnabled(string $sLevel, string $sChannel): bool
+	{
+		$sMinLogLevel = self::GetMinLogLevel($sChannel);
+
+		if ($sMinLogLevel === false || $sMinLogLevel === 'false') {
+			return false;
+		}
+		if (!is_string($sMinLogLevel)) {
+			return false;
+		}
+
+		if (!isset(self::$aLevelsPriority[$sMinLogLevel])) {
+			throw new ConfigException("invalid configuration for log_level '{$sMinLogLevel}' is not within the list: ".implode(',', array_keys(self::$aLevelsPriority)));
+		} elseif (self::$aLevelsPriority[$sLevel] < self::$aLevelsPriority[$sMinLogLevel]) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * @param string $sChannel
 	 *
-	 * @return string one of the LEVEL_* const value
+	 * @return string one of the LEVEL_* const value : the one configured it if exists, otherwise default log level for this channel
+	 *       Config can be done globally : `'log_level_min' => LogAPI::LEVEL_TRACE,`
+	 *       Or per channel : `'log_level_min' => ['InlineImage' => LogAPI::LEVEL_TRACE, 'UserRequest' => LogAPI::LEVEL_TRACE],`
+	 *
+	 * @uses \LogAPI::GetConfig()
 	 * @uses \LogAPI::GetLevelDefault
 	 */
 	protected static function GetMinLogLevel($sChannel)
@@ -789,9 +803,9 @@ class DeadLockLog extends LogAPI
  */
 class DeprecatedCallsLog extends LogAPI
 {
-	public const ENUM_CHANNEL_PHP = 'deprecated-method';
+	public const ENUM_CHANNEL_PHP_METHOD = 'deprecated-php-method';
 	public const ENUM_CHANNEL_FILE = 'deprecated-file';
-	public const CHANNEL_DEFAULT = self::ENUM_CHANNEL_PHP;
+	public const CHANNEL_DEFAULT = self::ENUM_CHANNEL_PHP_METHOD;
 
 	public const LEVEL_DEFAULT = self::LEVEL_ERROR;
 
@@ -815,8 +829,17 @@ class DeprecatedCallsLog extends LogAPI
 		return static::LEVEL_DEFAULT;
 	}
 
+	/**
+	 * @throws \ConfigException
+	 * @link https://www.php.net/debug_backtrace
+	 * @uses \debug_backtrace()
+	 */
 	public static function NotifyDeprecatedFile(?string $sAdditionalMessage = null): void
 	{
+		if (!static::IsLogLevelEnabled(self::LEVEL_WARNING, self::ENUM_CHANNEL_FILE)) {
+			return;
+		}
+
 		$aStack = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
 		$sCallerFile = $aStack[0]['file'];
 
@@ -832,12 +855,17 @@ class DeprecatedCallsLog extends LogAPI
 	/**
 	 * @param string|null $sAdditionalMessage
 	 *
-	 * @uses \debug_backtrace()
+	 * @throws \ConfigException
 	 * @link https://www.php.net/debug_backtrace
+	 * @uses \debug_backtrace()
 	 */
-	public static function NotifyDeprecatedMethod(?string $sAdditionalMessage = null): void
+	public static function NotifyDeprecatedPhpMethod(?string $sAdditionalMessage = null): void
 	{
-		$aStack = debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT, 3);
+		if (!static::IsLogLevelEnabled(self::LEVEL_WARNING, self::ENUM_CHANNEL_PHP_METHOD)) {
+			return;
+		}
+
+		$aStack = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 3);
 		$sCallerObject = $aStack[2]['class'];
 		$sCallerMethod = $aStack[2]['function'];
 		$sCallerLine = $aStack[2]['line'];
@@ -851,7 +879,7 @@ class DeprecatedCallsLog extends LogAPI
 			$sMessage .= ' : '.$sAdditionalMessage;
 		}
 
-		static::Warning($sMessage, static::ENUM_CHANNEL_PHP);
+		static::Warning($sMessage, static::ENUM_CHANNEL_PHP_METHOD);
 	}
 
 	public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = array()): void
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -765,9 +765,9 @@ class DeadLockLog extends LogAPI
  */
 class DeprecatedCallsLog extends LogAPI
 {
-	public const CHANNEL_PHP = 'deprecated-method';
-	public const CHANNEL_FILE = 'deprecated-file';
-	public const CHANNEL_DEFAULT = self::CHANNEL_PHP;
+	public const ENUM_CHANNEL_PHP = 'deprecated-method';
+	public const ENUM_CHANNEL_FILE = 'deprecated-file';
+	public const CHANNEL_DEFAULT = self::ENUM_CHANNEL_PHP;
 
 	public const LEVEL_DEFAULT = self::LEVEL_WARNING;
 
@@ -802,7 +802,7 @@ class DeprecatedCallsLog extends LogAPI
 			$sMessage .= ' : '.$sAdditionalMessage;
 		}
 
-		static::Warning($sMessage, static::CHANNEL_FILE);
+		static::Warning($sMessage, static::ENUM_CHANNEL_FILE);
 	}
 
 	/**
@@ -827,7 +827,7 @@ class DeprecatedCallsLog extends LogAPI
 			$sMessage .= ' : '.$sAdditionalMessage;
 		}
 
-		static::Warning($sMessage, static::CHANNEL_PHP);
+		static::Warning($sMessage, static::ENUM_CHANNEL_PHP);
 	}
 
 	public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = array()): void
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -844,10 +844,17 @@ class DeprecatedCallsLog extends LogAPI
 			return;
 		}
 
-		$aStack = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
-		$sCallerFile = $aStack[0]['file'];
+		$aStack = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2);
+		$sDeprecatedFile = $aStack[0]['file'];
+		if (array_key_exists(1, $aStack)) {
+			$sCallerFile = $aStack[1]['file'];
+			$sCallerLine = $aStack[1]['line'];
+		} else {
+			$sCallerFile = 'N/A';
+			$sCallerLine = 'N/A';
+		}
 
-		$sMessage = $sCallerFile;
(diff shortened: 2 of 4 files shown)


piRGoif:
PR is now ready for review !

Two consumers were added : 

Location | Type | How to test
--------- | ----- | -------------
`core/coreexception.class.inc.php` | file | for example http://localhost/itop-dev/core/coreexception.class.inc.php
`\WebPage::outputCollapsibleSectionInit` | PHP | Any page in iTop

Sample output for both examples (in `log/deprecated-calls.log`) : 

```
2021-03-05 19:22:43 | Error   | C:\Dev\wamp64\www\itop-dev\core\coreexception.class.inc.php | deprecated-file
2021-03-12 11:55:35 | Error   | WebPage::outputCollapsibleSectionInit L170 | deprecated-php
```

piRGoif:
Presentation to the dev team was done.
~~Just 2~~ things to change : 

* change log level from ERROR to WARNING (so change the default log level too)
* rename helpers methods : 
  * ErrorFile => NotifyDeprecatedFile
  * ErrorPhp => NotifyDeprecatedMethod
* change trigger_error call to send deprecated info to PHP

Also, check with Guillaume after modifications, and if he's ok then we're good to go !

(added one other point on 06/04)

piRGoif:
This morning took a few minutes to :
* rebase branch
* add fixes asked during dev team review
* change the log done for deprecated methods

Hipska (dismissed):

piRGoif:
Waiting for Molkobain review

core/log.class.inc.php
```diff
@@ -765,10 +760,92 @@ public static function Log($iMySQLErrNo, $sMessage, $iMysqlErroNo = null, $aCont
 }
 
 
+/**
+ * @since 3.0.0 N°3731
+ */
+class DeprecatedCallsLog extends LogAPI
```
Molkobain, line 804: Can we put this on a dedicated file?

core/log.class.inc.php
```diff
@@ -765,10 +760,92 @@ public static function Log($iMySQLErrNo, $sMessage, $iMysqlErroNo = null, $aCont
...
+{
+	public const CHANNEL_PHP = 'deprecated-method';
```
Molkobain: To keep consistency with other constant convention accross the app, can you rename them (and their usages) to ENUM_CHANNEL_XXX?

core/log.class.inc.php
```diff
@@ -765,10 +760,92 @@ public static function Log($iMySQLErrNo, $sMessage, $iMysqlErroNo = null, $aCont
...
+	public const CHANNEL_FILE = 'deprecated-file';
+	public const CHANNEL_DEFAULT = self::CHANNEL_PHP;
```
Molkobain: Rename to DEFAULT_CHANNEL to keep consistency on the constant convention

core/log.class.inc.php
```diff
@@ -649,33 +649,28 @@ public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = arr
 	 * @return string one of the LEVEL_* const value
 	 * @uses \LogAPI::LEVEL_DEFAULT
 	 */
-	private static function GetMinLogLevel($sChannel)
+	protected static function GetMinLogLevel($sChannel)
```
Molkobain, line 669:
```suggestion
	protected static function GetMinLogLevel(string $sChannel)
```

core/log.class.inc.php
```diff
@@ -765,10 +760,92 @@ public static function Log($iMySQLErrNo, $sMessage, $iMysqlErroNo = null, $aCont
...
+	public const CHANNEL_PHP = 'deprecated-method';
+	public const CHANNEL_FILE = 'deprecated-file';
+	public const CHANNEL_DEFAULT = self::CHANNEL_PHP;
+
+	public const LEVEL_DEFAULT = self::LEVEL_WARNING;
+
+	/** @var \FileLog we want our own instance ! */
+	protected static $m_oFileLog = null;
+
+	public static function Enable($sTargetFile = null): void
```
Molkobain, line 815:
```suggestion
	public static function Enable(?string $sTargetFile = null): void
```

core/log.class.inc.php
```diff
@@ -765,10 +760,92 @@ public static function Log($iMySQLErrNo, $sMessage, $iMysqlErroNo = null, $aCont
...
+	{
+		if (empty($sTargetFile)) {
+			$sTargetFile = APPROOT.'log/deprecated-calls.log';
+		}
+		parent::Enable($sTargetFile);
+	}
+
+	protected static function GetMinLogLevel($sChannel): string
```
Molkobain:
```suggestion
	protected static function GetMinLogLevel(string $sChannel): string
```

core/log.class.inc.php
```diff
@@ -765,10 +760,92 @@ public static function Log($iMySQLErrNo, $sMessage, $iMysqlErroNo = null, $aCont
...
+		$sMessage = "{$sCallerObject}::{$sCallerMethod} L{$sCallerLine} calling {$sDeprecatedObject}:{$sDeprecatedMethod}";
+
+		if (!is_null($sAdditionalMessage)) {
+			$sMessage .= ' : '.$sAdditionalMessage;
+		}
+
+		static::Warning($sMessage, static::CHANNEL_PHP);
+	}
+
+	public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = array()): void
```
Molkobain, line 892:
```suggestion
	public static function Log(string $sLevel, string $sMessage, string $sChannel = null, $aContext = array()): void
```

bootstrap.inc.php
```diff
@@ -50,18 +50,19 @@
...
 }
 
-if (! isset($GLOBALS['bBypassAutoload']) || $GLOBALS['bBypassAutoload'] == false)
-{
+if (!isset($GLOBALS['bBypassAutoload']) || $GLOBALS['bBypassAutoload'] == false) {
 	require_once APPROOT.'/lib/autoload.php';
 }
 
+require_once APPROOT.'core/log.class.inc.php';
+DeprecatedCallsLog::Enable();
```
Molkobain, line 58: To me this should be conditioned to either the DevelopmentEnvironment and/or a config parameter (disabled by default), the main reason is that you are using the `debug_backtrace` function which isn't meant for performances. 
Also it is most likely that this logging is going to be done on very few environments (probably only devs workstations & combodo internal iTop).

(I didn't see any disablign mechanism, sorry if there is one 😁)

Molkobain (changes requested):

core/log.class.inc.php
piRGoif: There are 15 classes for now in core/log.class.inc.php
I agree this would be a good idea to split them and handle them using autoloader, but this is a distinct development to my opinion. But I'll be glad to do it as I did for CoreException hierarchy ! I'll add a ticket in Combodo DB.

core/log.class.inc.php
piRGoif: Will do !

core/log.class.inc.php
piRGoif: This constant is an override of \LogAPI::CHANNEL_DEFAULT. This later was introduced in 2.7.0 with e5c49e3bd49f08ae3fb365602ddd15be567af16c, and is already used in an extension (Brute Force Protection). So would be hard to rename :(

core/log.class.inc.php
piRGoif: Override of a LogAPI's method... According to our convention we shouldn't add type hinting here right ?

core/log.class.inc.php
piRGoif: Override of a LogAPI's method... According to our convention we shouldn't add type hinting here right ?

core/log.class.inc.php
piRGoif: Override of a LogAPI's method... According to our convention we shouldn't add type hinting here right ?

core/log.class.inc.php
piRGoif: Override of a LogAPI's method... According to our convention we shouldn't add type hinting here right ?

bootstrap.inc.php
piRGoif: There aren't any disabling mechanism, because we agreed twice during team meetings that we wouldn't need one - you were present on the first of those meeting.
We should have a chat together about this

Hipska: Also agree to have this enabled at all times. The logging of those warnings should only occur in rare situations, so the arugument that it uses debug_backtrace is not really relevant. If users experience performance issues because of this, they should update their extensions/modules.

core/log.class.inc.php
piRGoif: By the way, there is also a \LogAPI::LEVEL_DEFAULT, but this one is overrided & used only in 2 core iTop classes so we should be ok to rename it

core/log.class.inc.php
piRGoif: Done : N°3940

bootstrap.inc.php
Molkobain: > ... you were present on the first of those meeting. ...

I don't recall that we said we will use the `debug_backtrace` method for this, that's the only point that is bothering me.
But okay @Hipska has a point 👍

Hipska: Is there a possibility to only use `debug_backtrace` when it is in the Dev env? Might be the best solution then?

piRGoif: > I don't recall that we said we will use the `debug_backtrace` method for this, that's the only point that is bothering me.

You're right, this wasn't part of the specs explicitly, I added it later on so we can trace the callee of the deprecated method/file : this is actually what will be interesting to the developer. My apologies for my previous answer, I was a little disappointed, but catching possible performance issues is important indeed.

Note that the debug_backtrace call impact is mitigated as we provide a specific flag (this can be tuned by the way) and a limit argument, and of course there shouldn't be much calls to the notify deprecated call methods as Thomas argued.

If you think debug_backtrace impact is still to risky, I would advise not to use \utils::IsDevelopmentEnvironment but instead use the log levels : add a log only notifying of the call at a higher level, then if the lower level is enabled do call debug_backtrace and logs the results at this lower log level.
There isn't yet in LogAPI any method to check if a log level is enabled, but we already have what it would need (\LogAPI::$aLevelsPriority), and it will be useful for other needs too !

core/log.class.inc.php
piRGoif: Done in 4414a331

core/log.class.inc.php
piRGoif: Discussed with Molkobain : as this is in LogAPI we're not changing it

core/log.class.inc.php
piRGoif: Discussed with Molkobain : as this is in LogAPI we're not changing it

core/log.class.inc.php
piRGoif: Discussed with Molkobain : as this is in LogAPI we're not changing it

core/log.class.inc.php
piRGoif: Discussed with Molkobain : as this is in LogAPI we're not changing it

core/log.class.inc.php
piRGoif: Discussed with Molkobain : as this is in LogAPI we're not changing it

bootstrap.inc.php
piRGoif: Discussed with Molkobain, the solution will be to :
* have a computed default log level : if on a dev env then something like DEBUG, otherwise ERROR (so by default we will log on a dev env and won't on other env)
* in the NotifyDeprecated* methods, exit if the corresponding log level isn't enabled

piRGoif:
After last modifications, here is an updated documentation : 

Consumers : 

Location | Type | How to test
--------- | ----- | -------------
`core/coreexception.class.inc.php` | file | For example add `require_once APPROOT.'/core/coreexception.class.inc.php';` in WebPage, then load any iTop page
`\WebPage::outputCollapsibleSectionInit` | PHP | Any page in iTop

Sample output for both examples (in `log/deprecated-calls.log`) : 

```
2021-04-13 10:31:12 | Warning | C:\Dev\wamp64\www\itop-dev\sources\application\WebPage\WebPage.php L32 including/requiring C:\Dev\wamp64\www\itop-dev\core\coreexception.class.inc.php : Classes were moved to /application/exceptions | deprecated-file
2021-04-13 10:31:12 | Warning | AjaxPage::output L2857 calling WebPage:outputCollapsibleSectionInit | deprecated-php-method
```

bootstrap.inc.php
Hipska: > if on a dev env then something like DEBUG, otherwise ERROR (so by default we will log on a dev env and won't on other env)

Am I correct if I don't see this implemented?

core/log.class.inc.php
```diff
@@ -765,10 +802,113 @@ public static function Log($iMySQLErrNo, $sMessage, $iMysqlErroNo = null, $aCont
...
+			$sMessage .= ' : '.$sAdditionalMessage;
+		}
+
+		static::Warning($sMessage, static::ENUM_CHANNEL_PHP_METHOD);
+	}
+
+	public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = array()): void
+	{
+		if (utils::IsDevelopmentEnvironment()) {
+			trigger_error($sMessage, E_USER_DEPRECATED);
```
Hipska, line 899: Nicely done! But why not also on `NotifyDeprecatedFile`?

bootstrap.inc.php
piRGoif: * have a computed default log level : look at \DeprecatedCallsLog::GetLevelDefault
* in the NotifyDeprecated* methods, exit if the corresponding log level isn't enabled : at the beginning of each NotifyDeprecated* methods, a test was added, we call \LogAPI::IsLogLevelEnabled

Is this clearer ?

Hipska: Thanks, didn't see the `DeprecatedCallsLog::GetLevelDefault` at first.

Molkobain (dismissed):

Molkobain:
Looks ok to me, ready to merge !

Hipska:
> Discussed with Molkobain, the solution will be to :
> 
> * have a computed default log level : if on a dev env then something like DEBUG, otherwise ERROR (so by default we will log on a dev env and won't on other env)
> * in the NotifyDeprecated* methods, exit if the corresponding log level isn't enabled

Why ERROR otherwise? In the case of Dev, the user will already see the deprecation message because of the `trigger_error`, but someone examining a regular setup will not see these messages as they are not logged by default?

Another thing, Extension developers will only see the message when they use a version from git and not when they test their modules on a released version? (This is what I always do)

piRGoif:
> In the case of Dev, the user will already see the deprecation message because of the `trigger_error`, but someone examining a regular setup will not see these messages as they are not logged by default?

Not sure to understand your question ?

If the log is enabled then we will log to the file. 
If in dev env, then we will also do a `trigger_error()` : this later will surely breaks AJAX calls, that's why we added this condition.

If log is disabled, we will do nothing : neither logging to the file, neither call `trigger_error()`

If the default level isn't satisfying, then it can be modified with the `log_level_min` config parameter (see https://www.itophub.io/wiki/page?id=2_7_0%3Aadmin%3Alog#filter_logs)
What will happen if a NotifyDeprecated* is done during the setup, that' a really good question we need to answer before merging... I have some doubts on the fact the the config would be really used ?

> Another thing, Extension developers will only see the message when they use a version from git and not when they test their modules on a released version? (This is what I always do)

Indeed, as "in dev env" test is done by calling `\utils::IsDevelopmentEnvironment`, which relies on the `ITOP_REVISION` constant value. Any Combodo build will have this constant having a value not matching a dev env...
A solution would be to modify this constant manually, but indeed it's not very nice. 

Adding others conditions in `\utils::IsDevelopmentEnvironment` check is the key, but which one ? This could be the subject of an evolution request, or a new PR :)

Hipska:
I'm saying that default level `ERROR` is not satisfying indeed, should be `WARNING`.

PS with 'setup' I mean an iTop instance, not the setup process 😉

piRGoif:
> I'm saying that default level `ERROR` is not satisfying indeed, should be `WARNING`.

ERROR is meant to avoid running possibly costly `debug_backtrace` in non dev env. But by modifying the config a non dev env can still add the file logging.

> PS with 'setup' I mean an iTop instance, not the setup process 😉

Oh ok ! The setup wizard is something I didn't think about so thanks anyway :D

Hipska:
Yeah, I have given a solution for that as well. For example, don't use `debug_backtrace` when not in dev env. But please, log the warning by default on each instance.

piRGoif:
>> ERROR is meant to avoid running possibly costly `debug_backtrace` in non dev env. But by modifying the config a non dev env can still add the file logging.
>
> Yeah, I have given a solution for that as well. For example, don't use `debug_backtrace` when not in dev env. But please, log the warning by default on each instance.

I don't agree as : 

* to me there is nothing useful to log if we don't have the call stack : you'll just see that something did a call to a deprecated thingy, with no details to solution the problem
* this log is meant for developers using iTop API, and should not exist in a production env by default. But having the possibility to enable it anyway in the config is a nice to have
* doing `debug_backtrace()` if and only if `\utils::IsDevelopmentEnvironment` returns true mean you won't ever be able to get this information on a non dev env - and as we saw "non dev env" definition is far from ideal for now, so anything that is not a direct clone won't have this

Can you elaborate on your proposition : surround the `debug_backtrace()` by a `\utils::IsDevelopmentEnvironment` call, but then log what ?

piRGoif:
About setup calls : there is another bug, as we are getting the config using `\MetaModel::GetConfig` : in the setup wizard this returns null, so we're getting the default level.
If we replace this call by `\utils::GetConfig` then this works as expected : on an instance having a config file then we're indeed getting the expected options.

I added another bug : N°3946

core/log.class.inc.php
```diff
@@ -603,85 +606,115 @@ public static function Trace($sMessage, $sChannel = null, $aContext = array())
...
-		if (isset($sLogLevelMin[$sChannel]))
-		{
+		if (isset($sLogLevelMin[$sChannel])) {
 			return $sLogLevelMin[$sChannel];
 		}
 
-		if (isset($sLogLevelMin[static::CHANNEL_DEFAULT]))
-		{
+		if (isset($sLogLevelMin[static::CHANNEL_DEFAULT])) {
 			return $sLogLevelMin[$sChannel];
```
Hipska, line 691: Is this intended? If I'm correct, `$sLogLevelMin[$sChannel]` would not even exist.
```suggestion
			return $sLogLevelMin[static::CHANNEL_DEFAULT]];
```

Hipska (changes requested):

Hipska:
> Can you elaborate on your proposition : surround the `debug_backtrace()` by a `\utils::IsDevelopmentEnvironment` call, but then log what ?

You have good points! 😂
`$sAdditionalMessage`? (if it would be mandatory) But you are right, I agree.

piRGoif:
> But you are right, I agree.

Ok, thanks
Sorry, I didn't gave much details earlier : most of this PR was discussed with the dev team members verbally, and I just wrote down a synthesis lacking the details.

core/log.class.inc.php
piRGoif: It is not, you're right !
This is part of the first bug I created : N°3944. This will be discussed with the products team and fixed independently, so I'm closing this conversation.

Molkobain (approved):

bootstrap.inc.php
piRGoif: Marking this conversation as resolved !

core/log.class.inc.php
```diff
@@ -765,10 +798,113 @@ public static function Log($iMySQLErrNo, $sMessage, $iMysqlErroNo = null, $aCont
...
+	}
+
+	protected static function GetLevelDefault(): string
+	{
+		if (utils::IsDevelopmentEnvironment()) {
+			return static::LEVEL_DEBUG;
+		}
+
+		return static::LEVEL_DEFAULT;
+	}
```
Hipska, line 830: Couldn't this be a candidate for the parent class?

piRGoif: Parent class is the LogAPI abstract class. This one is parent to 5 classes in iTop core, and some others in Combodo extensions. Some of those children classes are only for logging purpose (like IssueLog), some are used to report system errors (like DeadLockLog), some for dev info (like DeprecatedCallsLog).
The variety of LogAPI children usage make me think there is no need for now to pull up the trigger_error call mechanism. Maybe in the future if new needs appear ?

Hipska: The LogAPI class does not introduce a trigger_error. I was only talking about this mechanism in the `GetLevelDefault` method. Seems useful to me to auto set level to DEBUG on Dev Env?

piRGoif: Ok :)
I don't have a position on this right now, I think we would need to take a time to have a clear view of what is logged in which level. Both levels and channels were introduced in 2.7.0 so they are quite new and not used so much for now...
I'll make the proposition to the dev team, discussing this with multiple people will help :o).

bruno-ds: IMHO: 
 - opting in for debug or even trace level is quite easy
 - these level produce a lot of noise, that, in general is counterproductive (this is why they can be opt-in on a per channel basis)

so, from my point of vue, they should not change, even in dev. env.

piRGoif: Discussed with Bruno : as this PR is already merged, this will be discussed during the dev team meeting.
I'm closing the conversation.

piRGoif: Hello, an update about changing the default log level for all LogAPI impl on dev env : we discussed this with the team.

The conclusion is we won't change the current situation, that is almost all impl having OK as default log level, with some exceptions like DeprecatedCallsLog (for this one, the reason is none of those errors should exist and the developer should be aware of them as soon as possible - plus it's logged in a dedicated file).

Indeed for now changing default level in all impl would just give a few more logs, but levels are quite young in iTop history and we will certainly use them much more in the future ! DEBUG but also TRACE levels. We can imagine that in some versions there will be vast amount of logs in low levels : we don't want to overflood devs with logs they won't need all the time.

Though we made some efforts on the [log configuration documentation](https://www.itophub.io/wiki/page?id=2_7_0%3Aadmin%3Alog) to make changing log level super easy.

Hipska: Thanks for reporting back!

PR #198, opened 2021-03-30
🌐 improved ZH translation for 3.0

purplegrape:


piRGoif (approved):
Ok for me !
Translated existing keys were already translated by the author.

piRGoif:
Thanks @purplegrape !
You're right on time, as 3.0.0-beta should be packaged this week :) By the way, we are ok to include this PR in this release.

If you have time, can you test the beta version once it is out ? We'd be glad to have some screenshots in chinese so that we can see if the new keys fits well in the GUI for this language ! You can create a topic in the forums ? Thanks !

PR #202, opened 2021-03-31
[WIP][POC] the PhpUnit tests can now build and use various environments

bruno-ds:
**Objective**
Provide iTop core developers with a mean to implement unit test that require a specific datamodel.

**Usage**

1. Browse the directory test/testUtils/conf/targets, to identify the environment that has the piece of required datamodel. If necessary, further enrich its datamodel or even create a new environment.
2. Implement your test in class derived from iTopTestCase, that overrides the constant TEST_ITOP_ENV_DEFAULT

`class MyTestCaseTest extends ItopDataTestCase
{
	const TEST_ITOP_ENV_DEFAULT = 'myEnvironment';
	...
}`

Note: the compilation will occur each time your test is executed.

**Definition of a test environment**

A test environment is implemented into a dedicated directory under /test/testUtils/conf/targets.
Directory structure:

- Optional: delta.xml. A delta that will automatically be loaded on top of all the installed modules
- Mandatory?: target.ini. A recipe for building the environment. This recipe contains:
   - compile_from: the name of an environment it derives from
   - config_file: the full path to the configuration file
   - modules.... installed and selected modules to install

**About the proposed implementation**
Every changes are under the test directory, so as to minimize the risk of regressions.
On the other hand, a few refactorings occured with the test directory.

**Improvements to be discussed**

1. Do not compile systematically. Instead, provide the developpers with a mean to force compilation. Later: recompile automatically based on file dates.
2. Implement 2 environments: default and enriched
  - default: same as a basic installation (default choices in every setup screen). Every existing test will implicitely use that environment, decoupling tests from your installation settings (production environment) 
  - enriched: a datamodel with corner cases that are not found in the default environment. The datamodel maybe implemented in a specific extension maintained under /test/testUtils, or simply as a delta file.


piRGoif:
Discussed with Bruno : switching this PR to draft :)

test/testUtils/ItopDataTestCase.php
```diff
@@ -91,6 +88,37 @@ protected function setUp()
 		}
 	}
 
+	/**
+	 * At the time of the writting, je startup process isn't compatible with a custom env declared within the CLI
+	 * @throws \CoreException
+	 * @throws \DictExceptionUnknownLanguage
+	 * @throws \MySQLException
+	 */
+	public function EmulateApplicationStartup()
```
bruno-ds, line 97: This part is awful, but it's required since I didn't want to commit above the `test`  dir.

(the standard one forces us to make use of a session var, but the session_start within it overwrite every attempts so set the aforementioned var)

rquetiez:
I've just discussed with Bruno. Here is a new description and thoughts for the future of this PR.

**Objective**
Provide iTop core developers with a mean to implement unit test that require a specific datamodel.

**Usage**

1. Browse the directory test/testUtils/conf/targets, to identify the environment that has the piece of required datamodel. If necessary, further enrich its datamodel or even create a new environment.
2. Implement your test in class derived from iTopTestCase, that overrides the constant TEST_ITOP_ENV_DEFAULT

`class MyTestCaseTest extends ItopDataTestCase
{
	const TEST_ITOP_ENV_DEFAULT = 'myEnvironment';
	...
}`

Note: the compilation will occur each time your test is executed.

**Definition of a test environment**

A test environment is implemented into a dedicated directory under /test/testUtils/conf/targets.
Directory structure:

- Optional: delta.xml. A delta that will automatically be loaded on top of all the installed modules
- Mandatory?: target.ini. A recipe for building the environment. This recipe contains:
   - compile_from: the name of an environment it derives from
   - config_file: the full path to the configuration file
   - modules.... installed and selected modules to install

**About the proposed implementation**
Every changes are under the test directory, so as to minimize the risk of regressions.
On the other hand, a few refactorings occured with the test directory.

**Improvements to be discussed**

1. Do not compile systematically. Instead, provide the developpers with a mean to force compilation. Later: recompile automatically based on file dates.
2. Implement 2 environments: default and enriched
  - default: same as a basic installation (default choices in every setup screen). Every existing test will implicitely use that environment, decoupling tests from your installation settings (production environment) 
  - enriched: a datamodel with corner cases that are not found in the default environment. The datamodel maybe implemented in a specific extension maintained under /test/testUtils, or simply as a delta file.

bruno-ds:
> **Improvements to be discussed**
> 
> 1. Do not compile systematically. Instead, provide the developpers with a mean to force compilation. Later: recompile automatically based on file dates.

fyi, the "compile systematically" behavior is coded using a phpunit's event listener : test/testUtils/Listener/PurgeTestEnvDir.php

piRGoif:
Updated the description with the new one, so that it will be clearer for next reviewer !

piRGoif:
I'm closing the PR as it is very old.
But we still have this modification in mind... So this effort won't have be done for nothing at all !

PR #204, opened 2021-04-12
🐛 Decimal values with aggregate functions (dashlet group by)

jbostoen:
Opening this for discussion.

Why is the returned value of an aggregate function automatically converted to an Integer? It makes sense for count, which was the (only) initial aggregate function if I remember correctly.

However, when applied to an AttributeDecimal; this will be a float.
Explicitly converting this to a float is also not recommended: 1.10 would become 1.1 

So not sure if there's a reason for this explicit conversion/enforcement.
Perhaps a check to see if it's a null/empty value (if that is a possibility) and then simply putting "0" could be enough.

Otherwise it may need more tweaking.

https://sourceforge.net/p/itop/tickets/1961/

### DIFF
--- a/application/displayblock.class.inc.php
+++ b/application/displayblock.class.inc.php
@@ -478,7 +478,7 @@ class DisplayBlock
 					$aValues[$iRow] = $sValue;
 					$sHtmlValue = $oGroupByExp->MakeValueLabel($this->m_oFilter, $sValue, $sValue);
 					$aLabels[$iRow] = $sHtmlValue;
-					$aGroupBy[$iRow] = (int) $aRow[$sFctVar];
+					$aGroupBy[$iRow] = $aRow[$sFctVar];
 					$iTotalCount += $aRow['_itop_count_'];
 				}
 


Molkobain:
This would have to be checked with @eespie as he designed the group by. Will ask when he is back.

jbostoen:
Still in progress? :)

Hipska (approved):

Molkobain:
Completely forgot to ask him sorry.
I've put a reminder for tomorrow morning.

Molkobain:
Hello Jeffrey,

After discussing it internally, this was indeed an explicit limitation because at the time only the "count" aggregation function was available (even though you could want to account for .5 value 🤷‍♀️). This PR makes sense but it would require a few modifications for us to accept it:
- Rebase it on the develop branch so it will be part of the next major release
- Fix it for all the "rendering styles" (see DisplayBlock::RenderChartAjax), not only the "count" (table)
- Add some before / after screenshots for all concerned chart types (table, pie, bars)

Thanks!
Guillaume

jbostoen:
So rebase on iTop 3.0?
Is it best to wait (is Combodo still planning work in that area)?

Molkobain:
Yep rebase on 3.0 please. I think the dashlet internal stuff will stay stable for now, so you can get on it whenever you want; we might only do some polishing on the surface :)

jbostoen:
While I'm at it: would it be safe to pull today's develop branch?
I could also check the issue with linked sets in the meanwhile.

Hipska:
> While I'm at it: would it be safe to pull today's develop branch?

why not?

jbostoen:
> 
> 
> > While I'm at it: would it be safe to pull today's develop branch?
> 
> why not?

I actually meant if it would be a "working" version, or if there's some work in progress right now which might cause known issues

Molkobain:
I should be ok yep, we can go for it :)

PR #208, opened 2021-04-20
Feature/3968 mutex timeout

dflaven:
iTop Mutexes are based on the MySQL GET_LOCK() command. This command creates a server-side lock which is released either by calling RELEASE_LOCK() or when the connection to the MySQL server ends (which is handy in case of a PHP crash).

However, when a MySQL connection remains idle for a certain time (defined by the server parameter **wait_timeout**), it gets automatically closed by MySQL.

Since each iTop mutex uses its own MySQL connection (which sees no further commands once the lock is acquired), a "locked" Mutex is automatically  (and silently) released after wait_timeout seconds.

This pull request ensures that the wait_timeout _for the connection used by the mutex_ is at least 1 day (86400 seconds).

### DIFF
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -756,6 +756,9 @@ class DeadLockLog extends LogAPI
 	 *
 	 * @throws \Exception
 	 * @noinspection PhpParameterNameChangedDuringInheritanceInspection
+	 *
+	 * @since 2.7.1 method creation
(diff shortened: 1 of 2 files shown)


dflaven:
Je ne comprends pas comment ça fonctionne, ces pull requests. Inutilisable.

piRGoif:
> Je ne comprends pas comment ça fonctionne, ces pull requests. Inutilisable.

We'll be happy to help ?
I see this one was replaced by #209, I guess you had a branch problem ?

PR #209, opened 2021-04-20
N°3968 - Fixed mutex being silently released after wait_timeout seconds

dflaven:
iTop Mutexes are based on the MySQL GET_LOCK() command. This command creates a server-side lock which is released either by calling RELEASE_LOCK() or when the connection to the MySQL server ends (which is handy in case of a PHP crash).

However, when a MySQL connection remains idle for a certain time (defined by the server parameter **wait_timeout**), it gets automatically closed by MySQL.

Since each iTop mutex uses its own MySQL connection (which sees no further commands once the lock is acquired), a "locked" Mutex is automatically  (and silently) released after wait_timeout seconds.

Thi spull request ensures that the wait_timeout _for the connection used by the mutex_ is at least 86400 seconds (1 day).

### DIFF
--- a/core/mutex.class.inc.php
+++ b/core/mutex.class.inc.php
@@ -258,6 +258,38 @@ class iTopMutex
 		{
 			throw new Exception("Could not connect to the DB server (host=$sServer, user=$sUser): ".mysqli_connect_error().' (mysql errno: '.mysqli_connect_errno().')');
 		}
+
+		// Make sure that the server variable wait_timeout is at least 86400 seconds for this connection,
+		// since the lock will be released if/when the connection times out.
+
+		// BEWARE: If you want to check the value of this variable, when run from an interactive console "SHOW VARIABLES LIKE 'wait_timeout'" actually
+		// returns the value of the variable 'interactive_timeout' which may be quite different.
+
+		$sSql = "SHOW VARIABLES LIKE 'wait_timeout'";
+		$result = mysqli_query($this->hDBLink, $sSql);
+		if (!$result)
+		{
+		    throw new Exception("Failed to issue MySQL query '".$sSql."': ".mysqli_error($this->hDBLink).' (mysql errno: '.mysqli_errno($this->hDBLink).')');
+		}
+		if ($aRow = mysqli_fetch_array($result, MYSQLI_BOTH))
+		{
+		    $iTimeout = (int)$aRow[1];
+		}
+		else
+		{
+		    mysqli_free_result($result);
+		    throw new Exception("No result for query '".$sSql."'");
+		}
+		mysqli_free_result($result);
+
+		if ($iTimeout < 86400)
+		{
+		    $result = mysqli_query($this->hDBLink, 'SET SESSION wait_timeout=86400');
+		    if ($result === false)
+		    {
+		        throw new Exception("Failed to issue MySQL query '".$sSql."': ".mysqli_error($this->hDBLink).' (mysql errno: '.mysqli_errno($this->hDBLink).')');
+		    }
+		}
 	}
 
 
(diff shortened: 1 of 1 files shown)


Hipska:
Maybe not needed if it wouldn't have its own connection? Is there any conflict why they have their own connection?

core/mutex.class.inc.php
```diff
@@ -258,6 +258,38 @@ public function InitMySQLSession()
...
+		// BEWARE: If you want to check the value of this variable, when run from an interactive console "SHOW VARIABLES LIKE 'wait_timeout'" actually
+		// returns the value of the variable 'interactive_timeout' which may be quite different.
+
+		$sSql = "SHOW VARIABLES LIKE 'wait_timeout'";
+		$result = mysqli_query($this->hDBLink, $sSql);
+		if (!$result)
+		{
+		    throw new Exception("Failed to issue MySQL query '".$sSql."': ".mysqli_error($this->hDBLink).' (mysql errno: '.mysqli_errno($this->hDBLink).')');
+		}
+		if ($aRow = mysqli_fetch_array($result, MYSQLI_BOTH))
```
Hipska: Using the default (both) while using only numeric value.
```suggestion
		if ($aRow = mysqli_fetch_array($result, MYSQLI_NUM))
```

core/mutex.class.inc.php
```diff
@@ -258,6 +258,38 @@ public function InitMySQLSession()
...
+		{
+		    $iTimeout = (int)$aRow[1];
+		}
+		else
+		{
+		    mysqli_free_result($result);
```
Hipska: Not needed here, since it is executed anyway after the else block.

core/mutex.class.inc.php
```diff
@@ -258,6 +258,38 @@ public function InitMySQLSession()
...
+		    throw new Exception("No result for query '".$sSql."'");
```
Hipska: Would this ever occur? And does it need to kill the application? What about setting `$iTimeout` to zero then?

Hipska (changes requested):

piRGoif:
> Maybe not needed if it wouldn't have its own connection? Is there any conflict why they have their own connection?

I wasn't absolutely sure why, but I think this is because : 

> Before MySQL 5.7, only a single simultaneous lock can be acquired
Source https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html

Also on the same page : 

> A lock obtained with GET_LOCK() is released explicitly by executing RELEASE_LOCK() or implicitly when your session terminates

The connection in CMDBSource shouldn't be closed, but by having independent connections this is something less to worry about ;)

We should add a comment in the code explaining why indeed !

Hipska:
Okay, and 2.7 still asks for minimal MySQL 5.6 so indeed, that can't be changed yet. Might be something to consider for 3.0 though 😉

core/mutex.class.inc.php
piRGoif: @eespie and I are OK to leave it like that
Closing this conversation

core/mutex.class.inc.php
piRGoif: Good question
Indeed if this occurs we are in trouble :D 
We prefer to keep it this way

core/mutex.class.inc.php
piRGoif: Thanks for the suggestion, but I'm not absolutly sure of the real benefits, and as there was already manual tests done with the current code I'm worried changing this could cause regressions : I'm keeping code like this for now.

Hipska: Memory optimisation: `$aRow`
before:
```
array(2) {
  [0]=>
  string(5) "12345"
  ["wait_timeout"]=>
  string(5) "12345"
}
```
after:
```
array(1) {
  [0]=>
  string(5) "12345"
}
```

See [docs](https://www.php.net/manual/en/mysqli-result.fetch-array.php)

piRGoif: Thanks !
Keeping this in mind for next similar operations

PR #201, opened 2021-03-31
Add apcu as optional extension

Hipska:
[SF#1286](https://sourceforge.net/p/itop/tickets/1286/)
The description text might need to include that a fallback to file-based cache will be used.

PS: note that there are a lot of un-needed `if(function_exists('apc_` calls still existing in the code. Do you want me to find them and fix?

PR #212, opened 2021-05-07
N°3996 Add max_input_vars check in setup

piRGoif:
Check that max_input_vars is aligned with recommendation

If not, returns a warning : 

image

If equal or above recommended value then adds an info : 

image

### DIFF
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -332,6 +332,17 @@ class SetupUtils
 			$aResult[] = new CheckResult(CheckResult::INFO,  "Loaded php.ini files: $sPhpIniFile");
 		}
 
+		$iMaxInputVarsCurrent = ini_get('max_input_vars');
+		$iMaxInputVarsRecommended = 5000;
+		if (!is_numeric($iMaxInputVarsCurrent)) {
+			$iMaxInputVarsCurrent = 0;
+		}
+		if ($iMaxInputVarsCurrent < $iMaxInputVarsRecommended) {
+			$aResult[] = new CheckResult(CheckResult::WARNING,  "max_input_vars is set to $iMaxInputVarsCurrent : this is below recommended value of $iMaxInputVarsRecommended");
+		} else {
+			$aResult[] = new CheckResult(CheckResult::INFO,  "max_input_vars is set to $iMaxInputVarsCurrent (recommended value is $iMaxInputVarsRecommended)");
+		}
+
 		// Check the configuration of the sessions persistence, since this is critical for the authentication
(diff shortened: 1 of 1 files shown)


Hipska:
Wow, adding an additional warning in the setup? 😱 Be careful to not frighten users: https://github.com/Combodo/iTop/pull/201#issuecomment-814995968

Molkobain:
> Wow, adding an additional warning in the setup? 😱 Be careful to not frighten users: [#201 (comment)](https://github.com/Combodo/iTop/pull/201#issuecomment-814995968)

I was so waiting for you to comment with something like this 🤣🤣🤣

Molkobain:
Regarding the PR, I think it's a good idea to put a warning when server parameters do not match with th recommendations.

But in case they do, maybe show nothing (instead of an OK message) ? This way, the list does not grow too much in the coming versions. This is not a stand, rather a question / debate.

Hipska (approved):

piRGoif:
@azana38 did review and validate this PR.

About previous comments : 

> Wow, adding an additional warning in the setup? 😱 

@Hipska : when this max_input_vars option has an inappropriate value, this could cause a vast variety of problems in lots iTop screens : the cause is hard to understand, the problem seems to be random, and it is not easy to debug :( So this is indeed a very important option to keep iTop smooth and running.

> maybe show nothing (instead of an OK message) ?

@Molkobain : there are already some "ok" messages, and as this option is really a big deal I thought it would be a good idea to include it in the OK message. Can you check with the product team ?

Molkobain:
Discussed: PR will be reviewed next week with the Product team.

piRGoif:
About the "ok" message : will be discussed with the dev team in the coming weeks

piRGoif:
Woops sorry for the duplicate message, I didn't see yours Guillaume :/

piRGoif:
Discussed with @etaloc : this PR is rejected. Although we can have strange behaviors that are hard to debug with a `max_input_vars` inappropriate value, this doesn't happen when using the default datamodel : you need to have some customization and lots of attributes on a sole object to get bugs.
And also, modifying this option add risks to the server (the more data the webserver accept, the more it is subject to denial of service attacks).

As a workaround we will : 
* update the PHP/MySQL parameters wiki page
* add some logs on identified screens (pr #211)

Ideally we should detect that data were truncated, but this is far from trivial :/

piRGoif:
Wiki page updated : https://www.itophub.io/wiki/page?id=2_7_0%3Ainstall%3Aphp_and_mysql_configuration#max_input_vars

PR #214, opened 2021-05-19
Protect \iApplicationUIExtension::EnumAllowedActions uses

piRGoif:
Some impl just return null while we expect to have an array... This is causing PHP notices in lots of iTop instances with modules implementing this method incorrectly !

This modification get rid of the notice and add a log (warning level) indicating the impl class.

Sample XML to reproduce this error :

```xml
  <snippets>
    <snippet id="invalid_EnumAllowedActions_1" _delta="define">
      <placement>core</placement>
      <rank>0</rank>
      <content><![CDATA[
class InvalidEnumAllowedActions1 extends AbstractApplicationUIExtension {
    public function EnumAllowedActions(DBObjectSet $oSet){
     return null;
    }
}
      ]]></content>
    </snippet>
    <snippet id="invalid_EnumAllowedActions_2" _delta="define">
      <placement>core</placement>
      <rank>0</rank>
      <content><![CDATA[
class InvalidEnumAllowedActions2 extends AbstractApplicationUIExtension {
    public function EnumAllowedActions(DBObjectSet $oSet){
     return null;
    }
}
      ]]></content>
    </snippet>
    <snippet id="correct_EnumAllowedActions" _delta="define">
      <placement>core</placement>
      <rank>0</rank>
      <content><![CDATA[
class CorrectEnumAllowedActions extends AbstractApplicationUIExtension {
    public function EnumAllowedActions(DBObjectSet $oSet){
     return ['🏢 Combodo website' => 'https://www.combodo.com'];
    }
}
      ]]></content>
    </snippet>
  </snippets>
</itop_design>
```

This would generate 2 PHP notice in object list, and 4 in object details.

After the modification, no more PHP notice and we get those logs (plus an exception) : 
```
2021-05-19 12:17:21 | Warning | Some extensions returned non array value for EnumAllowedActions() method impl | IssueLog
array (
  'extensions' => 
  array (
    0 => 'InvalidEnumAllowedActions1',
    1 => 'InvalidEnumAllowedActions2',
  ),
)
2021-05-19 12:17:21 | Error   | Exception during GetDisplay: Some extensions returned non array value for EnumAllowedActions() method impl: 0 = InvalidEnumAllowedActions1, 1 = InvalidEnumAllowedActions2 | IssueLog
```
On a non dev env, we would get only the first log, and no exception.

### DIFF
--- a/application/displayblock.class.inc.php
+++ b/application/displayblock.class.inc.php
@@ -2327,19 +2327,34 @@ class MenuBlock extends DisplayBlock
 	 */
 	private function GetEnumAllowedActions(DBObjectSet $oSet, callable $callback)
 	{
+		$aInvalidExtensions = [];
+
 		/** @var \iApplicationUIExtension $oExtensionInstance */
 		foreach (MetaModel::EnumPlugins('iApplicationUIExtension') as $oExtensionInstance) {
 			$oSet->Rewind();
 			$aExtEnumAllowedActions = $oExtensionInstance->EnumAllowedActions($oSet);
+
 			if (!is_array($aExtEnumAllowedActions)) {
-				$sExtensionClass = get_class($oExtensionInstance);
-				IssueLog::Warning(
-					"Extension '{$sExtensionClass}' returned non array value for EnumAllowedActions() method impl"
-				);
+				$aInvalidExtensions[] = get_class($oExtensionInstance);
 				continue;
 			}
+
 			foreach ($aExtEnumAllowedActions as $sLabel => $data) {
 				$callback($sLabel, $data);
+			}
+		}
+
+		if (!empty($aInvalidExtensions)) {
+			$sMessage = 'Some extensions returned non array value for EnumAllowedActions() method impl';
+
+			IssueLog::Warning(
+				$sMessage,
+				null,
+				['extensions' => $aInvalidExtensions]
+			);
+
+			if (utils::IsDevelopmentEnvironment()) {
+				throw new CoreUnexpectedValue($sMessage, $aInvalidExtensions);
 			}
 		}
 	}
--- a/application/displayblock.class.inc.php
+++ b/application/displayblock.class.inc.php
@@ -1992,7 +1992,7 @@ class MenuBlock extends DisplayBlock
 
 						$this->AddMenuSeparator($aRegularActions);
 
-						$this->GetEnumAllowedActions($oSet, function ($sLabel, $data) use ($aRegularActions, $aActionParams) {
+						$this->GetEnumAllowedActions($oSet, function ($sLabel, $data) use (&$aRegularActions, $aActionParams) {
 							$aRegularActions[$sLabel] = array('label' => $sLabel, 'url' => $data) + $aActionParams;
 						});
 					}
@@ -2096,7 +2096,7 @@ class MenuBlock extends DisplayBlock
 
 			$this->AddMenuSeparator($aRegularActions);
 
-			$this->GetEnumAllowedActions($oSet, function ($sLabel, $data) use ($aRegularActions, $aActionParams) {
+			$this->GetEnumAllowedActions($oSet, function ($sLabel, $data) use (&$aRegularActions, $aActionParams) {
 				if (is_array($data)) {
 					// New plugins can provide javascript handlers via the 'onclick' property
 					//TODO: enable extension of different menus by checking the 'target' property ??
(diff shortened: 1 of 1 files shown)


Molkobain:
This is a very good idea. Could we imagine a more strict behavior when in development environement? Like an exception being thown or something that the dev cannot not see. Because I don't think many checks the warning logs when developing, or maybe it's just me? 😁

I don't recall which approach we took for the deprecated calls, but maybe something similar?

bruno-ds (dismissed):
Brillant!

piRGoif:
Thanks to both of you for your reviews !
Very good suggestion Guillaume, I'm going to do that !

Also both calls are almost identical, that's a pitty to have duplicate code for this, I'm going to try factorizing those !

Hipska:
Might as well be implemented for other methods that expect an array.

Couldn't the `iApplicationUIExtension` be updated yet to specify the return type? I don't know if minimal php version supports this.

On the other hand, I don't like the fact I need to add a `return array()` to this method when implementing this interface for another method.. IMHO `null` should be accepted without errors/notices.

piRGoif:
> Might as well be implemented for other methods that expect an array.

Yes, this PR won't resolve all the use cases, I've just seen a real life problem, I'm trying to at least solve this one for future versions.

> Couldn't the iApplicationUIExtension be updated yet to specify the return type?

In 2.7.0 we added some abstract classes to ease implementations !
Like for this case, instead of implementing all methods of iApplicationUIExtension, simply extends AbstractApplicationUIExtension

application/displayblock.class.inc.php
```diff
@@ -1992,13 +1992,9 @@ public function GetRenderContent(WebPage $oPage, array $aExtraParams = [], strin
...
-						/** @var \iApplicationUIExtension $oExtensionInstance */
-						foreach (MetaModel::EnumPlugins('iApplicationUIExtension') as $oExtensionInstance) {
-							$oSet->Rewind();
-							foreach ($oExtensionInstance->EnumAllowedActions($oSet) as $sLabel => $sUrl) {
-								$aRegularActions[$sLabel] = array('label' => $sLabel, 'url' => $sUrl) + $aActionParams;
-							}
-						}
+						$this->GetEnumAllowedActions($oSet, function ($sLabel, $data) use ($aRegularActions, $aActionParams) {
+							$aRegularActions[$sLabel] = array('label' => $sLabel, 'url' => $data) + $aActionParams;
+						});
```
Hipska, line 1997: Shouldn't this one also check if the array value is an array or url string like below?

piRGoif: This would be the subject of another PR, not the use case I'm trying to address here
I'm trying to keep focus as much as possible to ease review and allow this modification to make it to the develop branch...

piRGoif:
> I don't recall which approach we took for the deprecated calls, but maybe something similar?

In \DeprecatedCallsLog::Log we are doing a trigger_error with the E_USER_DEPRECATED flag.
Here, I'll go with an exception.

application/displayblock.class.inc.php
Hipska: Yes indeed, I understand. But make it an internal ticket about this potential bug..

Hipska (dismissed):

Molkobain:
> > I don't recall which approach we took for the deprecated calls, but maybe something similar?
> 
> In \DeprecatedCallsLog::Log we are doing a trigger_error with the E_USER_DEPRECATED flag.
> Here, I'll go with an exception.

Review ok for me when the exception will be in place 👍

piRGoif:
I've just committed the version with an exception throwned on dev env !
@bruno-ds and @Molkobain what do you think ? (and @Hipska also of course O:) )

I've updated the description to add a sample and details on the new result

Hipska (dismissed):

application/displayblock.class.inc.php
piRGoif: We have a whole bunch of ideas to improve the extensibility API, robustness like this is already referenced :)
I'm closing this conversation.

Molkobain (dismissed):

Molkobain:
Nice work 👍

piRGoif:
Discussed with @Molkobain : both calls are now using anonymous functions... But I didn't add a `&` prefix to pass by reference in the use statement the variable we need to modify :(
I've just fix this !
Also updated the XML sample in the PR description.

Molkobain (approved):

piRGoif:
Discussed with @Molkobain, OK to merge !

piRGoif:
Combodo internal ref : N°4024

PR #215, opened 2021-05-20
N°3806 If php memory_limit is lower than 256M, set 256M. If it's greater do …

BGdu38:
My memory_limit is set to 512M, but for csv import memory limit is hard coded in iTop.

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -608,7 +608,7 @@ class utils
 	 */
 	public static function IsMemoryLimitOk($iMemoryLimit, $iRequiredLimit)
 	{
-		return ($iMemoryLimit >= $iRequiredLimit) || ($iMemoryLimit == -1);
+		return ($iMemoryLimit >= $iRequiredLimit) || ($iMemoryLimit === -1);
 	}
 
 	/**
@@ -635,7 +635,7 @@ class utils
 			// In that case, ini_set will not work
 			return false;
 		}
-		$iMemoryLimit = utils::ConvertToBytes($sMemoryLimit);
+		$iMemoryLimit = static::ConvertToBytes($sMemoryLimit);
 
 		if (static::IsMemoryLimitOk($iMemoryLimit, $iRequiredLimit)) {
 			return null;
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -563,48 +563,86 @@ class utils
 
 	public static function ReadFromFile($sFileName)
 	{
-		if (!file_exists($sFileName)) return false;
+		if (!file_exists($sFileName)) {
+			return false;
+		}
+
 		return file_get_contents($sFileName);
 	}
 
 	/**
 	 * Helper function to convert a value expressed in a 'user friendly format'
 	 * as in php.ini, e.g. 256k, 2M, 1G etc. Into a number of bytes
+	 *
 	 * @param mixed $value The value as read from php.ini
+	 *
 	 * @return number
 	 */
-	public static function ConvertToBytes( $value )
+	public static function ConvertToBytes($value)
 	{
 		$iReturn = $value;
-	    if ( !is_numeric( $value ) )
-		{
-	        $iLength = strlen( $value );
-	        $iReturn = substr( $value, 0, $iLength - 1 );
-	        $sUnit = strtoupper( substr( $value, $iLength - 1 ) );
-	        switch ( $sUnit )
-			{
-	            case 'G':
-	                $iReturn *= 1024;
-	            case 'M':
-	                $iReturn *= 1024;
-	            case 'K':
-	                $iReturn *= 1024;
-	        }
-	    }
-        return $iReturn;
-    }
-  
-  /**
-   * Checks if the memory limit is at least what is required
-   *
-   * @param int $memoryLimit set limit in bytes
-   * @param int $requiredLimit required limit in bytes
-   * @return bool
-   */
-  public static function IsMemoryLimitOk($memoryLimit, $requiredLimit)
-  {
-      return ($memoryLimit >= $requiredLimit) || ($memoryLimit == -1);
-  }
+		if (!is_numeric($value)) {
+			$iLength = strlen($value);
+			$iReturn = substr($value, 0, $iLength - 1);
+			$sUnit = strtoupper(substr($value, $iLength - 1));
+			switch ($sUnit) {
+				case 'G':
+					$iReturn *= 1024;
+				case 'M':
+					$iReturn *= 1024;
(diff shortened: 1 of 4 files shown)


Hipska (changes requested):
iTop already has a helper method for that, so the code should be like:
```php
if (utils::ConvertToBytes(ini_get('memory_limit')) < '268435456') {
```
But it should be moved to after the includes in that case.

Question to combodo: Should `ini_set` actually happen?

piRGoif:
Hello,
Many thanks for this contribution !
This is a known limitation, that is referenced in Combodo internal db as N°3806. The fix is planned for 2.7.5 but as low priority, so it might not make its way to this version ? We'll see in the upcoming weeks as the development has just started.
I'm putting this PR on hold for now, as this may not be the solution that will be chosen ?

Molkobain:
> iTop already has a helper method for that, so the code should be like:
> 
> ```
> if (utils::ConvertToBytes(ini_get('memory_limit')) < '268435456') {
> ```

Indeed using the `utils::ConvertToBytes()` method is indeed a good practice, good catch.

> But it should be moved to after the includes in that case.

Not sure to follow? It was already before them.

Hipska:
> > But it should be moved to after the includes in that case.
> 
> Not sure to follow? It was already before them.

Well, the helper method needs to be loaded in php before it can be used. Currently the `ini_set` is before all the loading.

Molkobain:
> > Not sure to follow? It was already before them.
> 
> Well, the helper method needs to be loaded in php before it can be used. Currently the `ini_set` is before all the loading.

Right! I thought you were saying the `ini_set` itself had to be before.

piRGoif:
Using utils::ConvertToBytes would be a good idea indeed ! But Hipska you're right, we need to mind to call it AFTER the required calls, whereas the ini_set is currently before them.

Searching for other memory_limit occurrences, I found all of those ini_set calls : 
* csvimport.php of course O:)
* pages/ajax.render.php (operation "xlsx_run")
* setup/ajax.dataloader.php
* webservices/backoffice.dataloader.php

I will reference this in our internal ticket !

Also, in the setup (\SetupUtils::CheckPhpAndExtensions) we have a call to `\utils::IsMemoryLimitOk`, I think this should be the one to call !

piRGoif:
Also this PR is based and develop... We will certainly do a cherry pick manually to integrate it :/

piRGoif:
I've just wrote a proposition, I've contacted @BGdu38, awaiting for his answer.

application/utils.inc.php
```diff
@@ -658,48 +658,86 @@ public static function GetUploadTempId($sTransactionId = null)
 
 	public static function ReadFromFile($sFileName)
 	{
-		if (!file_exists($sFileName)) return false;
+		if (!file_exists($sFileName)) {
+			return false;
+		}
+
```
Hipska, line 569: Seems not relevant for this PR?

application/utils.inc.php
```diff
@@ -658,48 +658,86 @@ public static function GetUploadTempId($sTransactionId = null)
...
+	 *
+	 * @param int $memoryLimit set limit in bytes
+	 * @param int $requiredLimit required limit in bytes
+	 *
+	 * @return bool
+	 */
+	public static function IsMemoryLimitOk($memoryLimit, $requiredLimit)
+	{
+		return ($memoryLimit >= $requiredLimit) || ($memoryLimit == -1);
+	}
```
Hipska:
```suggestion
	 * @param int $iMemoryLimit set limit in bytes
	 * @param int $iRequiredLimit required limit in bytes
	 *
	 * @return bool
	 */
	public static function IsMemoryLimitOk($iMemoryLimit, $iRequiredLimit)
	{
		return ($iMemoryLimit >= $iRequiredLimit) || ($iMemoryLimit == -1);
	}
```

application/utils.inc.php
```diff
@@ -658,48 +658,86 @@ public static function GetUploadTempId($sTransactionId = null)
...
+		$sMemoryLimit = trim(ini_get('memory_limit'));
+		if (empty($sMemoryLimit)) {
+			// On some PHP installations, memory_limit does not exist as a PHP setting!
+			// (encountered on a 5.2.0 under Windows)
+			// In that case, ini_set will not work
+			return false;
+		}
+		$iMemoryLimit = utils::ConvertToBytes($sMemoryLimit);
+
+		if (($iMemoryLimit >= $iRequiredLimit) || ($iMemoryLimit === -1)) {
```
Hipska:
```suggestion
		if (static::IsMemoryLimitOk($iMemoryLimit, $iRequiredLimit)) {
```

test/application/UtilsTest.php
```diff
@@ -552,4 +552,23 @@ public function GetMentionedObjectsFromTextProvider(): array
...
+	public function testConvertToBytes($sExpressionToConvert, $iExpectedConvertedValue)
+	{
+		$iCurrentConvertedValue = utils::ConvertToBytes($sExpressionToConvert);
+		self::assertEquals($iExpectedConvertedValue, $iCurrentConvertedValue, 'Converted value wasn\'t the one expected !');
+	}
+
+	public function ConvertToBytesProvider()
+	{
+		return [
+			'512M' => ['512M', 512 * 1024 * 1024],
```
Hipska, line 447: Also add test for other options:
```suggestion
			'123' => ['123', 123],
			'56k' => ['56k', 56 * 1024],
			'512M' => ['512M', 512 * 1024 * 1024],
			'2G' => ['2G', 2 * 1024 * 1024 * 1024],
```

application/utils.inc.php
piRGoif: You're right ! The formatter usually does this kind of things, modifying only modified lines... but just a few extra more. That was the case here also, it just added some formatting on existing code just above modified lines.

piRGoif:
I've pushed my modifications.
Comments & reviews are welcome !

Hipska (changes requested):

application/utils.inc.php
piRGoif: Wooops !! I feel quite shamy huhu
Thanks, good catch !

piRGoif: Also havin a strict comparison with -1 is a good think, I'll add it in another commit

Hipska (dismissed):

Molkobain (dismissed):

piRGoif:
Rebased the branch using `git rebase --onto support/2.7 develop PR/215_memory_limit`.
Then changed the base branch for the PR.

Ready to merge !

Hipska (approved):

PR #220, opened 2021-06-14
N°4518 Add config parameter to set platform as dev env

piRGoif:
Since 2.6.0 we have the `\utils::IsDevelopmentEnvironment` that is used by lots of behaviors.

This method is based on the `ITOP_REVISION` constant value, so that we can suppose we're using a repo clone directly or a package.

But now we want to extend its use to other use cases : for example, shared dev platform based on a package...
This would benefits especially on Twig cache, symlinks in setup, DeprecatedCallsLog, ...

That's the goal of this new `developer_mode.enabled` config parameter. If set to true, it will unlocks dev env functionalities. False will disabled those functionalities.
Default value is null : that will keep existing behavior.

### DIFF
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -1415,7 +1415,7 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => true,
 		],
-		'is_dev_env' => [
+		'developer_mode.enabled' => [
 			'type' => 'bool',
 			'description' => 'If true then uncloks dev env functionnalities, see \utils::IsDevelopmentEnvironment',
 			'default' => null,
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -1415,6 +1415,14 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => true,
 		],
+		'is_dev_env' => [
+			'type' => 'bool',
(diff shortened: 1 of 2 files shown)


piRGoif:
Created after the review of #190 by the dev team

piRGoif:
Hello,
I've updated the description. Did you had something else in mind to modify ?

piRGoif:
PR reviewed by the dev team. We agreed to simply rename the config parameter to developer_mode.enabled

Also, modified the description to add why we want this :)

piRGoif:
Rebased branch

piRGoif:
Ready to merge ! 
But I need to add a ticket and document the parameter in the wiki O:)

Hipska (dismissed):

core/config.class.inc.php
```diff
@@ -1415,6 +1415,14 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => true,
 		],
+		'developer_mode.enabled' => [
+			'type' => 'bool',
+			'description' => 'If true then uncloks dev env functionnalities, see \utils::IsDevelopmentEnvironment',
```
Molkobain: Typo "unlocks"

piRGoif: Good catch thanks  ! And there was a typo also in "functionalities"
Fixed in 5d4e81bc

PR #221, opened 2021-07-08
N°3863 - exec.php : allow subdirectories in the page parameter

piRGoif:
For now, exec.php can only call pages at the root level of modules. This is caused by basename() calls on both page and module parameters.

This PR brings the possibility to call pages in the whole module hierarchy, still using `\utils::RealPath` to ensure we're not trying to access something outside of the module.

### DIFF
--- a/pages/exec.php
+++ b/pages/exec.php
@@ -29,7 +29,7 @@ utils::InitTimeZone();
 
 
 /**
- * @param string $sPagePath relative path
+ * @param string $sPagePath full path (if symlink, it will be resolved)
  * @param array $aPossibleBasePaths list of possible base paths
  *
  * @return string|bool false if invalid path
@@ -39,7 +39,7 @@ function CheckPageExists(string $sPagePath, array $aPossibleBasePaths)
 {
 	$sTargetPage = false;
 	foreach ($aPossibleBasePaths as $sBasePath) {
-		$sTargetPage = utils::RealPath($sBasePath.'/'.$sPagePath, $sBasePath);
+		$sTargetPage = utils::RealPath($sPagePath, $sBasePath);
 		if ($sTargetPage !== false) {
 			return $sTargetPage;
 		}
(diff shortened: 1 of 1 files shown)


Hipska (dismissed):

pages/exec.php
```diff
@@ -25,32 +25,64 @@
...
-if (!file_exists($sTargetPage))
-{
-	// Do not recall the parameters (security takes precedence)
+$sEnvFullPath = APPROOT.'env-'.$sEnvironment;
+$sPageRelativePath = $sModule.'/'.$sPage;
+$sPageEnvFullPath = $sEnvFullPath.'/'.$sPageRelativePath;
+if (is_link($sPageEnvFullPath)) {
+	// in case module was compiled to symlink, we need to check against real linked path as symlink is resolved
+	$aPossibleBasePaths = [
+		APPROOT.'datamodels/2.x', // warning, won't work for datamodels/1.x for example !
```
Hipska: Not `3.x` for new major iTop release?

piRGoif: No we didn't changed the directory name, see https://github.com/Combodo/iTop/tree/develop/datamodels
Indeed there was just some little refactoring (like the itop-structure module creation) but no revolution : same iTop objects and attributes.

piRGoif:
Dev team review conclusions : 
* [ ] the new behavior should be explicitly enabled (white list of the modules, or better an option in the module.*.php file !)
* [ ] there is already an existing provider for the extension locations : \ModelFactory::GetRootDirs
* [ ] does a test exist for utils::RealPath ?
* [ ] add a test for CheckPageExists !

Molkobain:
Did you ask @dflaven about why he didn't enable this in the beginning? It might have been on purpose.

piRGoif:
I talk about this with @rquetiez who made this originally.
This code is very old, and now we have \utils::RealPath \o/

pages/exec.php
```diff
@@ -25,32 +25,64 @@
...
+	return $sTargetPage;
+}
+
+
+$sModule = utils::ReadParam('exec_module', '');
+if ($sModule == '') {
 	echo "Missing argument 'exec_module'";
 	exit;
 }
-$sModule = basename($sModule); // protect against ../.. ...
```
Molkobain, line 36: Did we loose this protection?
I understand the need to drill through the folders of the module, but we *must not* be able to execute something outside of the iTop folder (eg. `../../../home/foo/malicious.php`)

pages/exec.php
```diff
@@ -25,32 +25,64 @@
...
 
 $sPage = utils::ReadParam('exec_page', '', false, 'raw_data');
-if ($sPage == '')
-{
+if ($sPage == '') {
 	echo "Missing argument 'exec_page'";
 	exit;
 }
-$sPage = basename($sPage); // protect against ../.. ...
```
Molkobain, line 44: Same comment as above

pages/exec.php
piRGoif: We are now calling CheckPageExists function, which calls utils::RealPath
This later ensure we won't allow to access files in the directories above in the hierarchy

Molkobain: Thanks! 👌
I saw the call of `utils::RealPath` but didn't noticed that it was actually checking that the filepath was contained in the basepath; preventing that kind a vulnerability; my bad 🙈

Molkobain (approved):

pages/exec.php
Molkobain: This means that the extensions made using this feature won't work on the iTop used internally at Combodo (poke @v-dumas ) and on other legacy packages. It might worth it to test for both 1.x and 2.x folders?

v-dumas: Do not put any code inside iTop which would not be compatible with legacy packages, that would be too much of a nightmare for the legacies...

Molkobain:
```suggestion
APPROOT.'datamodels/1.x', // Compatibility with legacy packages
APPROOT.'datamodels/2.x',
```

Molkobain (changes requested):

pages/exec.php
piRGoif: Good point, thanks ! This should be solved by using \ModelFactory::GetRootDirs

pages/exec.php
piRGoif: No prb, better check twice this kind of things O:)

piRGoif:
Rebased branch

pages/exec.php
Hipska: But it still isn't

piRGoif: Should be solved in 53a2fe2e

Hipska (approved):

Molkobain (approved):

Molkobain:

Molkobain:

piRGoif:
This new feature needs to be documented, but this will be done later on, and the follow-up will be done using the Combodo private ticketing system.
I'm merging !

PR #222, opened 2021-07-09
N°3200 Add "filter list" menu item for any datatable widget

piRGoif:
There was already two existing actions : 
* Add To Dashboard
* Create a Shortcut

This adds a new icon "Filter list".
This options is present in any datatable, except when we are in a search (UI.php operation=search, OQLMenuNode). That means this icon will appear for example in : 
* list dashlet
* 1-n or n-n attributes
* run query

For object relations, the query is rewritten otherwise we won't be having the correct advanced search criteria. It will also ease the "configure this list" option.
Unfortunately, there are no mean for now to pass fields list to UI.php, so we will display the default. Also we won't pass the table id so we won't get the datatable configuration set in the object.

Screenshot in the Team object's members tab :

image

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -1542,20 +1542,20 @@ class utils
 	}
 
 	/**
-	 * We cannot use iMenuId (corresponding values in iPopupMenuExtension constants) as value is always \iPopupMenuExtension::MENU_OBJLIST_TOOLKIT
+	 * We cannot use iMenuId (corresponding values in {@see \iPopupMenuExtension} constants) as value is always {@see \iPopupMenuExtension::MENU_OBJLIST_TOOLKIT}
 	 * whenever we are in a datatable, whereas it is included in a object tab, a dashlet or a search.
 	 *
 	 * So a {@see \ContextTag} is set on the corresponding calls.
 	 *
 	 * @return bool true if we are in a search page context, either directly or by the datatable ajax call
 	 *
-	 * @since 3.0.0
+	 * @since 3.1.0 N°3200
 	 *
-	 * @uses \ContextTag::TAG_SEARCH
+	 * @uses \ContextTag::TAG_OBJECT_SEARCH
 	 */
 	public static function IsCurrentPageASearch(): bool
 	{
-		if (ContextTag::Check(ContextTag::TAG_SEARCH)) {
+		if (ContextTag::Check(ContextTag::TAG_OBJECT_SEARCH)) {
 			return true;
 		}
 
@@ -1564,28 +1564,111 @@ class utils
 
 	/**
 	 * @param \DBObjectSearch $oFilter object list
+	 * @param array $aExtraParams
 	 *
 	 * @return string|null null if we are already in a search, otherwise the URL to open this list in a search
 	 *
 	 * @throws \ArchivedObjectException
 	 * @throws \CoreException
-	 *
 	 * @uses utils::IsCurrentPageASearch()
 	 *
-	 * @since 3.0.0
+	 * @since 3.1.0 N°3200
 	 */
-	public static function GetDataTableSearchUrl(DBSearch $oFilter): ?string
+	public static function GetDataTableSearchUrl(DBSearch $oFilter, array $aExtraParams): ?string
 	{
 		if (static::IsCurrentPageASearch()) {
 			// we don't want to add the link when already in a search page !
 			return null;
 		}
 
+		$bIsObjectRelation = isset($aExtraParams['object_id'], $aExtraParams['target_attr']);
+		if ($bIsObjectRelation) {
+			[$oDataTableSearchFilter, $aParams] = static::GetDataTableSearchForRelations($oFilter, $aExtraParams);
+		} else {
+			$oDataTableSearchFilter = $oFilter;
+			$aParams = [];
+		}
+
+		if (isset($aExtraParams['table_id'])) {
+			$aParams['table_id'] = $aExtraParams['table_id'];
+		}
+		$sParams = json_encode($aParams);
+
 		$sAppRootUrl = static::GetAbsoluteUrlAppRoot();
 		$oAppContext = new ApplicationContext();
-		$sUrl = $sAppRootUrl.'pages/UI.php?operation=search&'.$oAppContext->GetForLink().'&filter='.rawurlencode($oFilter->serialize());
+
+		$sUrl = $sAppRootUrl
+			.'pages/UI.php?operation=search&'
+			.$oAppContext->GetForLink()
+			.'&filter='.rawurlencode($oDataTableSearchFilter->serialize());
+		$sUrl .= '&aParams='.rawurlencode($sParams); // Not working... yet, cause not handled by UI.php
 
 		return $sUrl;
+	}
+
+	/**
+	 * Rewrites filter for object relations, so that in the search page we will have the correct criteria and will be able to use "configure this list"
+	 *
+	 * @param \DBSearch $oFilter object list
+	 * @param array{link_attr: string, target_attr: string, object_id: string} $aExtraParams
+	 *
+	 * @return array{\DBObjectSearch, string[]}
+	 * @throws \CoreException
+	 * @throws \OQLException
+	 *
+	 * @since 3.1.0 N°3200
+	 */
+	private static function GetDataTableSearchForRelations(DBSearch $oFilter, array $aExtraParams): array
+	{
+		$sObjectId = $aExtraParams['object_id'];
+		$bIsLinkedSetIndirect = isset($aExtraParams['link_attr']);
+		if ($bIsLinkedSetIndirect) {
+			//--- AttributeLinkedSetIndirect (n,n => lnk class)
+			$sLnkClass = $oFilter->GetClass();
+			$sExtKeyToObjectClass = $aExtraParams['link_attr'];
+			$sExtKeyToRemoteClass = $aExtraParams['target_attr'];
+
+			/** @var \AttributeExternalKey $oLnkExtKeyToRemote */
+			$oLnkExtKeyToRemote = MetaModel::GetAttributeDef($sLnkClass, $sExtKeyToRemoteClass);
+			$sRemoteClass = $oLnkExtKeyToRemote->GetTargetClass();
+
+			/** @var \AttributeExternalKey $oLnkExtKeyToRemote */
+			$oLnkExtKeyToObject = MetaModel::GetAttributeDef($sLnkClass, $sExtKeyToObjectClass);
+			$sObjectClass = $oLnkExtKeyToObject->GetTargetClass();
+
+			/** @var \AttributeExternalKey $oLnkExtKeyToRemote */
+			$oObjectExtKeyToLnk = $oLnkExtKeyToObject->GetMirrorLinkAttribute();
+			$sObjectExtKeyToLnkClass = $oObjectExtKeyToLnk->GetCode();
+
+			$sRemoteClassAliasName = ormLinkSet::REMOTE_ALIAS;
+			$sLnkClassAliasName = ormLinkSet::LINK_ALIAS;
+			$sOql = <<<SQL
+SELECT {$sRemoteClassAliasName},{$sLnkClassAliasName}
+	FROM {$sRemoteClass} AS {$sRemoteClassAliasName}
+	JOIN {$sLnkClass} AS {$sLnkClassAliasName} ON {$sLnkClassAliasName}.$sExtKeyToRemoteClass = {$sRemoteClassAliasName}.id
+	WHERE {$sLnkClassAliasName}.$sExtKeyToObjectClass = $sObjectId
+SQL;
+
+			$aAttCodesToDisplay = MetaModel::GetAttributeLinkedSetIndirectDatatableAttCodesToDisplay($sObjectClass, $sObjectExtKeyToLnkClass, $sRemoteClass, $sExtKeyToRemoteClass);
+			/** @noinspection PhpUnnecessaryLocalVariableInspection */
+			$sAttCodesToDisplay = implode(',', $aAttCodesToDisplay);
+			$aParams = [
+				'zlist'        => false,
+				'extra_fields' => $sAttCodesToDisplay,
+			];
+		} else {
+			//--- AttributeLinkedSet (1,n => AttributeExternalKey)
+			$sClass = $oFilter->GetClass();
+			$sExtKeyCode = $aExtraParams['target_attr'];
+
+			$sOql = "SELECT $sClass WHERE $sExtKeyCode = $sObjectId";
+
+			$aParams = [];
+		}
+
+		$oDataTableSearchFilter = DBSearch::FromOQL($sOql);
+
+		return [$oDataTableSearchFilter, $aParams];
 	}
 
 	/**
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -1462,6 +1462,7 @@ class utils
 				}
 				$aResult[] = new JSPopupMenuItem('UI:Menu:AddToDashboard', Dict::S('UI:Menu:AddToDashboard'), "DashletCreationDlg('$sOQL', '$sContext')");
 				$aResult[] = new JSPopupMenuItem('UI:Menu:ShortcutList', Dict::S('UI:Menu:ShortcutList'), "ShortcutListDlg('$sOQL', '$sDataTableId', '$sContext')");
+
 				break;
 
 			case iPopupMenuExtension::MENU_OBJDETAILS_ACTIONS:
@@ -1515,37 +1516,76 @@ class utils
 			default:
 				// Unknown type of menu, do nothing
 		}
-		foreach ($aResult as $oMenuItem)
-		{
+		foreach ($aResult as $oMenuItem) {
 			$aActions[$oMenuItem->GetUID()] = $oMenuItem->GetMenuItem();
 		}
 
 		// Invoke the plugins
 		//
 		/** @var \iPopupMenuExtension $oExtensionInstance */
-		foreach (MetaModel::EnumPlugins('iPopupMenuExtension') as $oExtensionInstance)
-		{
-			if (is_object($param) && !($param instanceof DBObject))
-			{
+		foreach (MetaModel::EnumPlugins('iPopupMenuExtension') as $oExtensionInstance) {
+			if (is_object($param) && !($param instanceof DBObject)) {
 				$tmpParam = clone $param; // In case the parameter is an DBObjectSet, clone it to prevent alterations
-			}
-			else
-			{
+			} else {
 				$tmpParam = $param;
 			}
-			foreach($oExtensionInstance->EnumItems($iMenuId, $tmpParam) as $oMenuItem)
-			{
-				if (is_object($oMenuItem))
-				{
+			foreach ($oExtensionInstance->EnumItems($iMenuId, $tmpParam) as $oMenuItem) {
+				if (is_object($oMenuItem)) {
 					$aActions[$oMenuItem->GetUID()] = $oMenuItem->GetMenuItem();
-					
-					foreach($oMenuItem->GetLinkedScripts() as $sLinkedScript)
-					{
+
+					foreach ($oMenuItem->GetLinkedScripts() as $sLinkedScript) {
 						$oContainerBlock->AddJsFileRelPath($sLinkedScript);
 					}
 				}
 			}
 		}
+	}
+
+	/**
+	 * We cannot use iMenuId (corresponding values in iPopupMenuExtension constants) as value is always \iPopupMenuExtension::MENU_OBJLIST_TOOLKIT
+	 * whenever we are in a datatable, whereas it is included in a object tab, a dashlet or a search.
+	 *
+	 * So a {@see \ContextTag} is set on the corresponding calls.
+	 *
+	 * @return bool true if we are in a search page context, either directly or by the datatable ajax call
+	 *
+	 * @since 3.0.0
+	 *
+	 * @uses \ContextTag::TAG_SEARCH
+	 */
+	public static function IsCurrentPageASearch(): bool
+	{
+		if (ContextTag::Check(ContextTag::TAG_SEARCH)) {
+			return true;
+		}
+
+		return false;
+	}
+
+	/**
+	 * @param \DBObjectSearch $oFilter object list
+	 *
+	 * @return string|null null if we are already in a search, otherwise the URL to open this list in a search
+	 *
+	 * @throws \ArchivedObjectException
+	 * @throws \CoreException
+	 *
+	 * @uses utils::IsCurrentPageASearch()
+	 *
+	 * @since 3.0.0
+	 */
+	public static function GetDataTableSearchUrl(DBSearch $oFilter): ?string
+	{
+		if (static::IsCurrentPageASearch()) {
+			// we don't want to add the link when already in a search page !
+			return null;
+		}
+
+		$sAppRootUrl = static::GetAbsoluteUrlAppRoot();
+		$oAppContext = new ApplicationContext();
+		$sUrl = $sAppRootUrl.'pages/UI.php?operation=search&'.$oAppContext->GetForLink().'&filter='.rawurlencode($oFilter->serialize());
+
+		return $sUrl;
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


Hipska (dismissed):

Hipska:
Wasn't there any other way to workaround that ugly hack?

piRGoif:
Yes it's really dirty :(
We'll have a chat with the rest of the dev team, I might have missed something obvious.
At least we could add some ContextTag...

Molkobain:
When making a PR formatting all the files makes it very difficult to see what you actually changed 😕

Hipska:
It's not all the files and not even the complete file, but it's true, I had a hard time to review the actual changes as well..

Molkobain:
Well you know what I meant, in this file there are tones of changes that you have to check to find that only the 4 last lines have been modified. 🤯 (And I didn't capture the rest of the file!)

image

dictionaries/en.dictionary.itop.ui.php
```diff
@@ -1373,6 +1373,7 @@
 	'Calendar-FirstDayOfWeek' => 0, // 0 = Sunday, 1 = Monday, etc...
 
 	'UI:Menu:ShortcutList' => 'Create a Shortcut...',
+	'UI:Menu:OpenSearch' => '🔍 Open list in search...',
```
Molkobain: Please don't use font for icons as they render different depending the browser / OS. Use FontAwesome icons like in the rest of the app. You can have a look at the "hyperlink configurator" for an example on how to do it.

application/utils.inc.php
```diff
@@ -1469,50 +1471,60 @@ public static function GetPopupMenuItemsBlock(iUIBlock &$oContainerBlock, $iMenu
...
+
+					foreach ($oMenuItem->GetLinkedScripts() as $sLinkedScript) {
 						$oContainerBlock->AddJsFileRelPath($sLinkedScript);
 					}
 				}
 			}
 		}
 	}
 
+	private static function GetDataTableSearchUrl(DBObjectSearch $oFilter): ?string
```
Molkobain: Some PHPDoc, at least the `@since` please?

Molkobain: Maybe this method could rather detect if we are currently in a search page or not? The method would then have more sense here in utils and it could be used more widely.

Molkobain (changes requested):

Molkobain:
Whereas for the feature itself, @v-dumas might be interested by it.

application/utils.inc.php
Hipska: Isn't that exactly what is done here?
https://github.com/Combodo/iTop/blob/85729d2ea88204faa01d431078c16b70e4061a67/application/utils.inc.php#L1502-L1505

Molkobain: I meant not call it `GetDataTableSearchUrl` but rather `IsSearchPage` or someting similar.

Hipska: Indeed, that makes sense.

Hipska:
I see that the same URL already is generated and is available as `$sURL`:
https://github.com/Combodo/iTop/blob/85729d2ea88204faa01d431078c16b70e4061a67/application/utils.inc.php#L1389-L1390

v-dumas:
> 
> 
> Whereas for the feature itself, @v-dumas might be interested by it.

Yes indeed, this is excellent if we could have it in LinkedSet and DashletList and if it can prefilter the search automatically with the appropriate criteria that would be perfect.
Then, assuming that action has an id, it could be displayed outside of the menu drop-down just by its icon, the day iTop knows how to do that, your challenge @steffunky !

dictionaries/en.dictionary.itop.ui.php
piRGoif: Totally agree ! Will do
I'm closing this as this is part of the dev team feedbacks

application/utils.inc.php
piRGoif: Good idea, having 2 methods is far better. Will do !
Closing this as this is part of the dev team feedbacks

piRGoif:
About formatting, well, this is not easy for a file like this, as I needed to add lines in a non well formatted switch case... I could have done better maybe (disabling the save action plugin, change the lines, commit, re-enable the plugin), but when you're in middle of your dev it's not that easy to forsee this...
I will push a commit reverting those comsetic changes, that will make the diff easier to read in GitHub.

I really advise to review code inside PHPStorm : PR are easy to browse and open, diffs can be seen without even doing a branch checkout, and indentations changes (actually white spaces) can be removed from display.

piRGoif:
I had a quick chat with v-dumas : a possible Combodo's internal ref is N°3200. We will discuss face to face next week on this subject :)

piRGoif:
Dev team reviewed this PR. Conclusion : 
* [x] no emoji in the menu items as the rendering will be platform dependent, we should rather use Font Awesome icons, and add icons also to existing keys (dashlet, shortcut) (see Hyperlink configurator extension)
* [x] the code that state if we are already in a search should be extracted in its own method
* [x] rather than testing on the current URL if we're in a search screen, use a aExtraDataParam if possible, or of not create new ContextTag ?
* [x] add phpdoc for new methods :)

application/utils.inc.php
piRGoif: > Some PHPDoc, at least the @since please?

Woops forgot to answer on this sorry
I'm a big supporter of adding those phpdoc tags in our API, but here this is a private method. What would be the reasons to add this tag here ?

piRGoif:
> I see that the same URL already is generated and is available as `$sURL`:

I'll have a look, thanks !

application/utils.inc.php
Hipska: Some small description so others directly know what it is used for would be handy ;) Even if private. (Maybe better change to `protected`?)

piRGoif: Agree on adding phpdoc :)
About protected, there shouldn't be any child of the utils class.

Molkobain:
> About formatting, well, this is not easy for a file like this, as I needed to add lines in a non well formatted switch case... I could have done better maybe (disabling the save action plugin, change the lines, commit, re-enable the plugin), but when you're in middle of your dev it's not that easy to forsee this...
> I will push a commit reverting those comsetic changes, that will make the diff easier to read in GitHub.
> 
> I really advise to review code inside PHPStorm : PR are easy to browse and open, diffs can be seen without even doing a branch checkout, and indentations changes (actually white spaces) can be removed from display.

No need to push another commit for the formatting (IMHO) as we already reviewed the changes. It was more a thought for next PRs.

application/utils.inc.php
Molkobain: > > Some PHPDoc, at least the @SInCE please?
> 
> Woops forgot to answer on this sorry
> I'm a big supporter of adding those phpdoc tags in our API, but here this is a private method. What would be the reasons to add this tag here ?

Even if it is private, it might help prevent a developer to use it when developing an extension that needs compatiblity with iTop 2.7 or older. (Some do the development on the develop branch, which can be tricky if they don't pay attention)

Hipska: So, you keep them private until release?

Molkobain: > So, you keep them private until release?

Sorry, I was making a point for Combodo's developers and the utility of the `@since` even in the internal classes / methods.

But you are totally right, as we rename the method it makes sense to change the visibility to `public` as it might be used around the app, both internally and by extension developers.

piRGoif:
Reviewed this today with @v-dumas. 

Conclusions : 

* [x] remove emoji indeed, but don't use FA icons yet (this will be subject of another FAF maybe, there are multiples issues around this, one is the icon that needs to be kept when transforming a menu item to a button using the `shortcut_actions` config parameter - N°4078 in Combodo internal DB)
* [x] change label to "Filter list..."
* [x] see what's happening when adding this menu item id in the `shortcut_actions` config parameter ?
* [x] in list dashlets we have a "menu" option that adds some icons (new object, refresh). Try to see if we could add the filter icon ?
* 🤔 in iPopupMenuExtension::EnumItems we have a $iMenuId parameter that gives where the menu is located, that could be an hint to add the filter option only on screens that are not the search itself

piRGoif:
Rebased branch

application/utils.inc.php
piRGoif: This is part of the dev team feebacks below.
I'm closing this conversation

piRGoif:
Updated description after label change

piRGoif:
Test adding the new menu item in the `shortcut_actions` config parameter. But this isn't working :/

Rendering is done in \MenuBlock::GetRenderContent, which is called twice. First time the item is not present, and on the second time it is but in the $aToolkitActions variables. The config param only has effect on the items present in the $aRegularActions or $aTransitionActions variables...

We may create a ticket to fix that in the future...

piRGoif:
Adding the menu item in dashlet menu : this would be a separate dev, I prefer not to do it now.
The job seems to be done in \DisplayBlock::RenderList (search for $bDisplayMenu var)

=> PR ready for new review

piRGoif:
Dev team review done : ok to merge !

piRGoif:
Waiting for a review with @v-dumas

piRGoif:
Rebase branch.

dictionaries/en.dictionary.itop.ui.php
```diff
@@ -1377,6 +1377,7 @@
 	'Calendar-FirstDayOfWeek' => 0, // 0 = Sunday, 1 = Monday, etc...
 
 	'UI:Menu:ShortcutList' => 'Create a Shortcut...',
+	'UI:Menu:OpenSearch' => 'Filter list...',
```
Molkobain: Shouldn't we remove the "..."? In UX the trailing dots should only be present when the user is going to be asked for another information/action in a popup/modal window. As this is redirecting it to another page, I would remove them.

piRGoif: I don't have an opinion on this really, my guess is that there wasn't a clear rule established from the beginning and now we have constants with or without trailing "..." :D 
For example in this menu both modify and delete (not visible in the description screenshot, but can be found for example in a list dashlet's menu) don't opens up a dialog...

I'll ask @v-dumas what he think about this ?

v-dumas: I would not consider that ... always means that there will be a dialog, but there will be another step before the action is truly performed, such as in the case of creation and modification, you will have to submit before the action is done, while here, I suspect that the search and so the result of the list will be executed immediately, if not, so if the user will have to enter some additional criteria, then press the refresh button. In that case only, the 3 dots can be considered as making sense.

Molkobain: https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#using-ellipses

piRGoif: Not sure what was your last argument Guillaume ? The document you linked say almost the same as Vincent does ? In this specific use cas, "filter list" menu item will open the same list, to filter it you'll need to add more criteria ?

Molkobain: It wasn't a final argument sorry, more a "hey check out what the good practices are" 😁

From what I read, I would say either label it
- "Filter this search..." with the ellipsis as it is intended that the user will make some actions on the criteria
- "Open in search" without the ellipsis as we don't ask the user for more actions on the next page

So the label can stay as-is I think, sorry for the disturbance! 😁

Could we just change the dict entry code then? For example `UI:Menu:OpenSearch` => `UI:Menu:FilterList`, this way:
- It will be consistent with the other one
- We won't have ambiguous entries when we a menu entry that opens a generic search form

v-dumas:
We reviewed it with @piRGoif yesterday:

- icon used could well be the already used magnifier as it does the same opening of a class search
- the menu should always be there on any list, as a simple icon always displayed
- when we open in a search, a many to many relationship, then the search executed should be a SELECT R,L FROM RemoteClass AS R JOIN LinkClass AS L ON ... thus it allow the user to filter both on the remote class and the link attributes if they are some. Bonus, if they aren't any attribute (_apart from the 2 ids, magic attributes and ExternalFields_) in the LinkedClass, we could limit the search to the remote class.
- when opening a n:1 relationship such as the Ticket of a Person, we have noticed that due to the way the query was build, the preset criteria is not easy to read and cannot be changed. Instead of SELECT Ticket AS T WHERE T.caller_id='xx', it does SELECT Ticket AS T JOIN Person AS P ON T.caller_id=P.id WHERE P.id='xx'. Perform those 2 queries in a Run Query, save as Shortcut and open the Shortcut, you will see the difference in terms of preset criteria in the search.

Molkobain:
> * the menu should **always be there on any list**, as a simple icon always displayed

When you say this, you mean always but not in a search right? Otherwise it will be confusing.

dictionaries/en.dictionary.itop.ui.php
piRGoif: Ok, thanks for the clarification :)

And yes I agree, I should rename the dict key

piRGoif:
> > * the menu should **always be there on any list**, as a simple icon always displayed
> 
> When you say this, you mean always but not in a search right? Otherwise it will be confusing.

Yes this is the idea : instead of adding it everywhere but in the search page as a menu entry, add it as a permanent icon like create and refresh.

piRGoif:
Wrote the TODO in the description, with checkboxes !

piRGoif:
Rebased branch

dictionaries/en.dictionary.itop.ui.php
piRGoif: Done in d5dc6fa8
Closing conversation !

piRGoif:
Rebased branch !

piRGoif:
Person.tickets_list

For now we are generating this query : 

```
SELECT `Ticket` FROM Ticket AS `Ticket` JOIN Person AS `Person` ON `Ticket`.caller_id = `Person`.id WHERE (`Person`.`id` = :id)
```

instead of : 

```
SELECT `Ticket` FROM Ticket AS `Ticket` JOIN Person AS `Person` ON `Ticket`.caller_id = `Person`.id WHERE (`Ticket`.`caller_id` = :id)
```

The method doing this is \AttributeLinkedSet::GetDefaultValue
It is : 
- creating a $oMyselfSearch DBObjectSearch on the current object class with its id
- creation a $oLinkSearch DBObjectSearch on the remote object
- doing `$oLinkSearch->AddCondition_PointingTo($oMyselfSearch, $sExtKeyToMe);`

I've asked the dev team if this could be changed...

piRGoif:
n-n attributes : same as before, this is done using the same means in \AttributeLinkedSet::GetDefaultValue (no impl in the child class AttributeLinkedSetIndirect, but in this impl we have a test : `if ($this->IsIndirect())`), and we are also calling \DBObjectSearch::AddCondition_PointingTo.

piRGoif:
Discussed with Romain about the DBSearch generation in \AttributeLinkedSet::GetDefaultValue : 
- this was like this from the very start (see b756db09 in... 2009 !!!)
- browsing the code history, we didn't find any specific reason for why the condition is set on Person and not Ticket ?

There are two choices : 
- either changing the code in \AttributeLinkedSet::GetDefaultValue
- or generating our own filter when creating the "filter list" icon ?

I'll have to check if the later is possible (do we have all the context we need when we generate the icon ? )

piRGoif:
Rebased branch on the current develop HEAD

piRGoif:
Now the link will use specific queries : 

* for AttributeLinkedSet (1,n) : 
```sql
SELECT ... WHERE extkey = ...
```

* for AttributeLinkedSetIndirect (n,n) : 
```sql
SELECT r,l
FROM ... AS R
JOIN lnk... AS l ON l.extkey2 = r...
WHERE l.extKey1 = ...
```

PR is ready for a new review.

@v-dumas we might do a test together to check everything is fine ?

Molkobain:
Ongoing developments on the linkedsets might have introduce conflicts with your branch 😕
Once you chatted with Vincent, see Benjamin for how to fit your branch in the refactoring.

piRGoif:
> Ongoing developments on the linkedsets might have introduce conflicts with your branch

Yes, that was almost certain this will happen :/
We will work with Vincent tomorrow, I'll see after if the branch can be rebased.

piRGoif:
We reviewed the current dev state with @v-dumas and @Molkobain this morning !

TODO : 
- ~~"More information about the query" isn't displayed on first page, but after refreshing the search (refresh button on the adv search criteria bar) then it appears ?!?? There might be a little something missing ?~~ => already known bug (N°4937), planned for 2.7.9 as WANT/4
- n-n relations : 
  - [x] keep the existing filters (for example in dashlets or DisplayBareRelations impl)
  - ~~add in the datatable actions menu a new item : "manage links". This will open a SELECT L,R (instead of R,L)~~ done in another PR
  - ~~see if we can open the SELECT R,L search without the L friendlyname and ext key friendlyname ? (was done in 3.0.0 in n-n relations displayed in object details, see N°2334)~~ ⚠ Not implemented yet in UI.php :(
  - [x] change the icon from magnifier to a filter ?
- Before getting back to review : 
  - [x] rebase on develop
  - [ ] review with @Molkobain 

Will be done in another context : 
- allow bulk edit and delete on multiple classes search

Note : I opened this PR because I needed to add filters in the search.
But bulk modify is also important, as it address use case like moving at once all items in Rack A to Rack B for example.

piRGoif:
Setting back to pending contributor

piRGoif:
Rebased and force pushed

piRGoif:
The "more information about the query" missing was a known bug. Edited the toto list !

piRGoif:
Settings fields to display in UI.php : actually this isn't possible for now, and I didn't manage to implement it :( Main problems are the aExtraParams datatable hell and the advanced search having an ajax load + rewriting the URL.

piRGoif:
About the columns list to pass to UI.php : 

We are calling ajax.searchform.php to display table data, using the seach_form_handler widget (definition in search_form_handler.js, call in \Combodo\iTop\Application\Search\SearchForm::GetSearchFormUIBlock).
The ajax endpoint then calls \DisplayBlock::GetDisplay (a if can also switch to \DisplayBlock::RenderContent which calls \DisplayBlock::GetRenderContent - but I don't know when ?)

With N°2334 we modified \Combodo\iTop\Application\UI\Links\AbstractBlockLinksViewTable::GetExtraParam
The table is then generated using \DisplayBlock::GetRenderContent (see \Combodo\iTop\Application\UI\Links\AbstractBlockLinksViewTable::InitTable)

Adding manually the same aExtraParams entries in ajax.searchform.php doesn't produced the same results :(
Both methods (GetDisplay VS GetRenderContent) are using different parameters...

piRGoif:
Saw and fixed a bug in OQLMenuNode : we were trying to add the action there as well, and it was generating notices !

piRGoif:
@v-dumas and @Molkobain the PR is ready for the new review ! 

Unfortunately passing fields list or table id isn't possible for now in UI.php, and it seems like a very big task to find out what parameter to set... This evolution should be planned and can't be in the same scope as this PR :(

However I fixed a lot of bugs, and as asked the query is rewritten for any 1-n or n-n object relation.

application/utils.inc.php
```diff
@@ -1515,50 +1516,170 @@ public static function GetPopupMenuItemsBlock(iUIBlock &$oContainerBlock, $iMenu
...
+					foreach ($oMenuItem->GetLinkedScripts() as $sLinkedScript) {
 						$oContainerBlock->AddJsFileRelPath($sLinkedScript);
 					}
 				}
 			}
 		}
 	}
 
+	/**
+	 * We cannot use iMenuId (corresponding values in iPopupMenuExtension constants) as value is always \iPopupMenuExtension::MENU_OBJLIST_TOOLKIT
```
```suggestion
	 * We cannot use iMenuId (corresponding values in {@see \iPopupMenuExtension} constants) as value is always {@see \iPopupMenuExtension::MENU_OBJLIST_TOOLKIT}
```

application/utils.inc.php
```diff
@@ -1515,50 +1516,170 @@ public static function GetPopupMenuItemsBlock(iUIBlock &$oContainerBlock, $iMenu
...
+	 * whenever we are in a datatable, whereas it is included in a object tab, a dashlet or a search.
+	 *
+	 * So a {@see \ContextTag} is set on the corresponding calls.
+	 *
+	 * @return bool true if we are in a search page context, either directly or by the datatable ajax call
+	 *
+	 * @since 3.0.0
```
```suggestion
	 * @since 3.1.0
```

core/contexttag.class.inc.php
```diff
@@ -52,17 +52,19 @@
...
-	const TAG_SETUP = 'Setup';
-	const TAG_SYNCHRO = 'Synchro';
-	const TAG_REST = 'REST/JSON';
+	public const TAG_PORTAL        = 'GUI:Portal';
+	public const TAG_CRON          = 'CRON';
+	public const TAG_CONSOLE       = 'GUI:Console';
+	public const TAG_SETUP         = 'Setup';
+	public const TAG_SYNCHRO       = 'Synchro';
+	public const TAG_REST          = 'REST/JSON';
+	public const TAG_OBJECT_SEARCH = 'ObjectSearch';
```
Molkobain, line 65:
```suggestion
	/**
	 * @var string
	 * @since 3.1.0
	 */
	public const TAG_OBJECT_SEARCH = 'ObjectSearch';
```

application/utils.inc.php
```diff
@@ -1515,50 +1516,170 @@ public static function GetPopupMenuItemsBlock(iUIBlock &$oContainerBlock, $iMenu
...
+	 * @param \DBObjectSearch $oFilter object list
+	 * @param array $aExtraParams
+	 *
+	 * @return string|null null if we are already in a search, otherwise the URL to open this list in a search
+	 *
+	 * @throws \ArchivedObjectException
+	 * @throws \CoreException
+	 * @uses utils::IsCurrentPageASearch()
+	 *
+	 * @since 3.0.0
```
```suggestion
	 * @since 3.1.0
```

application/utils.inc.php
```diff
@@ -1515,50 +1516,170 @@ public static function GetPopupMenuItemsBlock(iUIBlock &$oContainerBlock, $iMenu
...
+
+	/**
+	 * Rewrites filter for object relations, so that in the search page we will have the correct criteria and will be able to use "configure this list"
+	 *
+	 * @param \DBSearch $oFilter object list
+	 * @param array{link_attr: string, target_attr: string, object_id: string} $aExtraParams
+	 *
+	 * @return array{\DBObjectSearch, string[]}
+	 * @throws \CoreException
+	 * @throws \OQLException
```
Molkobain, line 1617:
```suggestion
	 * @return array{\DBObjectSearch, string[]}
	 * @throws \CoreException
	 * @throws \OQLException
	 * @since 3.1.0
```

core/metamodel.class.php
```diff
@@ -2064,6 +2064,40 @@ public static function GetZListAttDefsFilteredForIndirectLinkClass(string $sClas
...
+	 * @return string[] attcodes to display, containing aliases
+	 * @throws \CoreException
+	 */
+	public static function GetAttributeLinkedSetIndirectDatatableAttCodesToDisplay(string $sObjectClass, string $sObjectLinkedSetIndirectAttCode, string $sRemoteClass, string $sLnkExternalKeyToRemoteClassAttCode):array
+	{
+		$aLnkAttDefsToDisplay = MetaModel::GetZListAttDefsFilteredForIndirectLinkClass($sObjectClass, $sObjectLinkedSetIndirectAttCode);
+		$aRemoteAttDefsToDisplay = MetaModel::GetZListAttDefsFilteredForIndirectRemoteClass($sRemoteClass);
+		$aLnkAttCodesToDisplay = array_map(
+			function ($oLnkAttDef) {
+				return \ormLinkSet::LINK_ALIAS.'.'.$oLnkAttDef->GetCode();
```
Molkobain: Add a use statement for the ormLinkSet class please (several instances below)

Molkobain (commented):
Comments are not mandatory before merge, I can make them afterwards.
@piRGoif how do you want to proceed, do you (squash &) merge it yourself or do you want me to do it?

piRGoif:
> piRGoif how do you want to proceed, do you (squash &) merge it yourself or do you want me to do it?

Discussed with Guillaume : I will do it myself as it might be better to have more than 1 commit.

application/utils.inc.php
piRGoif: Fixed in 5c40081e

core/contexttag.class.inc.php
piRGoif: Fixed in 5c40081e

application/utils.inc.php
piRGoif: Fixed in 5c40081e

application/utils.inc.php
piRGoif: Fixed in 5c40081e

core/metamodel.class.php
piRGoif: To be discussed with Guillaume...

piRGoif: Discussed with Guillaume, will simply remove the "\"

piRGoif: Done in 5daaa2bf

PR #224, opened 2021-07-28
Feature/dict apcu bug

Molkobain:
Creating this PR on behalf of @odain-cbd for an easier review

### DIFF
--- a/core/apc-service.class.inc.php
+++ b/core/apc-service.class.inc.php
@@ -1,5 +1,9 @@
 <?php
 
+/**
+ * Class ApcService
+ * @since 2.7.6 N°4125
+ */
 class ApcService {
 	public function __construct() {
 	}
@@ -11,6 +15,7 @@ class ApcService {
 	/**
 	 * @param $key string|array
 	 * @return mixed
+	 * @see apc_fetch
 	 */
 	function apc_fetch($key)
 	{
@@ -22,6 +27,7 @@ class ApcService {
 	 * @param $var
 	 * @param int $ttl
 	 * @return array|bool
+	 * @see apc_store
 	 */
 	function apc_store($key, $var = NULL, $ttl = 0)
 	{
--- a/core/apc-service.class.inc.php
+++ b/core/apc-service.class.inc.php
@@ -5,7 +5,7 @@ class ApcService {
 	}
 
 	public function function_exists($function_name) {
-		return $this->function_exists($function_name);
+		return function_exists($function_name);
 	}
 
 	/**
--- a/core/apc-service.class.inc.php
+++ b/core/apc-service.class.inc.php
@@ -39,3 +39,4 @@ class ApcService {
 		return apc_store($key, $var, $ttl);
 	}
 }
+?>
--- a/core/apc-service.class.inc.php
+++ b/core/apc-service.class.inc.php
@@ -8,14 +8,19 @@ class ApcService {
 	public function __construct() {
 	}
 
+	/**
+	 * @param string $function_name
+	 * @return bool
+	 * @see function_exists()
+	 */
 	public function function_exists($function_name) {
 		return function_exists($function_name);
 	}
 
 	/**
-	 * @param $key string|array
+	 * @param string|array $key
 	 * @return mixed
-	 * @see apc_fetch
+	 * @see apc_fetch()
 	 */
 	function apc_fetch($key)
 	{
@@ -27,7 +32,7 @@ class ApcService {
 	 * @param $var
 	 * @param int $ttl
 	 * @return array|bool
-	 * @see apc_store
+	 * @see apc_store()
 	 */
 	function apc_store($key, $var = NULL, $ttl = 0)
 	{
--- a/core/apc-service.class.inc.php
+++ b/core/apc-service.class.inc.php
@@ -1,0 +1,30 @@
+<?php
+
+class ApcService {
+	public function __construct() {
+	}
+
+	public function function_exists($function_name) {
+		return $this->function_exists($function_name);
+	}
+
+	/**
+	 * @param $key string|array
+	 * @return mixed
+	 */
+	function apc_fetch($key)
+	{
+		return apc_fetch($key);
+	}
+
+	/**
+	 * @param array|string $key
+	 * @param $var
+	 * @param int $ttl
+	 * @return array|bool
+	 */
+	function apc_store($key, $var = NULL, $ttl = 0)
+	{
+		return apc_store($key, $var, $ttl);
+	}
+}
--- a/core/dict.class.inc.php
+++ b/core/dict.class.inc.php
@@ -273,6 +273,7 @@ class Dict
 			self::$m_aData[$sLangCode] = self::GetApcService()->apc_fetch(self::$m_sApplicationPrefix.'-dict-'.$sLangCode);
 			if (! is_array(self::$m_aData[$sLangCode]))
 			{
+				IssueLog::Warning("APCu corrupted data (with $sLangCode dictionnary). APCu configuration and running version should be troubleshooted...");
 				unset(self::$m_aData[$sLangCode]);
 			}
 			else
--- a/core/dict.class.inc.php
+++ b/core/dict.class.inc.php
@@ -152,7 +152,7 @@ class Dict
 
 		if (!array_key_exists($sLangCode, self::$m_aData))
 		{
-			IssueLog::Error("Cannot find $sLangCode in dictionnaries. default labels displayed", LogChannels::APC);
+			IssueLog::Warning("Cannot find $sLangCode in dictionnaries. default labels displayed");
 			// It may happen, when something happens before the dictionaries get loaded
 			return $sStringCode;
 		}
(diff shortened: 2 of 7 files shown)


core/apc-service.class.inc.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+class ApcService {
+	public function __construct() {
+	}
+
+	public function function_exists($function_name) {
+		return $this->function_exists($function_name);
```
Hipska: That doesn't look good? (seems like an endless loop, no?)

core/dict.class.inc.php
```diff
@@ -241,20 +262,22 @@ public static function SetLanguagesList($aLanguagesList)
 	public static function InitLangIfNeeded($sLangCode)
 	{
 		if (array_key_exists($sLangCode, self::$m_aData)) return true;
-		
+
 		$bResult = false;
-		
-		if (function_exists('apc_fetch') && (self::$m_sApplicationPrefix !== null))
+
+		if (self::GetApcService()->function_exists('apc_fetch')
```
Hipska, line 270: Why still using `function_exists`. The function will always exist because of `/core/apc-emulation.php`. This is one of the cleanup actions to be done..

core/apc-service.class.inc.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+class ApcService {
```
Hipska, line 7: What is the added value for this class?

odain-cbd: nothing fonctional. only testability. 

with this class we are able to isolate the piece of code to test and mock the outside services (here ApcService) used by it.

core/dict.class.inc.php
odain-cbd: I am not aware of cleanup action. thanks for the reminder.

but still I prefer to make something robust and clear for everybody. i still will keep function_exists to make clear that refactoring was made for testability purpose only. 3.0 is coming and we cannot afford any regression. 

when we will start cleanup actions this one will disappear with the others.

core/apc-service.class.inc.php
```diff
@@ -0,0 +1,30 @@
...
+	public function __construct() {
+	}
+
+	public function function_exists($function_name) {
+		return $this->function_exists($function_name);
+	}
+
+	/**
+	 * @param $key string|array
+	 * @return mixed
```
Hipska:
```suggestion
	 * @see apc_fetch
```

core/apc-service.class.inc.php
```diff
@@ -0,0 +1,30 @@
...
+	function apc_fetch($key)
+	{
+		return apc_fetch($key);
+	}
+
+	/**
+	 * @param array|string $key
+	 * @param $var
+	 * @param int $ttl
+	 * @return array|bool
```
Hipska, line 34:
```suggestion
	 * @see apc_store
```

Hipska (changes requested):

core/apc-service.class.inc.php
odain-cbd: good point. it could not work.... i will fix

test/core/dictTest.php
```diff
@@ -38,20 +38,82 @@
...
 {
+	private $sEnvName;
 	protected function setUp()
 	{
 		parent::setUp();
 		require_once (APPROOT.'core/coreexception.class.inc.php');
 		require_once (APPROOT.'core/dict.class.inc.php');
-		require_once 'mockDict.incphp';
+		require_once (APPROOT.'core/apc-service.class.inc.php');
+		$this->sEnvName = date("c");
```
piRGoif: This isn't working on Windows, and I assume on many FS.
Sample directory name generated on my Windows env : `C:\Dev\wamp64\www\itop-27/env-2021-11-24T14:25:55+00:00/dictionaries/fr-fr.dict.php`

test/core/dictApcuTest.php
```diff
@@ -0,0 +1,388 @@
...
+	private $sDictionaryFolder;
+
+	protected function setUp()
+	{
+		parent::setUp();
+		require_once (APPROOT.'core/coreexception.class.inc.php');
+		require_once (APPROOT.'core/dict.class.inc.php');
+		require_once (APPROOT.'core/apc-service.class.inc.php');
+
+		$this->sEnvName = date("c");
```
piRGoif: Same problem on both test : this isn't working on Windows

piRGoif (changes requested):
Please fix the PHPUnit tests

core/apc-service.class.inc.php
Hipska: Still need to remove the old `@param` and `@return` no?

piRGoif: I like the `@see` ! +1 to add it
I would keep the `@param` anyway, as it helps the IDE a lot :)

@odain warning, your `@param` order is reversed. See https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/param.html#param

I'm re-opening this thread

Hipska: I was under the impression, the IDE would fill them in that case. But if not, then please reuse the same description as the referenced function.

According to phpdoc reference it should even be `@see api_fetch()`.

test/core/dictTest.php
odain-cbd: i will use DIRECTORY_SEPARATOR for path compliancy with Windows

test/core/dictApcuTest.php
odain-cbd: idem

core/apc-service.class.inc.php
odain-cbd: phpdoc fixed based on your remarks. thank you.

core/dict.class.inc.php
```diff
@@ -241,20 +264,22 @@ public static function SetLanguagesList($aLanguagesList)
...
-			if (self::$m_aData[$sLangCode] === false)
-			{
+			self::$m_aData[$sLangCode] = self::GetApcService()->apc_fetch(self::$m_sApplicationPrefix.'-dict-'.$sLangCode);
+			if (self::$m_aData[$sLangCode] === false) {
 				unset(self::$m_aData[$sLangCode]);
-			}
-			else
-			{
+			} else if (! is_array(self::$m_aData[$sLangCode])) {
+				IssueLog::Warning("APCu corrupted data (with $sLangCode dictionnary). APCu configuration and running version should be troubleshooted...");
```
piRGoif:
```suggestion
				IssueLog::Error("APCu corrupted data (with $sLangCode dictionnary). APCu configuration and running version should be troubleshooted...", LogChannels::APC);
```

core/dict.class.inc.php
```diff
@@ -145,15 +147,17 @@ public static function S($sStringCode, $sDefault = null, $bUserLanguageOnly = fa
...
 		// Attempt to find the string in the user language
 		//
-		self::InitLangIfNeeded(self::GetUserLanguage());
+		$sLangCode = self::GetUserLanguage();
+		self::InitLangIfNeeded($sLangCode);
 
-		if (!array_key_exists(self::GetUserLanguage(), self::$m_aData))
+		if (!array_key_exists($sLangCode, self::$m_aData))
 		{
+			IssueLog::Error("Cannot find $sLangCode in dictionnaries. default labels displayed", LogChannels::APC);
```
piRGoif: We made a mistake, this isn't the log that needs to be changed. The correct one is in InitLangIfNeeded (see my comment below).

Now I am wondering what is the purpose of this log, as it may happen that we are loading a dict key before the local cache is populated ?

piRGoif (changes requested):

core/dict.class.inc.php
odain-cbd: this log can appear when there is an inconsistent dictionnary file. like declaring lables for french in a ru-ru dictionnary file path...

Molkobain:
All good for me, I'm more confortable with not changing the app behavior for a third-party failure that can be upgrade. The choosen solution (logging) seems nice 👍

I can't approve the PR as I created it for @odain-cbd but merge in support/2.7 is ok for me.

core/apc-service.class.inc.php
```diff
@@ -0,0 +1,41 @@
...
+	 * @see apc_store()
+	 */
+	function apc_store($key, $var = NULL, $ttl = 0)
+	{
+		return apc_store($key, $var, $ttl);
+	}
+}
```
Hipska: Fix file ending?

core/dict.class.inc.php
```diff
@@ -145,15 +147,17 @@ public static function S($sStringCode, $sDefault = null, $bUserLanguageOnly = fa
...
 		// Attempt to find the string in the user language
 		//
-		self::InitLangIfNeeded(self::GetUserLanguage());
+		$sLangCode = self::GetUserLanguage();
+		self::InitLangIfNeeded($sLangCode);
 
-		if (!array_key_exists(self::GetUserLanguage(), self::$m_aData))
+		if (!array_key_exists($sLangCode, self::$m_aData))
 		{
+			IssueLog::Warning("Cannot find $sLangCode in dictionnaries. default labels displayed");
```
Hipska, line 155: Isn't this going to give a lot of warnings in the log file? Maybe better to check in `InitLangIfNeeded`?

Hipska (changes requested):

core/dict.class.inc.php
Molkobain: Maybe we could dispatch it to the APC channel like for the error message.

Hipska: Could be done indeed, but even then it would still be a massive amount of messages. I would log it in the init method once when it cannot load the lang..

odain-cbd: 1) this log is generated when there is a broken dictionnary.  it is not linked to APC.

2) iTop will generate one log per label to display on a specific page. so yes, it could be huge!!

but normally when we package iTop we have a test (./integration/DictionariesConsistencyTest.php) that prevents from releasing broken dictionnaries (Romain's work). it does not happen anymore....

for this reasons we can leave as is....

core/dict.class.inc.php
piRGoif: Ok... 
This may generate lots of noise in the log... We'll have to monitor this ! After next iTop Combodo upgrade !

Hipska (commented):
@odain-cbd No I meant that there is no newline at the end of the file. I (and php) don't care about the `?>`.

core/apc-service.class.inc.php
Hipska: Not resolved!

core/dict.class.inc.php
Hipska: That's what I said as well.. https://github.com/Combodo/iTop/pull/224#discussion_r760868804

PR #225, opened 2021-08-05
N°5370 - Improve feedback when updating objects via REST without changes

larhip:
When updating an object via REST the feedback is always "updated", doesn't matter whether there is a real change.

It might be interesting to give a feedback via rest whether the object has been changed or not, just like it is in the UI.

This PR checks whether an object is modified and set message to "unchanged" if not.

### DIFF
--- a/core/restservices.class.inc.php
+++ b/core/restservices.class.inc.php
@@ -366,8 +366,13 @@ class CoreServices implements iRestServiceProvider
 			{
 				$oObject = RestUtils::FindObjectFromKey($sClass, $key);
 				RestUtils::UpdateObjectFromFields($oObject, $aFields);
-				$oObject->DBUpdate();
-				$oResult->AddObject(0, 'updated', $oObject, $aShowFields, $bExtendedOutput);
+				$sMessage = 'unchanged';
+				if($oObject->IsModified())
+				{
+					$oObject->DBUpdate();
+					$sMessage = 'updated';
+				}
+				$oResult->AddObject(0, $sMessage, $oObject, $aShowFields, $bExtendedOutput);
 			}
 			break;
 	


Hipska (approved):

Molkobain:
Hello Lars, I'll check this PR on friday, would you update/rebase you branch on develop in the meantime? Thanks

Molkobain:
Guillaume will created the corresponding bug in our internal bug tracker.

Remaining question for technical / functional teams: Should we add v1.4 of the API for this, if so in which iTop versions.

Molkobain:
Hello Lars,

The functional aspect was accepted during today's review.
We are still discussing about the technical implementation as it might be a breaking change for some people interpreting the result of the call in the v1.3 of the API.

I'm checking if there are other REST/API related bugs that we could pack to this future v1.4 version of the API.

Cheers,
Guillaume

Molkobain:
For now iTop doesn't handle multiple versions of it own API meaning that one cannot call API v1.2 on iTop 3.0 for example.
If we increase the version to 1.4 for this change in iTop 3.1, all existing webservices querying iTop with the v1.3 of the API will break after upgrading to iTop 3.1... This is quite ugly. 😕

We should think to a proper way to handle this better, do you have any proposition?

_PS: Except for a better handling of tag attributes, there were not many requests on the REST/API enhancements_

Hipska:
@Molkobain there is the `$sVersion` variable where it can do a `version_compare` to do the current behaviour or the new behaviour..

Molkobain:
Yes, what I meant is that I don't feel like adding "if / version_compare()" conditions everywhere in the `ExecOperation()` method each time we change something.
Even though there would be only one for now, it would eventually become a mess. That's why I would like to find a more elegant solution; do you know how other APIs manage it?

Hipska:
This might be useful to read: 
https://stackoverflow.com/questions/29871744/how-do-you-manage-the-underlying-codebase-for-a-versioned-api

PR #228, opened 2021-08-18
Update install.txt

lazki:
Change old URL for new, linking to iTop Hub.

### DIFF
--- a/install.txt
+++ b/install.txt
@@ -1,3 +1,3 @@
 For installation instructions, please refer to:
 
-https://wiki.openitop.org/doku.php?id=latest:install:start
+https://www.itophub.io/wiki/page?id=latest:install:start


Hipska:
Good catch, although the old URL still works and gives you a redirect to the new one.

Hipska (approved):

lazki:
The old URL still works, but... openitop.org will be phased out in the future, isn't it?

Molkobain:
Good catch indeed! Thanks Federico. ✌

Molkobain (approved):

Molkobain:
> The old URL still works, but... openitop.org will be phased out in the future, isn't it?

Yep, we don't know when exactly, but the sooner we fix the URL the better, that way we minimize the number of packages with an outdated URL out there.

Hipska:
@Molkobain be noted there are still a lot of links to openitop.org in readme.txt (Why is that file still there, since it has been replaced by README.md?)

Molkobain:
@Hipska indeed yes those would need to be fixed as well! (if nobody makes a PR, I'll change them later).

Regarding the readme.txt it is packaged in the distributed zip file. We don't put the markdown files in the package as people are more familiar with .txt than .md, but that might change in the future.

Hipska:
Actually, the readme.txt is already markdown format. So it might even be more clear for everyone if a file with the .md extension would be present in the package.

Molkobain:
Hu, you are right. Then I'm clueless. Will raise the point, thanks!

PR #229, opened 2021-08-20
N°5066 - Clean CMDBSource methods

Hipska:
This improves the function calls to use more relevant `QueryToScalar` instead of the `QueryToCol`.

This also deprecates `GetServerInfo` since the info is not always reliable. (Related to https://github.com/Combodo/itop-system-information/pull/5) (See https://www.php.net/manual/en/mysqli.get-server-info.php#118822)


core/cmdbsource.class.inc.php
```diff
@@ -350,14 +350,20 @@ public static function IsDB($sSource)
...
 	{
-		$aVersions = self::QueryToCol('SELECT Version() as version', 'version');
-		return $aVersions[0];
+		return static::QueryToScalar('SELECT VERSION()', 0);
 	}
 
 	/**
-	 * @return string
+	 * @deprecated Use `CMDBSource::GetDBVersion` instead.
+	 * @see mysqli_get_server_info
```
piRGoif:
```suggestion
	 * @uses mysqli_get_server_info
```

We prefer to write `@uses` in such case (see https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/uses.html#uses-used-by)

piRGoif (changes requested):
Thanks for those improvements, seems good to me ! 
Just a little remark about the phpdoc

core/cmdbsource.class.inc.php
Hipska: Thanks, didn't know that one.

piRGoif (dismissed):

piRGoif (approved):

piRGoif (changes requested):
After a team review, we found that the behavior has changed : before, if the result was 0 we were returning an empty string... Can you modify that ? 

I've checked exceptions, seems to me that we are getting the same with \CMDBSource::QueryToArray and \CMDBSource::QueryToScalar

Hipska:
Fixed the conflicts

piRGoif:
Hello,

> After a team review, we found that the behavior as changed : before, if the result was 0 we were returning an empty string... Can you modify that ?

@Hipska you asked for a new review, but this wasn't changed right ?

Hipska:
Ah, you mean for `GetServerVariable`? You want me to return this instead?

```php
return static::QueryToScalar($sSql, 0) ?: '';
```

piRGoif:
> Ah, you mean for `GetServerVariable`? You want me to return this instead?
> 
> ```
> return static::QueryToScalar($sSql, 0) ?: '';
> ```

Would be perfect, thanks !

Hipska:
Applied, please review.

piRGoif (approved):

piRGoif:
Sorry, we missed technical review today (too much PR to review). Will be reviewed next month !

Molkobain (approved):

Molkobain:

Molkobain:
Accepted in functional review, will be part of 3.1.0

PR #231, opened 2021-09-09
Properly init objects array

Hipska:
This caused a warning in data-collector-base/toolkit/dump_tasks.php when there are no DataSources yet.

### DIFF
--- a/core/restservices.class.inc.php
+++ b/core/restservices.class.inc.php
@@ -137,6 +137,12 @@ class RestResultWithObjects extends RestResult
 {
 	public $objects;
 
+	public function __construct()
+	{
+		parent::__construct();
+		$this->objects = array();
+	}
+
 	/**
 	 * Report the given object
 	 * 	 
--- a/core/restservices.class.inc.php
+++ b/core/restservices.class.inc.php
@@ -135,6 +135,7 @@ class ObjectResult
  */
 class RestResultWithObjects extends RestResult
 {
+	/** @var DBObject[] */
 	public $objects;
 
 	public function __construct()
@@ -191,6 +192,7 @@ class RestResultWithObjects extends RestResult
 
 class RestResultWithRelations extends RestResultWithObjects
 {
+	/** @var array */
 	public $relations;
 	
 	public function __construct()


piRGoif:
Hello Thomas, 
The code you added makes perfectly sense to me.
But can you detail how to reproduce the notice with the current code base ?

Hipska:
Run the toolkit dump_tasks.php against a iTop instance that has no `SynchroDataSource` objects. See also https://github.com/Combodo/itop-data-collector-base/pull/12 for a fix on that side.

The collector (and probably other clients as well) assumes the `objects` field is always an array. But it seems that when there are no fields, the field is just `null`, which throws a warning when you try to `count()` it.

Molkobain:

Molkobain:
@Hipska could you rebase the PR on the support/3.0 branch so it will be part of the 3.0.1 version please?

Thanks,
Guillaume

Hipska:
Is the proposed method good, or would you rather want to see `public $objects = array();`. Or even the new `[]` notation?

Molkobain:
Using short array synthax is ok yes. As for the proposed method, we don't have guidelines for this, so do as you please, current proposal is fine by me :)

piRGoif (approved):

Hipska:
@Molkobain Done, I left it like it was since that is similar as how it is done in `RestResultWithRelations`. I did add some phpdoc to help IDE's a bit better.

piRGoif (approved):

piRGoif:
After merge we had a notification about a failed test : RestTest#testDeleteApi
I'll have a look...

piRGoif:
Ok, the test is KO because this PR changed the value returned when no objects.
Before : `{"objects":null,"code":0,"message":"Found: 0"}`
After : `{"objects":[],"code":0,"message":"Found: 0"}`
We are discussing what to do in the dev team...

Hipska:
Oh, that test wasn't good in the first place then 😆

piRGoif:
Test updated, and documentation also : https://www.itophub.io/wiki/page?id=latest%3Aadvancedtopics%3Arest_json#response

piRGoif:
2.7.0 documentation : https://www.itophub.io/wiki/page?id=2_7_0%3Aadvancedtopics%3Arest_json#response

Molkobain:
Careful, this change can break existing usages of the REST/JSON API! Its must be discussed with the product team and put in the migration notes. Typically, depending on how "case exchange" integrations have been set up, some of their scenario will not work anymore.

Molkobain:
@piRGoif whats is the corresponding bug number in our internal bug tracker?

piRGoif:
PR updated to be reviewed on next meeting.
No ticket created yet as we're not sure to keep this modification as is.

Molkobain:
We need to create a ticket for anything that might change the behavior of the app. It's already merged, so the ticket is necessary to me, it would also allow a better follow up internally then the PR I think.

Molkobain:
Internal bug tracker: N°4671

Molkobain:
Discussion within the R&D team:
  * Too much consequences in all existing CEX integrations
  * The API should not change it behavior / response without a version change (1.3 => 1.4 for example)

So in conclusion, we will revert both the PR and the change on the unit test.
Fix should rather be in the collector base for now.

@piRGoif can you manage the revert please as you handled the PR / test / documentation? Thanks

Hipska:
After reading the updated wiki, I also came to the realisation that when there are objects, it is also not an (json) array, but an object. So in that case, `null` is maybe still the better option for when there are no objects in the result.

Would someone be able to have a look on https://github.com/Combodo/itop-data-collector-base/pull/12 then? This solves a problem when dumping from an instance without any SynchroDataSource defined.

Molkobain:
> Would someone be able to have a look on [Combodo/itop-data-collector-base#12](https://github.com/Combodo/itop-data-collector-base/pull/12) then? This solves a problem when dumping from an instance without any SynchroDataSource defined.

Yes, we will have a look on that PR, just asked a question there :)

piRGoif:
Yes I will tomorrown : revert code + revert phpunit test + fix 3_0_0 wiki

piRGoif:
Code was reverted (see attached commit)
And [3_0_0 wiki also](https://www.itophub.io/wiki/page?id=latest%3Aadvancedtopics%3Arest_json#response).

Indeed we will fix instead collector base, certainly with Thomas' other PR (Combodo/itop-data-collector-base#12)

PR #233, opened 2021-09-17
Portal ExceptionHandler : add concerned file and line in IssueLog

piRGoif:
When an exception occurs, a message is sent to IssueLog but containing only the message and no other info.
This adds concerned file and line, which will help greatly for debug.
Also, adding the portal channel (introduced in 2.7.5)


piRGoif:
wooops didn't create this PR with the correct base branch !
Closing and recreating it :/

PR #235, opened 2021-09-17
Portal ExceptionHandler : add concerned file and line in IssueLog

piRGoif:
When an exception occurs, a message is sent to IssueLog but containing only the message and no other info.
This adds concerned file and line, which will help greatly for debug.
Also, adding the portal channel (introduced in 2.7.5)

IssueLog before : 

```
2021-09-17 13:09:08 | Error   | Oups ! Une erreur est survenue.: Key "portals" for array with keys "properties, forms, bricks, bricks_total_width, lists" does not exist. | IssueLog
```

And after this PR : 

```
2021-09-17 13:47:48 | Error   | Oups ! Une erreur est survenue.: Key "portals" for array with keys "properties, forms, bricks, bricks_total_width, lists" does not exist. | portal
array (
  'filename' => '/path/to/faulty/template.html.twig',
  'line' => 222,
)
```

As we are using the standard Exception getters, and because IssueLog does simply a `var_export()` call on its aContext param, we should be protected if file and line are null.

### DIFF
--- a/datamodels/2.x/itop-portal-base/portal/src/EventListener/ExceptionListener.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/EventListener/ExceptionListener.php
@@ -24,6 +24,7 @@ namespace Combodo\iTop\Portal\EventListener;
 
 use Dict;
 use IssueLog;
+use LogChannels;
 use Symfony\Component\Debug\Exception\FlattenException;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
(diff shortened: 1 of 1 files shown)


piRGoif:
Discussed with @Molkobain : 

In fact we have all the info we need in the 2.7 branch : the file and line are displayed when using the querystring param `&debug=true`. For now the error message isn't, but it will be in 2.7.6 (known and fixed bug).
So indeed the urgency on the support/2.7 branch isn't that big...

And in the develop branch, @bruno-ds already pushed a new ExceptionLog class that is called in this exact scenario. I will review his code and give some advices like adding line number and file ;o)

In conclusion I'm closing this PR !

PR #236, opened 2021-09-17
N°4766 - DataSynchro: Allow binary data to be imported/synchronised with the synchro_import.php

Hipska:
Currently it is not possible to fill the binary (LONGBLOB) fields of the synchro_data tables by csv upload using synchro_import.php

When the provided data is base64 encoded, this becomes possible, but the tool needs to support it.

See the following example synchronising a Person object with picture using synchro_import.php:
SynchroReplica example

### DIFF
--- a/synchro/synchro_import.php
+++ b/synchro/synchro_import.php
@@ -435,6 +435,7 @@ try
 
 	$aIsDateToTransform = array();
 	$aDateToTransformReport = array();
+	$aIsBinaryToTransform = array();
 	foreach ($aInputColumns as $iFieldId => $sInputColumn)
 	{
 		if (array_key_exists($sInputColumn, $aDateColumns))
@@ -456,6 +457,7 @@ try
 		{
 			throw new ExchangeException("Unknown column '$sInputColumn' (class: '$sClass')");
 		}
+		$aIsBinaryToTransform[$iFieldId] = $aColumns[$sInputColumn] === 'LONGBLOB';
 	}
 	if (!isset($iPrimaryKeyCol))
 	{
@@ -535,6 +537,10 @@ try
 							$aValues[] = $sDate;
 						}
 					}
+					elseif ($aIsBinaryToTransform[$iCol])
+					{
+						$aValues[] = base64_decode($value);
+					}
 					else
 					{
 						$aValues[] = $value;
@@ -599,6 +605,10 @@ try
 						{
 							$aValuePairs[] = "`$sCol` = ".CMDBSource::Quote($sDate);
 						}
+					}
+					elseif ($aIsBinaryToTransform[$iCol])
+					{
+						$aValuePairs[] = "`$sCol` = FROM_BASE64(".CMDBSource::Quote($aRow[$iCol], true).")";
 					}
 					else
 					{


Hipska:
Data collectors can be adapted by overriding the `Collector::AddHeader` method, so it won't give a warning on `_data`, `_mimetype` and `_filename` fields.

piRGoif:
Updated PR as this is currently reviewed by Erwan

Molkobain:
@Molkobain will see with some devs for a technical review before GO.

Molkobain:
Technical review planned for 2022-01-25

synchro/synchro_import.php
```diff
@@ -456,6 +457,7 @@ function ChangeDateFormat($sProposedDate, $sFormat, $bDateOnly)
 		{
 			throw new ExchangeException("Unknown column '$sInputColumn' (class: '$sClass')");
 		}
+		$aIsBinaryToTransform[$iFieldId] = $aColumns[$sInputColumn] === 'LONGBLOB';
```
rquetiez, line 460: Two comments:

1. The identification of Blob inputs is based on the type of column in the database whereas date inputs are identified by the mean of the attribute type. For the sake of consistency, I would rather have one single way of identifying column types. Though I'm the author of the original implementation, I agree that the new proposed way is more relevant to what is intended: synchro_import aims at writing into a synchro table.
2. Do we have BLOBS (instead of LONGBLOBS) as well?

rquetiez: This improvement must be covered by a test. See /test/synchro/DataSynchroTest.php to add the case.

Hipska: You are right, do you want me to move the check to this part?
https://github.com/Combodo/iTop/blob/1495ab992f03a8bee3a553ac959a20552df3ee5b/synchro/synchro_import.php#L424-L434

rquetiez: Yes, doing it once for all (before proceeding with the processing of the whole input) makes sense.
My expectation is also that you will modify the preexisting conditions in order to check the SQL column type rather than checking the attribute type. This will make it more adaptable against the introduction of new attribute types.

Hipska: Only problem with your wanted approach is that there is no direct relation to attribute code and column name, since it becomes something like `attcode_data`. So I don't see how the clearness of the code could be improved with the way you are requesting..

rquetiez: Okay, I wanted to base the detection on the columns in the data table, BUT I had forgotten that the data table column types are TEXT for most attribute types and dates in particular (see AttributeDefinition::GetImportColumns), and I admit that using AttributeDefinition::GetSQLColumns is not relevant there.
Therefore, I now consider your implementation as okay.
To avoid the confusion I've just made, a comment at line 424 would be welcome : "Detection cannot be based on the data table column type here because dates attributes are imported as TEXT".

Molkobain: @Hipska we seem OK to merge the PR once the comment is added. Is that OK for you?

Hipska: I don't understand why I need to put that comment? That's a part of the code I didn't touch and has no relation to allowing binary data for import..

Molkobain: From what Romain said, it's to avoid having the same debate you just had in the future. To explain why the approach you choose is actually ok.

Hipska: No, he wants me to put that comment on line 424 where they check which attributes/columns/fields are a date type..

This PR has nothing to do with dates, only with blobs/binary data.

rquetiez: Yes, I would like you to comment the legacy code. This to prevent people like me to make wrong interpretations of the code. (Though I am the author of that stuff, I felt into the trap!)

Moreover, a test is a must too.

Molkobain (approved):

Molkobain:
Checking with Erwan tomorrow about his tests.
Then merge, add comment (myself).

Molkobain:
Hello Thomas, 

I was about to merge it, but actually it would be great to have this on the support/3.0 branch so it would be part of the upcoming 3.0.1. Would you mind rebasing the PR? If I do from the GH GUI, it will mess everything.

Thanks,
Guillaume

Hipska:
Done: #269 (no rebase needed, all done from GH UI 😉 )

Molkobain:
Thanks!

(I meant rebase locally on your computer then force push to avoid creating another PR, but that's fine)

Hipska:
No, that's not possible because you can't change the branch where you want to merge to on GitHub UI. 

(While thinking on it, it does work sometimes indeed. Not sure why)

piRGoif:
Yes indeed changing the brase branch in GitHub UI does weird stuff most of the time. It works sometimes indeed, but that's the exception :(

To rebase locally you can use something like : 

```
git rebase --onto support/3.0 develop mybranch
```

Hipska:
Rebase wasn't needed, since I forked before the split between master and support branch was made 😉

DarkNight97boss:
@Hipska 
Hello Thomas!
The profile picture data is correctly retrieved and visible in the synchro_replica table, but the image is not actually updated in the Person object. There is no transposition of the image fields from the synchronization table to the final object.

tomrss:
Hi @Hipska ! Me and @DarkNight97boss found the problem, we need to add the `picture_downloads_count` column as well in addition to the 3 columns stated above, due to this code:

https://github.com/Combodo/iTop/blob/4d88ef6ed418ac7db140f1273303df8a84e1f67e/synchro/synchrodatasource.class.inc.php#L2696-L2713

the `*_downloads_count` column is part of the `AttributeBlob` as well, so if it is missing, that code returns null as the attribute value (`picture` of `Person` in this case).

PR #237, opened 2021-09-23
N°4325 CMDBSource wrap both mysqli attributes

piRGoif:
In 2.7.5 with N°3513 we added a second mysqli attribute in CMDBSource, so that we can test transactions (see TransactionsTest).

But this wasn't documented, and was really causing confusion.

This refactor wraps both attributes in a dedicated object so that the logic is clearer.

### DIFF
--- a/core/DbConnectionWrapper.php
+++ b/core/DbConnectionWrapper.php
@@ -17,8 +17,10 @@ use mysqli;
  *
  * @used-by \CMDBSource
  *
- * @since 2.7.5 N°3513 new mockable {@see mysqli} object in {@see CMDBSource} //FIXME
- * @since 3.0.0 N°4325 add this object to avoid confusions and document the wanted behavior
+ * @since 3.0.0 N°4325 Object creation
+ *          This wrapper handles the 2 {@mysqli myqsli} attributes that were previously in {@see CMDBSource}
+ *          To allow testing we added a second mysqli object (N°3513 in 2.7.5) and code became a bit confusing :/
+ *          With this wrapper everything is in the same place, and we can express the intention more clearly !
  */
 class DbConnectionWrapper
 {
--- a/core/DbConnectionWrapper.php
+++ b/core/DbConnectionWrapper.php
@@ -26,27 +26,40 @@ class DbConnectionWrapper
 	protected static $oDbCnxStandard;
 
 	/**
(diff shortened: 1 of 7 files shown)


piRGoif:
Tests are showing "Fatal error: Uncaught Error: Class name must be a valid object or a string" errors :/
Will continue work on this...

piRGoif:
Build is back to normal, PR is ready to review \o/

core/DbConnectionWrapper.php
```diff
@@ -0,0 +1,52 @@
...
+
+	public static function GetDbConnection(bool $bIsMockable = false): ?mysqli
+	{
+		if ($bIsMockable) {
+			return self::$oDbCnxMockable;
+		}
+
+		return self::$oDbCnxStandard;
+	}
+
```
odain-cbd, line 51: SetDbConnection is ok like this. but code would be simpler to read if there was two distinct setters. one for mockable object and this one.

odain-cbd (commented):
PR changes are fine for me. except the composer ones that may be explained by another version of composer?

core/DbConnectionWrapper.php
piRGoif: Yes, good idea, will do !

piRGoif:
> 
> 
> PR changes are fine for me. except the composer ones that may be explained by another version of composer?

Thanks for the review !
About the autoloader, this is under discussion in the dev team and will be updated aside

core/DbConnectionWrapper.php
piRGoif: Done in e4e7c8fd

piRGoif:
Forced push to resolve conflicts on the autoloader
Also new commit to add code reviews feedbacks

odain-cbd (dismissed):

eespie (dismissed):

Molkobain:
As discussed with @piRGoif , I'm not confortable with having to manually pass the `$bIsForQuery` parameter when consuming the API (as well as propagating it to other consumers, eg. `CMDBSource::LogDeadlock()`. Ideally we would rework it a bit so that the wrapper automatically returns the right object depending on the called method, but as this should only be used internally we consider it good enough for now.

Molkobain (approved):

piRGoif:
Merging !

PR #238, opened 2021-09-30
Try removing config file writing in OQL PHPUnit tests

piRGoif:


piRGoif:
utils::GetConfig refactoring was done in 7f156e96 : my guess is that those tests were done before...

PR #239, opened 2021-09-30
N°4261 Refactor ExceptionLog

piRGoif:
N°4261 was a way to : 

* Log more info on user portal exceptions (this is done in \Combodo\iTop\Portal\EventListener\ExceptionListener::onKernelException) : before we only logged the exception message using IssueLog
* Allow to log as EventIssue objects so that we can query those. Default level to log as EventIssue objects must be ERROR
* Configuration is done by class hierarchy. For example, defining Trace level for CoreException hierarchy (this class and all of its children)
* When logging, the channel used is the one defined in the configuration. The exception class is logged but in aContext data.

Doing a code review with Bruno, we agreed to do some little refactoring : 

* Level per exception class
  - Before the whole ExceptionLog::Log method was a total rewrite of its parent, with some code duplicates... not a good idea : we should better improve LogAPI to make other similar uses possible in the future !
  - The logic to get level from config must be in a GetMinLogLevel override
* Write to DB
  - Pull up this functionnality in LogAPI
  - Add a sCode parameter in GetLevelDefault

Doing this refactoring, I also improved : 

* Test the attributes set when creating the EventIssue object : during my dev I had crashes because I didn't filled all the mandatory fields... Having a PHPUnit test checking this will prevent future bugs to happen if attributes are modified in the class or in the object creation method
* Use Throwable instead of Exception : this was added in PHP 7.0 and will allow to catch both Exception and Error
* Because we need to have 2 statements on the same line in \Combodo\iTop\Test\UnitTest\Core\Log\ExceptionLogTest::testLogInFile, I modified the editorConfig file to allow disabling the formatter using comments.

### DIFF
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -586,6 +586,8 @@ abstract class LogAPI
 	 */
 	protected static $m_oMockMetaModelConfig = null;
 
+	protected static $oLastEventIssue = null;
+
 	public static function Enable($sTargetFile)
 	{
 		// m_oFileLog is not defined as a class attribute so that each impl will have its own
@@ -707,29 +709,34 @@ abstract class LogAPI
 	 */
 	protected static function GetMinLogLevel($sChannel, $sCode = self::ENUM_CONFIG_PARAM_FILE)
 	{
-		$sConfiguredLevelForChannel = static::GetMinLogLevelFromChannel($sChannel, $sCode);
+		$sLogLevelMin = static::GetLogConfig($sCode);
+
+		$sConfiguredLevelForChannel = static::GetMinLogLevelFromChannel($sLogLevelMin, $sChannel);
 		if (!is_null($sConfiguredLevelForChannel)) {
 			return $sConfiguredLevelForChannel;
 		}
 
-		return static::GetMinLogLevelFromDefault($sChannel, $sCode);
+		return static::GetMinLogLevelFromDefault($sLogLevelMin, $sChannel);
 	}
 
-	/**
-	 * @param string $sChannel
-	 * @param string $sCode
-	 *
-	 * @return string|null null if not defined
-	 */
-	protected static function GetMinLogLevelFromChannel($sChannel, $sCode = self::ENUM_CONFIG_PARAM_FILE)
+	final protected static function GetLogConfig($sCode = self::ENUM_CONFIG_PARAM_FILE)
 	{
 		$oConfig = static::GetConfig();
 		if (!$oConfig instanceof Config) {
 			return static::GetLevelDefault();
 		}
 
-		$sLogLevelMin = $oConfig->Get($sCode);
+		return $oConfig->Get($sCode);
+	}
 
+	/**
+	 * @param string|array $sLogLevelMin log config parameter value
+	 * @param string $sChannel
+	 *
+	 * @return string|null null if not defined
+	 */
+	protected static function GetMinLogLevelFromChannel($sLogLevelMin, $sChannel)
+	{
 		if (empty($sLogLevelMin)) {
 			return static::GetLevelDefault();
 		}
@@ -745,7 +752,7 @@ abstract class LogAPI
 		return null;
 	}
 
-	protected static function GetMinLogLevelFromDefault($sChannel, $sCode = self::ENUM_CONFIG_PARAM_FILE)
+	protected static function GetMinLogLevelFromDefault($sLogLevelMin, $sChannel)
 	{
 		if (isset($sLogLevelMin[static::CHANNEL_DEFAULT])) {
 			return $sLogLevelMin[static::CHANNEL_DEFAULT];
@@ -1208,8 +1215,6 @@ class ExceptionLog extends LogAPI
 	public const CHANNEL_DEFAULT = 'Exception';
 	public const CONTEXT_EXCEPTION = '__exception';
 
-	private static $oLastEventIssue = null;
-
 	protected static $m_oFileLog = null;
 
 	/**
@@ -1258,22 +1263,24 @@ class ExceptionLog extends LogAPI
 	 */
 	protected static function GetMinLogLevel($sExceptionClass, $sCode = self::ENUM_CONFIG_PARAM_FILE)
 	{
-		$sConfiguredLevelForChannel = null;
+		$sLogLevelMin = static::GetLogConfig($sCode);
+
+		$sConfiguredLevelForExceptionClass = null;
 		$sExceptionClassInHierarchy = $sExceptionClass;
 		while ($sExceptionClassInHierarchy !== false) {
-			$sConfiguredLevelForChannel = static::GetMinLogLevelFromChannel($sExceptionClassInHierarchy, $sCode);
-			if (!is_null($sConfiguredLevelForChannel)) {
+			$sConfiguredLevelForExceptionClass = static::GetMinLogLevelFromChannel($sLogLevelMin, $sExceptionClassInHierarchy);
+			if (!is_null($sConfiguredLevelForExceptionClass)) {
 				break;
 			}
 
 			$sExceptionClassInHierarchy = get_parent_class($sExceptionClassInHierarchy);
 		}
 
-		if (!is_null($sConfiguredLevelForChannel)) {
-			return $sConfiguredLevelForChannel;
+		if (!is_null($sConfiguredLevelForExceptionClass)) {
+			return $sConfiguredLevelForExceptionClass;
 		}
 
-		return static::GetMinLogLevelFromDefault($sExceptionClass, $sCode);
+		return static::GetMinLogLevelFromDefault($sLogLevelMin, $sExceptionClass);
 	}
 
 	protected static function GetEventIssue(string $sMessage, string $sChannel, array $aContext): EventIssue
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -804,12 +804,11 @@ abstract class LogAPI
 		}
 
 		// Protect against reentrance
-		static $aWriteToDbReentrance = array();
-		$sKey = $sChannel.$sMessage;
-		if (array_key_exists($sKey, $aWriteToDbReentrance)) {
+		static $bWriteToDbReentrance;
+		if ($bWriteToDbReentrance === true) {
 			return;
 		}
-		$aWriteToDbReentrance[$sKey] = true;
+		$bWriteToDbReentrance = true;
 
 		try {
 			self::$oLastEventIssue = static::GetEventIssue($sMessage, $sChannel, $aContext);
@@ -823,7 +822,7 @@ abstract class LogAPI
 			]);
 		}
 		finally {
-			unset($aWriteToDbReentrance[$sKey]);
+			$bWriteToDbReentrance = false;
 		}
 	}
 
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -648,7 +648,7 @@ abstract class LogAPI
 		}
 
 		if (
-			(false === is_null(static::$m_oFileLog))
+			(null !== static::$m_oFileLog)
 			&& static::IsLogLevelEnabled($sLevel, $sChannel, static::ENUM_CONFIG_PARAM_FILE)
 		) {
 			static::$m_oFileLog->$sLevel($sMessage, $sChannel, $aContext);
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -562,12 +562,19 @@ abstract class LogAPI
 	public const LEVEL_DEBUG = 'Debug';
 	public const LEVEL_TRACE = 'Trace';
 	/**
-	 * @var string default log level
-	 * @see GetMinLogLevel
+	 * @see     GetMinLogLevel
 	 * @used-by GetLevelDefault
+	 * @var string default log level
 	 * @since 2.7.1 N°2977
 	 */
 	public const LEVEL_DEFAULT = self::LEVEL_OK;
+	/**
+	 * @see     GetMinLogLevel
+	 * @used-by GetLevelDefault
+	 * @var string default log level when writing to DB
+	 * @since 3.0.0 N°4261
+	 */
+	public const LEVEL_DEFAULT_DB = self::LEVEL_ERROR;
 
 	protected static $aLevelsPriority = array(
 		self::LEVEL_ERROR   => 400,
@@ -664,9 +671,9 @@ abstract class LogAPI
 	 * @throws \ConfigException if log wrongly configured
 	 * @uses GetMinLogLevel
 	 */
-	final public static function IsLogLevelEnabled(string $sLevel, string $sChannel, string $sCode = self::ENUM_CONFIG_PARAM_FILE): bool
+	final public static function IsLogLevelEnabled(string $sLevel, string $sChannel, string $sLogConfigKey = self::ENUM_CONFIG_PARAM_FILE): bool
 	{
-		$sMinLogLevel = self::GetMinLogLevel($sChannel, $sCode);
+		$sMinLogLevel = self::GetMinLogLevel($sChannel, $sLogConfigKey);
 
 		if ((is_bool($sMinLogLevel) && ($sMinLogLevel === false)) || $sMinLogLevel === 'false') {
 			return false;
@@ -686,7 +693,7 @@ abstract class LogAPI
 
 	/**
 	 * @param string $sChannel
-	 * @param string $sCode
+	 * @param string $sLogConfigKey
 	 *
 	 * @return string one of the LEVEL_* const value : the one configured it if exists, otherwise default log level for this channel
 	 *       Config can be set :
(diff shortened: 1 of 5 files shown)


core/log.class.inc.php
```diff
@@ -645,22 +645,26 @@ public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = arr
 			$sChannel = static::CHANNEL_DEFAULT;
 		}
 
-		if (!static::IsLogLevelEnabled($sLevel, $sChannel)) {
-			return;
+		if (
+			(false === is_null(static::$m_oFileLog))
```
bruno-ds: Since you're using a strict comparison, I propose to use `null !== static::$m_oFileLog` 
Or if you prefer to employ `is_null`, since it always return a bool: `! is_null(static::$m_oFileLog)` 

PS: This is not an important topic, but hey, I am a little obsessed...

core/log.class.inc.php
```diff
@@ -645,22 +645,26 @@ public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = arr
...
-
-		static::$m_oFileLog->$sLevel($sMessage, $sChannel, $aContext);
 	}
 
 	/**
 	 * @throws \ConfigException if log wrongly configured
 	 * @uses GetMinLogLevel
 	 */
-	final public static function IsLogLevelEnabled(string $sLevel, string $sChannel, string $sCode = 'log_level_min'): bool
+	final public static function IsLogLevelEnabled(string $sLevel, string $sChannel, string $sCode = self::ENUM_CONFIG_PARAM_FILE): bool
```
bruno-ds: This is off topic, bu I think taht the `final`  keyword should be removed, since it made the initial implementation of ExceptionLog really cumbersome !

core/log.class.inc.php
```diff
@@ -719,18 +740,63 @@ protected static function GetMinLogLevel($sChannel, $sCode = 'log_level_min')
 			return $sLogLevelMin[$sChannel];
 		}
 
+		return null;
+	}
+
+	protected static function GetMinLogLevelFromDefault($sChannel, $sCode = self::ENUM_CONFIG_PARAM_FILE)
+	{
 		if (isset($sLogLevelMin[static::CHANNEL_DEFAULT])) {
```
bruno-ds, line 785: `$sLogLevelMin` isn't initialised, or did I missed something?

core/log.class.inc.php
```diff
@@ -719,18 +740,63 @@ protected static function GetMinLogLevel($sChannel, $sCode = 'log_level_min')
 			return $sLogLevelMin[$sChannel];
 		}
 
+		return null;
+	}
+
+	protected static function GetMinLogLevelFromDefault($sChannel, $sCode = self::ENUM_CONFIG_PARAM_FILE)
```
bruno-ds: I don't really understand why you did broke the previous code into two separate functions since both have many branches that lead to "default" return values.

core/log.class.inc.php
```diff
@@ -1169,110 +1235,72 @@ public static function LogException(Exception $oException, $aContext = array(),
...
-		}
-
-		if (!isset(self::$aLevelsPriority[$sLevel])) {
-			IssueLog::Error("invalid log level '{$sLevel}'");
-
-			return;
-		}
-
+		$sConfiguredLevelForChannel = null;
+		$sExceptionClassInHierarchy = $sExceptionClass;
```
bruno-ds, line 1414: I thought that you were going to move this logic into `IssueLog ` did I misunderstood your intentions?

core/log.class.inc.php
piRGoif: I like your first proposition !
About the second, we avoid it in the dev team as it's easy to miss the "!", whereas "false ===" is clear and can't be missed.

core/log.class.inc.php
piRGoif: Wooops, I'll have to take a look...

piRGoif:
Rebased branch

core/log.class.inc.php
piRGoif: Done in 31ae5f3a

core/log.class.inc.php
piRGoif: I added the final keyword from the very start, when I added the method in f6d9d0f
As this method is an utility one that is calling really \LogAPI::GetMinLogLevel, I thought there shouldn't be any necessity to override it.
Do you have any use case in mind ?

piRGoif: Discussed with Bruno, I'm closing this discussion.

bruno-ds: We debated this with @piRGoif , my suggestion was wrong, sorry :(

core/log.class.inc.php
piRGoif: Should be fixed by now, I'w solving this conversation

piRGoif:
Tests for log are back to OK now \o/
But... TransactionsTest throws an error, I need to solve this.

Also I'm going to write some info in the description to explain the idea behind this PR !

piRGoif:
2 more commits this morning : 
* use Throwable instead of Exception in \ExceptionLog::LogException signature
* test that \LogAPI::GetEventIssue really sets all the mandatory fields in the object

The PR description was updated accordingly !

core/log.class.inc.php
```diff
@@ -645,22 +654,27 @@ public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = arr
 			$sChannel = static::CHANNEL_DEFAULT;
 		}
 
-		if (!static::IsLogLevelEnabled($sLevel, $sChannel)) {
-			return;
+		if (
+			(null !== static::$m_oFileLog)
+			&& static::IsLogLevelEnabled($sLevel, $sChannel, static::ENUM_CONFIG_PARAM_FILE)
```
bruno-ds, line 669: about `static::ENUM_CONFIG_PARAM_FILE`: For sure, using constants provide robustness, and it's not an error to use them, but personally I'm really not fond of them for a readability POV.

For instance, an alternative would have been to implement  `IsLogLevelEnabledForFiles` and `IsLogLevelEnabledForDb` (the names are lame, but I like the idea).

core/log.class.inc.php
```diff
@@ -645,22 +654,27 @@ public static function Log($sLevel, $sMessage, $sChannel = null, $aContext = arr
...
 	 * @uses GetMinLogLevel
 	 */
-	final public static function IsLogLevelEnabled(string $sLevel, string $sChannel, string $sCode = 'log_level_min'): bool
+	final public static function IsLogLevelEnabled(string $sLevel, string $sChannel, string $sLogConfigKey = self::ENUM_CONFIG_PARAM_FILE): bool
 	{
-		$sMinLogLevel = self::GetMinLogLevel($sChannel, $sCode);
+		$sMinLogLevel = self::GetMinLogLevel($sChannel, $sLogConfigKey);
 
-		if ($sMinLogLevel === false || $sMinLogLevel === 'false') {
+		// the is_bool call is to remove a IDE O:) warning as $sMinLogLevel is typed as string
```
bruno-ds, line 697: I'm not even sure the `$sMinLogLevel === 'false'` is required, maybe I coded it preventatively (because I don't know well how iTop enforce type correctness in the config)

core/log.class.inc.php
```diff
@@ -719,16 +756,65 @@ protected static function GetMinLogLevel($sChannel, $sCode = 'log_level_min')
...
+
+		try {
+			self::$oLastEventIssue = static::GetEventIssue($sMessage, $sChannel, $aContext);
+			self::$oLastEventIssue->DBInsertNoReload();
+		}
+		catch (Exception $e) {
+			// calling low level methods : if we would call Error() for example we would try to write to DB again...
+			static::$m_oFileLog->Error('Failed to log issue into the DB', LogChannels::CORE, [
+				'exception message' => $e->getMessage(),
+				'exception stack'   => $e->getTraceAsString(),
```
bruno-ds, line 821: Writing the call stack is a hard choice to make, I think this should be debated, maybe with molkobain or eric?
(I think it's a good idea, but fear an stratospheric pollution of the logs in case of certain circumstances such as read_only mode)

core/log.class.inc.php
```diff
@@ -719,16 +756,65 @@ protected static function GetMinLogLevel($sChannel, $sCode = 'log_level_min')
...
+	{
+		if (false === MetaModel::IsLogEnabledIssue()) {
+			return;
+		}
+		if (false === MetaModel::IsValidClass('EventIssue')) {
+			return;
+		}
+
+		try {
+			self::$oLastEventIssue = static::GetEventIssue($sMessage, $sChannel, $aContext);
```
bruno-ds, line 814: extracting the instantiation really improve the readability, congrats!

core/log.class.inc.php
```diff
@@ -719,16 +756,65 @@ protected static function GetMinLogLevel($sChannel, $sCode = 'log_level_min')
...
+	{
+		$sDate = date('Y-m-d H:i:s');
+		$aStack = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 5);
+		$sCurrentCallStack = var_export($aStack, true);
+
+		$oEventIssue = new EventIssue();
+		$oEventIssue->Set('issue', $sMessage);
+		$oEventIssue->Set('message', $sMessage);
+		$oEventIssue->Set('date', $sDate);
+		$oEventIssue->Set('userinfo', UserRights::GetUserFriendlyName());
```
bruno-ds: This may result in additional queries to the database, I'm a little afraid by the impact it mayu have for corner cases.

core/log.class.inc.php
```diff
@@ -744,15 +830,26 @@ protected static function GetConfig(): ?Config
...
 	 * @uses    \LogAPI::LEVEL_DEFAULT
+	 * @uses    \LogAPI::LEVEL_DEFAULT_DB
 	 *
 	 * @since 3.0.0 N°3731
+	 * @since 3.0.0 N°4261 add specific default level for DB write
 	 */
-	protected static function GetLevelDefault(): string
+	protected static function GetLevelDefault(string $sLogConfigKey): string
 	{
+		if ($sLogConfigKey === self::ENUM_CONFIG_PARAM_DB) {
```
bruno-ds: while totally producing the same result, maybe a switch would improve the readability, or may just a 2d if `if ($sLogConfigKey === self::ENUM_CONFIG_PARAM_FILE) {` .

(I'm not 100% percent sure it would be better, I just think that the current code do not explcitely expose the 3 handled cases with juste one if)

core/log.class.inc.php
```diff
@@ -965,8 +1063,12 @@ public static function DeprecatedNoticesErrorHandler(int $errno, string $errstr,
 		return true;
 	}
 
-	protected static function GetLevelDefault(): string
+	protected static function GetLevelDefault(string $sLogConfigKey): string
 	{
+		if ($sLogConfigKey === self::ENUM_CONFIG_PARAM_DB) {
```
bruno-ds: I'm surprised that `utils::IsDevelopmentEnvironment()` has impact only for the file backend and not the db backend.

core/log.class.inc.php
```diff
@@ -1133,147 +1235,171 @@ private function GetLogFileNameBuilderClassName()
...
 
 	/**
 	 * This method should be used to write logs.
 	 *
 	 * As it encapsulate the operations performed using the Exception, you should prefer it to the standard API inherited from LogApi `ExceptionLog::Error($oException->getMessage(), get_class($oException), ['__exception' => $oException]);`
-	 * The parameter order is not standard, but in our use case, the resulting API is way more convenient this way.
+	 * The parameter order is not standard, but in our use case, the resulting API is way more convenient this way !
 	 */
-	public static function LogException(Exception $oException, $aContext = array(), $sLevel = self::LEVEL_WARNING)
+	public static function LogException(Throwable $oException, $aContext = array(), $sLevel = self::LEVEL_WARNING): void
```
bruno-ds, line 1307: Thanks for the change for `Throwable` type hint!

core/log.class.inc.php
```diff
@@ -1133,147 +1235,171 @@ private function GetLogFileNameBuilderClassName()
...
 	{
+		if (!isset(self::$aLevelsPriority[$sLevel])) {
+			IssueLog::Error("invalid log level '{$sLevel}'");
+
+			return;
+		}
+
+		$sExceptionClass = get_class($oException);
+
 		if (empty($aContext[self::CONTEXT_EXCEPTION])) {
```
bruno-ds, line 1317: I am the faulty one on this one again!
=> using array_merge() would treat all 4 array keys at once, and, I think, improve overall readability. 

ie: 
```
$aDefaultContext = [
    self::CONTEXT_EXCEPTION => $oException,
    'exception class' => $sExceptionClass, 
    'file' => $oException->getFile(),
    'line' => $oException->getLine(),
]
$aContext = array_merge($aDefaultContext, $aContext);
```

core/log.class.inc.php
```diff
@@ -1133,147 +1235,171 @@ private function GetLogFileNameBuilderClassName()
...
+	 * 3. OQLException
+	 * 4. CoreException
+	 * 5. Exception
+	 *
+	 * @param string $sExceptionClass
+	 * @param string $sLogConfigKey
+	 *
+	 * @return string|null the current or parent class name defined in the config, otherwise null if no class of the hierarchy found in the config
+	 */
+	protected static function GetExceptionClassInConfig($sExceptionClass, $sLogConfigKey = self::ENUM_CONFIG_PARAM_FILE)
```
bruno-ds: 1/ I'm not fond of the method  name as I find it hard to understand.
Maybe replace `Get` by `Find` as the result is computed, but also the class hierarchy traversal is not really guessable just by the method name.

2/ Maybe this method would benefit from being moved into it's parent, as the use case may be implemented by other classes than ExceptionLog?

bruno-ds (commented):
I found it really hard to read it with PHPstorm's PR review UI, maybe we should read the code together and exchange oraly.

core/log.class.inc.php
piRGoif: And also improves extensibility :) Both were the goals to reach here O:)

piRGoif:
We did a face to face review with @bruno-ds :)

Here are the conclusions / TODO : 

- [x] LogAPITest : there was a line removed in a dataprovider, why ? 
- LogAPI 
  - [x] rename $sLogConfigKey parameters to $sConfigKey
  - [x] WriteToDb : add a static attribute to avoid re-entrance (similar to what is done in \DBObject::DBUpdate)
  - [x] GetEventIssue : \UserRights::GetUserFriendlyName is generating queries to the DB, we should find another better way to print the current user info (Session::Get('auth_user'); that is set in \LoginWebPage::OnLoginSuccess or \UserRights::$m_oUser maybe ?)
  - [x] GetLevelDefault : to improve readability use switch or if instead
- [x] \DeprecatedCallsLog::GetLevelDefault : add a PHPDoc to explain the behavior
- Exceptionlog
  - [ ] maybe pull up in LogAPI the search in the class hierarchy ? (this may be used when logging inline image as we're using the DM class as channel...)
  - [ ] the exception is stored in aContext and this is bad :/ We avoided an extra param but this isn't nice. Maybe change the parameter to have 2 root keys : data (current content) and metadata (the exception instance we need to keep) ?
  - [x] in the unit test use a Throwable / Error
  - [x] Log() : instead in LogAPI extract write operations in a dedicated method and override only this one
  - [ ] ExceptionClassFromHierarchy : to improve readability extract to a dedicated method

core/log.class.inc.php
piRGoif: Reviewed with Bruno directly, closing this conversation

core/log.class.inc.php
piRGoif: Reviewed with Bruno directly, closing this conversation

core/log.class.inc.php
piRGoif: Reviewed with Bruno directly, closing this conversation

core/log.class.inc.php
piRGoif: Reviewed with Bruno directly, closing this conversation

core/log.class.inc.php
piRGoif: Reviewed with Bruno directly, closing this conversation

core/log.class.inc.php
piRGoif: Reviewed with Bruno directly, closing this conversation

core/log.class.inc.php
piRGoif: Reviewed with Bruno directly, closing this conversation

piRGoif:
Removal of a dataprovider line : in \Combodo\iTop\Test\UnitTest\Core\Log\LogAPITest::LogApiProvider
To show it, use the git history, open diff between current version and the one in 3abcd59b0369480faf8323f528d4e4503475d989 :

image

This line was causing the test to fail, because the current test impl uses mock for the config and so we had : 
`TypeError : Return value of LogAPI::GetConfig() must be an instance of Config or null, array returned`

I replaced the `array()` by null : OK.
Committing this ! 357254f2

core/log.class.inc.php
```diff
@@ -782,6 +782,14 @@ protected static function WriteToDb(string $sMessage, string $sChannel, array $a
 			return;
 		}
 
+		// Protect against reentrance
+		static $aWriteToDbReentrance = array();
```
bruno-ds: I think that a boolean would have been sufficient as it would have mean
> while you're writing a log to DB, then the code inside this write cannot write a log to db

your implementation mean:
> while you're writing a log to DB, then you are forbidden to write the exact same message to db

piRGoif: Indeed you're right !

piRGoif: Fixed in 4cc77608

PR #240, opened 2021-10-14
Restore prototypes of cmdbAbstractObject methods in which we added $sMode to preserve compatibility with existing extensions

Molkobain:
During 3.0.0 development I added an $sMode parameter to some of the cmdbAbstractObject class' methods.
It broke the compatibility with some extensions that were overloading the methods in their DM classes.

When discussed a few months ago, we decided to put that "display mode" in a class property instead, this PR aims at that.

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -101,6 +101,12 @@ abstract class cmdbAbstractObject extends CMDBObject implements iDisplay
 	 * @since 3.0.0
 	 */
 	public const ENUM_DISPLAY_MODE_PRINT = 'print';
+	/**
+	 * @var string
+	 * @see static::$sDisplayMode
+	 * @since 3.0.0
+	 */
+	public const ENUM_DISPLAY_MODE_BULK_EDIT = self::ENUM_DISPLAY_MODE_EDIT;
 
 	// N°3750 rendering used
 	/** @var string */
@@ -201,6 +207,8 @@ abstract class cmdbAbstractObject extends CMDBObject implements iDisplay
 			static::ENUM_DISPLAY_MODE_EDIT,
 			static::ENUM_DISPLAY_MODE_CREATE,
 			static::ENUM_DISPLAY_MODE_STIMULUS,
+			static::ENUM_DISPLAY_MODE_PRINT,
+			static::ENUM_DISPLAY_MODE_BULK_EDIT,
 		];
 	}
 
@@ -4971,7 +4979,7 @@ HTML
 			// Now create an object that has values for the homogeneous values only
 			/** @var \cmdbAbstractObject $oDummyObj */
 			$oDummyObj = new $sClass(); // @@ What if the class is abstract ?
-			$oDummyObj->SetDisplayMode(static::ENUM_DISPLAY_MODE_EDIT);
+			$oDummyObj->SetDisplayMode(static::ENUM_DISPLAY_MODE_BULK_EDIT);
 			$aComments = array();
 			function MyComparison($a, $b) // Sort descending
 			{
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -2766,7 +2766,7 @@ JS
 		$this->SetDisplayMode(($iKey > 0) ? static::ENUM_DISPLAY_MODE_EDIT : static::ENUM_DISPLAY_MODE_CREATE);
 		$sDisplayMode = $this->GetDisplayMode();
 
-		if ($sDisplayMode === static::ENUM_DISPLAY_MODE_EDIT)
+		if ($this->GetDisplayMode() === static::ENUM_DISPLAY_MODE_EDIT)
 		{
 			// The concurrent access lock makes sense only for already existing objects
 			$LockEnabled = MetaModel::GetConfig()->Get('concurrent_lock_enabled');
@@ -2816,7 +2816,7 @@ JS
 		if (isset($aExtraParams['custom_button'])) {
 			$sApplyButton = $aExtraParams['custom_button'];
 		} else {
-			if ($sDisplayMode === static::ENUM_DISPLAY_MODE_EDIT) {
+			if ($this->GetDisplayMode() === static::ENUM_DISPLAY_MODE_EDIT) {
 				$sApplyButton = Dict::S('UI:Button:Apply');
 			} else {
 				$sApplyButton = Dict::S('UI:Button:Create');
@@ -2826,7 +2826,7 @@ JS
 		if (isset($aExtraParams['custom_operation'])) {
 			$sOperation = $aExtraParams['custom_operation'];
 		} else {
-			if ($sDisplayMode === static::ENUM_DISPLAY_MODE_EDIT) {
+			if ($this->GetDisplayMode() === static::ENUM_DISPLAY_MODE_EDIT) {
 				$sOperation = 'apply_modify';
 			} else {
 				$sOperation = 'apply_new';
@@ -2841,7 +2841,7 @@ JS
 			->SetOnSubmitJsCode("return OnSubmit('form_{$this->m_iFormId}');");
 		$oContentBlock->AddSubBlock($oForm);
 
-		if ($sDisplayMode === static::ENUM_DISPLAY_MODE_EDIT) {
+		if ($this->GetDisplayMode() === static::ENUM_DISPLAY_MODE_EDIT) {
 			// The object already exists in the database, it's a modification
 			$oForm->AddSubBlock(InputUIBlockFactory::MakeForHidden('id', $iKey, "{$sPrefix}_id"));
 		}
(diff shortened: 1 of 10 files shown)


application/cmdbabstract.class.inc.php
```diff
@@ -4861,6 +4920,7 @@ public static function DisplayBulkModifyForm($oP, $sClass, $aSelectedObj, $sCust
 			// Now create an object that has values for the homogeneous values only
 			/** @var \cmdbAbstractObject $oDummyObj */
 			$oDummyObj = new $sClass(); // @@ What if the class is abstract ?
+			$oDummyObj->SetDisplayMode(static::ENUM_OBJECT_MODE_EDIT);
```
rquetiez: This line could be removed: as this instance is not created for display purposes, why not leave the default behavior (which gives currently the same value) ?

application/cmdbabstract.class.inc.php
```diff
@@ -174,6 +180,30 @@ public static function EnumObjectModes(): array
 		];
 	}
 
+	/**
+	 * @see static::$sDisplayMode
+	 * @return string
+	 * @since 3.0.0
+	 */
+	public function GetDisplayMode(): string
```
rquetiez, line 220: The name is not consistent with EnumObjectModes() and ENUM_OBJECT_MODE...
I would prefer SetObjectMode, or even align everything with the word "Display" which is more significant.

Molkobain: Good point, I'll check if the constants are already used in migrated extensions.

application/cmdbabstract.class.inc.php
Molkobain: We are calling `$oDummyObj->DisplayModifyForm()` later at line 5063, so we need the mode to be set as it will be used there.

application/cmdbabstract.class.inc.php
Molkobain: No usage (yet) in our extensions.
@xtophe38 as used them in TeemIP, I'll let him know when the rename is done.

application/cmdbabstract.class.inc.php
```diff
@@ -282,8 +312,7 @@ public static function SetSessionMessage($sClass, $iKey, $sMessageId, $sMessage,
 	 * To insert something IN the panel, we now need to add UIBlocks in either the "subtitle" or "toolbar" sections of the array that will be returned.
 	 *
 	 * @param \WebPage $oPage
-	 * @param bool $bEditMode
-	 * @param string $sMode Mode in which the object is displayed (see static::ENUM_OBJECT_MODE_XXX)
+	 * @param bool $bEditMode Note that this parameter is no longer used in this method, {@see static::$sDisplayMode} is used instead
```
piRGoif:
```suggestion
	 * @param bool $bEditMode Note that this parameter is no longer used in this method. Use {@see static::$sDisplayMode} instead !
```

application/cmdbabstract.class.inc.php
```diff
@@ -293,10 +322,10 @@ public static function SetSessionMessage($sClass, $iKey, $sMessageId, $sMessage,
 	 * @throws \CoreUnexpectedValue
 	 * @throws \MySQLException
 	 * @throws \OQLException
-	 * @todo 3.0.0: This has to be discussed within the R&D when I come back in 10 days
 	 *
+	 * @since 3.0.0 $bEditMode is deprecated and no longer used
```
piRGoif, line 358:
```suggestion
	 * @since 3.0.0 $bEditMode is deprecated and no longer used. Use {@see static::$sDisplayMode} instead !
```

application/cmdbabstract.class.inc.php
```diff
@@ -476,12 +505,14 @@ public function DisplayBareHistory(WebPage $oPage, $bEditMode = false, $iLimitCo
...
 	 * @param \WebPage $oPage
-	 * @param bool $bEditMode
+	 * @param bool $bEditMode Note that this parameter is no longer used in this method, {@see static::$sDisplayMode} is used instead
 	 * @param string $sPrefix
 	 * @param array $aExtraParams
 	 *
 	 * @return array
 	 * @throws \CoreException
+	 *
+	 * @since 3.0.0 $bEditMode is deprecated and no longer used
```
piRGoif, line 547:
```suggestion
	 * @since 3.0.0 $bEditMode is deprecated and no longer used. Use {@see static::$sDisplayMode} instead !
```

application/cmdbabstract.class.inc.php
```diff
@@ -848,6 +881,8 @@ public function GetRelatedTriggersIDs(): array
 	 * @throws \MySQLException
 	 * @throws \OQLException
 	 * @throws \Exception
+	 *
+	 * @since 3.0.0 $bEditMode is deprecated and no longer used
```
piRGoif, line 927:
```suggestion
	 * @since 3.0.0 $bEditMode is deprecated and no longer used. Use {@see static::$sDisplayMode} instead !
```

application/cmdbabstract.class.inc.php
```diff
@@ -1041,16 +1075,18 @@ public function GetBareProperties(WebPage $oPage, $bEditMode, $sPrefix, $aExtraP
 	 * @throws \MySQLException
 	 * @throws \MySQLHasGoneAwayException
 	 * @throws \OQLException
+	 *
+	 * @since 3.0.0 $bEditMode is deprecated and no longer used
```
piRGoif, line 1121:
```suggestion
	 * @since 3.0.0 $bEditMode is deprecated and no longer used. Use {@see static::$sDisplayMode} instead !
```

application/cmdbabstract.class.inc.php
```diff
@@ -1068,14 +1104,14 @@ public function DisplayDetails(WebPage $oPage, $bEditMode = false, $sMode = self
 		}
 
 		// Object's details
-		$oObjectDetails = ObjectFactory::MakeDetails($this);
+		$oObjectDetails = ObjectFactory::MakeDetails($this, $this->GetDisplayMode());
```
piRGoif, line 1149: The called method was modified but outside of this PR ?

application/cmdbabstract.class.inc.php
```diff
@@ -2665,10 +2720,11 @@ public function DisplayModifyForm(WebPage $oPage, $aExtraParams = array())
 		$sOwnershipToken = null;
 		$iKey = $this->GetKey();
 		$sClass = get_class($this);
-		$sMode = ($iKey > 0) ? static::ENUM_OBJECT_MODE_EDIT : static::ENUM_OBJECT_MODE_CREATE;
 
+		$this->SetDisplayMode(($iKey > 0) ? static::ENUM_OBJECT_MODE_EDIT : static::ENUM_OBJECT_MODE_CREATE);
```
piRGoif: I know this feels a bit harsh, but feels more readable to me ? 

```suggestion
		if ($iKey > 0) {
				$this->SetDisplayMode(static::ENUM_OBJECT_MODE_EDIT);
		} else {
				 $this->SetDisplayMode(static::ENUM_OBJECT_MODE_CREATE);
		}
```

application/cmdbabstract.class.inc.php
```diff
@@ -2665,10 +2720,11 @@ public function DisplayModifyForm(WebPage $oPage, $aExtraParams = array())
...
+		$sDisplayMode = $this->GetDisplayMode();
```
piRGoif, line 2767: Why not call `GetDisplayMode()` everywhere, instead of using a variable ?

application/cmdbabstract.class.inc.php
piRGoif: Ok, there was another call L2910 that was already using the second parameter. This one was certainly just forgotten on the first modification ?

application/cmdbabstract.class.inc.php
```diff
@@ -3145,9 +3202,11 @@ public static function DisplayCreationForm(WebPage $oPage, $sClass, $oSourceObje
 	 */
 	public function DisplayStimulusForm(WebPage $oPage, $sStimulus, $aPrefillFormParam = null, $bDisplayBareProperties = true)
 	{
+		$this->SetDisplayMode(static::ENUM_OBJECT_MODE_STIMULUS);
+
 		$sClass = get_class($this);
 		$iKey = $this->GetKey();
-		$sMode = static::ENUM_OBJECT_MODE_STIMULUS;
+		$sDisplayMode = $this->GetDisplayMode();
```
piRGoif, line 3268: I would move this variable declaration just above where it is used, so around L3363

application/cmdbabstract.class.inc.php
piRGoif: Event if they are doing the same thing, as the UI is quite different between object edition and bulk edit, shouldn't we create a ENUM_OBJECT_MODE_BULK_EDIT constant ?

piRGoif (dismissed):
Just some little comments, but seems OK for me :)

application/cmdbabstract.class.inc.php
Molkobain: Exactly yes

application/cmdbabstract.class.inc.php
Molkobain: I thought it would be more efficient than to make multiple calls / easier to read, but maybe it is too much?

piRGoif: When using a variable, it's always a pain to know where it is defined and if the value wasn't overrided...
About calling multiple times the method, it's just a getter to an attribute so the performance difference would be minimal I think ?

Molkobain:
Sorry, I had to force push due to a wrong operation on my side. :/

application/cmdbabstract.class.inc.php
Molkobain: Done

application/cmdbabstract.class.inc.php
Molkobain: Yep, done

PR #245, opened 2021-11-18
N°4498 - Introduce new APIs due to iPageUIExtension deprecation

Molkobain:
Since we deprecated the \iPageUIExtension API in iTop 3.0.0, there is no simple way to add JS/CSS files/snippets to a backoffice page anymore. The only way is to use the \iPageUIBlockExtension but it requires to make a UIBlock which is not intuitive when you only want to include some files as a UIBlock is designed to be a specific UI element.

Unlike the previous interface, we decided to make more atomic one to improve evolution through the upcoming versions and avoid the necessity to implement all methods when you only need a few.

### DIFF
--- a/application/applicationextension.inc.php
+++ b/application/applicationextension.inc.php
@@ -1278,7 +1278,7 @@ interface iBackofficeLinkedScriptsExtension
  * @api
  * @since 3.0.0
  */
-interface iBackofficeEarlyScriptExtension
+interface iBackofficeEarlyScriptsExtension
 {
 	/**
 	 * @see \iTopWebPage::$a_early_scripts
--- a/application/applicationextension.inc.php
+++ b/application/applicationextension.inc.php
@@ -1101,7 +1101,9 @@ class JSButtonItem extends JSPopupMenuItem
  * @api
  * @package     Extensibility
  * @since 2.0
- * @deprecated since 3.0.0 use iPageUIBlockExtension instead
+ * @deprecated 3.0.0
+ *   * If you need to include JS/CSS files/snippets use {@see \iBackofficeLinkedScriptsExtension}, {@see \iBackofficeLinkedStylesheetsExtension}, etc instead
+ *   * If you need to include HTML (and optionally JS/CSS) use {@see \iPageUIBlockExtension} to manipulate {@see \Combodo\iTop\Application\UI\Base\UIBlock} instead
  */
 interface iPageUIExtension
 {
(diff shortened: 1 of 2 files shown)


application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+
+/**
+ * Implement this interface to add inline script (JS) to the backoffice pages' head.
+ * Will be executed first, BEFORE the DOM interpretation.
+ *
+ * @see \iTopWebPage::$a_early_scripts
+ * @api
+ * @since 3.0.0
+ */
+interface iBackofficeEarlyScriptExtension
```
steffunky, line 1281: `Script` is singular in this interface name but but plural in others

application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+{
+	/**
+	 * @see \iTopWebPage::$a_early_scripts
+	 * @return string
+	 */
+	public function GetEarlyScript(): string;
```
steffunky, line 1287: `Script` is singular in this method name but but plural in others

application/applicationextension.inc.php
Molkobain: Good catch, indeed it should be plural as the interface could contain several methods. I'll change it.

application/applicationextension.inc.php
Molkobain: In other interface the method name is single, only the interface name is plural.
This method only return 1 script, so I think it stay single, what do you think?

application/applicationextension.inc.php
```diff
@@ -1101,7 +1101,9 @@ class JSButtonItem extends JSPopupMenuItem
  * @api
  * @package     Extensibility
  * @since 2.0
- * @deprecated since 3.0.0 use iPageUIBlockExtension instead
+ * @deprecated 3.0.0
+ *   * If you need to include JS/CSS files/snippets use {@see \iBackofficeLinkedScriptsExtension}, {@see \iBackofficeLinkedStylesheetsExtension}, etc instead
+ *   * If you need to include HTML (and optionally JS/CSS) use {@see \iPageUIBlockExtension} to manipulate {@see \Combodo\iTop\Application\UI\Base\UIBlock} instead
```
piRGoif:
```suggestion
 * @deprecated 3.0.0 If you need to include  :
 *   * JS/CSS files/snippets : use {@see \iBackofficeLinkedScriptsExtension}, {@see \iBackofficeLinkedStylesheetsExtension}, etc instead
 *   * HTML (and optionally JS/CSS) : use {@see \iPageUIBlockExtension} to manipulate {@see \Combodo\iTop\Application\UI\Base\UIBlock} instead
```

application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+ *
+ * @see \iTopWebPage::$a_linked_scripts
+ * @api
+ * @since 3.0.0
+ */
+interface iBackofficeLinkedScriptsExtension
+{
+	/**
+	 * @see \iTopWebPage::$a_linked_scripts
+	 * @return array An array of absolute URLs to the files to include
```
piRGoif, line 1268:
```suggestion
	 * @return array An array containing for each script to include its absolute URL. Deduplication will be done by iTop itself
```

application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+ * Implement this interface to add script (JS) files to the backoffice pages
+ *
+ * @see \iTopWebPage::$a_linked_scripts
+ * @api
+ * @since 3.0.0
+ */
+interface iBackofficeLinkedScriptsExtension
+{
+	/**
+	 * @see \iTopWebPage::$a_linked_scripts
```
piRGoif:
```suggestion
	 * @see \iTopWebPage::$a_linked_scripts Each script will be included using this method
```

application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+
+/**
+ * Implement this interface to add inline script (JS) to the backoffice pages' head.
+ * Will be executed first, BEFORE the DOM interpretation.
+ *
+ * @see \iTopWebPage::$a_early_scripts
+ * @api
+ * @since 3.0.0
+ */
+interface iBackofficeEarlyScriptsExtension
```
piRGoif: I'm not very fan of this naming... 
Maybe something like HeadBefore ? This will speak to any web developer ?
But the core iTop methods are already called like this, so you may prefer to stick with this naming to avoid confusion ?
On my opinion the naming for iTop extensions API should be the easiest to understand for anyone : and this public isn't the one that will use core methods so having 2 distinct names wouldn't be that bad ?

piRGoif (dismissed):
Thanks for the proposition !

I made some comments / suggestions on little details.

But on the principle, Why adding so much interfaces with only one method each ? Maybe we could group all JS methods, and all CSS methods, and add for each an abstract class ? Stating interfaces aren't part of the API but the abstract classes are ?

Also, most of the comments are on the interface PHPDoc. I would better add those comments to the `@return` tag on each method's phpdoc ?

application/applicationextension.inc.php
steffunky: You're right!

application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+}
+
+/**
+ * Implement this interface to add inline script (JS) to the backoffice pages that will be executed slightly AFTER the DOM is ready (just after the init. scripts).
+ *
+ * @see \iTopWebPage::$a_ready_scripts
+ * @api
+ * @since 3.0.0
+ */
+interface iBackofficeReadyScriptsExtension
```
rquetiez: Singular would be more relevant : iBackofficeReadyScriptExtension

application/applicationextension.inc.php
rquetiez: Singular would be more relevant : iBackofficeEarlyScriptExtension

application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+}
+
+/**
+ * Implement this interface to add inline script (JS) to the backoffice pages that will be executed immediately, without waiting for the DOM to be ready.
+ *
+ * @see \iTopWebPage::$a_scripts
+ * @api
+ * @since 3.0.0
+ */
+interface iBackofficeScriptsExtension
```
rquetiez: Singular...

application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+}
+
+/**
+ * Implement this interface to add inline script (JS) to the backoffice pages that will be executed right when the DOM is ready.
+ *
+ * @see \iTopWebPage::$a_init_scripts
+ * @api
+ * @since 3.0.0
+ */
+interface iBackofficeInitScriptsExtension
```
rquetiez: Singular...

application/applicationextension.inc.php
```diff
@@ -1252,6 +1254,119 @@ public function GetFooterBlock()
...
+}
+
+/**
+ * Implement this interface to add inline style (CSS) to the backoffice pages' head.
+ *
+ * @see \iTopWebPage::$a_styles
+ * @api
+ * @since 3.0.0
+ */
+interface iBackofficeStylesExtension
```
rquetiez: Singular...

application/applicationextension.inc.php
Molkobain: Done

application/applicationextension.inc.php
Molkobain: Done

application/applicationextension.inc.php
Molkobain: I rather have the info about removing duplicates only in the property documentation, it avoids maintaining 2x documentation and minimize the risk that they diverge over time.

application/applicationextension.inc.php
Molkobain: Changed to singular as indeed this will return / include only 1 (might it be big / have several sections) script.

application/applicationextension.inc.php
Molkobain: Changed to singular as indeed this will return / include only 1 (might it be big / have several sections) script.

application/applicationextension.inc.php
Molkobain: Changed to singular as indeed this will return / include only 1 (might it be big / have several sections) script.

application/applicationextension.inc.php
Molkobain: Changed to singular as indeed this will return / include only 1 (might it be big / have several sections) script.

application/applicationextension.inc.php
Molkobain: Changed to singular as indeed this will return / include only 1 (might it be big / have several sections) style.

PR #247, opened 2021-12-03
:bug: N°4665 - Fix notice in logs when uploading an SVG image in an AttributeImage

annProg:
Try to fix notice log when upload svg image to `AttributeImage` field.

version: 2.7.5-1
this is the log
```
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP Notice:  Trying to access array offset on value of type bool in /var/www/itop/application/cmdbabstract.class.inc.php on line 3802
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP Stack trace:
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP   1. {main}() /var/www/itop/pages/UI.php:0
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP   2. Server->UpdateObjectFromPostedForm($sFormPrefix = *uninitialized*, $aAttList = *uninitialized*, $aAttFlags = *uninitialized*) /var/www/itop/pages/UI.php:970
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP   3. Server->PrepareValueFromPostedForm($sFormPrefix = '', $sAttCode = 'logo', $sClass = *uninitialized*, $aPostedData = *uninitialized*) /var/www/itop/application/cmdbabstract.class.inc.php:3732
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP Notice:  Trying to access array offset on value of type bool in /var/www/itop/application/cmdbabstract.class.inc.php on line 3803
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP Stack trace:
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP   1. {main}() /var/www/itop/pages/UI.php:0
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP   2. Server->UpdateObjectFromPostedForm($sFormPrefix = *uninitialized*, $aAttList = *uninitialized*, $aAttFlags = *uninitialized*) /var/www/itop/pages/UI.php:970
[03-Dec-2021 10:15:24 Asia/Shanghai] PHP   3. Server->PrepareValueFromPostedForm($sFormPrefix = '', $sAttCode = 'logo', $sClass = *uninitialized*, $aPostedData = *uninitialized*) /var/www/itop/application/cmdbabstract.class.inc.php:3732
```

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -4222,7 +4222,7 @@ HTML;
 				if (!is_null($oImage->GetData()))
 				{
 					$aSize = utils::GetImageSize($oImage->GetData());
-					if (is_array($aSize) && count($aSize) > 1)
+					if (is_array($aSize) && $aSize[0] > 0 && $aSize[1] > 0)
 					{
 						$oImage = utils::ResizeImageToFit(
 							$oImage,
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -4219,16 +4219,19 @@ HTML;
 			case 'Image':
 				$value = null;
 				$oImage = utils::ReadPostedDocument("attr_{$sFormPrefix}{$sAttCode}", 'fcontents');
-				if (!is_null($oImage->GetData()) && $oImage->GetMimeType() != "image/svg+xml")
+				if (!is_null($oImage->GetData()))
 				{
 					$aSize = utils::GetImageSize($oImage->GetData());
-					$oImage = utils::ResizeImageToFit(
-						$oImage,
-						$aSize[0],
-						$aSize[1],
-						$oAttDef->Get('storage_max_width'),
-						$oAttDef->Get('storage_max_height')
-					);
+					if (is_array($aSize))
+					{
+						$oImage = utils::ResizeImageToFit(
+							$oImage,
+							$aSize[0],
+							$aSize[1],
+							$oAttDef->Get('storage_max_width'),
+							$oAttDef->Get('storage_max_height')
+						);
+					}
 				}
 				$aOtherData = utils::ReadPostedParam("attr_{$sFormPrefix}{$sAttCode}", null, 'raw_data');
 				if (is_array($aOtherData))
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -4219,7 +4219,7 @@ HTML;
 			case 'Image':
 				$value = null;
 				$oImage = utils::ReadPostedDocument("attr_{$sFormPrefix}{$sAttCode}", 'fcontents');
-				if (!is_null($oImage->GetData()))
+				if (!is_null($oImage->GetData()) && $oImage->GetMimeType() != "image/svg+xml")
 				{
 					$aSize = utils::GetImageSize($oImage->GetData());
 					$oImage = utils::ResizeImageToFit(
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -4234,7 +4234,7 @@ HTML;
 					}
 					else
 					{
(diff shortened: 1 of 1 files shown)


piRGoif (changes requested):
Good idea, thanks !
I would change the test though : do not call `ResizeImageToFit()` if `$aSize === false`
This would be more generic

annProg:
@piRGoif how about `if (is_array($aSize))` then call `ResizeImageToFit()`

Hipska (dismissed):
Looks good, what about adding an `IssueLog::Warning('Image could not be resized ...` when not an array?

piRGoif:
Yes, good suggestion for is_array... You may also add a count() > 2 ?
And I agree with Hipska, adding a log if resize can't be done would make perfect sense... if the image isn't SVG (again...). That might add too much complexity (adding a ne method to test if the image is a vector format or a bitmap) ?

Hipska:
As per [php documentation](https://www.php.net/manual/en/function.getimagesize.php), the function will always return the first 2 indexes when it is an array, so the count check is not needed as extra. Although, an extra check if those values are zero might need to be done!

application/cmdbabstract.class.inc.php
```diff
@@ -4222,13 +4222,20 @@ protected function PrepareValueFromPostedForm($sFormPrefix, $sAttCode, $sClass =
...
+							$oImage,
+							$aSize[0],
+							$aSize[1],
+							$oAttDef->Get('storage_max_width'),
+							$oAttDef->Get('storage_max_height')
+						);
+					}
+					else
+					{
+						IssueLog::Warning($sClass . ":" . $this->GetKey() . '/' . $sAttCode . ': Image could not be resized. Mimetype: ' . $oImage->GetMimeType());
```
piRGoif: Nice log, thanks ! You might also add the file name ? \ormDocument::GetFileName

application/cmdbabstract.class.inc.php
```diff
@@ -4222,13 +4222,20 @@ protected function PrepareValueFromPostedForm($sFormPrefix, $sAttCode, $sClass =
...
 				{
 					$aSize = utils::GetImageSize($oImage->GetData());
-					$oImage = utils::ResizeImageToFit(
-						$oImage,
-						$aSize[0],
-						$aSize[1],
-						$oAttDef->Get('storage_max_width'),
-						$oAttDef->Get('storage_max_height')
-					);
+					if (is_array($aSize) && count($aSize) > 1)
```
piRGoif: @Hipska pointed out that the documentation stated that if you have an array no extra check is needed : https://github.com/Combodo/iTop/pull/247#issuecomment-987946211
But hey, we could leave the test anyway O:)

About testing the values to be > 0, to my opinion this is utils::ResizeImageToFit() job to do it ?

piRGoif (changes requested):
Many thanks ! A little change again, sorry

application/cmdbabstract.class.inc.php
Hipska: This check is indeed fixed @piRGoif, see latest version.

The `utils::ResizeImageToFit()` already checks if the size is already smaller than the max allowed, but having width and height as zero is a specific behaviour of the `utils::GetImageSize()`, so IMHO it is checked at the right place.

application/cmdbabstract.class.inc.php
piRGoif: That was done later on : closing the conversation

application/cmdbabstract.class.inc.php
piRGoif: Well, I disagree but indeed this is done and it won't do no any harm :o)
Closing this conversation

piRGoif (approved):

piRGoif:
Ok for me !
Will ask other developers to validate this PR and we will be good to merge.

Hipska (approved):

Molkobain:

Molkobain:
@annProg could you rebase on support/3.0 branch so it will be part of the 3.0.1 release?

Thanks,
Guillaume

piRGoif:
The rebase was done in #257

PR #248, opened 2021-12-06
N°4894 - Improve AttributeDecimal validation during CSV import

Hipska:
The RegEx could be simplified, and adding the `CheckFormat` method improved validation when doing for example CSV import.

Current situation: Bulk import validation does not give warning on invalid decimal notation. Proceeding the import results in possible data corruption (if using `,` instead of `.` for example, the decimal value is cut off) and if php/webserver is configured to show php warnings, user will get an ugly debug screen instead of a nice, descriptive error message:
image


core/attributedef.class.inc.php
```diff
@@ -2843,9 +2843,20 @@ public function GetValidationPattern()
 	{
 		$iNbDigits = $this->Get('digits');
 		$iPrecision = $this->Get('decimals');
-		$iNbIntegerDigits = $iNbDigits - $iPrecision - 1; // -1 because the first digit is treated separately in the pattern below
+		$iNbIntegerDigits = $iNbDigits - $iPrecision;
 
-		return "^[-+]?[0-9]\d{0,$iNbIntegerDigits}(\.\d{0,$iPrecision})?$";
+		return "^[-+]?\d{1,$iNbIntegerDigits}(\.\d{0,$iPrecision})?$";
```
Molkobain: Hello Thomas,

With this changed regexp, am I correct that decimals starting with a point are not allowed anymore?
For example, you cannot put `.25`, you have to put `0.25`.

Hipska: This was the same before as well, because of the `[0-9]` part.

Meaning, first optional character could be + or - and then 1 character in the range of 0-9 and then between 0 and x-1 decimals. My RegEx simplifies that to first optional +/- and then between 1 and x decimals.

Molkobain:
Technical review:
  * @Molkobain will discuss with Alexandre / Vincent about possible consequences throughout the app.
  * @Molkobain will create a modernization ticket to use the AttributeDefinition::GetValidationPattern() globally in the base AttributeDefinition::CheckFormat() method so all attribute types will be covered.

Molkobain:
> Technical review:
> 
> * @Molkobain will discuss with Alexandre / Vincent about possible consequences throughout the app.
> * @Molkobain will create a modernization ticket to use the AttributeDefinition::GetValidationPattern() globally in the base AttributeDefinition::CheckFormat() method so all attribute types will be covered.

Referenced under N°4894

Molkobain:
@Hipska do you mind if we rename the PR "Improve AttributeDecimal validation during CSV import" as the main impact of it is not the RegExp change but the change in `CheckFormat()`

Hipska:
Ok

Molkobain:
Planned for 3.2:
  * Introduces a behavior change in the CSV import as it will be blocking (hence not in 3.0.x)
  * Along with the fix in the PR, a global pass is planned to fix it for other attributes as well
  * 3.2 is supposed to have several changes / fixes around the import / synchro

core/attributedef.class.inc.php
```diff
@@ -2843,9 +2843,20 @@ public function GetValidationPattern()
...
+	}
+
+	/**
+	 * @param $value
+	 *
+	 * @return bool|int
+	 */
+	public function CheckFormat($value)
```
```suggestion
	/**
	 * @inheritDoc
	 * @since 3.2.0
	 */
	public function CheckFormat($value)
```

Molkobain (dismissed):

Molkobain (approved):

Molkobain:
Thanks Thomas.
Merge will be done when internal ticket will be processed.

Hipska:
When will internal ticket be processed?

Molkobain:
It is planned for iTop 3.2 (see milestone section) so not before '23 spring.

Hipska:
@Molkobain can you do a new review on this PR and also the line https://github.com/Combodo/iTop/blob/5d20b76476561e8345065ffcb30184159df2074e/core/attributedef.class.inc.php#L3092

I came to another problem by a customer and I don't think the `assert` should be there. Let me know then I can remove that line in this PR..

Molkobain:
What symptom do you have in your customer's environment and how can we reproduce?
I don't know why the assert was introduced in the first place, I'll ask Denis if he remembers.

Hipska:
Some decimal fields are auto-calculated and because of that have more digits after comma. The assert seems not relevant as the digits are cut anyway in the next lines..

Molkobain:
_This PR is not forgotten, it will be handled with iTop 3.2 starts (in the coming weeks / months)_

core/attributedef.class.inc.php
```diff
@@ -3246,7 +3256,7 @@ public function ScalarToSQL($value)
```
Hipska, line 3256: This assert should be removed. Agree or not?

Hipska:
@Molkobain 3.2 is started.. (Also even future develop)

Molkobain:
Indeed, but we (R&D) are still waiting for the functional scope to be defined for 3.2 and to see if there will be some global rework around the synchro feature. Depending on that, I'll discuss with the Product team how to merge the PR.

Hipska:
@Molkobain is there a functional scope for 3.2 in the meantime?

Molkobain:
CSV import won't be in the functional scope, so the global fix for all types of attributes won't be done in this version. That being said, as this PR already fix 1 type of attribute, we merge it anyway as it is still 1 attribute that is better handled.

Hipska:
Many thanks @Molkobain, but what about https://github.com/Combodo/iTop/pull/248#issuecomment-1468429624?

Molkobain:
Can you give a use case or a reproduction procedure where this would fail? I don't know why the assert was put in the first place, but what it checks seems ok doesn't it?

Hipska:
You already asked that before, and I already responded that:

> Some decimal fields are auto-calculated and because of that have more digits after comma. The assert seems not relevant as the digits are cut anyway in the next lines..

Molkobain:
You are right indeed, I didn't remember.
This could be reviewed in another PR I think, it might be ok to remove the assertion. If you make a PR, please attach an XML delta so it can be reproduced easily, thanks.

PR #250, opened 2021-12-07
:globe_with_meridians: Person Name for chinese user

annProg:
In Chinese it is customary to put the surname before the given name


piRGoif:
Hello,
Thanks for this contribution !

The Class/Name synthax wasn't documented, this is now fixed thanks to @Molkobain and @v-dumas (not published yet but it will be either with 3.0.0 RC or 2.7.6)

We will see with the product team if they are OK with your change.

Hipska (approved):

Molkobain (approved):
From my PoV, dict entries of the friendlyname are here for that exact reason, so we don't have to alter the datamodel to change the attributes order. So nice proposition 👍

Molkobain:
Merged, thanks!

PR #253, opened 2021-12-31
N°4756 Feature/faf event service

eespie:
FAF #8 Event Service


Hipska:
Hi @eespie what is the goal of "event service" ?

eespie:
Hi @Hipska, the problem is well described in PR 121 (https://github.com/Combodo/iTop/pull/121) but the solution proposed here is different. This PR provides an implementation of https://www.php-fig.org/psr/psr-14/ integrated into the datamodel, with additional filtering.

PR #255, opened 2022-01-05
:bulb: Document versions constants

piRGoif:
To clarify things, add phpdoc to know when to use which constant between `ITOP_DESIGN_LATEST_VERSION` and `ITOP_VERSION`

### DIFF
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -25,7 +25,7 @@ define('ITOP_APPLICATION_SHORT', 'iTop');
 
 /**
  * Constant containing the application version
- * Warning : this might be different from iTop core version !!
+ * Warning: this might be different from iTop core version!
  *
  * @see     ITOP_DESIGN_LATEST_VERSION for iTop core version
  *
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -24,7 +24,7 @@ define('ITOP_APPLICATION', 'iTop');
 define('ITOP_APPLICATION_SHORT', 'iTop');
 
 /**
- * Constant containing the package version
+ * Constant containing the application version
  * Warning : this might be different from iTop core version !!
  *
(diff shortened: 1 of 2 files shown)


core/config.class.inc.php
```diff
@@ -22,7 +22,17 @@
 
 define('ITOP_APPLICATION', 'iTop');
 define('ITOP_APPLICATION_SHORT', 'iTop');
-define('ITOP_VERSION', '2.7.0-dev'); // @see utils::GetItopVersionShort() and utils::GetItopVersionWikiSyntax()
+
+/**
+ * Constant containing the package version
```
```suggestion
 * Constant containing the application version
```

As we refer to "application" in the constants above.

core/config.class.inc.php
```diff
@@ -22,7 +22,17 @@
...
+ * Warning : this might be different from iTop core version !!
```
```suggestion
 * Warning: this might be different from iTop core version!
```

setup/itopdesignformat.class.inc.php
```diff
@@ -17,7 +17,14 @@
  * You should have received a copy of the GNU Affero General Public License
  */
 
-define('ITOP_DESIGN_LATEST_VERSION', '1.7'); // iTop >= 2.7.0
+
+/**
+ * iTop Datamodel XML format version
+ * And also iTop core version
```
Molkobain, line 23:
```suggestion
 * And also iTop core version (since 2020-10-16)
```

Should we add this to remember or not?

core/config.class.inc.php
piRGoif: Woops, sorry, thanks... I will get it one day :/

setup/itopdesignformat.class.inc.php
piRGoif: Not sure... Besides it's not that clear to me when reading the meeting minutes which decision was made ? 
I've added a subject for the dev team meeting, we may add this information afterwards ?

PR #258, opened 2022-01-15
:bug: Resolve JQuery error when service description is not empty

alally:
The addition of $(item.tooltip).text() and $(model.tooltip).text() in the service rendering twigs resulted in a jquery JS error when text was present in the description of a service. This caused the rendering of the service catalog in the portal to fail.


alally:
Sorry, it looks like I made the pull request to the wrong branch.

Hipska:
I think you could just change the brand where you want to push to here. But fine, you already created another PR.

Molkobain:
No problem :)

piRGoif:
Was replaced by #259

PR #260, opened 2022-01-18
N°4760 TwigBase : add possibility to control BreadCrumb

piRGoif:
Asked by @xtophe38 for TeemIP
For now when using TwigBase you can't set a breadcrumb at all. It is set to the page title, with a wrench icon.

After this PR each operation inside a Controller impl will be able to :

* disable the breadcrumb by calling `DisableBreadCrumb()`
* set its own custom breadcrumb entry by calling `SetBreadCrumbEntry()`

### DIFF
--- a/sources/application/TwigBase/Controller/Controller.php
+++ b/sources/application/TwigBase/Controller/Controller.php
@@ -572,13 +572,7 @@ abstract class Controller
 			$aBreadCrumbEntry = $this->GetOperationBreadCrumbEntry();
 			if ((false === empty($aBreadCrumbEntry)) && (is_array($aBreadCrumbEntry)) && (count($aBreadCrumbEntry) >= 3)) {
 				@list($sId, $sTitle, $sDescription, $sUrl, $sIcon) = $aBreadCrumbEntry;
-				if (false === isset($sUrl)) {
-					$sUrl = '';
-				}
-				if (false === isset($sIcon)) {
-					$sIcon = '';
-				}
-				$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl, $sIcon);
+				$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl ?: '', $sIcon ?: '');
 			}
 		} else {
 			$this->m_oPage->DisableBreadCrumb();
--- a/sources/application/TwigBase/Controller/Controller.php
+++ b/sources/application/TwigBase/Controller/Controller.php
@@ -571,9 +571,7 @@ abstract class Controller
 		if ($this->IsOperationBreadCrumbEnabled()) {
 			$aBreadCrumbEntry = $this->GetOperationBreadCrumbEntry();
 			if ((false === empty($aBreadCrumbEntry)) && (is_array($aBreadCrumbEntry)) && (count($aBreadCrumbEntry) >= 3)) {
-				list($sId, $sTitle, $sDescription) = $aBreadCrumbEntry;
-				$sUrl = (count($aBreadCrumbEntry) >= 4) ? $aBreadCrumbEntry[3] : '';
-				$sIcon = (count($aBreadCrumbEntry) >= 5) ? $aBreadCrumbEntry[4] : '';
+				@list($sId, $sTitle, $sDescription, $sUrl, $sIcon) = $aBreadCrumbEntry;
 				$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl, $sIcon);
 			}
 		} else {
--- a/sources/application/TwigBase/Controller/Controller.php
+++ b/sources/application/TwigBase/Controller/Controller.php
@@ -581,8 +581,8 @@ abstract class Controller
 
 				if ($this->m_bIsBreadCrumbEnabled) {
 					if (count($this->m_aBreadCrumbEntry) > 0) {
-						@list($sId, $sTitle, $sDescription, $sUrl, $sIcon) = $this->m_aBreadCrumbEntry;
-						$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl ?: '', $sIcon ?: '');
+						list($sId, $sTitle, $sDescription, $sUrl, $sIcon) = $this->m_aBreadCrumbEntry;
+						$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl, $sIcon);
 					}
 				} else {
 					$this->m_oPage->DisableBreadCrumb();
--- a/sources/application/TwigBase/Controller/Controller.php
+++ b/sources/application/TwigBase/Controller/Controller.php
@@ -520,12 +520,11 @@ abstract class Controller
 	}
 
 	/**
-	 * @param bool $bIsBreadCrumbEnabled true to display the breadcrumb, false to hide it
 	 * @since 2.7.7 3.0.1 3.1.1 method creation
 	 * @see Controller::SetBreadCrumbEntry() to set breadcrumb content (by default will be title)
 	 */
-	public function SetIsOperationBreadCrumbEnabled($bIsBreadCrumbEnabled) {
-		$this->m_bIsBreadCrumbEnabled = $bIsBreadCrumbEnabled;
+	public function DisableBreadCrumb() {
+		$this->m_bIsBreadCrumbEnabled = false;
 	}
 
 	/**
--- a/sources/application/TwigBase/Controller/Controller.php
+++ b/sources/application/TwigBase/Controller/Controller.php
@@ -59,6 +59,10 @@ abstract class Controller
 	private $m_aLinkedStylesheets;
 	private $m_aSaas;
 	private $m_aAjaxTabs;
+	/** @var boolean false to disable breadcrumb */
+	private $m_bIsBreadCrumbEnabled = true;
+	/** @var array contains same parameters as {@see iTopWebPage::SetBreadCrumbEntry()} */
+	private $m_aBreadCrumbEntry = [];
 
 
 	/**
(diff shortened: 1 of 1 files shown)


sources/application/TwigBase/Controller/Controller.php
```diff
@@ -579,6 +591,28 @@ public function GetOperationTitle()
...
+	 * @see Controller::IsOperationBreadCrumbEnabled()
+	 * @return string[] empty array, or one containing 3 to 5 items containing :
+	 *   0. id
+	 *   1. title
+	 *   2. description
+	 *   3. url
+	 *   4. icon
+	 */
+	public function GetOperationBreadCrumbEntry() {
+		return [];
```
Hipska: Other parts of iTop make an associative array for it, see https://github.com/Combodo/iTop/blob/5f575d524a7bcc7567cbc8b904204416e9b445e3/sources/Application/WebPage/iTopWebPage.php#L580-L587

But I think the product would benefit from having a `BreadCrumbEntry` class type

sources/application/TwigBase/Controller/Controller.php
```diff
@@ -567,6 +567,18 @@ private function CreatePage($sPageType)
 				$this->m_oPage = new SetupPage($this->GetOperationTitle());
 				break;
 		}
+
+		if ($this->IsOperationBreadCrumbEnabled()) {
+			$aBreadCrumbEntry = $this->GetOperationBreadCrumbEntry();
+			if ((false === empty($aBreadCrumbEntry)) && (is_array($aBreadCrumbEntry)) && (count($aBreadCrumbEntry) >= 3)) {
+				list($sId, $sTitle, $sDescription) = $aBreadCrumbEntry;
+				$sUrl = (count($aBreadCrumbEntry) >= 4) ? $aBreadCrumbEntry[3] : '';
+				$sIcon = (count($aBreadCrumbEntry) >= 5) ? $aBreadCrumbEntry[4] : '';
```
Hipska: What about this?
```suggestion
				@list($sId, $sTitle, $sDescription, $sUrl, $sIcon) = $aBreadCrumbEntry;
```

Hipska (commented):
I assume these new methods are intended to be overriden? Otherwise this code doesn't add anything. Also, wouldn't the default of `IsOperationBreadCrumbEnabled` be `false`?

sources/application/TwigBase/Controller/Controller.php
piRGoif: I tried this in PHP 7.1 and this produces a notice :(
See https://3v4l.org/QeIVr#v7.1.33

sources/application/TwigBase/Controller/Controller.php
piRGoif: I agree, we rely far too much on PHP arrays instead of DTO !
But I hesitate to do it for this small modification... 

Let's see what other dev team members think of this ?

sources/application/TwigBase/Controller/Controller.php
Hipska: Note the `@` 😉

piRGoif:
> I assume these new methods are intended to be overriden? 

Yes exactly !

> Also, wouldn't the default of `IsOperationBreadCrumbEnabled` be `false`?

Currently the breadcrumb is enabled, and contains the page title. I didn't want to change this behavior, so by default a Controller will still has its own breadcrumb entry.

Note that the corresponding method is \iTopWebPage::DisableBreadCrumb which is confusing... But I rather create the opposite meaning for readability (when dealing with boolean having double false is always very confusing : I prefer to have Enabled returning true than Disabled returning false...)

sources/application/TwigBase/Controller/Controller.php
piRGoif: Wooops didn't see that sorry :D
I'm integrating your suggestion :)

piRGoif: To get exactly the same behavior I've pushed 0e63ad60

Hipska: Could be much simpler with `$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl ?: '', $sIcon ?: '');` https://3v4l.org/cPjZJ

piRGoif: I didn't find [in the doc](https://www.php.net/manual/fr/language.operators.comparison.php#language.operators.comparison.ternary) when this was introduced ?
iTop 2.7.* must support PHP 5.6

Hipska: See https://3v4l.org/20lfG Seems to be since 5.3

piRGoif: Thanks ! Pushed with d7e28ea4

sources/application/TwigBase/Controller/Controller.php
```diff
@@ -567,6 +567,16 @@ private function CreatePage($sPageType)
...
 		}
+
+		if ($this->IsOperationBreadCrumbEnabled()) {
+			$aBreadCrumbEntry = $this->GetOperationBreadCrumbEntry();
+			if ((false === empty($aBreadCrumbEntry)) && (is_array($aBreadCrumbEntry)) && (count($aBreadCrumbEntry) >= 3)) {
+				@list($sId, $sTitle, $sDescription, $sUrl, $sIcon) = $aBreadCrumbEntry;
+				$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl ?: '', $sIcon ?: '');
+			}
+		} else {
+			$this->m_oPage->DisableBreadCrumb();
```
Molkobain: I might be missing something on the PR, but `$m_oPage` can be either an iTopWebPage or an AjaxPage.
The latest does not have breadcrumbs, so it makes no sense to manipulate it, and moreover the `$m_oPage->DisableBreadCrumb()` call will crash for AjaxPage.

Molkobain:
Could you include an example of how you need to do currently on iTop and an example of how it would be done after? So we can see the improvement.
For now I'm not in favor of this as it is hard-coded for iTopWebPages and does not consider AjaxPages.

Molkobain:
Accepted in technical team review.
Small rework will be done but globally accepted as-is as there are already some tradeoffs / coupling depending on the page class (eg. AddAjaxPage).
Will be reviewed in functional review next week.

sources/application/TwigBase/Controller/Controller.php
piRGoif: You are perfectly right !
Will be fixed in next iteration

piRGoif:
Discussion summary : 

* Flaws
  - we are calling iTopWebPage breadcrumb methods even if we haven't got an iTopWebPage instance :(
  - this allows to customize the breadcrumb for each Controller impl, but we want to change it for each operation !
* Solutions 
  - add 2 breadcrumb methods in Controller that will be called by operation impl, and save passed parameters to a Controller structure
  - use the saved parameters in Controller::CreatePage

Adding breadcrumb methods in Controller, which should be able to handle multiple WebPage impl isn't very elegant :/ But this is out of scope for this PR, and furthermore having one Controller class per WebPage impl would bring breaking changes... So we are postponing this for now.

sources/application/TwigBase/Controller/Controller.php
piRGoif: I'm closing this conversation, as the solution discussed with the rest of the dev team is less ugly. It could still benefit of a dedicated DTO object, but won't be as necessary.

piRGoif:
Did a rebase + force push

sources/application/TwigBase/Controller/Controller.php
Hipska: So you will end up with sometimes having an associative array and sometimes an indexed array?

I don't think that's the intention of the dev team or product owners?

piRGoif:
Dev team review comments implemented in the last commit !

sources/application/TwigBase/Controller/Controller.php
piRGoif: Closing this conversation as I pushed the modification

sources/application/TwigBase/Controller/Controller.php
Hipska: Oh I see, it is only used internally now.

sources/application/TwigBase/Controller/Controller.php
```diff
@@ -557,6 +578,16 @@ private function CreatePage($sPageType)
 			case 'html':
 				$this->m_oPage = new iTopWebPage($this->GetOperationTitle());
 				$this->m_oPage->add_xframe_options();
+
+				if ($this->m_bIsBreadCrumbEnabled) {
+					if (count($this->m_aBreadCrumbEntry) > 0) {
+						@list($sId, $sTitle, $sDescription, $sUrl, $sIcon) = $this->m_aBreadCrumbEntry;
+						$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl ?: '', $sIcon ?: '');
```
Hipska: The internal member will always have the 5 items..
```suggestion
						list($sId, $sTitle, $sDescription, $sUrl, $sIcon) = $this->m_aBreadCrumbEntry;
						$this->m_oPage->SetBreadCrumbEntry($sId, $sTitle, $sDescription, $sUrl, $sIcon);
```

Hipska (changes requested):

sources/application/TwigBase/Controller/Controller.php
piRGoif: Perfectly right, thanks !

Hipska (approved):

piRGoif:
> Could you include an example of how you need to do currently on iTop and an example of how it would be done after? So we can see the improvement. For now I'm not in favor of this as it is hard-coded for iTopWebPages and does not consider AjaxPages.

Already discussed together, but I've changed the description to add some details.

piRGoif:
Accepted during functionnal review

piRGoif:
Woops sorry, I did a rebase & merge instead of a squash & merge
commits pushed to support/2.7 are : 3ef3166b, 41f6e856, 1709082e, dd7e73e4, 2a3e6384, ce36c00b, 3000659e

PR #261, opened 2022-01-19
N°4761 Fix license.xml content not displayed in setup with multi modules extensions

piRGoif:
Extensions can contain a `license.<module_code>.xml` file to specify licenses for embedded resources (libraries, images, fonts, ...)
The setup was displaying this file content only if the file was directly one level below `/extensions` or `/data/*`. This wasn't the case for multi module extensions, for example TeemIP where we have `/extensions/teemip-core-ip-mgmt/teemip-ip-mgmt/license.teemip-ip-mgmt.xml`

Note the problem isn't present when displaying licenses in the about box in the console : in such scenario the files are searched in the env-* folder, where every modules are copied on the first level.

This PR changed the glob patterns used in \SetupUtils::GetLicenses so that we are now also searching at the second level.

### DIFF
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -1878,9 +1878,9 @@ JS
 		$aLicenceFiles = glob(APPROOT.'setup/licenses/*.xml');
 		if (empty($sEnv))
 		{
-			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'datamodels/*/{*,*/*}/license.*.xml', GLOB_BRACE));
+			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'datamodels/*/*/license.*.xml'));
 			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'extensions/{*,*/*}/license.*.xml', GLOB_BRACE));
-			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'data/*-modules/*/license.*.xml'));
+			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'data/*-modules/{*,*/*}/license.*.xml', GLOB_BRACE));
 		}
 		else
 		{
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -1878,8 +1878,8 @@ JS
 		$aLicenceFiles = glob(APPROOT.'setup/licenses/*.xml');
 		if (empty($sEnv))
 		{
-			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'datamodels/*/*/license.*.xml'));
-			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'extensions/*/license.*.xml'));
+			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'datamodels/*/{*,*/*}/license.*.xml', GLOB_BRACE));
+			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'extensions/{*,*/*}/license.*.xml', GLOB_BRACE));
 			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'data/*-modules/*/license.*.xml'));
 		}
 		else


setup/setuputils.class.inc.php
```diff
@@ -1878,8 +1878,8 @@ static public function GetLicenses($sEnv = null)
 		$aLicenceFiles = glob(APPROOT.'setup/licenses/*.xml');
 		if (empty($sEnv))
 		{
-			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'datamodels/*/*/license.*.xml'));
-			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'extensions/*/license.*.xml'));
+			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'datamodels/*/{*,*/*}/license.*.xml', GLOB_BRACE));
```
Hipska: are there also multi modules in the iTop core data models?

setup/setuputils.class.inc.php
```diff
@@ -1878,8 +1878,8 @@ static public function GetLicenses($sEnv = null)
...
+			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'extensions/{*,*/*}/license.*.xml', GLOB_BRACE));
 			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'data/*-modules/*/license.*.xml'));
```
Molkobain: Why not presume that any folder that can have modules might one day have extensions with nested modules? In that cas you could simplify the processing and do the same for all of them (loop)

Molkobain: Also, can we use the current environment (eg. production, test, ...) instead of scanning all of them? I don't know if it is set there (but it should). That it would be ready for @jbostoen PR about environments.

jbostoen: Isn't this precompilation anyway?

I haven't checked the entire flow, just quickly here because of the mention.  I noticed it's in the setup utils so there the environment is not much of an issue I think?

If it's called from within the iTop console too, it may be worth just checking the contents of certain directories instead of everything in "extensions" (as a user might simply also have extensions there which were not selected for install)

Molkobain: Actually in the setup the environment is already set. Yesterday I ran the setup with the "toolkit" environment set by mistake, it was displayed in the conf file path etc.

So it really depends where this method is called, if it is also called in the "About iTop" in the backoffice, it's an issue.

setup/setuputils.class.inc.php
piRGoif: Not yet, and this shouldn't happen
I'm not sure this is worse it to add the pattern here as well ?

Hipska: Well, its a bit strange to support it while it should not happen. If you support it, you imply it might happen.

piRGoif: The datamodels/2.x folder has always contained modules.
Extensions are a package of one or many modules. This package is intended for easier delivery by the iTop Hub or ITSM Designer, or to have just one download for manual installation. 
Installing manually extensions in datamodels/2.x is not something we want to support.

setup/setuputils.class.inc.php
piRGoif: Thomas opened another thread on the same subject, let's continue the discussion there : https://github.com/Combodo/iTop/pull/261#discussion_r787813352
I'm closing this one.

piRGoif: \SetupUtils::GetLicenses is called from : 
* \WizStepLicense::Display : this step is only displayed when installing iTop for the first time ("install" setup choice instead of "update")
* ajax.render.php with operation=about_box : there we are calling it with the current env as parameter

Does it answers your concerns Guillaume ?

setup/setuputils.class.inc.php
Hipska: So the pattern should be stayed as
```suggestion
			$aLicenceFiles = array_merge($aLicenceFiles, glob(APPROOT.'datamodels/*/*/license.*.xml'));
```

setup/setuputils.class.inc.php
Molkobain: Yes thanks. In the second use case (with the current implementation) it will display too much modules license, we should restrict it to the current environment only.

piRGoif: Closing this conversation :)

setup/setuputils.class.inc.php
piRGoif: Sorry I did too much tasks and didn't get right your comments :/
I made a mistake in the initial commit, I fixed it with 59465b50

setup/setuputils.class.inc.php
Molkobain: Sorry, but I don't see why we should close this, there is no answer to the pointed issue.

piRGoif: Oh so I didn't understand your previous answer. Let's have a direct chat together, shall we ?

Hipska (approved):

setup/setuputils.class.inc.php
piRGoif: Discussed with Guillaume, there were misunderstandings on both sides, we have now clear minds, we can close this discussion :)

setup/setuputils.class.inc.php
piRGoif: closing this conversation as well after Thomas reaction + approval

Molkobain (approved):

Molkobain:
Accepted in technical team review.
Will be reviewed next week in functional review.

Molkobain:
Accepted during functional review, next steps before merge:
  * Create corresponding bug in internal bug tracker
  * Merge PR

PR #263, opened 2022-01-28
N°3129 PHP 8.0 compatibility

piRGoif:
Created this PR to share ongoing work !


piRGoif:
About Twig incompatibilities see : 
https://github.com/twigphp/Twig/pull/3331
https://github.com/symfony/symfony/issues/36872

setup/modelfactory.class.inc.php
```diff
@@ -2294,8 +2294,11 @@ public function IsClassNode()
 	 * Replaces a node by another one, making sure that recursive nodes are preserved
 	 *
 	 * @param MFElement $oNewNode The replacement
+	 *
+	 * @since 2.7.7 3.0.1 3.1.0 N°3129 rename method (from `ReplaceWith` to `MFReplaceWith`) to avoid collision with parent `DOMNode::replaceChild` (different method modifier and parameters :
```
Hipska: I don't get it, how did `ReplaceWith` collide with `replaceChild`?

core/sqlobjectquery.class.inc.php
```diff
@@ -491,7 +491,7 @@ private function PrepareRendering()
 		}
 	}
 
-	private function PrepareSingleTable(SQLObjectQuery $oRootQuery, &$aFrom, $sCallerAlias = '', $aJoinData)
+	private function PrepareSingleTable(SQLObjectQuery $oRootQuery, &$aFrom, $sCallerAlias, $aJoinData)
```
Hipska: Isn't this supported anymore in PHP 8? Or is it because it's not the last item?

setup/modelfactory.class.inc.php
piRGoif: Thanks, I did a wrong paste :/
It collides with \DOMElement::replaceWith of course !
Fixed in 8cd664ae

core/sqlobjectquery.class.inc.php
piRGoif: 2nd option :)

The error message is (note that I removed the full path) : 

```
Deprecated: Required parameter $aJoinData follows optional parameter $sCallerAlias in (...)core\sqlobjectquery.class.inc.php on line 494
```

piRGoif:
From my dev env point of view, everything seems to be fine now.
I'm rebasing/squashing/force push and this PR will exit draft state.

Molkobain:
You can try https://github.com/phan/phan locally to ensure that all the code base is safe if you want to check compatibility with PHP8 and older versions (backward compatibility). It's pretty quick to configure, I tried it on my computer this morning.

piRGoif:
Thanks, will do :)

We will need to switch our CI to PHP 8.0 at least and check for results also (phpunit + BeHat) before announcing compatibility

Hipska:
Yeah, let the CI run on every php version you support 😉

piRGoif:
History rewrote and forced push.
Removing draft flag.

Molkobain:
> Yeah, let the CI run on every php version you support 😉

That's the idea yes, especially as we have a large range of PHP versions supported!

piRGoif:
Phan : I tried... but far too many errors thrown. I give up for now, will certainly be made with another ticket (N°4678 ?)
Merging this PR for now !

PR #265, opened 2022-02-03
N°4698 setup/phpinfo.php : when no valid iTop installation, display a comprehensive message instead of an exception

piRGoif:
When calling the `setup/phpinfo.php` page on a iTop that has not been installed yet, you're getting this exception: 

image

Indeed due to security reasons this page is only accessible to admin users, and this authentication needs a valid iTop instance...
In such a scenario, there is still the `setup/phpcheck.php` page (thanks @BenGrenoble !) that can be used though, but you'll need to know it...

This PR adds a comprehensive message when calling the page in such scenario: 

image

Also a comment was added to the file beginning to recall the existence of `setup/phpcheck.php`

### DIFF
--- a/setup/phpinfo.php
+++ b/setup/phpinfo.php
@@ -17,11 +17,27 @@
  * You should have received a copy of the GNU Affero General Public License
  */
 
+/**
+ * Since N°1934 this page is accessible only to iTop admin users
+ * @since 2.5.2 2.6.1 2.7.0 N°1934 must login as admin user to use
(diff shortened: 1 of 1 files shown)


piRGoif:
Technical review : OK !

Molkobain (approved):

Molkobain:
Accepted during functional review for 2.7.9 / 3.0.4 / 3.1.0

PR #270, opened 2022-02-16
N°4816 - Fix broken setup on fresh install with existing configuration file

Hipska:
This solves issue when trying to setup a new iTop instance with our Docker image.

Fixes Super-Visions/docker-itop#4

I am not completely sure if the config file is a requirement as well, I don't think so.


Molkobain:
Thanks for the PR Thomas, ticket tracked under N°4816

Molkobain (approved):

Molkobain:

Hipska:
@Molkobain could you verify if this code also needs to check if the config file exists or not?

Molkobain:

Molkobain:
> @Molkobain could you verify if this code also needs to check if the config file exists or not?

It might not, but we consider a valid environment when it has both a config file and a compiled env-xxx folder.

PR #273, opened 2022-02-18
N°4921 - Add support for attcode & attvalue parameters in URL to access an object

piRGoif:
## Problem

Since iTop 2.7.0, we are using a different method to init the ref field. Since then in certain cases we can have a different value stored in the reference and in the id field. For more technical details see [Ticket Ref generation](https://www.itophub.io/wiki/page?id=2_7_0%3Arelease%3A2_7_whats_new#ticket_ref_generation)

This can become problematic for the user when he has a ref field containing "R-000100" and an id=99
This means we cannot build an URL to access the object using the ref field value, as accessing object details is done using UI.php with parameters `class` and `id`. For example : `http://localhost/itop-27/pages/UI.php?operation=details&class=UserRequest&id=99`

## Existing workaround

A first workaround is already implemented : the id parameter can contain the friendly name value : in UI.php the object is then seeked using `\MetaModel::GetObjectByName`.
But in certain datamodels the friendlyname can be quite complex...

### Admin Console
This PR brings new `UI.php` URL parameters : `attcode` and `attvalue`.

With the previous example, that means the same object can be accessed using one of those URL : 
`http://localhost/itop-27/pages/UI.php?operation=details&class=UserRequest&id=99`
`http://localhost/itop-27/pages/UI.php?operation=details&class=UserRequest&attcode=ref&attvalue=R-000100`

When using the new URLs, the object retrieval is done using `\MetaModel::GetObjectByColumn`, with the `bMustBeFound` parameter set to `true` : in consequence an exception will be thrown if no object is found, as if multiple instances are.

### User portal
New route : `/object/view/{sObjectClass}/{sObjectAttCode}/{sObjectAttValue}`

For example : 
`http://localhost/itop-27/pages/exec.php/object/view/UserRequest/99?exec_module=itop-portal-base&exec_page=index.php&portal_id=itop-portal`
`http://localhost/itop-27/pages/exec.php/object/view/UserRequest/ref/R-000100?exec_module=itop-portal-base&exec_page=index.php&portal_id=itop-portal`

If none or multiple objects found, then we will get the standard 404 error page (as of the standard object view route).

### Admin Console : unchanged

- Invalid `class` param (`UI.php?operation=details&class=NotExistingClass&id=1`)
image

- Invalid `id` param : non existing id or friendlyname (`UI.php?operation=details&class=UserRequest&id=-1`)
image

### Admin Console : updated

- No `class` and `id` params (`UI.php?operation=details`)
Before PR : 
image
After PR : 
image

- No `id` param (`UI.php?operation=details&class=UserRequest`)
Before PR:
image
After PR:
image

### Admin Console : new

- multiple matches when using `attcode` and `attvalue` params
image

- no matches when using `attcode` and `attvalue` params
image

### DIFF
--- a/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
@@ -69,8 +69,8 @@ class ObjectController extends BrickController
 	 * Displays an cmdbAbstractObject if the connected user is allowed to.
 	 *
 	 * @param \Symfony\Component\HttpFoundation\Request $oRequest
-	 * @param string                                    $sObjectClass (Class must be instance of cmdbAbstractObject)
-	 * @param string                                    $sObjectId
+	 * @param string $sObjectClass (Class must be an instance of cmdbAbstractObject)
+	 * @param string $sObjectId
 	 *
 	 * @return \Symfony\Component\HttpFoundation\Response
 	 *
@@ -83,29 +83,19 @@ class ObjectController extends BrickController
 	 */
 	public function ViewAction(Request $oRequest, $sObjectClass, $sObjectId)
 	{
-		/** @var \Combodo\iTop\Portal\Helper\RequestManipulatorHelper $oRequestManipulator */
-		$oRequestManipulator = $this->get('request_manipulator');
-		/** @var \Combodo\iTop\Portal\Routing\UrlGenerator $oUrlGenerator */
-		$oUrlGenerator = $this->get('url_generator');
-		/** @var \Combodo\iTop\Portal\Helper\ObjectFormHandlerHelper $oObjectFormHandler */
-		$oObjectFormHandler = $this->get('object_form_handler');
 		/** @var \Combodo\iTop\Portal\Helper\SecurityHelper $oSecurityHelper */
 		$oSecurityHelper = $this->get('security_helper');
 		/** @var \Combodo\iTop\Portal\Helper\ScopeValidatorHelper $oScopeValidator */
 		$oScopeValidator = $this->get('scope_validator');
-		/** @var \Combodo\iTop\Portal\Brick\BrickCollection $oBrickCollection */
-		$oBrickCollection = $this->get('brick_collection');
 
 		// Checking parameters
-		if ($sObjectClass === '' || $sObjectId === '')
-		{
+		if ($sObjectClass === '' || $sObjectId === '') {
 			IssueLog::Info(__METHOD__.' at line '.__LINE__.' : sObjectClass and sObjectId expected, "'.$sObjectClass.'" and "'.$sObjectId.'" given.');
 			throw new HttpException(Response::HTTP_INTERNAL_SERVER_ERROR, Dict::Format('UI:Error:2ParametersMissing', 'class', 'id'));
 		}
 
 		// Checking security layers
-		if (!$oSecurityHelper->IsActionAllowed(UR_ACTION_READ, $sObjectClass, $sObjectId))
-		{
+		if (!$oSecurityHelper->IsActionAllowed(UR_ACTION_READ, $sObjectClass, $sObjectId)) {
 			IssueLog::Warning(__METHOD__.' at line '.__LINE__.' : User #'.UserRights::GetUserId().' not allowed to read '.$sObjectClass.'::'.$sObjectId.' object.');
 			throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
 		}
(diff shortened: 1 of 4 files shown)


jbostoen:
Question: does it simply take the first object, if the attribute value is not unique?

Hipska:
It throws exception:
> in consequence an exception will be thrown if no object is found, as if multiple instances are.

pages/UI.php
```diff
@@ -385,30 +385,35 @@ function DisplayNavigatorGroupTab($oP)
...
+			if (($id === '') || (false === is_numeric($id))) {
+				$sAttCode = utils::ReadParam('attcode', '');
+				$sAttValue = utils::ReadParam('attvalue', '');
+
+				if (empty($sAttCode) || empty($sAttValue)) {
+					throw new ApplicationException(Dict::Format('UI:Error:3ParametersMissing', 'class', 'attcode', 'attvalue'));
+				}
+
+				$oObj = MetaModel::GetObjectByColumn($sClass, $sAttCode, $sAttValue, true);
+			} else if (is_numeric($id)) {
```
Molkobain: I would rather keep the test on the ID as the first test, it would insist on the fact that it's the main / best way to access an object, especially for performance reason. To me, accessing the object via another attribute is a last resort thing.

pages/UI.php
```diff
@@ -385,30 +385,35 @@ function DisplayNavigatorGroupTab($oP)
...
 			}
 
-			if (is_numeric($id))
-			{
+			if (($id === '') || (false === is_numeric($id))) {
+				$sAttCode = utils::ReadParam('attcode', '');
+				$sAttValue = utils::ReadParam('attvalue', '');
+
+				if (empty($sAttCode) || empty($sAttValue)) {
+					throw new ApplicationException(Dict::Format('UI:Error:3ParametersMissing', 'class', 'attcode', 'attvalue'));
```
Hipska:
```suggestion
					throw new ApplicationException(Dict::Format('UI:Error:2ParametersMissing', 'attcode', 'attvalue'));
```

pages/UI.php
piRGoif: I changed the way param handling is done in a9eaba21
Ok for you @Molkobain  ?

piRGoif:
Updated description to add error messages

pages/UI.php
piRGoif: Just pushed 92329bd9 also : if only the class param is provided, we will now display that id is missing (instead of attcode/attvalue)

piRGoif:
Missed technical review today (too much PR to review). Will be reviewed next month !

pages/UI.php
```diff
@@ -384,31 +384,37 @@ function DisplayNavigatorGroupTab($oP)
...
+				if (is_numeric($id)) {
+					$oObj = MetaModel::GetObject($sClass, $id, false /* MustBeFound */);
+				} else {
+					$oObj = MetaModel::GetObjectByName($sClass, $id, false /* MustBeFound */);
+				}
+			} else {
+				$sAttCode = utils::ReadParam('attcode', '');
+				$sAttValue = utils::ReadParam('attvalue', '');
+
+				if (empty($sAttCode) || empty($sAttValue)) {
```
Molkobain:
```suggestion
				if ( (strlen($sAttCode) === 0) || (strlen(empty($sAttValue) === 0) ) {
```

We need to use `strlen()` otherwise accessing an object with $sAttValue at `"0"` will not work.

Molkobain (changes requested):

pages/UI.php
piRGoif: Damn :( Thanks !
Fixed in 217e796c (your suggestion kept the empty() call for the second param)

Molkobain: Woops, sorry! 😬

piRGoif:
Just pushed 2 new commits adding a new route in the user portal.
Description updated accordingly.

Thanks to @Molkobain for the precious help !
And @v-dumas for the suggestion :)

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
```diff
@@ -68,8 +68,8 @@ class ObjectController extends BrickController
 	 * Displays an cmdbAbstractObject if the connected user is allowed to.
```
Molkobain:
```suggestion
	 * Displays an cmdbAbstractObject (from its ID) if the connected user is allowed to.
```

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
```diff
@@ -82,53 +82,123 @@ class ObjectController extends BrickController
...
 			// We should never be there as the secuirty helper makes sure that the object exists, but just in case.
 			IssueLog::Info(__METHOD__.' at line '.__LINE__.' : Could not load object '.$sObjectClass.'::'.$sObjectId.'.');
 			throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
 		}
 
+		return $this->PrepareViewObjectResponse($oRequest, $oObject);
+	}
+
+	/**
+	 * @param \Symfony\Component\HttpFoundation\Request $oRequest
```
Molkobain, line 118:
```suggestion
	 * Displays an cmdbAbstractObject (if the connected user is allowed to) from a specific attribute. If several or none objects are found with the attribute value, an exception is thrown.
	 * 
	 * @param \Symfony\Component\HttpFoundation\Request $oRequest
```

Molkobain (dismissed):

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
piRGoif: This is the old code, so this suggestion is an error right ?

Molkobain: As we introduced a new method to display an object, it makes sense (to me) to explain how the old one is now different than the new one.

piRGoif:
Approved during functional review, for 3.1.0

piRGoif:
Rebased branch on develop

Molkobain (approved):

Molkobain:
Ok to merge @piRGoif !

PR #274, opened 2022-02-19
:bug: N°4834 - Add the 'pattern' parameter with the unicode Cyrillic block regex to the CKEditor mentions config

vbkunin:
So that mentions work with Russian, Ukrainian, Belarusian, Bulgarian and other languages.

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -2654,10 +2654,11 @@ HTML;
 				$aDefaultConf['mentions'][] = [
 					'feed' => $sMentionEndpoint,
 					'marker' => $sMentionMarker,
-					'minChars' => MetaModel::GetConfig()->Get('min_autocomplete_chars'),
+					'minChars' => $iMinChars = MetaModel::GetConfig()->Get('min_autocomplete_chars'),
 					'itemTemplate' => $sMentionItemTemplate,
 					'outputTemplate' => $sMentionOutputTemplate,
 					'throttle' => 500,
+					'pattern' => "\\{$sMentionMarker}[_a-zA-Z0-9À-žЀ-Ӿ]{{$iMinChars},}$"
 				];
 			}
 		}
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -2654,11 +2654,10 @@ HTML;
 				$aDefaultConf['mentions'][] = [
 					'feed' => $sMentionEndpoint,
 					'marker' => $sMentionMarker,
-					'minChars' => $iMinChars = MetaModel::GetConfig()->Get('min_autocomplete_chars'),
+					'minChars' => MetaModel::GetConfig()->Get('min_autocomplete_chars'),
 					'itemTemplate' => $sMentionItemTemplate,
 					'outputTemplate' => $sMentionOutputTemplate,
 					'throttle' => 500,
-					'pattern' => "\\{$sMentionMarker}[_a-zA-Z0-9À-žЀ-Ӿ]{{$iMinChars},}$"
 				];
 			}
 		}
--- a/js/ckeditor.on-init.js
+++ b/js/ckeditor.on-init.js
@@ -21,3 +21,24 @@ if ((CKEDITOR !== undefined) && (CKEDITOR.plugins.registered['disabler'] === und
 
 		});
 }
+
+// Rewrite the CKEditor Mentions plugin regexp to make it suitable for all Unicode alphabets.
+if (CKEDITOR !== undefined && CKEDITOR.plugins.registered['mentions']) {
+	// from https://github.com/ckeditor/ckeditor4/blob/a3786007fb979d7d7bff3d10c34a2d422935baed/plugins/mentions/plugin.js#L147
+	function createPattern(marker, minChars) {
+		let pattern = marker + '[\\p{L}\\p{N}_-]';
+		if ( minChars ) {
+			pattern += '{' + minChars + ',}';
+		} else {
+			pattern += '*';
+		}
+		pattern += '$';
+		return new RegExp(pattern, 'u');
+	}
+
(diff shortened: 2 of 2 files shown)


Molkobain:
Hello Vladimir,

Thanks for pointing this out! Are you saying that the standard `\w` regexp symbol does not work for cyrillic alphabet (and other languages such as asian / arabic probably) ? It's what is used in the [default pattern](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_plugins_mentions.html#property-pattern): `/^@+\w*$/`

vbkunin:
No, `\w` doesn't work because it's an equivalent of `[a-zA-Z0-9_]`. [By default](https://github.com/ckeditor/ckeditor4/blob/a3786007fb979d7d7bff3d10c34a2d422935baed/plugins/mentions/plugin.js#L149) the plugin uses `[_a-zA-Z0-9À-ž]` pattern, so this is why it works not only for English but also for other Latin-script alphabets.

Molkobain:
Damn! I really thought the \w worked for all alphabets... 😕 Thanks for the info 

In that case I'll check if we can improve it for asian alphabets as well. At least Chinese as there are several contributors on the project.

Molkobain:
From what I saw [here](https://www.regular-expressions.info/unicode.html#prop) we could try the `\p{L}` which matches any kind of letter for all unicode alphabets.

Molkobain:
_TL;DR: Need to find if CKEditor works with the `u` (unicode) flag on RegExp._

The `\p{L}` seems to be supported by ECMAScript since v6 ([source](https://caniuse.com/mdn-javascript_builtins_regexp_property_escapes)), so it could be a great step forward. But for that to work we need to set the `u` (unicode) flag on the RegExp, I still need to check if this is possible within CKEditor;

On a side note, this also shows that we currently don't support "-" in names either (eg. "Anne-Catherine") 😅

Here is the tested RegExp `^@[\p{L}\p{N}-_]*$` ("@" being replaced by the marker, "*" by the min. chars. param.)
And the list it was tested against:
```
@Guillaume          // Latin chars: Passed
@Clémentine         // Latin accent: Passed
@Anne-Catherine     // With dash: Passed
@Foo123             // With number: Passed
@Александра         // Cyrillic chars: Passed
@Alexandra          // - Latin equivalent: Passed
@姓氏               // Chinese chars: Passed
@성씨               // Korean chars: Passed
@عائشة
@Aïcha              // - Latin equivalent: Passed
@:_#                // Should not match: Passed (no match)
```

For the arabic name, it doesn't pass yet. I don't know if it is a LTR vs RTL issue.

_Good source for copying names in various alphabets: [behindthename.com](https://www.behindthename.com/)_

vbkunin:
> The `\p{L}` seems to be supported by ECMAScript since v6 ([source](https://caniuse.com/mdn-javascript_builtins_regexp_property_escapes), so it could be a great step forward. But for that to work we need to set the `u` (unicode) flag on the RegExp, I still need to check if this is possible within CKEditor;

I was just going to say the same things)
In the CKEditor config the pattern is expected to be a RegExp object so it works with flags when using literal notation `/\p{L}/u`. But it doesn't work with a string pattern `"/\p{L}/u"`, when a string argument passed to match()  it is implicitly converted to a RegExp using `new RegExp("/\p{L}/u")` and flargs are treated as a part of the  pattern.

Molkobain:
Using the RegExp object would be great as it would allow to set the flag, but due to how the CKEditor integration is done within iTop I'm not sure we can find a way to use it.

  * CKEditor options are defined in a [PHP array](https://github.com/Combodo/iTop/blob/support/3.0/application/utils.inc.php#L2654),
  * then [JSON encoded](https://github.com/Combodo/iTop/blob/support/3.0/application/cmdbabstract.class.inc.php#L2362) to be passed in the [JS instantiation](https://github.com/Combodo/iTop/blob/support/3.0/application/cmdbabstract.class.inc.php#L2365).

But have can we put something in a PHP array that will be encoded as `"pattern": new RegExp(...)` and not `"pattern": "new RegExp(...)"`?

Edit: If we find no way to do it, we will stick with your initial proposal.

vbkunin:
I do not know any way to pass an object through json_encode, I think this may have unpredictable consequences.  Instead, there is another option. We can handle the CKEditor 'instanceLoaded' event and then rewrite the pattern in the plugin config.
See this commit: https://github.com/vbkunin/iTop/commit/01ccd1593483dadd4200b34a534784ec1963609b

Molkobain:
> [...] See this commit: [vbkunin@01ccd15](https://github.com/vbkunin/iTop/commit/01ccd1593483dadd4200b34a534784ec1963609b)

This seems like a very nice workaround, I'll present it during the next PR review, thanks Vladimir!

Molkobain:
Technical review: @Molkobain will check if we can use the regexp modes internally (eg. [here](https://www.regular-expressions.info/modifiers.html)) to activate the unicode flag within the string and stay with the original approach. If not, we will proceed with the alternative from Vladimir.

Molkobain:
For reference, a chinese user complains about it on [SourceForge](https://sourceforge.net/p/itop/discussion/922361/thread/cabea00009/?limit=25)

Molkobain:
Unfortunately we cannot use the regex modes internally in ECMAScript, so we'll havre to go with your second approach @vbkunin .
Do you mind updating the PR to reflect that? Thanks ✌

vbkunin:
Hello @Molkobain
I've reverted changes in the utils.inc.php and cherry picked from the commit  https://github.com/vbkunin/iTop/commit/01ccd1593483dadd4200b34a534784ec1963609b.

Molkobain:
Accepted in functional review, will be part of the 3.0.2.

PR #275, opened 2022-02-23
N°4849 - Enable notification emails grouping in threads in email clients

Molkobain:
This PR is an attempt and a call for help 😁

### Issue
When we receive notification emails from iTop, they are displayed as different conversations in the email client (eg. Outlook) even though they have the same subject and are about the same iTop object. One of the main drawback is that the mailbox can feel spammed / flooded.

_The first 3 emails are notifications about the same object, but they are already flooding the mailbox folder_
image

### Expectation
Find a way to group notification emails for a same object in a thread like email clients do when receiving emails from other platforms.

_Example with comments on a PR on GitHub here_ 
image

_Example of desired behavior with iTop notification emails_
image

#### Specs
From what I saw [here](https://en.wikipedia.org/wiki/Email#Header_fields) and on other sites, it seems that the email needs to have the `In-Reply-To` header, the spec says that it should contain the `Message-ID` of the email to chain to. But we can't do that when it comes to notifications as they are not replies, they have no previous `Message-ID`.

For instance, GitHub is using the same value as in the `References` header in the `In-Reply-To`, this seems to allow emails to group around the same "object", here a PR, but it can be an issue or a commit.
```
[...]

From: xxxx <notifications@github.com>
Reply-To: Combodo/iTop <noreply@github.com>
To: Combodo/iTop <iTop@noreply.github.com>
Cc: Push <push@noreply.github.com>
Message-ID: <Combodo/iTop/pull/262/push/9043873881@github.com>
In-Reply-To: <Combodo/iTop/pull/262@github.com>
References: <Combodo/iTop/pull/262@github.com>
Subject: =?UTF-8?Q?Re:_[Combodo/iTop]_=F0=9F=8C=90Updated_Dutch_transl?=
 =?UTF-8?Q?ations_=28PR_#262=29?=

[...]
```

It is pretty clear that the `Message-ID` is unique to the email, but that `References` and `In-Reply-To` are common to all emails regarding this PR.

#### Implementation
Previously, both `Message-ID` and `References` headers had the same value, it contained the object class, the object ID, the current timestamp and an identifier of the current instance.

```
[...]

Message-ID: <iTop_UserRequest_542_1645547692.537156@8c43a16d628776c7a7098fae22c1e880-production.openitop.org>
Date: Tue, 22 Feb 2022 17:35:00 +0100
Subject: The ticket R-000561 has been updated
From: <guillaume.lajarige@combodo.com>
To: <guillaume.lajarige@combodo.com>
References: <iTop_UserRequest_542_1645547692.537156@8c43a16d628776c7a7098fae22c1e880-production.openitop.org>

[...]
```

With this PR the following things have been done:
  * Change the `References` header so it is common to all emails about the triggering object by removing the timestamp part.
  * Add a `In-Reply-To` header with the same value as the `References`

```
Message-ID: <iTop_UserRequest_542_1645604682@8c43a16d628776c7a7098fae22c1e880-production.openitop.org>
Date: Wed, 23 Feb 2022 09:24:42 +0100
Subject: The ticket R-000561 has been updated
From: <guillaume.lajarige@combodo.com>
To: <guillaume.lajarige@combodo.com>
References: <iTop_UserRequest_542@8c43a16d628776c7a7098fae22c1e880-production.openitop.org>
In-Reply-To: <iTop_UserRequest_542@8c43a16d628776c7a7098fae22c1e880-production.openitop.org>
```

#### Results
  * Grouping works for same subject in GMail, but it was actually already working before as it seems to match on the subject anyway.
  * Grouping works for same subject in Evolution since the PR.
  * Grouping works for same subject in Apple's Mail since this PR.
  * Grouping does NOT works in Outlook 2013 or Outlook Online.
    * Workaround is to prefix notifications with "RE: "

### What's next
  - [X] Find why it's not working on Outlook
  - [x] Add a unit test for the \ActionEmail::GenerateIdentifierForHeaders() method

### DIFF
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -575,17 +575,16 @@ class ActionEmail extends ActionNotification
 				if ($sHeaderName === static::ENUM_HEADER_NAME_REFERENCES) {
 					$sIdentifier = "<$sIdentifier>";
 				}
-				break;
 
-			default:
-				$sErrorMessage = sprinf('%s: Could not generate identifier for header "%s", only %s are supported', static::class, $sHeaderName, implode(' / ', [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES]));
-				IssueLog::Error($sErrorMessage, LogChannels::NOTIFICATIONS, [
-					'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->GetRawName().')',
-					'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->GetRawName().')',
-				]);
-				throw new Exception($sErrorMessage);
+				return $sIdentifier;
 		}
 
-		return $sIdentifier;
+		// Requested header name invalid
+		$sErrorMessage = sprinf('%s: Could not generate identifier for header "%s", only %s are supported', static::class, $sHeaderName, implode(' / ', [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES]));
+		IssueLog::Error($sErrorMessage, LogChannels::NOTIFICATIONS, [
+			'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->GetRawName().')',
+			'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->GetRawName().')',
+		]);
+		throw new Exception($sErrorMessage);
 	}
 }
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -579,8 +579,8 @@ class ActionEmail extends ActionNotification
 			default:
 				$sErrorMessage = sprinf('%s: Could not generate identifier for header "%s", only %s are supported', static::class, $sHeaderName, implode(' / ', [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES]));
 				IssueLog::Error($sErrorMessage, LogChannels::ACTION, [
-					'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->Get('friendlyname').')',
-					'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->Get('friendlyname').')',
+					'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->GetRawName().')',
+					'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->GetRawName().')',
 				]);
 				throw new Exception($sErrorMessage);
 		}
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -579,7 +579,7 @@ class ActionEmail extends ActionNotification
 
 			default:
 				$sErrorMessage = sprinf('%s: Could not generate identifier for header "%s", only %s are supported', static::class, $sHeaderName, implode(' / ', [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES]));
-				IssueLog::Error($sErrorMessage, LogChannels::ACTION, [
+				IssueLog::Error($sErrorMessage, LogChannels::NOTIFICATIONS, [
 					'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->GetRawName().')',
 					'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->GetRawName().')',
 				]);
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -201,19 +201,30 @@ abstract class ActionNotification extends Action
 class ActionEmail extends ActionNotification
 {
 	/**
+	 * @var string
+	 * @since 3.0.1
+	 */
+	const ENUM_HEADER_NAME_MESSAGE_ID = 'Message-ID';
+	/**
+	 * @var string
+	 * @since 3.0.1
+	 */
+	const ENUM_HEADER_NAME_REFERENCES = 'References';
+
+	/**
 	 * @inheritDoc
 	 */
 	public static function Init()
 	{
 		$aParams = array
 		(
-			"category" => "grant_by_profile,core/cmdb,application",
-			"key_type" => "autoincrement",
-			"name_attcode" => "name",
-			"state_attcode" => "",
-			"reconc_keys" => array('name'),
-			"db_table" => "priv_action_email",
-			"db_key_field" => "id",
+			"category"            => "grant_by_profile,core/cmdb,application",
+			"key_type"            => "autoincrement",
+			"name_attcode"        => "name",
+			"state_attcode"       => "",
+			"reconc_keys"         => array('name'),
+			"db_table"            => "priv_action_email",
+			"db_key_field"        => "id",
 			"db_finalclass_field" => "",
 		);
 		MetaModel::Init_Params($aParams);
@@ -416,9 +427,8 @@ class ActionEmail extends ActionNotification
 			$sBody = MetaModel::ApplyParams($this->Get('body'), $aContextArgs);
 
 			$oObj = $aContextArgs['this->object()'];
-			$sMessageId = sprintf('iTop_%s_%d_%f@%s.openitop.org', get_class($oObj), $oObj->GetKey(), microtime(true /* get as float*/),
-				MetaModel::GetEnvironmentId());
-			$sReference = '<'.$sMessageId.'>';
+			$sMessageId = $this->GenerateIdentifierForHeaders($oObj, static::ENUM_HEADER_NAME_MESSAGE_ID);
+			$sReference = $this->GenerateIdentifierForHeaders($oObj, static::ENUM_HEADER_NAME_REFERENCES);
 		}
 		catch (Exception $e) {
 			/** @noinspection PhpUnhandledExceptionInspection */
@@ -456,8 +466,7 @@ class ActionEmail extends ActionNotification
 
 		$oEmail = new EMail();
 
-		if ($this->IsBeingTested())
-		{
+		if ($this->IsBeingTested()) {
 			$oEmail->SetSubject('TEST['.$sSubject.']');
 			$sTestBody = $sBody;
 			$sTestBody .= "<div style=\"border: dashed;\">\n";
@@ -467,8 +476,8 @@ class ActionEmail extends ActionNotification
 			$sTestBody .= "<li>TO: $sTo</li>\n";
 			$sTestBody .= "<li>CC: $sCC</li>\n";
 			$sTestBody .= "<li>BCC: $sBCC</li>\n";
-			$sTestBody .= empty($sFromLabel) ? "<li>From: $sFrom</li>\n": "<li>From: $sFromLabel &lt;$sFrom&gt;</li>\n";
-			$sTestBody .= empty($sReplyToLabel) ? "<li>Reply-To: $sReplyTo</li>\n": "<li>Reply-To: $sReplyToLabel &lt;$sReplyTo&gt;</li>\n";
+			$sTestBody .= empty($sFromLabel) ? "<li>From: $sFrom</li>\n" : "<li>From: $sFromLabel &lt;$sFrom&gt;</li>\n";
+			$sTestBody .= empty($sReplyToLabel) ? "<li>Reply-To: $sReplyTo</li>\n" : "<li>Reply-To: $sReplyToLabel &lt;$sReplyTo&gt;</li>\n";
 			$sTestBody .= "<li>References: $sReference</li>\n";
 			$sTestBody .= "</ul>\n";
 			$sTestBody .= "</p>\n";
@@ -478,9 +487,9 @@ class ActionEmail extends ActionNotification
 			$oEmail->SetRecipientFrom($sFrom, $sFromLabel);
 			$oEmail->SetReferences($sReference);
 			$oEmail->SetMessageId($sMessageId);
-		}
-		else
-		{
+			// Note: N°4849 We pass the "References" identifier instead of the "Message-ID" on purpose as we want notifications emails to group around the triggering iTop object, not just the users' replies to the notification
+			$oEmail->SetInReplyTo($sReference);
+		} else {
 			$oEmail->SetSubject($sSubject);
 			$oEmail->SetBody($sBody, 'text/html', $sStyles);
 			$oEmail->SetRecipientTO($sTo);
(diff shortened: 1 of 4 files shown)


steffunky:
Outlook seems to use `Thread-Topic` and `Thread-Index` to group multiple emails (https://bugzilla.mozilla.org/show_bug.cgi?id=411601 , https://docs.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxomsg/9e994fbb-b839-495f-84e3-2c8c02c7dd9b?redirectedfrom=MSDN).

If GMail uses subject match to group iTop notifications, would this still work if an attribute value used in the notification subject changes ?

An other question would be, do we want different notifications (change, new caslog entry, state change, mention) for the same object to group ?

core/action.class.inc.php
```diff
@@ -516,26 +527,64 @@ protected function _DoExecute($oTrigger, $aContextArgs, &$oLog)
...
+	protected function GenerateIdentifierForHeaders(DBObject $oObject, string $sHeaderName): string
+	{
+		$sObjClass = get_class($oObject);
+		$sObjId = $oObject->GetKey();
+
+		switch ($sHeaderName) {
+			case static::ENUM_HEADER_NAME_MESSAGE_ID:
+			case static::ENUM_HEADER_NAME_REFERENCES:
+				// Prefix
+				$sPrefix = sprintf('iTop_%s_%d', $sObjClass, $sObjId);
```
Hipska: What about
```suggestion
				$sPrefix = sprintf(ITOP_APPLICATION_SHORT.'_%s_%d', $sObjClass, $sObjId);
```
or
```suggestion
				$sPrefix = sprintf('%s_%s_%d', ITOP_APPLICATION_SHORT, $sObjClass, $sObjId);
```

core/action.class.inc.php
```diff
@@ -516,26 +527,64 @@ protected function _DoExecute($oTrigger, $aContextArgs, &$oLog)
...
+				if ($sHeaderName === static::ENUM_HEADER_NAME_REFERENCES) {
+					$sIdentifier = "<$sIdentifier>";
+				}
+				break;
+
+			default:
+				$sErrorMessage = sprinf('%s: Could not generate identifier for header "%s", only %s are supported', static::class, $sHeaderName, implode(' / ', [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES]));
+				IssueLog::Error($sErrorMessage, LogChannels::ACTION, [
+					'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->Get('friendlyname').')',
+					'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->Get('friendlyname').')',
```
Hipska:
```suggestion
					'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->GetRawName().')',
					'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->GetRawName().')',
```

Also, the `DBObject` class should have a `__toString`-like method that does this in one go. Unfortunately, the existing one returns html representation. (a `ToRawString` or something)

Molkobain:
> If GMail uses subject match to group iTop notifications, would this still work if an attribute value used in the notification subject changes ?

After a quick test, this doesn't work, a new subject means a new thread on GMail.
 
> An other question would be, do we want different notifications (change, new caslog entry, state change, mention) for the same object to group ?

Ideally yes, I would like to have all emails about the same object to be on the same thread, but if you could just group on similar mail subject it would already be a great step forward.

Molkobain:
After playing a bit, Stephen found that simply prefixing the mail subject with "Re: " in the action allowed the grouping in Outlook.
So we could imagine that iTop would add it automatically to ActionEmail but we haven't found a great condition yet.

Hipska (changes requested):
Hereby I confirm grouping is currently also not working for Apple's Mail (macOS,iOS)

core/action.class.inc.php
Molkobain: Exactly my thought, we repeat this pattern on several occasion and it would be great to have it, but I'm didn't find a good name for it yet.

core/action.class.inc.php
Molkobain: I'm not sure of the consequences if an app has white-space or a special char within this constant (it is redefined by people building a different application upon iTop Community)

Hipska: Maybe sanitise it first?

Molkobain: Yes, done in b22d8dce7a1d19c568388eeb5f4287f67241eb6f (I was lazy, I admit it 🙈)

Hipska (approved):

Hipska:
@Molkobain If you want me to test this behaviour on Apple's mail, just send me 2 notification mails and I'll let you know if it worked..

core/action.class.inc.php
Hipska: Yeah, looking at the other existing methods, `ToRawString` or `GetAsRawString` would suffice. I also had a look and maybe we can just adapt the existing `__toString` method:
```php
	/**
	 * String representation of the object
	 *
	 * Only a few meaningful information will be returned.
	 * This representation is for debugging purposes, and is subject to change.
	 *
	 * @param bool $bHTML The returned string is raw HTML
	 *
	 * @return string
	 * @throws \ArchivedObjectException
	 * @throws \CoreException
	 * @throws \DictExceptionMissingString
	 */
	public function __toString($bHTML = true)
	{
        $sRet = '';
        $sClass = static::GetClassName(get_class($this));
        $iPKey = $this->GetKey();
        if ($bHTML) {
	        $sRootClass = MetaModel::GetRootClass($sClass);
	        $sFriendlyName = $this->GetAsHTML('friendlyname');
	        $sRet .= "<b title=\"$sRootClass\">$sClass</b>::$iPKey ($sFriendlyName)<br/>\n";
        } else {
        	$sFriendlyName = $this->GetRawName();
        	$sRet .= sprintf('%s::%d (%s)', $sClass, $iPKey, $sFriendlyName);
        }
        return $sRet;
	}
```

Molkobain:
> @Molkobain If you want me to test this behaviour on Apple's mail, just send me 2 notification mails and I'll let you know if it worked..

Sent, thanks!

Hipska:
Works perfectly:
image

Molkobain:
> Works perfectly [...]

For real?!

Well, I guess this isn't perfect as it does not group all messages about an object on the same thread, but this could already be a step forward. Thanks Thomas for the test! 😊

Molkobain:
Accepted in technical review.

Hipska (approved):

core/action.class.inc.php
```diff
@@ -516,26 +527,65 @@ protected function _DoExecute($oTrigger, $aContextArgs, &$oLog)
...
+	 * @throws \Exception
+	 * @since 3.0.1 N°4849
+	 */
+	protected function GenerateIdentifierForHeaders(DBObject $oObject, string $sHeaderName): string
+	{
+		$sObjClass = get_class($oObject);
+		$sObjId = $oObject->GetKey();
+		$sAppName = utils::Sanitize(ITOP_APPLICATION_SHORT, '', utils::ENUM_SANITIZATION_FILTER_VARIABLE_NAME);
+
+		switch ($sHeaderName) {
```
piRGoif, line 563: I found the switch syntax kinda difficult to read

I would rather move the exit condition directly in the method first lines.
Then having the regular code below...

piRGoif (changes requested):
Very nice phpdoc and comments, bravo !
I'm just asking for a small change in GenerateIdentifierForHeaders method as I find the current code not very readable ?

core/action.class.inc.php
Molkobain: With the current implementation, when new identifier types will be added, we'll only need to add code (new case), no refactor will be necessary (change of code of the exit condition), reducing the risk of regressions.

What I can propose to simplify a bit is:
  * Move the exception after the switch as the fallback behavior
  * Move the return condition in the case block

That way we keep the "code addition" principle and get a bit of simplification.

piRGoif: YAGNI and KISS ?
When we'll add new header types, it will be time to refactor.
You're introducing now an unnecessary complexity for something we're not sure we will even need one day... I don't like it much :/

Molkobain: _Just pushed a commit to move the invalid condition after the switch_

To me it's way more simpler to understand/read the logic as-is than to use the header type constants in several IF conditions, but if the rest of the team feels like you I'll change it. (For logic changes, I think we should tend to something most of the devs are confortable with, rather than just choose between the two of us)

PS: Is the following what you had in mind?

```
	protected function GenerateIdentifierForHeaders(DBObject $oObject, string $sHeaderName): string
	{
		$sObjClass = get_class($oObject);
		$sObjId = $oObject->GetKey();
		$sAppName = utils::Sanitize(ITOP_APPLICATION_SHORT, '', utils::ENUM_SANITIZATION_FILTER_VARIABLE_NAME);

		if (! in_array($sHeaderName, [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES])) {
			$sErrorMessage = sprinf('%s: Could not generate identifier for header "%s", only %s are supported', static::class, $sHeaderName, implode(' / ', [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES]));
			IssueLog::Error($sErrorMessage, LogChannels::NOTIFICATIONS, [
				'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->GetRawName().')',
				'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->GetRawName().')',
			]);
			throw new Exception($sErrorMessage);
		}
		
		if (in_array($sHeaderName, [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES])) {
			// Prefix
			$sPrefix = sprintf('%s_%s_%d', $sAppName, $sObjClass, $sObjId);
			if ($sHeaderName === static::ENUM_HEADER_NAME_MESSAGE_ID) {
				$sPrefix .= sprintf('_%f', microtime(true /* get as float*/));
			}
			// Suffix
			$sSuffix = sprintf('@%s.openitop.org', MetaModel::GetEnvironmentId());
			// Identifier
			$sIdentifier = $sPrefix.$sSuffix;
			if ($sHeaderName === static::ENUM_HEADER_NAME_REFERENCES) {
				$sIdentifier = "<$sIdentifier>";
			}
		}
		
		return $sIdentifier;
	}
```

core/action.class.inc.php
```diff
@@ -516,26 +527,64 @@ protected function _DoExecute($oTrigger, $aContextArgs, &$oLog)
...
+			case static::ENUM_HEADER_NAME_MESSAGE_ID:
+			case static::ENUM_HEADER_NAME_REFERENCES:
+				// Prefix
+				$sPrefix = sprintf('%s_%s_%d', $sAppName, $sObjClass, $sObjId);
+				if ($sHeaderName === static::ENUM_HEADER_NAME_MESSAGE_ID) {
+					$sPrefix .= sprintf('_%f', microtime(true /* get as float*/));
```
Hipska, line 569: You're adding the time as part of the identifier? Not sure if that is a good idea, there are better and more secure ways to add some random data.

core/action.class.inc.php
piRGoif: Here is what I suggest : 

```php
	/**
	 * @param \DBObject $oObject
	 * @param string $sHeaderName one of {@see \ActionEmail::ENUM_HEADER_NAME_REFERENCES} or {@see \ActionEmail::ENUM_HEADER_NAME_MESSAGE_ID}
	 *
	 * @return string The formatted identifier for $sHeaderName based on $oObject
	 * 
	 * @throws CoreUnexpectedValue if sHeaderName is not one of the expected value
	 * @throws \CoreException
	 * 
	 * @since 3.0.1 N°4849
	 */
	protected function GenerateIdentifierForHeaders(DBObject $oObject, string $sHeaderName): string
	{
		$sObjClass = get_class($oObject);
		$sObjId = $oObject->GetKey();

		if (false === in_array($sHeaderName, [static::ENUM_HEADER_NAME_MESSAGE_ID, self::ENUM_HEADER_NAME_REFERENCES], true)) {
			$sErrorMessage = sprinf('%s: Could not generate identifier for header "%s", only %s are supported', static::class, $sHeaderName, implode(' / ', [static::ENUM_HEADER_NAME_MESSAGE_ID, static::ENUM_HEADER_NAME_REFERENCES]));
			IssueLog::Error($sErrorMessage, LogChannels::NOTIFICATIONS, [
				'Object' => $sObjClass.'::'.$sObjId.' ('.$oObject->GetRawName().')',
				'Action' => get_class($this).'::'.$this->GetKey().' ('.$this->GetRawName().')',
			]);
			throw new CoreUnexpectedValue($sErrorMessage);
		}

		$sAppName = utils::Sanitize(ITOP_APPLICATION_SHORT, '', utils::ENUM_SANITIZATION_FILTER_VARIABLE_NAME);

		// Prefix
		$sPrefix = sprintf('%s_%s_%d', $sAppName, $sObjClass, $sObjId);
		if ($sHeaderName === static::ENUM_HEADER_NAME_MESSAGE_ID) {
			$sPrefix .= sprintf('_%f', microtime(true /* get as float*/));
		}
		// Suffix
		$sSuffix = sprintf('@%s.openitop.org', MetaModel::GetEnvironmentId());
		// Identifier
		$sIdentifier = $sPrefix.$sSuffix;
		if ($sHeaderName === static::ENUM_HEADER_NAME_REFERENCES) {
			$sIdentifier = "<$sIdentifier>";
		}

		return $sIdentifier;
	}
```

core/action.class.inc.php
piRGoif: This is what was implemented before the PR... Changing this should be done in another PR I think ?

Hipska: Correct, but before it was also using "iTop" instead of `ITOP_APPLICATION_SHORT`.

Next to that, I also think the app name should be part of the suffix, not the prefix. (semantically more correct)

Molkobain: Indeed microtime was already part of the `Message-ID` and `References`.
With this PR we remove the microtime from `References` to enable the grouping as intended. It is now only in `Message-ID`.

We could find something else maybe but in another PR I agree.
Replacing "iTop" with the constant doesn't introduce a change of the generated header for standard products (iTop Community / Essential / Professional); it might change behavior for people repackaging / rebranding the app, but it's out of our scope (should be documented in migration notes though).

Molkobain:
Accepted in functional review, will be part of 2.7.7 / 3.0.2

core/action.class.inc.php
piRGoif: As Hipska agrees, closing this conversation

PR #278, opened 2022-03-11
N°4947 Fix \EMail::LoadConfig

piRGoif:
In `\EMail::SendSynchronous` we are calling `\EMail::LoadConfig`.
This method uses the `ITOP_DEFAULT_CONFIG_FILE` constant, and as a consequence is always loading config using the production environment.

This PR changes the loading and now uses the standard `\utils::GetConfig` method !

Config was saved in the `\EMail::$m_oConfig` attribute : it is only written there.
The `ITOP_DEFAULT_CONFIG_FILE` constant is only used in the toolkit and `sources/application/status/status.inc.php`.  All of those usages seems legit to me so I'm not changing them in this PR.

Thanks to @jbostoen for the bug report (see https://github.com/Combodo/iTop/pull/223#issuecomment-1064854430) !

### DIFF
--- a/core/email.class.inc.php
+++ b/core/email.class.inc.php
@@ -161,8 +161,6 @@ class EMail
 		// If the body of the message is in HTML, embed all images based on attachments
 		$this->EmbedInlineImages();
 		
-		$this->LoadConfig();
-
 		$sTransport = self::$m_oConfig->Get('email_transport');
 		switch ($sTransport)
 		{
--- a/core/email.class.inc.php
+++ b/core/email.class.inc.php
@@ -54,7 +54,13 @@ class EMail
 	{
 		$this->m_aData = array();
 		$this->m_oMessage = Swift_Message::newInstance();
-		$this->SetRecipientFrom(MetaModel::GetConfig()->Get('email_default_sender_address'), MetaModel::GetConfig()->Get('email_default_sender_label'));
+
+		$this->LoadConfig();
+		$oConfig = self::$m_oConfig;
+		$this->SetRecipientFrom(
+			$oConfig->Get('email_default_sender_address'),
+			$oConfig->Get('email_default_sender_label')
+		);
 	}
 
 	/**
--- a/core/email.class.inc.php
+++ b/core/email.class.inc.php
@@ -41,11 +41,10 @@ class EMail
 	protected static $m_oConfig = null;
 	protected $m_aData; // For storing data to serialize
 
-	public function LoadConfig($sConfigFile = ITOP_DEFAULT_CONFIG_FILE)
+	public function LoadConfig()
 	{
-		if (is_null(self::$m_oConfig))
-		{
-			self::$m_oConfig = new Config($sConfigFile);
+		if (is_null(self::$m_oConfig)) {
+			self::$m_oConfig = utils::GetConfig();
 		}
 	}
 
--- a/core/email.class.inc.php
+++ b/core/email.class.inc.php
@@ -56,7 +56,7 @@ class EMail
 		$this->m_oMessage = Swift_Message::newInstance();
 
 		$this->LoadConfig();
-		$oConfig = self::$m_oConfig;
+		$oConfig = static::$m_oConfig;
 		$this->SetRecipientFrom(
 			$oConfig->Get('email_default_sender_address'),
 			$oConfig->Get('email_default_sender_label')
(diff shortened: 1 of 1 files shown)


core/email.class.inc.php
```diff
@@ -41,11 +41,10 @@ class EMail
...
 	protected $m_aData; // For storing data to serialize
 
-	public function LoadConfig($sConfigFile = ITOP_DEFAULT_CONFIG_FILE)
+	public function LoadConfig()
 	{
-		if (is_null(self::$m_oConfig))
-		{
-			self::$m_oConfig = new Config($sConfigFile);
+		if (is_null(self::$m_oConfig)) {
+			self::$m_oConfig = utils::GetConfig();
```
Hipska: Why using `utils::GetConfig()` while the constructor uses `MetaModel::GetConfig()`? This seems inconsistent..

Hipska (changes requested):

core/email.class.inc.php
piRGoif: Ouch, didn't see that, thanks !

piRGoif: Fixed in 12951f8a

core/email.class.inc.php
```diff
@@ -55,7 +54,13 @@ public function __construct()
...
 		$this->m_aData = array();
 		$this->m_oMessage = Swift_Message::newInstance();
-		$this->SetRecipientFrom(MetaModel::GetConfig()->Get('email_default_sender_address'), MetaModel::GetConfig()->Get('email_default_sender_label'));
+
+		$this->LoadConfig();
+		$oConfig = self::$m_oConfig;
+		$this->SetRecipientFrom(
+			$oConfig->Get('email_default_sender_address'),
+			$oConfig->Get('email_default_sender_label')
+		);
```
Hipska:
```suggestion
		$this->SetRecipientFrom(
			static::$m_oConfig->Get('email_default_sender_address'),
			static::$m_oConfig->Get('email_default_sender_label')
		);
```

core/email.class.inc.php
Hipska:
```suggestion
		if (is_null(static::$m_oConfig)) {
			static::$m_oConfig = utils::GetConfig();
```

Hipska (changes requested):
When the LoadConfig is already done in the constructor, there is no need to do it on other places anymore..

core/email.class.inc.php
piRGoif: I found my solution easier to read...
Guess we'll see what other devs will say during the review

Hipska: What about removing the `LoadConfig` and `$m_oConfig` completely and do this in construct then: `$oConfig = utils::GetConfig()`. It doesn't make much sense to have it as a static var.

piRGoif: In the legacy code, there were lots of static attributes to save the config object (it was the case in the utils class before 2.7.0 and that was the main reason why I was calling it a mess :o) ). I guess this was done for performance reasons ? 
Removing `\EMail::$m_oConfig` is something we'll have to discuss during this PR dev review indeed !
I have some doubt personally, as the class can be called multiple times without the MetaModel loaded, so this would mean opening the config file each time...

There are no reasons to keep the `\EMail::LoadConfig` call in `\EMail::SendSynchronous` though, so I'm removing it ! (commit 1ab3a51f)

Hipska: It indeed makes sense to not to have to load the config multiple times if metamodel isn't loaded yet. I agree to keep it static for now.

PS; just make sure it's using `static::$m_oConfig` instead of `self::$m_oConfig` everywhere.

Hipska (approved):

piRGoif:
Indeed, there were quite a few remaining uses of the attribute with the "self" keyword. I've changed them in 8640918d
Thanks !

Hipska (approved):

core/email.class.inc.php
```diff
@@ -41,11 +41,10 @@ class EMail
 	protected static $m_oConfig = null;
 	protected $m_aData; // For storing data to serialize
 
-	public function LoadConfig($sConfigFile = ITOP_DEFAULT_CONFIG_FILE)
+	public function LoadConfig()
```
Molkobain, line 55: Can you just add an `@since` to notify of the param removal?

Molkobain (approved):

Molkobain:

core/email.class.inc.php
piRGoif: Waiting for the functional review so that we know in which version this will be merged

Molkobain:
Will be merged on support/2.7 branch **only once the 2.7.7 is out**, so we can test it on 3.0.2 first.

core/email.class.inc.php
Hipska: Sorry to come back on this, but what do you think of the following?
```php
$oConfig = $this->LoadConfig();
```

core/email.class.inc.php
piRGoif: Done in 9ea94415

core/email.class.inc.php
piRGoif: Done in 99ab128d :)
Closing this conversation

piRGoif:
The Email class was totally refactored in iTop 2.7.7 : 

* eb1d56f43942b3f49275005073919a701f13d109
* 754946bf62fcc9f62ddaf8120951b6c6127001e1
* 622f40c06cf9d3dbe170416498c4906dd1314318

Rebasing this branch cannot be done anymore. I'm closing this PR and opening a new one.

Molkobain:
Can we delete the branch?

piRGoif:
Now that I finished the new PR, yes :)

PR #279, opened 2022-03-11
N°5072 - Fix default priority to undefined (null) or default value if not nullable

Timmy38:
By default, if no urgency or impact have been selected in a user request, the priority shown is "Critical"
priority_01

It is misleading when the impact is set to only one person : changing the value of urgency from "undefined" to "critical" change the priority from "Critical" to "High" !
priority_02

I suggest changing the default priority (when no priority can be computed) to undefined ($iPriority is null)
priority_fixed

The provided implementation has been tested on the 4 cases : user request (all in one), user request (itil), incident and problem
This behavior was already present in previous versions.

### DIFF
--- a/datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
+++ b/datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
@@ -1142,43 +1142,40 @@
 					<static>false</static>
 					<access>public</access>
 					<type>LifecycleAction</type>
-					<code><![CDATA[        public function ComputePriority()
-        {
-                // priority[impact][urgency]
-                $aPriorities = array(
-                        // single person
-                        1 => array(
-                                        1 => 1,
-                                        2 => 1,
-                                        3 => 2,
-                                        4 => 4,
-                        ),
-                        // a group
-                        2 => array(
-                                1 => 1,
-                                2 => 2,
-                                3 => 3,
-                                4 => 4,
-                        ),
-                        // a departement!
-                        3 => array(
-                                        1 => 2,
-                                        2 => 3,
-                                        3 => 3,
-                                        4 => 4,
-                        ),
-                );
-                $iPriority = null;
-                $oAttDef = MetaModel::GetAttributeDef(get_class($this), 'priority');
-				if (!$oAttDef->IsNullAllowed()) {
-					$iPriority = $oAttDef->GetDefaultValue();
-				}
+					<code><![CDATA[public function ComputePriority()
+	{
+		// priority[impact][urgency]
+		$aPriorities = array(
+			// single person
+			1 => array(
+				1 => 1,
+				2 => 1,
+				3 => 2,
+				4 => 4,
+			),
+			// a group
+			2 => array(
+				1 => 1,
+				2 => 2,
+				3 => 3,
+				4 => 4,
+			),
+			// a departement!
+			3 => array(
+				1 => 2,
+				2 => 3,
+				3 => 3,
+				4 => 4,
+			),
+		);
+		$oAttDef = MetaModel::GetAttributeDef(get_class($this), 'priority');
+		$iPriority = $oAttDef->IsNullAllowed() ? null : $oAttDef->GetDefaultValue();
 
-                if (isset($aPriorities[(int)$this->Get('impact')]) && isset($aPriorities[(int)$this->Get('impact')][(int)$this->Get('urgency')])) {
-                        $iPriority = $aPriorities[(int)$this->Get('impact')][(int)$this->Get('urgency')];
-                }
-                return $iPriority;              
-        }]]></code>
+		if (isset($aPriorities[(int)$this->Get('impact')]) && isset($aPriorities[(int)$this->Get('impact')][(int)$this->Get('urgency')])) {
+			$iPriority = $aPriorities[(int)$this->Get('impact')][(int)$this->Get('urgency')];
+		}
+		return $iPriority;
+	}]]></code>
 				</method>
 				<method id="ComputeValues">
 					<static>false</static>
--- a/datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
+++ b/datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
@@ -1168,7 +1168,7 @@
                                         4 => 4,
                         ),
                 );
-                $iPriority = 1;
+                $iPriority = null;
                 if (isset($aPriorities[(int)$this->Get('impact')]) && isset($aPriorities[(int)$this->Get('impact')][(int)$this->Get('urgency')]))
                 {
                         $iPriority = $aPriorities[(int)$this->Get('impact')][(int)$this->Get('urgency')];
--- a/datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
+++ b/datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
@@ -1169,8 +1169,12 @@
                         ),
                 );
                 $iPriority = null;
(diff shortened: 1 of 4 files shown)


Hipska:
IMHO `null` should be selected when a non-existing mapping for impact & urgency combination is used. (for example when user added new options via extensions). And the default priority in all other cases could maybe be lowered to "medium" or even "low", but that actually doesn't matter much as both fields are required.

piRGoif:
Hello,
Can you modify this PR's description to add the info that this was always present ? As your screenshots represent a 3.0.0 version, there is an ambiguity that this is a regression fix :/

Hipska:
> Can you modify this PR's description to add the info that this was always present ? As your screenshots represent a 3.0.0 version, there is an ambiguity that this is a regression fix :/

Pierre, this doesn't seem more relevant than on any other PR IMHO. So I would either request this for all fix-PR's or none. Not just only this one because it happens to be the first one since 3.0 release.

piRGoif:
Thomas, believe me we will need this information during internal PR review meeting !
Especially cause this PR is based on the develop branch (not on support/2.7) and has iTop 3.0.0 screenshots only in its description.

If you have any other PR in mind where this should be necessary, just comment on those.
Do not add the needed information just because it might also be needed in other PR doesn't seems right to me.
There are 30 PR in Combodo org repos that were created after iTop 3.0.0 delivery (search with : `org:Combodo is:public is:pr created:>=2022-01-01`), 10 of them contains the "bug" label, none opened PR are based on the develop branch.

Hipska:
This is actually not really the place to discuss this, but I'm going to add one example:
PR's on branch develop created after release of 2.7 would have the same ambiguity: Screenshots could meant it was a bug in 2.7, 2.6, 2.5 or even older..

I do agree it's useful information to know on which versions a problem/bug exists, but not just because of there happens to be screenshots from the latest version only in the PR.

piRGoif:
> IMHO `null` should be selected when a non-existing mapping for impact & urgency combination is used. (for example when user added new options via extensions). And the default priority in all other cases could maybe be lowered to "medium" or even "low", but that actually doesn't matter much as both fields are required.

Not sure what you meant ? 
After this PR if no mapping exists in `$aPriorities` we will return null indeed ?

Adding new values will force to override the whole method just to specify new elements in `$aPriorities`, this is bad :/ We might change that in another PR.

piRGoif (dismissed):

Hipska:
What I was hinting to, was actually; If a person already selects an Impact (or Urgency) only, the Priority could be set to "low" or "medium" instead of "undefined". But as said, it doesn't matter much, since both the fields are required.

Hipska (dismissed):

piRGoif:
Oh ok, seems to make sense. We should talk about this in the functionnal review meeting

Molkobain (dismissed):

Molkobain:

Molkobain:
Functional review: Setting value to null can be an issue for someone who customized the datamodel and allowed 1 of the 2 attributes to not be mandatory anymore. In that case, the user will get an exception on save.

Code should be adjusted so it sets the default value to either:
  * null is attribute is nullable
  * its default value if attribute is NOT nullable

Molkobain:
Also @Timmy38 can you create the corresponding ticket in our internal tracker? If already existing, please prefix the PR title with its reference (eg. "N°1234 - Blah blah") thanks!

datamodels/2.x/itop-request-mgmt-itil/datamodel.itop-request-mgmt-itil.xml
```diff
@@ -1344,7 +1344,12 @@
 	{
 
 		// Compute the priority of the ticket
-		$this->Set('priority', $this->ComputePriority());
+		$iPriority = $this->ComputePriority();
+		$oAttDef = MetaModel::GetAttributeDef(get_class($this), 'priority');
+		if(is_null($iPriority) && !$oAttDef->IsNullAllowed()){
+			$iPriority = $oAttDef->GetDefaultValue();
+		}
+		$this->Set('priority', $iPriority);
```
Hipska: What about this?

```suggestion
		$oAttDef = MetaModel::GetAttributeDef(get_class($this), 'priority');
		$iDefaultPriority = $oAttDef->GetDefaultValue();
		$this->Set('priority', $this->ComputePriority($iDefaultPriority));
```
Then you always start with setting priority to the default value and apply the correct one if there is a mapping..

Hipska: @Timmy38 what do you think of my suggestion?

datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
```diff
@@ -1169,8 +1169,12 @@
                         ),
                 );
                 $iPriority = null;
-                if (isset($aPriorities[(int)$this->Get('impact')]) && isset($aPriorities[(int)$this->Get('impact')][(int)$this->Get('urgency')]))
-                {
+                $oAttDef = MetaModel::GetAttributeDef(get_class($this), 'priority');
+				if (!$oAttDef->IsNullAllowed()) {
+					$iPriority = $oAttDef->GetDefaultValue();
+				}
```
Molkobain: Mind the indentation please, same on the other files.

Also, now that the nullable property is tested in this method, we could make it even simpler, lines 1171-1175 could become the following. What do you think?
```
$oAttDef = MetaModel::GetAttributeDef(get_class($this), 'priority');
$iPriority = $oAttDef->IsNullAllowed() ? null : $oAttDef->GetDefaultValue();
```

Molkobain (changes requested):

datamodels/2.x/itop-request-mgmt-itil/datamodel.itop-request-mgmt-itil.xml
Timmy38: Hello, I think it's better to compute everything priority-related (including default values) inside ComputePriority()

Hipska: You got a point on that, but then I would also remove all datamodel dependent stuff from `ComputePriority` 😉

datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
Hipska: @Molkobain What about this?

```php
$iPriority = (int) $this->Get('priority');
```

I think that will also always return the default value.

Molkobain:
Accepted in technical review

datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
Molkobain: I guess that in the case the attribute is nullable and has a default value of let say '4', it will return '4' which is not what we aim at.

Hipska: 🤔 I think we do. If datamodel is set to `<default />` then we would get `null`.

Molkobain: Hum, you are right. The excepted returned value in that case will be discussed next Tuesday

Molkobain:
Accepted in functional review

datamodels/2.x/itop-incident-mgmt-itil/datamodel.itop-incident-mgmt-itil.xml
Hipska: What was the outcome, it seems to be merged as-is?

Molkobain: Keeping it as-is was the preferred solution as some attributes are not prefilled on init.
Now this is going to be tested by the quality team.

Hipska: 👍 Thanks, my suggestion wasn't good anyway as it would return `0` when value would be `null`.

PR #281, opened 2022-03-24
N°4998 Fix CSS for AttributeDuration in transition form

piRGoif:
Those fields are badly displayed in such context (notice the "s" that is wrapped on next line)

image

XML to reproduce : 

```xml
  <classes>
    <class id="UserRequest" _created_in="itop-request-mgmt" _delta="must_exist">
      <fields>
        <field id="timespentphase" xsi:type="AttributeDuration" _delta="define">
          <sql>timespentphase</sql>
          <default_value/>
          <tracking_level>all</tracking_level>
        </field>
      </fields>
      <lifecycle>
        <states>
          <state id="new" _delta="must_exist">
            <transitions>
              <transition id="ev_assign" _delta="must_exist">
                <flags _delta="define">
                  <attribute id="timespentphase">
                    <must_prompt/>
                  </attribute>
                </flags>
              </transition>
            </transitions>
          </state>
        </states>
      </lifecycle>
      <presentation>
        <details>
          <items>
            <item id="timespentphase" _delta="define">
              <rank>90</rank>
            </item>
          </items>
        </details>
      </presentation>
    </class>
  </classes>
  <dictionaries>
  </dictionaries>
  <menus>
  </menus>
  <user_rights _created_in="core">
  </user_rights>
  <module_designs>
  </module_designs>
</itop_design>
```

### DIFF
--- a/css/light-grey.scss
+++ b/css/light-grey.scss
@@ -2424,26 +2424,33 @@ fieldset .details>.field_container {
 
 								.selectize-dropdown,
 								.selectize-input,
-								.selectize-input input{
-									font-size: 12px;
-								}
-								.selectize-input{
-									padding: 2px 2px 0px 2px; /* padding-bottom = padding-top - item margin-bottom */
-									border: 1px solid #ABABAB;
-									border-radius: 0;
+                .selectize-input input {
+                  font-size: 12px;
+                }
 
-									.attribute-set-item.partial-code{
-										color: transparentize($gray-darker, 0.4);
-										background-color: lighten($gray-lighter, 5%);
-									}
-								}
-							}
-						}
-					}
-				}
-			}
-		}
-	}
+                .selectize-input {
+                  padding: 2px 2px 0px 2px; /* padding-bottom = padding-top - item margin-bottom */
+                  border: 1px solid #ABABAB;
+                  border-radius: 0;
+
+                  .attribute-set-item.partial-code {
+                    color: transparentize($gray-darker, 0.4);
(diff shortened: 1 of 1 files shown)


css/light-grey.scss
```diff
@@ -2424,26 +2424,33 @@ fieldset .details>.field_container {
...
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  &[data-attribute-type="AttributeDuration"] {
+    div.field_value_container {
```
Molkobain: Why use the "div" tag name in the selector, was is mandatory? It is introducing an unnecessary "weight" for further overloading of that rule.

piRGoif: Indeed I tried to add as many criteria as possible to increase the [selector's specificity](https://www.w3.org/TR/CSS22/cascade.html#specificity), as this rule should always be applied and using !important is a very bad idea.

Molkobain:
Reviewed today, @piRGoif will test a simpler selector (no div), then we will advise.

css/light-grey.scss
piRGoif: Works well without the div in the selector
Updated with f52a1e86

piRGoif:
> Reviewed today, @piRGoif will test a simpler selector (no div), then we will advise.

Done, it is working well without "div" in the selector. The SCSS file is updated.

Molkobain (approved):

Molkobain:
Accepted in technical review.

Molkobain:
Accepted in functional review, will be part of the 2.7.7.

PR #282, opened 2022-03-29
N°5414 Debug log for invalid placeholders

piRGoif:
This adds a debug log when a placeholder cannot be replaced.

This can happen for example in notifications, for a hyperlink() function for which we are specifying an invalid portal id : `$this->hyperlink('non_existing_portal')$`

Before the placeholder was just not replaced.
Now we can enable a debug log on the LogChannels::NOTIFICATION channel.

Unfortunately the catch is done where we don't have the trigger / action context... But at least we will have more info than before !

### DIFF
--- a/core/metamodel.class.php
+++ b/core/metamodel.class.php
@@ -7370,7 +7370,18 @@ abstract class MetaModel
 									}
 								}
 								catch (Exception $e) {
-									// No replacement will occur
+									$aContext = [
+										'placeholder'   => $sPlaceholderAttCode,
+										'replace class' => get_class($replace),
+									];
+									if ($replace instanceof DBObject) {
+										$aContext['replace id'] = $replace->GetKey();
+									}
+									IssueLog::Debug(
+										'Invalid placeholder in notification, no replacement will occur !',
+										LogChannels::NOTIFICATION,
+										$aContext
+									);
 								}
 							}
 						}
@@ -7391,7 +7402,14 @@ abstract class MetaModel
 								$aSearches[] = $aMatches[1][$idx].$sSearch.$aMatches[1][$idx];
 							}
 							catch (Exception $e) {
-								// No replacement will occur
+								IssueLog::Debug(
+									'Invalid placeholder in notification, no replacement will occur !',
+									LogChannels::NOTIFICATION,
+									[
+										'placeholder' => $sPlaceholderAttCode,
+										'replace'     => $replace,
+									]
+								);
 							}
 						}
 					}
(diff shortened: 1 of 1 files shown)


core/metamodel.class.php
```diff
@@ -7344,78 +7344,79 @@ public static function ApplyParams($sInput, $aParams)
...
+								catch (Exception $e) {
+									$aContext = [
+										'placeholder'   => $sPlaceholderAttCode,
+										'replace class' => get_class($replace),
+									];
+									if ($replace instanceof DBObject) {
+										$aContext['replace id'] = $replace->GetKey();
+									}
+									IssueLog::Debug(
+										'Invalid placeholder in notification, no replacement will occur !',
```
Molkobain:
```suggestion
										'Invalid placeholder in notification, no replacement will occur!',
```

core/metamodel.class.php
```diff
@@ -7344,78 +7344,79 @@ public static function ApplyParams($sInput, $aParams)
...
+									}
+								}
+								catch (Exception $e) {
+									$aContext = [
+										'placeholder'   => $sPlaceholderAttCode,
+										'replace class' => get_class($replace),
+									];
+									if ($replace instanceof DBObject) {
+										$aContext['replace id'] = $replace->GetKey();
+									}
```
Molkobain, line 7379: Reviewed last month, @piRGoif will encapsulate these lines in an IF statement to go there only in case debug is enabled.

piRGoif:
During technical review, we wanted to see if it was possible not to call Debug() method when the level isn't enabled.
Indeed it is... but starting from 3.0.0 only :/ The \LogAPI::IsLogLevelEnabled was added in that version with N°3731, and is already called when simply logging (in \LogAPI::WriteLog).
So nothing to be done in 3.0.0, and if we want to limit the logs in 2.7 we should simply update the LogAPI class in that branch.

Molkobain:
Sorry @piRGoif I just saw that comments from the last review where not submitted :/

core/metamodel.class.php
Molkobain: Reply from @piRGoif at https://github.com/Combodo/iTop/pull/282#issuecomment-1116226167

piRGoif:
Discussed during technical review !
We agreed to change the log level : use Error instead of Debug

Ideally we could replace the placeholder with the error message, but that should be done deeper in the call stack... 
Also, a specific behavior when beeing on a isDevEnv instance could be triggered ? 
For now we are ok on simply adding a log :)

piRGoif:
rebased & force pushed to solve conflicts

piRGoif:
Functional review : OK, but for 2.7.8 / 3.0.2

PR #283, opened 2022-04-01
:sparkles: Add support for AttributeClass from XML

GurneyHallack:
Hello,

When creating an attribute of type "AttributeClass" from PHP, it's working.

When creating an attribute of type "AttributeClass" from XML, there are two parameters that are not taken from the XML :
- class_category 
- more_values 

Thus, it generates an error on compilation (setup):
2022-02-26 08_55_39-Window_itop_error
Notice: Undefined index: class_category in /var/www/html/web/core/attributedef.class.inc.php
Notice: Undefined index: more_values in /var/www/html/web/core/attributedef.class.inc.php

I didn't found a testCase for testing the Compiler creating attributes. If you point me to it, I would gladly add the adequate test.

Have a good snowy day (April joke here :) )

### DIFF
--- a/setup/compiler.class.inc.php
+++ b/setup/compiler.class.inc.php
@@ -1725,6 +1725,16 @@ EOF
 						$aParameters['definition_file'] = "'".str_replace("'", "\\'", $sFileName)."'";
 					}
 				}
+				elseif ($sAttType == 'AttributeClass')
+				{
+                    $aParameters['class_category'] = $this->GetPropString($oField, 'class_category', '');
+                    $aParameters['more_values'] = $this->GetPropString($oField, 'more_values', '');
+                    $aParameters['allowed_values'] = 'null'; // or "new ValueSetEnum('SELECT xxxx')"
+                    $aParameters['sql'] = $this->GetMandatoryPropString($oField, 'sql');
+                    $aParameters['default_value'] = $this->GetPropString($oField, 'default_value', '');
+                    $aParameters['is_null_allowed'] = $this->GetPropBoolean($oField, 'is_null_allowed', false);
+                    $aParameters['depends_on'] = $sDependencies;
+				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


Hipska:
I'm not sure this is meant to be used with xml, as this type is also not documented in the xml data model reference wiki page.

GurneyHallack:
Then, why the compiler manages the case of "AttributeClassAttCodeSet", even if it is also not documented in the xml data model reference ? 
The case in the compiler for "AttributeClassAttCodeSet" was added with https://github.com/Combodo/iTop/commit/ec597f697a912d05dc5d5f8f8f1bd4d4436e4ddd, after the commit "N°917: AttributeClassAttCodeSet created" https://github.com/Combodo/iTop/commit/720d33405305e295b7afb1c81a54e5d49dc047b3

Hipska:
🤷 Don't ask me … 😂

piRGoif:
Hello,
Thanks for the PR !

`AttributeClass` was created a long time ago, I think before XML customization was added.
This was certainly forgotten when XML was added as a possible source...

Can you share a XML with such a field ?

Hipska:
@piRGoif could you in the meantime make sure the attribute is described in the documentations?

piRGoif:
> could you in the meantime make sure the attribute is described in the documentations?

Documenting or not will be decided during the functional review of this PR.

GurneyHallack:
[datamodel.pull-283-sample.xml.txt](https://github.com/Combodo/iTop/files/8538701/datamodel.pull-283-sample.xml.txt)
Hello,
Here's an xml file (.txt) with the AttributeClass field. 
The first class is the minimalist declaration, the second one is a simple one.
Would that be good ?

Molkobain:
Hello @GurneyHallack , sorry for the long response time. This was fixed in iTop 3.2.0 which will be released this year.

Hipska:
@Molkobain So will it now be documented?

Molkobain:
Yep, it will be documented on iTop 3.2 wiki which will be available along with the package.

PR #285, opened 2022-04-07
🎨 AttributeCaseLog index is not binary data, but serialised data

Hipska:
This improves compatibility and portability of raw database data.

Both data types have same length, so it should have no problems when migrating.

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -4667,7 +4667,7 @@ class AttributeCaseLog extends AttributeLongText
 		$aColumns = array();
 		$aColumns[$this->GetCode()] = 'LONGTEXT' // 2^32 (4 Gb)
 			.CMDBSource::GetSqlStringColumnDefinition();
-		$aColumns[$this->GetCode().'_index'] = 'BLOB';
+		$aColumns[$this->GetCode().'_index'] = 'TEXT';
 
 		return $aColumns;
 	}


Molkobain:
Hello Thomas,

Thanks for the PR, but in order to move this forward we need a bit more information:
  * What symptom did you had when migrating the DB?
  * How can we reproduce the said symptom?

Hipska:
I discovered this during testing of #286, I didn't had real actual reproducible problems, but I can imagine they could occur because of the incorrect storage type. It simply isn't binary data, so shouldn't be stored as such.

Molkobain:
I understand that, but with the lack of man power we have, investing time to reproduce/fix/test something that is perfectly working to potentially introduce side effects on some environments has a very low interest for us when there are plenty of other fixes to do around the app.

I'm mostly saying this as without further arguments to "the case" it will likely to be rejected by both technical / functional teams.

Hipska:
Well, at least you don't need to fix it anymore as I did already for you guys 😉 

Its their loss if they're going to reject this cleanup because of that..

Molkobain:
References for technical review: [MySQL](https://dev.mysql.com/doc/refman/8.0/en/blob.html), [MariaDB](https://mariadb.com/kb/en/blob-and-text-data-types/)

Molkobain:
During setup iTop run SQL queries to fix DB fields and align them to the correct specifications. We need to check on an high-volume DB instance if this has a significant impact on setup duration and if so what decision to make:
  * Accept the PR, document the SQL query so it can be run prior to the iTop upgrade, limiting the maintenance time of iTop (was done before for other DB changes)
  * Refuse the PR as the cons would overtake the pros

Status switch to "pending functional review" so we have the discussion next week. Benchmark will be done in the meantime hopefully.

Molkobain:
PR was rejected for now during functional review:
  * As stated this could cause setup to take a very long time during upgrade on high volume databases, like we experienced in the past.
  * Evaluating the performance impact would require spending 0.5/1 day on this, unfortunately we don't have the bandwidth for this as it does not have enough added value compared to the rest of the backlog.

Hipska:
Please also close on rejection..

piRGoif:
Indeed, it was forgotten ! Thanks

PR #269, opened 2022-02-11
N°4766 - DataSynchro: Allow binary data to be imported/synchronised with the synchro_import.php

Hipska:
See https://github.com/Combodo/iTop/pull/236#issuecomment-1036073741, this is the same PR as #236 but now to merge into `support/3.0` branch.

PR #286, opened 2022-04-07
N°5867 - Display binary data size in SynchroReplica details

Hipska:
Currently the SynchroReplica details page shows all data from the `synchro_data` table as unformatted/raw. With this change it doesn't try to display binary data directly anymore, but it now shows the size of the data.

Before:
image

After:
image

Disclaimer: The sample data is not an actual image, but just 128 sample bytes.

### DIFF
--- a/synchro/synchrodatasource.class.inc.php
+++ b/synchro/synchrodatasource.class.inc.php
@@ -2823,7 +2823,11 @@ class SynchroReplica extends DBObject implements iDisplay
 			);
 			$aRows = array();
 			foreach ($aData as $sKey => $value) {
-				$aRows[] = array('attcode' => $sKey, 'data' => $value);
+				if ((strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false))
+				{
+					$aRows[] = array('attcode' => $sKey, 'data' => chunk_split(base64_encode($value)));
+				}
+				else $aRows[] = array('attcode' => $sKey, 'data' => $value);
 			}
 			$oPage->Table($aHeaders, $aRows);
 			$oPage->add('</fieldset>');
--- a/synchro/synchrodatasource.class.inc.php
+++ b/synchro/synchrodatasource.class.inc.php
@@ -2818,8 +2818,8 @@ class SynchroReplica extends DBObject implements iDisplay
 			$aData = $this->LoadExtendedDataFromTable($sSQLTable);
 
 			$aHeaders = array(
-				'attcode' => array('label' => 'Attribute Code', 'description' => ''),
-				'data'    => array('label' => 'Value', 'description' => ''),
+				'attcode' => array('label' => Dict::S('UI:Form:Property'), 'description' => ''),
+				'data'    => array('label' => Dict::S('UI:Form:Value'), 'description' => ''),
 			);
 			$aRows = array();
 			foreach ($aData as $sKey => $value) {
--- a/synchro/synchrodatasource.class.inc.php
+++ b/synchro/synchrodatasource.class.inc.php
@@ -2825,7 +2825,7 @@ class SynchroReplica extends DBObject implements iDisplay
 			foreach ($aData as $sKey => $value) {
 				if ((strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false))
 				{
-					$aRows[] = array('attcode' => $sKey, 'data' => chunk_split(base64_encode($value)));
+					$aRows[] = array('attcode' => $sKey, 'data' => sprintf('<i>%s (%s)</i>', Dict::S('Core:AttributeBlob'), utils::BytesToFriendlyFormat(strlen($value))));
 				}
 				else $aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
 			}
--- a/synchro/synchrodatasource.class.inc.php
+++ b/synchro/synchrodatasource.class.inc.php
@@ -2823,11 +2823,11 @@ class SynchroReplica extends DBObject implements iDisplay
 			);
 			$aRows = array();
 			foreach ($aData as $sKey => $value) {
-				if ((strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false))
-				{
+				if (strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false) {
 					$aRows[] = array('attcode' => $sKey, 'data' => sprintf('<i>%s (%s)</i>', Dict::S('Core:AttributeBlob'), utils::BytesToFriendlyFormat(strlen($value))));
+				} else {
+					$aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
 				}
-				else $aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
 			}
 			$oPage->Table($aHeaders, $aRows);
 			$oPage->add('</fieldset>');
--- a/synchro/synchrodatasource.class.inc.php
+++ b/synchro/synchrodatasource.class.inc.php
@@ -2827,7 +2827,7 @@ class SynchroReplica extends DBObject implements iDisplay
 				{
 					$aRows[] = array('attcode' => $sKey, 'data' => chunk_split(base64_encode($value)));
 				}
-				else $aRows[] = array('attcode' => $sKey, 'data' => $value);
+				else $aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
 			}
 			$oPage->Table($aHeaders, $aRows);
 			$oPage->add('</fieldset>');


piRGoif:
Hello,
Do you need to see the base64 content ? Or should we replace it with the mention "blob" and value size maybe ?

Hipska:
No, but with it being base64 encoded, the user could copy and download/convert the data easier. But I like your suggestion as well.. Hmm 🤔

Hipska:
@piRGoif @Molkobain Do you have any update? What route should we take? Show content, show meta (size), show both?

piRGoif:
Hello,
Sorry I moved this PR to the "pending contributor" state and it was a bad idea...
I'll change the state again so that this PR will be processed on next reviews !

piRGoif:
Reviewed during PR technical review : actually the main problem is that we are getting raw data and display them as is... This will create problems also for HTML attributes for example...
We don't have any helper method for now that could help displaying raw values from a MySQL table (datasynchro data table)...

Can you also add a \utils::EscapeHtml call on value for attributes other than AttributeBlob ?
We all agreed displaying base64 data won't be very useful. What can be displayed instead ? I suggested the size, but there might be other metadata brought from the DB ? Having a download link would be ideal but that would mean create a new endpoint...

Hipska:
Hi @piRGoif I added the HTML escaping..

piRGoif:
Thanks Thomas !
Would you also replace the base64 value by something else, maybe the size ? To display the value properly you could use `\SetupUtils::HumanReadableSize`

Hipska:
I have done the change, but I feel like we still need to show the first x bytes (encoded)

piRGoif:
> I have done the change

Thanks !

> I feel like we still need to show the first x bytes (encoded)

Can you give some use cases where it would be useful ? We didn't find any when discussing about the feature ?

piRGoif (approved):

Hipska:
Yeah, what is the use case of this whole feature.. It gives the ability for the user to see if the data makes any sense or not.

But if you all agree that only showing something as follows is good enough, then that's fine by me.
> _234 kB_

piRGoif:
:D
I understand !
There's no doubt this is far from ideal now, and still after this PR. 
But your PR improves things a bit, for us it's better now !
The PR will be reviewed again next week.

Hipska:
Maybe it should say something like this instead of only the size?
> _Binary data (234kB)_

or

> _BLOB (234kB)_

piRGoif:
Very nice idea, OK for me ! I would maybe prefer "binary data", as "blob" is more of an implementation detail ?
Can you change it before the review on thursday ?

Hipska:
I would have used the dictionary for `Core:AttributeBlob`. But anyway, applied the requested change.

piRGoif:
Oh, I didn't knew `Core:AttributeBlob` : you're right, it's better to use it ! Sorry :/
Can you do this last change ?

Hipska:
It was also you that suggested to use "blob" anyway 😛 
> Hello,
> Do you need to see the base64 content ? Or should we replace it with the mention "blob" and value size maybe ?

Molkobain:
Accepted in technical review.

Molkobain:
Hello Thomas, next review in a few weeks, do you mind updating the picture in the OP to reflect the new result? It will ease convincing. Thanks!

synchro/synchrodatasource.class.inc.php
```diff
@@ -2818,12 +2818,16 @@ function DisplayBareProperties(WebPage $oPage, $bEditMode = false, $sPrefix = ''
...
+				'data'    => array('label' => Dict::S('UI:Form:Value'), 'description' => ''),
 			);
 			$aRows = array();
 			foreach ($aData as $sKey => $value) {
-				$aRows[] = array('attcode' => $sKey, 'data' => $value);
+				if ((strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false))
+				{
+					$aRows[] = array('attcode' => $sKey, 'data' => sprintf('<i>%s (%s)</i>', Dict::S('Core:AttributeBlob'), utils::BytesToFriendlyFormat(strlen($value))));
+				}
+				else $aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
```
Molkobain:
```suggestion
				else {
					$aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
				}
```

One line if statement are not part of the code conventions.

synchro/synchrodatasource.class.inc.php
```diff
@@ -2818,12 +2818,16 @@ function DisplayBareProperties(WebPage $oPage, $bEditMode = false, $sPrefix = ''
...
 			$aHeaders = array(
-				'attcode' => array('label' => 'Attribute Code', 'description' => ''),
-				'data'    => array('label' => 'Value', 'description' => ''),
+				'attcode' => array('label' => Dict::S('UI:Form:Property'), 'description' => ''),
+				'data'    => array('label' => Dict::S('UI:Form:Value'), 'description' => ''),
 			);
 			$aRows = array();
 			foreach ($aData as $sKey => $value) {
-				$aRows[] = array('attcode' => $sKey, 'data' => $value);
+				if ((strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false))
```
Molkobain:
```suggestion
				if (strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false)
```

Molkobain (changes requested):

synchro/synchrodatasource.class.inc.php
```diff
@@ -2818,12 +2818,18 @@ function DisplayBareProperties(WebPage $oPage, $bEditMode = false, $sPrefix = ''
...
 			$aRows = array();
 			foreach ($aData as $sKey => $value) {
-				$aRows[] = array('attcode' => $sKey, 'data' => $value);
+				if (strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false)
+				{
+					$aRows[] = array('attcode' => $sKey, 'data' => sprintf('<i>%s (%s)</i>', Dict::S('Core:AttributeBlob'), utils::BytesToFriendlyFormat(strlen($value))));
+				}
+				else {
+					$aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
+				}
```
Molkobain, line 2832:
```suggestion
				if (strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false) {
					$aRows[] = array('attcode' => $sKey, 'data' => sprintf('<i>%s (%s)</i>', Dict::S('Core:AttributeBlob'), utils::BytesToFriendlyFormat(strlen($value))));
				} else {
					$aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
				}
```

Molkobain:
_(Nevermind the notification, I thought I would be able to commit it directly myself. I'll do it aftermerge)_

Molkobain (approved):

Hipska:
Didn't got any notifications. If you still have suggestions, you can add them here..

synchro/synchrodatasource.class.inc.php
Molkobain:
```suggestion
				if (strpos(CMDBSource::GetFieldType($sSQLTable, $sKey), 'blob') !== false) {
					$aRows[] = array('attcode' => $sKey, 'data' => sprintf('<i>%s (%s)</i>', Dict::S('Core:AttributeBlob'), utils::BytesToFriendlyFormat(strlen($value))));
				} else {
					$aRows[] = array('attcode' => $sKey, 'data' => utils::EscapeHtml($value));
				}
```

(1 bracket inline, 3 on a new line)

Hipska: I already found your previous suggestion about the style a strange one 🙂

Molkobain: Indeed 😅

Molkobain:
Hello Thomas,

This was accepted in functional review. Exceptionally as the rest of the feature (#269) was merged in support/3.0 we would like to merge this one in support/3.0 as well, can you rebase the PR?

Thanks,
Guillaume

Hipska:
These two are not dependant on each other.. It was already possible to sync binary data by directly connecting the datasynchro tables..

Molkobain:
So you won't rebase it?

piRGoif:
> These two are not dependant on each other.. It was already possible to sync binary data by directly connecting the datasynchro tables..

Well, I think Guillaume meant that we would like to group those 2 blob sync fixes in the same release :)
Are you ok to rebase this pr branch ?

Molkobain:
Indeed, but it's too late anyway, 3.0.2 is on code freeze.
Having it on the support/3.0 should a good though.

Molkobain:
What is the status on this PR, I see it's on "pending contributor update". Did we just wanted to rebase on support/3.0 so we could merge it in the 3.0.2? If so, that train departed a long time ago, we could merge it in support/3.0 any time now.

@piRGoif @Hipska any memories?

Hipska:
I see in conversation history it was requested to base on 3.0, which I did. 

Oh, but also seem to have reverted.

piRGoif:
Hello,
Indeed we asked to rebase the branch on support/3.0.
Currently the branch start is 42599eae which is in develop...
Can you do the rebase Thomas ?

Molkobain:
Once rebased we can merge :)

Hipska:
Will look into it..

Hipska:
Whoa, that was a hard one, but it worked!

piRGoif:
Well done ! 
Merging !

PR #288, opened 2022-05-06
🌐 Add dictionary items for integrity checks

Hipska:
Added dictionary items for:

- Leading message for `DBObject::CheckValue` result
- Leading message for `DBObject::CheckConsistency` result

Also cleaned up code for synchro replica slave attribute checks.

### DIFF
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -2265,7 +2265,7 @@ abstract class DBObject implements iDisplay
 		if ($res !== true)
 		{
 			// $res contains the error description
-			$this->m_aCheckIssues[] = "Consistency rules not followed: $res";
+			$this->m_aCheckIssues[] = Dict::Format('Core:CheckConsistencyError', $res);
 		}
 
 		// Synchronization: are we attempting to modify an attribute for which an external source is master?
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -2279,12 +2279,10 @@ abstract class DBObject implements iDisplay
 				if ($iFlags & OPT_ATT_SLAVE)
 				{
 					// Note: $aReasonInfo['name'] could be reported (the task owning the attribute)
-					$oAttDef = MetaModel::GetAttributeDef(get_class($this), $sAttCode);
-					$sAttLabel = $oAttDef->GetLabel();
 					if (!empty($aReasons))
 					{
-						// Todo: associate the attribute code with the error
-						$this->m_aCheckIssues[] = Dict::Format('UI:AttemptingToSetASlaveAttribute_Name', $sAttLabel);
+						$sAttLabel = $this->GetLabel($sAttCode);
+						$this->m_aCheckIssues[] = Dict::Format('UI:AttemptingToSetASlaveAttribute_Name', $sAttLabel, $sAttCode);
 					}
 				}
 			}
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -2249,8 +2249,9 @@ abstract class DBObject implements iDisplay
 		foreach($aChanges as $sAttCode => $value) {
 			$res = $this->CheckValue($sAttCode);
 			if ($res !== true) {
+				$sAttLabel = $this->GetLabel($sAttCode);
 				// $res contains the error description
-				$this->m_aCheckIssues[] = "Unexpected value for attribute '$sAttCode': $res";
+				$this->m_aCheckIssues[] = Dict::Format('Core:CheckValueError', $sAttLabel, $sAttCode, $res);
 			}
 
 			$this->DoCheckLinkedSetDuplicates($sAttCode, $value);
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -2028,7 +2028,7 @@ abstract class DBObject implements iDisplay
 	 *
      * @overwritable-hook You can extend this method in order to provide your own logic.
      * 
-	 * @return bool 
+	 * @return true|string true if successful, the error description otherwise
 	 */
 	public function CheckConsistency()
 	{
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * @copyright   Copyright (C) 2010-2021 Combodo SARL
+ * @copyright   Copyright (C) 2010-2022 Combodo SARL
  * @license     http://opensource.org/licenses/AGPL-3.0
  */
 
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -2028,7 +2028,7 @@ abstract class DBObject implements iDisplay
 	 *
      * @overwritable-hook You can extend this method in order to provide your own logic.
      * 
-	 * @return true|string true if successful, the error description otherwise
+	 * @return bool 
 	 */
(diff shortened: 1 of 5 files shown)


core/dbobject.class.php
```diff
@@ -2249,8 +2249,9 @@ public function DoCheckToWrite()
 		foreach($aChanges as $sAttCode => $value) {
 			$res = $this->CheckValue($sAttCode);
 			if ($res !== true) {
+				$sAttLabel = $this->GetLabel($sAttCode);
 				// $res contains the error description
-				$this->m_aCheckIssues[] = "Unexpected value for attribute '$sAttCode': $res";
+				$this->m_aCheckIssues[] = Dict::Format('Core:CheckValueError', $sAttLabel, $sAttCode, $res);
```
Molkobain, line 2254: Why put 3 parameters if only 2 are used in the dict. entry? This will lead to having the other parameter displayed after the string, not very slick.

core/dbobject.class.php
```diff
@@ -2278,12 +2279,10 @@ public function DoCheckToWrite()
...
 				{
 					// Note: $aReasonInfo['name'] could be reported (the task owning the attribute)
-					$oAttDef = MetaModel::GetAttributeDef(get_class($this), $sAttCode);
-					$sAttLabel = $oAttDef->GetLabel();
 					if (!empty($aReasons))
 					{
-						// Todo: associate the attribute code with the error
-						$this->m_aCheckIssues[] = Dict::Format('UI:AttemptingToSetASlaveAttribute_Name', $sAttLabel);
+						$sAttLabel = $this->GetLabel($sAttCode);
+						$this->m_aCheckIssues[] = Dict::Format('UI:AttemptingToSetASlaveAttribute_Name', $sAttLabel, $sAttCode);
```
Molkobain, line 2285: Same comment about the extra param not being used in the dict. entry

Molkobain (changes requested):

Molkobain:
Skip technical review as it's just new dict entries.

core/dbobject.class.php
Hipska: See the removed ToDo, there seems to be the intention of Combodo to add both in the message. So now users can already use it with their extension and/or Combodo only needs to update the respective dictionary item.

core/dbobject.class.php
Hipska: Same here, I think it might be useful in the future, and then there is only a dictionary item change needed..

Hipska: > This will lead to having the other parameter displayed after the string, not very slick.

Whut? https://3v4l.org/pqZ5I

core/dbobject.class.php
Molkobain: For the PR either update the corresponding dict entry in EN / NL or remove this change.
We rather go all the way through an enhancement or keep as-is as the result in the UI will not be great : `The field start_date is not writable because it is mastered by the data synchronization. Value not set. - 2022-05-06 14:24:00`

Hipska: I really don't see how you get to that result.. https://3v4l.org/pqZ5I

Molkobain: I'm a bit confused, previously when we put a parameter in Dict::Format that was not used in the dict entry, it was appending it to the dict entry like on my previous comment. But I can't get this behavior now on a 3.0. Not sure if it is because of a PHP version change or something else.

Hipska: No, that only happens if a dict entry doesn't exist. But in this PR I clearly created those dict entries.

Molkobain: Anyway, either we update the dict entry or leave the TODO  as is; with the PR this is still a "TODO" as the dict entry does not use it.

core/dbobject.class.php
Molkobain: _See discussion on the other thread_

core/dbobject.class.php
Hipska: I checked the code and for `Dict::Format` it hasn't been changed for 11 years. So we can safely assume it has been always like this.

Molkobain: > I checked the code and for `Dict::Format` it hasn't been changed for 11 years. So we can safely assume it has been always like this.

Yeah, I might have seen this on dict entries for which I had not ran a setup yet I guess, my bad sorry! 👀 (as you said in https://github.com/Combodo/iTop/pull/288#discussion_r866796716)

> Anyway, either we update the dict entry or leave the TODO as is; with the PR this is still a "TODO" as the dict entry does not use it.

This is not resolved yet IMHO

Hipska: I don't know how Combodo wants this entry to be formatted..

dictionaries/en.dictionary.itop.core.php
```diff
@@ -29,6 +29,8 @@
 	'Core:UnknownObjectTip' => 'The object could not be found. It may have been deleted some time ago and the log has been purged since.',
 
 	'Core:UniquenessDefaultError' => 'Uniqueness rule \'%1$s\' in error',
+	'Core:CheckConsistencyError' => 'Consistency rules not followed: %1$s',
+	'Core:CheckValueError' => 'Unexpected value for attribute \'%1$s\': %3$s',
```
Molkobain: This changes the printed string, it's now displaying the attribute label instead of the attribute code.
Proposed change so it displays everything:
```suggestion
	'Core:CheckValueError' => 'Unexpected value for attribute \'%1$s\' (%2$s) : %3$s',
```

core/dbobject.class.php
Molkobain: `'The field %1$s (%2$s) is not writable because it is mastered by the data synchronization. Value not set.~~'`

Molkobain (changes requested):
Accepted in functional review once the 2 dict entries will be adjusted.

core/dbobject.class.php
Molkobain: _Note: Just to be clear we'll wait on that entry update before merging the PR_

core/dbobject.class.php
```diff
@@ -2028,7 +2028,7 @@ public function CheckValue($sAttCode, $value = null)
 	 *
      * @overwritable-hook You can extend this method in order to provide your own logic.
      * 
-	 * @return bool 
+	 * @return true|string true if successful, the error description otherwise
```
Molkobain: Why this change?

Hipska: See the commit message. My IDE was complaining return value would always be `false` if it isn't `true`. Which is not  correct, as in that case it's a string.

piRGoif: Indeed, the caller \DBObject::DoCheckToWrite does use yes or a string, as of CheckValue()...

piRGoif: I suggest we do a separate commit to update this PHPDoc ? 
@Molkobain if this modification is removed, the PR can be merged right ?

Molkobain: Yep

piRGoif: PHPDoc updated directly in iTop support/2.7 branch, see 424e2a57
@Hipska can you remove this modification, and we'll be good to merge ?

piRGoif (approved):

PR #289, opened 2022-05-06
Feature/5135 impersonated changelogs

odain-cbd:
**= N°5135 - Impersonate: history of changes versus log entries =**
Impersonation feature is nice and powerfull for supportability. this PR aims at tracking any edition made under impersonation.

when user 'John Doe' is impersonated by 'Isaac Asimov', any edition history from UI will indicate ''Isaac Asimov on behalf of John Doe" in history of changes (either private/public or activity logs).

in case of impersonation, there will not be any user avatar nor friendly name. an acronym will be also computed (utils::ToAcronym()).

**corner case:**
if an extension (ie approval extensions) set a specific current change user information (ie "John Doe Approved on behalf of Jane Doe"), activity log will contain both extension and impersonation mentions (ie "'Isaac Asimov on behalf of John Doe (John Doe Approved on behalf of Jane Doe)").

PR #290, opened 2022-05-06
N°5135 - Impersonate: history of changes versus log entries

odain-cbd:
= N°5135 - Impersonate: history of changes versus log entries =
Impersonation feature is nice and powerfull for supportability. this PR aims at tracking any edition made under impersonation.

when user 'John Doe' is impersonated by 'Isaac Asimov', any edition history from UI will indicate ''Isaac Asimov on behalf of John Doe" in history of changes (either private/public or activity logs).

in case of impersonation, there will not be any user avatar nor friendly name. an acronym will be also computed (utils::ToAcronym()).

corner case:
if an extension (ie approval extensions) set a specific current change user information (ie "John Doe Approved on behalf of Jane Doe"), activity log will contain both extension and impersonation mentions (ie "'Isaac Asimov on behalf of John Doe (John Doe Approved on behalf of Jane Doe)").

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -1685,7 +1685,7 @@ class AttributeLinkedSet extends AttributeDefinition
 					{
 						if ($sObjClass == $this->GetLinkedClass())
 						{
-							// Simplify the output if the exact class could be determined implicitely 
+							// Simplify the output if the exact class could be determined implicitely
 							continue;
 						}
 					}
@@ -2007,7 +2007,7 @@ class AttributeLinkedSet extends AttributeDefinition
 					{
 						if ($sObjClass == $this->GetLinkedClass())
 						{
-							// Simplify the output if the exact class could be determined implicitely 
+							// Simplify the output if the exact class could be determined implicitely
 							continue;
 						}
 					}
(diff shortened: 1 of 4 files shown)


core/attributedef.class.inc.php
```diff
@@ -4587,7 +4587,12 @@ public function MakeRealValue($proposedValue, $oHostObj)
 			{
 				if (strlen($proposedValue) > 0)
 				{
-					$oCaseLog->AddLogEntry($proposedValue);
+					//N°5135 - add impersonation information in caselog
+					if (UserRights::IsImpersonated()){
+						$oCaseLog->AddLogEntry($proposedValue, '', UserRights::GetConnectedUserId());
```
odain-cbd: I have been conservative to avoid any regression. but it means we may have to adapt extension that use AddLogEntry as well in case of Impersonation.

another approach could be to adapt AddLogEntry implementation in case of impersonation. R&D could see it as a cleaner design? 
once decided I will add a test on this part as well...

piRGoif:
This is replacing #289 right ? Is this a new implementation ?

piRGoif:
Oh ok I was confused by the comments displayed in 289 ! I assumed you made a mistake by rebasing in 289, and create this new PR to replace.

piRGoif:
In the description you're pointing out the existing behavior issues. Can you describe as well the implemented new behavior(s) ?

PR #291, opened 2022-05-17
N°5160 - Hiding a DashboardAttribute on the fly does not work

v-dumas:
Allowing a DashboardAttribute to be hidden by an iTop object method such as GetAttributeFlags


piRGoif:
Although this is still a work in progress, I'm adding this PR to our project so that we don't loose track of it !

PR #295, opened 2022-05-27
N°4280 Handle modules with non existing model.*.php files imported

piRGoif:
A module containing XML must add its corresponding model.php file in its "datamodel" section.

So for example, this module containing those files : 
datamodel.4280-datamodel-define.xml
module.4280-datamodel-define.php

In the module.4280-datamodel-define.php file you must have : 

```
		'datamodel'            => array(
			'model.4280-datamodel-define.php',
		),
```

Otherwise the XML compilation made by iTop won't be present in iTop.

But sometimes the module XML won't generate a model file, or another module will remove the XML nodes defined... 
This was crashing iTop.

Now the compiler, after compilation and before generating the env-*/autoload.php will check if the file really exists on disk.
If not : 

* if isDevEnv then an exception is throwned
* else a \SetupLog::Warning call will be made

In both cases, the context is filled with moduleId, moduleLocation, includedFile

### DIFF
--- a/setup/compiler.class.inc.php
+++ b/setup/compiler.class.inc.php
@@ -617,7 +617,20 @@ EOF;
 			// files to include (PHP datamodels)
 			foreach($oModule->GetFilesToInclude('business') as $sRelFileName)
 			{
-				$aDataModelFiles[] = "MetaModel::IncludeModule(MODULESROOT.'/$sRelativeDir/$sRelFileName');";
+				if (file_exists("{$sTempTargetDir}/{$sRelativeDir}/{$sRelFileName}")) {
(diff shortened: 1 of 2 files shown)


Hipska:
So that means the `model.*.php` file doesn't need to be present in the array anymore?

piRGoif:
> So that means the `model.*.php` file doesn't need to be present in the array anymore?

No, this is to address the cases where you add the model.php file but the compiler doesn't generate it.
I've created another PR so that you won't need to add model.php file yourself : #296

piRGoif:
PR technical review : remove the warning log (too verbose), instead only does something when isDevEnv.
Make the error message clearer by talking about the 'datamodel' key.

piRGoif:
Functional review : OK for 3.1 !
Don't forget to align datamodels/2.x modules after merge (no model.php on disk and in the module.php file)

PR #296, opened 2022-05-27
N°4875 : compiler now automatically adds the model.php file

piRGoif:
Before this fix, when creating a module for iTop that was creating PHP code (an iTop class for example, a snippet, ...) it was mandatory to add the model.php file in your module.php file ("datamodel" key). If you forgot this, then the compilation was completed OK but the result code wasn't included in iTop.

Now the compiler automatically adds the model.php file to the included files.

### DIFF
--- a/setup/compiler.class.inc.php
+++ b/setup/compiler.class.inc.php
@@ -592,15 +592,18 @@ EOF;
 				{
 					// Write the code into the given module as model.<module>.php
 					//
-					$sResultFile = $sTempTargetDir.'/'.$sRelativeDir.'/model.'.$sModuleName.'.php';
+					$sModelFileName = 'model.'.$sModuleName.'.php';
+					$sResultFile = $sTempTargetDir.'/'.$sRelativeDir.'/'.$sModelFileName;
 					$this->WritePHPFile($sResultFile, $sModuleName, $sModuleVersion, $sCompiledCode);
+					// In case the model file wasn't present in the module file, we're adding it ! (N°4875)
+					$oModule->AddFileToInclude('business', $sModelFileName);
 				}
 				else
 				{
 					// Write the code into core/main.php
 					//
 					$this->sMainPHPCode .=
-					<<<EOF
+						<<<EOF
 /**
  * Data model from the delta file
  */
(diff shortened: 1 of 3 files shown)


Hipska:
After this, the key named `datamodel` doesn't reflect what it is used for anymore. Maybe add an `include` key or something (to include those files if wanted) (which would do exactly the same "datamodel" is doing atm)

piRGoif:
> After this, the key named datamodel doesn't reflect what it is used for anymore.

Not sure to understand what you mean ? Can you elaborate on your point ? 
Today the `datamodel` key doesn't list all the files used by the compiler : the datamodel.*.xml file isn't listed in here.

Hipska:
It lists the files to be used for inclusion. Usually they are (generated) datamodel files. Since the generated datamodel doesn't need to be in there and might only be used for other types of files, the key name doesn't reflect it's goal anymore.

piRGoif:
It was already the case : if you needed to include any file for your php code then it was mandatory to include it there.

For example the 'datamodel' key [in module.itop-system-information.php](https://github.com/Combodo/itop-system-information/blob/d00580e50cd8d583d1bbd9d6fb2871f6b2cd6ae3/module.itop-system-information.php#L31) lists 5 files, only the first one is about datamodel (and it is indeed the model.php file) !

You are totally right, but for me this is out of scope of this PR.

Hipska:
Exactly, system-information is a perfect example of what I'm trying to explain. Imagine how that section would look like without the `model.itop-system-information.php` in it.

I agree it might be out of scope for this PR, I just wanted to point this out to you, in order to get it on the table internally for consideration. (It could simply even be an alias for the current `datamodel` key..)

piRGoif:
Ok :) I'll add it to our dev team meeting agenda !

piRGoif:
Hello,
We discussed with the rest of the dev team the adding of a 'include' alias key to the existing 'datamodel' key. We agree of course this isn't very nice for now ! Some little things that were said during the meeting : 

* for compatibility reasons of course we cannot remove the 'datamodel' key before at least next iTop LTS version
* order in which the files are added in the iTop env autoload is very important, so a rule should be chosen when there are both 'datamodel' and 'include' keys present in a module.php file
* we are aiming at using a single autload file for new modules (that means the naming inconsistency still exists but with only 1 file). Note that the module template generator produces an autoload.php for example
* if present, this module's autoload.php file might also be automatically added to the iTop env autoload ?

Combodo is not willing to make the effort for now... This is something you could submit as a PR though !

Hipska:
I love the idea to also automatically add the autoload.php! 👍

piRGoif:
Technical review : OK !

piRGoif:
> I love the idea to also automatically add the autoload.php! 👍

I'll suggest adding that on functional review next week !

piRGoif:
Functional review : OK for 3.1 !
Don't forget to align datamodels/2.x modules after merge (no model.php on disk and in the module.php file)

PR #297, opened 2022-06-07
N°5216 Invalid message-id when sending notification using cron on system with a specific locale set

piRGoif:
When sending notification using cron, we had one french configured VM throwing this error : 
`Error: Invalid ID given <iTop_UserRequest_41_1654590867,049049@4802a49edd657824f11970fb7330fbb0-production.openitop.org>`

The timestamp used was indeed locale dependent.
This PR fixes this behavior by removing the locale dependency using a better printf format (see https://www.php.net/manual/fr/function.sprintf.php)

### DIFF
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -328,8 +328,12 @@ class ActionEmail extends ActionNotification
 			$sBody = MetaModel::ApplyParams($this->Get('body'), $aContextArgs);
 
 			$oObj = $aContextArgs['this->object()'];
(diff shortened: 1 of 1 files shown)


Hipska (approved):

piRGoif:
Accepted in technical review.
But we should check if `%f` is used elsewhere ?

piRGoif:
Just used in \EmailReplica::MakeMessageId (combodo-email-synchro)
Will do a PR for this module, too

piRGoif:
Functional review : OK for 2.7.8 / 3.0.2

Molkobain (approved):

PR #298, opened 2022-06-08
Feature/merge 3 0 develop

bdalsass:


Hipska:
Whoa, why not do a regular merge like was usually done in the past? This misses a lot of historical information..

piRGoif:
I agree :)
Benjamin tried something, this is under discussion

Hipska:
Okay, but it's already merged 😕

piRGoif:
The PR is merged yes but in a git sense there are no merge between support/3.0 and develop : trying to merge right now produces dozens of conflicts...
We'll figure all of this out and keep everyone posted here.

Hipska:
Yes there are merges between support/3.0 and develop. For example one you did in 3d541b2a2da36323cd74c1aa0ec614c13b9b415a.

piRGoif:
I meant usually we did regular merge commits like the one you pointed out.
This PR adds a commit in develop but the branches remained with unmerged changes anyway...

PR #299, opened 2022-06-13
:bug: N°5369 - Fix BrowseBrick tree "opening_target" mode for "self" and "new" values

vbkunin:
Hi! I faced with a small bug in the Browse Brick tree more. The level actions didn't work when `<opening_target>` was set in `self` or `new` value. The reason is that the `href` attribute is attached to the `span` element and there are no listeners on it to handle click events. Modal mode works fine because it's based on the bootstrap modal functionality.
I swapped the `span` and `a` elements and made some changes to the scss/css to keep the current look.

### DIFF
--- a/datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
+++ b/datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
@@ -200,8 +200,8 @@
                 var nameElem = $('<a></a>').addClass('tree-item').text(item.name);
                 // Building node
                 $('ul[data-level-id="'+nodeId+'"]').append(liElem);
-                aElem.append(nameElem);
-                liElem.append(aElem);
+                spanElem.append(nameElem);
+                liElem.append(spanElem);
 
                 // Delegating a click on <span> to its child <a> element
                 // N°5396 - Fix BrowseBrick tree "opening_target" mode for "self" and "new" values
--- a/datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
+++ b/datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
@@ -196,8 +196,8 @@
 				var url = '';
 
                 var liElem = $('<li></li>').addClass('list-group-item');
-                var aElem = $('<span></span>').addClass('tree-item-wrapper').attr('data-item-id', item.id).attr('data-level-alias', item.level_alias);
-                var nameElem = $('<a></a>').addClass('tree-item').text(item.name);
+                var aElem = $('<a></a>').addClass('tree-item-wrapper').attr('data-item-id', item.id).attr('data-level-alias', item.level_alias);
+                var nameElem = $('<span></span>').addClass('tree-item').text(item.name);
                 // Building node
                 $('ul[data-level-id="'+nodeId+'"]').append(liElem);
                 aElem.append(nameElem);
--- a/datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
+++ b/datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
@@ -196,7 +196,7 @@
 				var url = '';
 
                 var liElem = $('<li></li>').addClass('list-group-item');
-                var aElem = $('<span></span>').addClass('tree-item-wrapper').attr('data-item-id', item.id).attr('data-level-alias', item.level_alias);
+                var spanElem = $('<span></span>').addClass('tree-item-wrapper').attr('data-item-id', item.id).attr('data-level-alias', item.level_alias);
                 var nameElem = $('<a></a>').addClass('tree-item').text(item.name);
                 // Building node
                 $('ul[data-level-id="'+nodeId+'"]').append(liElem);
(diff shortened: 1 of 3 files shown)


piRGoif:
Hello Vladimir, thanks for this contribution ! 

I applied this XML to a standard iTop on the develop branch : 

```xml
  <module_designs>
    <module_design xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="itop-portal" xsi:type="portal" _delta="must_exist">
      <bricks>
        <brick id="services" xsi:type="Combodo\iTop\Portal\Brick\BrowseBrick" _delta="must_exist">
          <levels>
            <level id="1" _delta="must_exist">
              <levels>
                <level id="1" _delta="must_exist">
                  <levels>
                    <level id="1" _delta="must_exist">
                      <actions>
                        <action id="create_from_this" xsi:type="create_from_this" _delta="must_exist">
                          <opening_target _delta="define">self</opening_target>
                        </action>
                      </actions>
                    </level>
                  </levels>
                </level>
              </levels>
            </level>
          </levels>
          <browse_modes>
            <default _delta="force">tree</default>
          </browse_modes>
        </brick>
      </bricks>
    </module_design>
  </module_designs>
</itop_design>
```

And easily reproduced the problem, thanks !
Indeed with your fix the link now works as expected, but I'm not sure if this is the best fix, and won't cause regressions (changing the DOM structure).
We'll see during the technical review next week !

Molkobain:
Discussed during technical review, @Molkobain will review it for next time.

Molkobain (approved):

Molkobain:
Nested anchors are illegal in HTML (see [here](https://www.w3.org/TR/html401/struct/links.html#h-12.2.2) and [here](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element)).

With your proposal, in the case the Service description is in HTML and contains an hyperlink (which we encountered in several occasions), we will have those nested anchors.

That being said, even though nested anchors were not working in previous versions of browsers, this seems to work on both Chrome and Firefox (to date versions) now.

With that in mind and the fact that the portal option is indeed not working as-is, we think that this should be considered. The PR will be discussed further more in functional review next week.

In the mean time I'll create the corresponding bug in our bug tracker.

Molkobain:
Hello Vladimir,

We discussed this during today's review; we don't really want to nest anchors as it is still prohibited by the HTML specs as mentioned on the previous post.

What we would agree to in order to restore that feature is to add an `onclick=""` attribute on the element with the corresponding JS code to open the href in either the current window or a new tab.

Cheers,
Guillaume

vbkunin:
Hi!

It seems that nested anchors are not only a problem of the proposed solution, since they also exist in the Mosaic view even at several levels (service/subcategory description and the hamburger menu with its items).

I figured that the code of the helper functions SetActionUrl() and SetActionOpeningTarget() assumes that the <a> element should be passed as an argument. And in all cases, this is true, except for the Tree view where <span> is passed. I'm not sure if the RFCs allow the use of href and target attributes for the <span> element, I've never seen this, and it's confusing in my opinion.

As a workaround to the current problem, I suggest adding href and target to <a> and hanging an onclick handler on <span> to delegate these events to <a>. I think a more complete solution requires more complex refactoring outside of this PR.

datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
```diff
@@ -203,6 +203,15 @@
...
                 liElem.append(aElem);
 
+                // Delegating a click on <span> to its child <a> element
+                // N°5396 - Fix BrowseBrick tree "opening_target" mode for "self" and "new" values
+                aElem.click(function (event) {
+                    var a = $(this).children('a.tree-item')[0];
+                    if (a !== event.target) {
+                        a.click();
+                    }
+                });
```
Molkobain:
```suggestion
                spanElem.on('click', function (event) {
                    var aElem = $(this).children('a.tree-item')[0];
                    if (aElem !== event.target) {
                        aElem.trigger('click');
                    }
                });
```

This modification to better match our conventions (use of `.on()` and `.trigger()` instead of the shortcut).
The renaming of the `aElem` into `spanElem` on lines 199, 203 and 204.
And we will be good to go!

datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
```diff
@@ -203,6 +203,15 @@
                 aElem.append(nameElem);
                 liElem.append(aElem);
```
Molkobain:
```suggestion
                spanElem.append(nameElem);
                liElem.append(spanElem);
```

Molkobain (changes requested):

Molkobain:
Hello Vladimir,

Thanks for the changes, your new approach seems good regarding the whole refactoring that it would require. Once suggestions are applied it will good for me.

Cheers,
Guillaume

Molkobain:
Hello @vbkunin , any update?

datamodels/2.x/itop-portal-base/portal/templates/bricks/browse/mode_tree.html.twig
vbkunin: It looks like the `aElem` on lines 255 and 232 should also be renamed.

vbkunin:
Hello @Molkobain ! Sorry for the long wait, I made changes.

Molkobain (approved):

Molkobain:
No worries Vladimir, thanks for the changes!

PR #301, opened 2022-06-15
N°5235 - Add tmp dir is writable check in setup

piRGoif:
If the tmp dir isn't writable then the compilation will crash, with a stack trace and no explicit information.

Example of the setup.log : 
```
2022-09-07 15:13:07 | Info    |       | ##### STEP copy start | SetupLog
2022-09-07 15:13:07 | Info    |       | ##### STEP copy duration: 0s | SetupLog
2022-09-07 15:13:07 | Info    |       | operation 'async_action', peak memory usage. 2729008 | SetupLog
2022-09-07 15:13:07 | Info    |       | ##### STEP compile start | SetupLog
2022-09-07 15:13:07 | Info    |       | Compiling data model. | SetupLog
2022-09-07 15:13:07 | Info    |       | ----> Entering maintenance mode | SetupLog
2022-09-07 15:13:20 | Error   |       | An exception occurred: movedir: the source directory '' is not a valid directory or cannot be read at line 808 in file C:\Dev\wamp64\www\itop-27\setup\setuputils.class.inc.php | SetupLog
2022-09-07 15:13:20 | Ok      |       | Call stack: | SetupLog
2022-09-07 15:13:20 | Ok      |       | #0 C:\Dev\wamp64\www\itop-27\setup\compiler.class.inc.php(186): SetupUtils::movedir(...) | SetupLog
2022-09-07 15:13:20 | Ok      |       | #1 C:\Dev\wamp64\www\itop-27\setup\applicationinstaller.class.inc.php(624): MFCompiler->Compile(...) | SetupLog
2022-09-07 15:13:20 | Ok      |       | #2 C:\Dev\wamp64\www\itop-27\setup\applicationinstaller.class.inc.php(292): ApplicationInstaller::DoCompile(...) | SetupLog
2022-09-07 15:13:20 | Ok      |       | #3 C:\Dev\wamp64\www\itop-27\setup\wizardsteps.class.inc.php(2464): ApplicationInstaller->ExecuteStep(...) | SetupLog
2022-09-07 15:13:20 | Ok      |       | #4 C:\Dev\wamp64\www\itop-27\setup\ajax.dataloader.php(177): WizStepSummary->AsyncAction(...) | SetupLog
2022-09-07 15:13:20 | Info    |       | ##### STEP compile duration: 12.39s | SetupLog
2022-09-07 15:13:20 | Info    |       | operation 'async_action', peak memory usage. 36346328 | SetupLog
```

This PR aims to check on the first step if the temp dir is writable and display a clear message to the iTop admin.

### DIFF
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -1817,7 +1817,7 @@ JS
 	 *
 	 * @param string[] $aWritableDirs list of dirs to check, relative to APPROOT (for example : `['log','conf','data']`)
 	 *
-	 * @return array full path as key, CheckResult error as value
+	 * @return array<string, \CheckResult> full path as key, CheckResult error as value
 	 *
 	 * @uses \CheckResult
 	 * @uses \is_dir()
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -96,7 +96,7 @@ class SetupUtils
 		if (is_writable($sTmpDir)) {
 			$aResult[] = new CheckResult(CheckResult::INFO, "The temp directory is writable by the application.");
 		} else {
-			$aResult[] = new CheckResult(CheckResult::ERROR, "The temp directory <b>'".$sTmpDir."'</b> is not writable for the application. Change its permission or use another dir (sys_temp_dir option in php.ini).");
+			$aResult[] = new CheckResult(CheckResult::ERROR, "The temp directory <b>'".$sTmpDir."'</b> is not writable by the application. Change its permission or use another dir (sys_temp_dir option in php.ini).");
 		}
 
 		$aMandatoryExtensions = array(
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -90,7 +90,7 @@ class SetupUtils
 		// Check the common directories
 		$aWritableApprootDirsErrors = self::CheckWritableDirs(array('log', 'env-production', 'env-production-build', 'conf', 'data'));
 		$aResult = array_merge($aResult, $aWritableApprootDirsErrors);
-		// check temp dir (N°5235)
+		// Check temp dir (N°5235)
 		$sTmpDir = static::GetTmpDir();
 		clearstatcache(true, $sTmpDir);
 		if (is_writable($sTmpDir)) {
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -1819,7 +1819,6 @@ JS
 	 *
 	 * @return array<string, \CheckResult> full path as key, CheckResult error as value
 	 *
-	 * @uses \CheckResult
 	 * @uses \is_dir()
 	 * @uses \is_writable()
 	 * @uses \file_exists()
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -90,7 +90,7 @@ class SetupUtils
 		// Check the common directories
 		$aWritableApprootDirsErrors = self::CheckWritableDirs(array('log', 'env-production', 'env-production-build', 'conf', 'data'));
 		$aResult = array_merge($aResult, $aWritableApprootDirsErrors);
-		// Check temp dir (N°5235)
+		// Check temp dir (N°5235) : as this path isn't under APPROOT we are doing a custom check and not using \SetupUtils::CheckWritableDirs
 		$sTmpDir = static::GetTmpDir();
 		clearstatcache(true, $sTmpDir);
 		if (is_writable($sTmpDir)) {
@@ -1805,21 +1805,31 @@ JS
 
 	public static function GetVersionManifest($sInstalledVersion)
 	{
-		if (preg_match('/^([0-9]+)\./', $sInstalledVersion, $aMatches))
-		{
+		if (preg_match('/^([0-9]+)\./', $sInstalledVersion, $aMatches)) {
 			return APPROOT.'datamodels/'.$aMatches[1].'.x/manifest-'.$sInstalledVersion.'.xml';
 		}
+
 		return false;
 	}
 
+	/**
+	 * Check paths relative to APPROOT : is existing, is dir, is writable
+	 *
+	 * @param string[] $aWritableDirs list of dirs to check, relative to APPROOT (for example : `['log','conf','data']`)
+	 *
+	 * @return array full path as key, CheckResult error as value
+	 *
+	 * @uses \CheckResult
+	 * @uses \is_dir()
+	 * @uses \is_writable()
+	 * @uses \file_exists()
+	 */
 	public static function CheckWritableDirs($aWritableDirs)
 	{
 		$aNonWritableDirs = array();
-		foreach($aWritableDirs as $sDir)
-		{
+		foreach ($aWritableDirs as $sDir) {
 			$sFullPath = APPROOT.$sDir;
-			if (is_dir($sFullPath) && !is_writable($sFullPath))
-			{
+			if (is_dir($sFullPath) && !is_writable($sFullPath)) {
 				$aNonWritableDirs[APPROOT.$sDir] = new CheckResult(CheckResult::ERROR, "The directory <b>'".APPROOT.$sDir."'</b> exists but is not writable for the application.");
 			}
 			else if (file_exists($sFullPath) && !is_dir($sFullPath))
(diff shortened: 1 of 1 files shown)


Molkobain:
Accepted in technical review. @piRGoif will test on Windows / Linux env to ensure it does not bring any regression before reviewing it in functional review.

piRGoif:
For now, \SetupUtils::CheckWritableDirs only checks under APPROOT :(

piRGoif:
Now the readonly is correctly detected !
Rebased the branch by the way O:)

piRGoif:
Error message : 

image

And success message : 

image

setup/setuputils.class.inc.php
```diff
@@ -81,16 +81,23 @@ public static function CheckPhpAndExtensions()
...
-		$aResult = array_merge($aResult, $aWritableDirsErrors);
+		$aWritableApprootDirsErrors = self::CheckWritableDirs(array('log', 'env-production', 'env-production-build', 'conf', 'data'));
+		$aResult = array_merge($aResult, $aWritableApprootDirsErrors);
+		// check temp dir (N°5235)
+		$sTmpDir = static::GetTmpDir();
+		clearstatcache(true, $sTmpDir);
+		if (is_writable($sTmpDir)) {
+			$aResult[] = new CheckResult(CheckResult::INFO, "The temp directory is writable by the application.");
+		} else {
+			$aResult[] = new CheckResult(CheckResult::ERROR, "The temp directory <b>'".$sTmpDir."'</b> is not writable for the application. Change its permission or use another dir (sys_temp_dir option in php.ini).");
```
Molkobain:
```suggestion
			$aResult[] = new CheckResult(CheckResult::ERROR, "The temp directory <b>'".$sTmpDir."'</b> is not writable by the application. Change its permission or use another dir (sys_temp_dir option in php.ini).");
```

To be consistent with the success message.

setup/setuputils.class.inc.php
```diff
@@ -81,16 +81,23 @@ public static function CheckPhpAndExtensions()
...
 		}
 
 		self::CheckPhpVersion($aResult);
 
 		// Check the common directories
-		$aWritableDirsErrors = self::CheckWritableDirs(array('log', 'env-production', 'env-production-build', 'conf', 'data'));
-		$aResult = array_merge($aResult, $aWritableDirsErrors);
+		$aWritableApprootDirsErrors = self::CheckWritableDirs(array('log', 'env-production', 'env-production-build', 'conf', 'data'));
+		$aResult = array_merge($aResult, $aWritableApprootDirsErrors);
+		// check temp dir (N°5235)
```
Molkobain:
```suggestion
		// Check temp dir (N°5235)
```

Molkobain (changes requested):

Molkobain:
Nice addition to the preliminary checks ✌

Molkobain (approved):

Molkobain:
Discussed during technical review: Why not include the directory test in the already existing `self::CheckWritableDirs()` method? Is it just for the specific error message?

piRGoif:
> Discussed during technical review: Why not include the directory test in the already existing `self::CheckWritableDirs()` method? Is it just for the specific error message?

That was what I wrote in my message dated of 5th july : this method is meant only to check dirs under approot for now... I tried first to refactor it, but it made the code complicated so I added the tmp dir check as a standalone test : simplier, much easier to read and understand.

Molkobain:
Ok thanks for the clarification, maybe we should add a comment above it to avoid someone to refactor this and put the tmp dir in the above array. What do you think?

piRGoif:
Yep, adding a phpdoc block to \SetupUtils::CheckWritableDirs would be a good idea !
I'll do it in the upcoming days

piRGoif:
Rebase branch  + force push

piRGoif:
PHPDoc added : 4a026469

piRGoif:
Processed today in technical review : good to go for functional review next week !

setup/setuputils.class.inc.php
```diff
@@ -1798,21 +1805,31 @@ public static function CheckVersion($sInstalledVersion, $sSourceDir)
...
 	}
 
+	/**
+	 * Check paths relative to APPROOT : is existing, is dir, is writable
+	 *
+	 * @param string[] $aWritableDirs list of dirs to check, relative to APPROOT (for example : `['log','conf','data']`)
+	 *
+	 * @return array full path as key, CheckResult error as value
+	 *
+	 * @uses \CheckResult
```
Hipska:
```suggestion
	 * @return \CheckResult[] full path as key, CheckResult error as value
	 *
```

Hipska (approved):

setup/setuputils.class.inc.php
piRGoif: I'm unsure this is right ? 
For me `\CheckResult[]` would mean there are no keys ? I mean we don't have an associative array ?

Hipska: It says it’s an array, nothing about keys. But that is in the description. 

To me it is more clear than just “array”. I’m only not sure about the correct syntax.

piRGoif: Not sure either :(

The phpdoc reference doesn't help as it does say nothing about associative arrays : https://docs.phpdoc.org/guide/references/phpdoc/types.html#arrays

In the PHPStan syntax I suppose the 3rd syntax is dedicated to associative arrays ? https://phpstan.org/writing-php-code/phpdoc-types#general-arrays
It says more in the Psalm reference : https://psalm.dev/docs/annotating_code/type_syntax/array_types/#generic-arrays

This article does sum up, and list indeed the type[] syntax as a standard indexed array (no keys), and describe both the phpstan and psalm syntax and array shapes : https://dev.to/suckup_de/modern-phpdoc-annotations-11d4

This psalm / phpstan syntax seems to be understood well by PHPStorm, I've pushed ca699a45

piRGoif:
Rebased & forced pushed !

setup/setuputils.class.inc.php
Hipska: Yup indeed, if you want to be that specific, then it should become this: `* @return array<string, \CheckResult>`. Otherwise `* @return \CheckResult[]` or `* @return array<\CheckResult>` are sufficient.

Molkobain:
Functional review: See with @Rorcha07 for possible complications on exotic setups (eg. lodabalancers, netwrk drives, ...)
If any concern raised will be part of 3.1.0, if green light from Stephane then 2.7.8

piRGoif:
Discussed with Stephane : we will switch the check to warning so it will become non blocking. 
The information will still be there, and we won't block setup wrongly !

piRGoif:
New message when dir isn't writable moved to warning : 

image

And after opening the details : 

image

setup/setuputils.class.inc.php
piRGoif: I've just updated our [coding standards wiki page](https://www.itophub.io/wiki/page?id=latest:customization:coding_standards#phpdoc)

setup/setuputils.class.inc.php
```diff
@@ -81,16 +81,23 @@ public static function CheckPhpAndExtensions()
...
-		$aResult = array_merge($aResult, $aWritableDirsErrors);
+		$aWritableApprootDirsErrors = self::CheckWritableDirs(array('log', 'env-production', 'env-production-build', 'conf', 'data'));
+		$aResult = array_merge($aResult, $aWritableApprootDirsErrors);
+		// Check temp dir (N°5235) : as this path isn't under APPROOT we are doing a custom check and not using \SetupUtils::CheckWritableDirs
+		$sTmpDir = static::GetTmpDir();
+		clearstatcache(true, $sTmpDir);
+		if (is_writable($sTmpDir)) {
+			$aResult[] = new CheckResult(CheckResult::INFO, "The temp directory is writable by the application.");
+		} else {
+			$aResult[] = new CheckResult(CheckResult::ERROR, "The temp directory <b>'".$sTmpDir."'</b> is not writable by the application. Change its permission or use another dir (sys_temp_dir option in php.ini).");
```
Molkobain, line 99: Did you push your last modifications for changing the error into a warning? Because it still looks an error here.

Molkobain: Nevermind I just saw it's done bdfe3a3b

Molkobain:
_Extra commit under bdfe3a3b to actually push the error to warning change_

setup/setuputils.class.inc.php
piRGoif: Yep, I made a mistake, forgot to push the modification before merging the PR (!!)
To make things a little clearer I've added a reference to this PR in the commit I've pushed after the pr merging.

piRGoif: See in the PR log : 

image

PR #305, opened 2022-07-07
N°5168 - Access to unauthorized contact information on Portal

bdalsass:
GlobalRequestMgmt issue


piRGoif (approved):

datamodels/2.x/itop-portal-base/portal/src/Form/ObjectFormManager.php
```diff
@@ -93,7 +93,7 @@ class ObjectFormManager extends FormManager
 	 * @since 2.7.6 3.0.0 N°4384 method creation : factorize as this is used twice now
 	 * @since 2.7.7 3.0.1 N°4867 now only used once, but we decided to keep this method anyway
```
piRGoif, line 94:
```suggestion
	 * @since 2.7.7 3.0.1 N°5168 made public as we need this helper method in other objects
```

piRGoif (changes requested):
As the `DecodeFormManagerData` method is static, maybe we should move it to ObjectFormHandlerHelper ? Shouldn't be a problem as there was only one call before this PR.

PR #306, opened 2022-07-08
N°3213 - Order transition attributes like in the "details" zlist instead of reordering them regarding the dependencies

accognet:
Transition screen before PR
 
image

Transition screen after PR:
 
image

The fields which should displayed in the same order than defined in xml in presentation/details
Only requied fields are displayed. Fields that are not in the display tag are displayed at the end


Molkobain:
@accognet as this bug was planned on iTop 3.x I'm assuming that it is not a development for the upcoming 3.1, is it a proposition that you want to pitch during the PRs review?

accognet:
No, this is a FAF

accognet:
Planned on iTop 3.1

Molkobain:
Discussed during technical review: Good PR but we need to discuss more deeply within the R&D team and with the Product team:
- Where to put the XML definition of this.
- How to make transition forms display as expected most of the time (ordered as the "details" zlist instead of ordered by dependencies) without forcing people to declare a zlist for a transition.

This will be discussed next week during functional review.

setup/compiler.class.inc.php
```diff
@@ -2068,19 +2067,32 @@ protected function CompileClass($oClass, $sTempTargetDir, $sFinalTargetDir, $sMo
...
+				$sStimulusCode = $oListNode->getAttribute('id');
+
+				$aAttributes = $oListNode->GetNodeAsArrayOfItems();
+				if (!is_array($aAttributes)) {
+					$aAttributes = array();
+				}
+				$this->ArrayOfItemsToZList($aAttributes);
+
+				$sZAttributes = var_export($aAttributes, true);
+				$sZlists .= "		MetaModel::Init_SetZListItems('$sStimulusCode', $sZAttributes);\n";
```
Hipska: What if a stimulus is named `details`, `list` or one of the search? There is no check and thus it will override them.

Hipska (changes requested):

setup/compiler.class.inc.php
Hipska: I also think a call to `MetaModel ::RegisterZList()` is needed as otherwise this line will also generate a `CoreException` for "Wrong value".

piRGoif:
Functional review : 

Only one zlist for all transitions ? Simpler, yes do it ! But don't forget the fallback for forgotten fields (not defined in XML but present in the transition anyway) 
What name for the XML node ? To be discussed with Romain + Vincent (we might add a possibility to specify each transition in the future, but a transition is defined both by the stimulus and the target state)

In the meantime, implement the fallback : return the order similar to the one defined in the details zlist. This will be the new default behavior, product team is OK to change it in 3.1 (major release + nobody likes the current one)

Note that in the portal we already have one form per stimulus (see forms/form/modes/mode apply_stimulus)
Also note that there is an ongoing bug on n-n relations in transition forms (not enough width for now)

setup/compiler.class.inc.php
Molkobain, line 1: No need to change that file, revert changes.

core/metamodel.class.php
Molkobain, line 1: Why this change? Please elaborate.

application/cmdbabstract.class.inc.php
```diff
@@ -3355,21 +3355,25 @@ public function DisplayStimulusForm(WebPage $oPage, $sStimulus, $aPrefillFormPar
...
 		}
-		$aList = $this->OrderDependentFields($aDeps);
+		$aListOrdered = $this->OrderDependentFields($aDeps);
+
+		//merge of the two list
+		foreach ($aListOrdered as $sAttCode) {
+			if (!in_array($sAttCode, $aList)) {
+				$aList[] = $sAttCode;
+			}
+		}
```
Molkobain: As the new specification is to just display required (msut_xxx and mandatory attributes with no value), you don't need to do all that, we retrieve the flatten "details" zlist.

Molkobain (changes requested):

Molkobain:
  * Code needs to be simplified / commented
  * Attributes list / ordering has to be extracted in a dedicated method so we can add a unit test to it. Moreover, this will be discussed with the product team as the specifications / use cases do not seem to be clear for everyone.

piRGoif:
Accepted during Functional Review.
Warning : the current screenshot of a transition with the fix isn't up to date !

Anne Catherine did some changes to handle Guillaume feedbacks, they need to be reviewed technically.
Also we need to figure out how to add an automatic test.

setup/compiler.class.inc.php
piRGoif: There are only formatting changes remaining now, so closing conversation

core/metamodel.class.php
piRGoif: Seems like the change you're referring to isn't there anymore. Closing conversation.

piRGoif (changes requested):
Ok changed reviewed !
Agree that we should create a test to document the ordering 
@accognet we might try to code it together ?

setup/compiler.class.inc.php
piRGoif: The changes were removed since then. Closing conversation.

application/cmdbabstract.class.inc.php
```diff
@@ -3354,22 +3354,13 @@ public function DisplayStimulusForm(WebPage $oPage, $sStimulus, $aPrefillFormPar
 		];
 
 		// The list of candidate fields is made of the ordered list of "details" attributes + other attributes
-		$aAttributes = array();
-		foreach ($this->FlattenZList(MetaModel::GetZListItems($sClass, 'details')) as $sAttCode) {
-			$aAttributes[$sAttCode] = true;
-		}
+		$aList = $this->FlattenZList(MetaModel::GetZListItems($sClass, 'details'));
+		//Add forgotten attributes to the list
```
Molkobain:
```suggestion
		// First attributes from the "details" zlist as they were sorted...
		$aList = $this->FlattenZList(MetaModel::GetZListItems($sClass, 'details'));

		// ... then append forgotten attributes
```

Molkobain (approved):

PR #308, opened 2022-07-08
N°5302 - Replace deprecated php strlen usages

bdalsass:


Hipska:
These are not deprecated usages as it is not using `strlen` but the wrapper `utils::StrLen`.

PR #309, opened 2022-07-08
N°2250 - DisplayObject with ormLinkSet ignore Removed

accognet:
PR in order to study the ticket internal after my holidays.
Fix which use ReadPostedParam in Display function. Not very nice but I don't have find another system without initial object in function.


Molkobain (approved):
Reviewed with AC

PR #311, opened 2022-07-19
N°6104 - Fix exception thrown if 'item_class' has 'org_id' attcode mapped to another attcode not present in Person class

vbkunin:
Hello!
If a class from Attachment.item_class has the MapContextParam method that returns an attribute code not present in the Person class, the execution fails on `$oCurrentPerson->Get($sAttCode)` with the unknown attribute exception and attachments are not saved.
In my case I faced with the problem in the UserRequest when mapping via MapContextParam the org_id to a newly created provider_org_id. The Person class obviously don't have the provider_org_id attcode but the second MetaModel::IsValidAttCode in the code checks the class from item_class again and not the Person class.

### DIFF
--- a/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
+++ b/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
@@ -176,18 +176,18 @@
 		$aCallSpec = array($sClass, 'MapContextParam');
 		if (is_callable($aCallSpec))
 		{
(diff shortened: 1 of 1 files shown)


Molkobain:
Hello Vladimir,

Your fix seems good, thanks! Could you share the XML / configuration to reproduce the issue so we can reproduce locally? It would help moving this PR forward.

Molkobain:

vbkunin:
Hello, @Molkobain, 
Thank you for accepting the PR! I'm afraid I can't find that XML anymore :(

Molkobain:
It's alright, thanks anyway!

PR #315, opened 2022-07-26
N°5281 - Symfony 5.4 extensions controllers registration

bdalsass:
Symfony upgrade to 5.4 improvements:

- Set default env var value (prod)
- Remove switch env by request param (&debug=true)
- Remove kernel deprecation calls

### DIFF
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -94,7 +94,7 @@ class CheckResult {
 class SetupUtils
 {
 	// -- Minimum versions (requirements : forbids installation if not met)
-	const PHP_MIN_VERSION             = '7.2.5'; // 7 will be supported until the end of 2019 (see http://php.net/supported-versions.php)
+	const PHP_MIN_VERSION             = '7.2.5';
 	const MYSQL_MIN_VERSION           = '5.7.0'; // 5.6 is no longer supported
 	const MYSQL_NOT_VALIDATED_VERSION = ''; // MySQL 8 is now OK (N°2010 in 2.7.0) but has no query cache so mind the perf on large volumes !
 
--- a/datamodels/2.x/itop-portal-base/portal/config/bootstrap.php
+++ b/datamodels/2.x/itop-portal-base/portal/config/bootstrap.php
@@ -56,7 +56,7 @@ if (file_exists(dirname(__DIR__).'/.env.local.php')) {
 
 	// load all the .env files
 	if (method_exists($oDotenv, 'loadEnv')) {
-		$oDotenv->loadEnv($sPath);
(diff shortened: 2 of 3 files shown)


Molkobain:
Benjamin, can you rename the PR's title with the corresponding bug friendlyname please? (eg. _N°132 - Bim bam boom_)

setup/setuputils.class.inc.php
```diff
@@ -94,7 +94,7 @@ public static function FromObjectsToStrings(array $aResults): array {
 class SetupUtils
 {
 	// -- Minimum versions (requirements : forbids installation if not met)
-	const PHP_MIN_VERSION             = '7.2.5'; // 7 will be supported until the end of 2019 (see http://php.net/supported-versions.php)
+	const PHP_MIN_VERSION             = '7.2.5';
```
Molkobain, line 97: For the record, this is about https://github.com/Combodo/iTop/pull/316#issuecomment-1195979947

Molkobain:
**Debug URL param**
Regarding the _"add the debug param to the URL and for each external key of a form, you will have the computed OQL (DM rules x Scopes) of available values in the log"_: It is no longer true since iTop 2.7... It was actually lost during the Silex to Symfony migrate, so my bad for pointing you in the wrong direction. 😅

The `IssueLog` is still present in the [portal ext. key renderer](https://github.com/Combodo/iTop/blob/develop/sources/Renderer/Bootstrap/FieldRenderer/BsSelectObjectFieldRenderer.php#L65) but is no longer activated as the _"debug"_ `ContextTag` is never positioned.

The `ContextTag` was previsouly positioned in the _itop-portal-base_ [frontal](https://github.com/Combodo/iTop/blob/2.6.0/datamodels/2.x/itop-portal-base/portal/web/index.php#L59) but this is no longer the case.

So:
  * As this is a regression from iTop 2.7, not your SF migration,
  * As this may not be used a lot by Combodo's support team, otherwise we would have had complains,
  * As we now have the LogChannels,
  * It's ok to go forward with your PR, we can drop the "debug" mode / param in the portal, this question can be close. (I'll create a bug in the bug tracker to switch all those IssueLog to more accurate channel / level).

**Symfony "dev" app env**
Regarding the dev. app. env. of SF, I was not able to activate it through the `.env.local` file, event though I uncommented both `APP_ENV` and `APP_TOKEN` params. This should still work as it is helpful both when developing and when troubleshooting. Can you get back to me on this? Thanks

image

Molkobain:
Discussed with Benjamin, dev env does not work if `.env.local` file is created directly in the env-xxx folder while using the symlinks.
This is due to a bug in the portal `bootstrap.php` file.

PR #316, opened 2022-07-26
N°4517 - Correct composer PHP max version and update setup utils constant

bdalsass:


Molkobain:
Ok for me. While you are at it, could you remove the comment on line 97 next to the PHP min. version? It's outdated.

Molkobain (approved):

Molkobain:
> Ok for me. While you are at it, could you remove the comment on line 97 next to the PHP min. version? It's outdated.

Was done in #315 at https://github.com/Combodo/iTop/pull/315/files#diff-91103f22b11a278c25c2d264d93af32996c30c6599a28f4a0b963b2cdc0adba6R97

PR #320, opened 2022-08-08
N°5270 - Move "apereo/phpcas" lib from "authent-cas" module to core composer.json

Molkobain:
For discussion about moving the lib up to the main dependancy manager.

Major cons:
  * Makes "cURL" PHP extension mandatory during setup, which changes prerequisites in a minor version, blocking the setup.

As 3.0.2 won't be PHP 8.1 compatible, there is no rush, I'll rather do this in iTop 3.1 than 3.0.2


Molkobain:
Discussed internally, will be only part of 3.1.0

Hipska:
Hey, then #171 can be revisited 😉 

Edit: Nevermind, it already has been added as optional ext..

PR #321, opened 2022-08-08
N°4342 - Improve generic bulk deletion function with memory limit handling

accognet:
Change function PurgeData in order to delete data by lot

### DIFF
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -1122,7 +1122,6 @@ class Config
 		'purge_data.max_chunk_size' => [
 			'type' => 'integer',
 			'description' => 'Maximum item deleted per loop. Used in function MetaModel::PurgeData',
-			// examples... not used
 			'default' => 1000,
 			'value' => 1000,
 			'source_of_value' => '',
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -70,6 +70,7 @@ define('DEFAULT_MIN_DISPLAY_LIMIT', 20);
 define('DEFAULT_MAX_DISPLAY_LIMIT', 30);
 define('DEFAULT_STANDARD_RELOAD_INTERVAL', 5 * 60);
 define('DEFAULT_FAST_RELOAD_INTERVAL', 1 * 60);
+define('DEFAULT_MAX_BUFFER_SIZE', 1000);
 define('DEFAULT_SECURE_CONNECTION_REQUIRED', false);
 define('DEFAULT_ALLOWED_LOGIN_TYPES', 'form|external|basic');
 define('DEFAULT_EXT_AUTH_VARIABLE', '$_SERVER[\'REMOTE_USER\']');
@@ -447,14 +448,14 @@ class Config
 			'show_in_conf_sample' => true,
 		],
 		'export_pdf_font' => [ // @since 2.7.0 PR #49 / N°1947
-			'type' => 'string',
-			'description' => 'Font used when generating a PDF file',
-			'default' => 'DejaVuSans', // DejaVuSans is a UTF-8 Unicode font, embedded in the TCPPDF lib we're using
-			// Standard PDF fonts like helvetica or times newroman are NOT Unicode
-			// A new DroidSansFallback can be used to improve CJK support (se PR #49)
-			'value' => '',
-			'source_of_value' => '',
-			'show_in_conf_sample' => false,
+		                       'type'                => 'string',
+		                       'description'         => 'Font used when generating a PDF file',
+		                       'default'             => 'DejaVuSans', // DejaVuSans is a UTF-8 Unicode font, embedded in the TCPPDF lib we're using
+		                       // Standard PDF fonts like helvetica or times newroman are NOT Unicode
+		                       // A new DroidSansFallback can be used to improve CJK support (se PR #49)
+		                       'value'               => '',
+		                       'source_of_value'     => '',
+		                       'show_in_conf_sample' => false,
 		],
 		'access_mode' => [
 			'type' => 'integer',
@@ -1324,9 +1325,9 @@ class Config
 		'draft_attachments_lifetime' => [
 			'type'                => 'integer',
 			'description'         => 'Lifetime (in seconds) of drafts\' attachments and inline images: after this duration, the garbage collector will delete them.',
-			'default' => 86400,
-			'value' => '',
-			'source_of_value' => '',
+			'default'             => 86400,
+			'value'               => '',
+			'source_of_value'     => '',
 			'show_in_conf_sample' => false,
 		],
 		'date_and_time_format' => [
@@ -1882,6 +1883,12 @@ class Config
 	 * @var integer Number of seconds between two reloads of the display (standard)
 	 */
 	protected $m_iStandardReloadInterval;
+
+	/**
+	 * @var integer Number of objects treated in a lot for bulk action (as anonymize, archival..)
+	 */
+	protected $m_iMaxBufferSize;
+
 	/**
 	 * @var integer Number of seconds between two reloads of the display (fast)
 	 */
@@ -1979,6 +1986,7 @@ class Config
 		$this->m_iMinDisplayLimit = DEFAULT_MIN_DISPLAY_LIMIT;
 		$this->m_iMaxDisplayLimit = DEFAULT_MAX_DISPLAY_LIMIT;
 		$this->m_iStandardReloadInterval = DEFAULT_STANDARD_RELOAD_INTERVAL;
+		$this->m_iMaxBufferSize = DEFAULT_MAX_BUFFER_SIZE;
 		$this->m_iFastReloadInterval = DEFAULT_FAST_RELOAD_INTERVAL;
 		$this->m_bSecureConnectionRequired = DEFAULT_SECURE_CONNECTION_REQUIRED;
 		$this->m_sDefaultLanguage = 'EN US';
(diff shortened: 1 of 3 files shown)


core/config.class.inc.php
```diff
@@ -1687,6 +1688,12 @@ public function IsCustomValue(string $sPropCode): bool
 	 * @var integer Number of seconds between two reloads of the display (standard)
 	 */
 	protected $m_iStandardReloadInterval;
+
+	/**
+	 * @var integer Number of objects treated in a lot for bulk action (as anonymize, archival..)
```
```suggestion
	 * @var integer Number of objects treated in a lot for bulk action (as anonymize, archival..)
	 * @since 3.1.0	 
```

core/config.class.inc.php
```diff
@@ -2058,6 +2067,11 @@ public function GetStandardReloadInterval()
 		return $this->m_iStandardReloadInterval;
 	}
 
+	public function GetMaxBufferSize()
```
```suggestion
    /*
     * @since 3.1.0
     */
	public function GetMaxBufferSize()
```

core/config.class.inc.php
```diff
@@ -2146,6 +2160,11 @@ public function SetStandardReloadInterval($iStandardReloadInterval)
 		$this->m_iStandardReloadInterval = $iStandardReloadInterval;
 	}
 
+	public function SetMaxBufferSize($iMaxBufferSize)
```
```suggestion
    /*
     * @since 3.1.0
     */
	public function SetMaxBufferSize($iMaxBufferSize)
```

core/metamodel.class.php
```diff
@@ -7095,32 +7095,41 @@ public static function BulkUpdate(DBObjectSearch $oFilter, array $aValues)
...
-			);
-			foreach($aTargetClasses as $sSomeClass)
-			{
-				$sTable = MetaModel::DBGetTable($sSomeClass);
-				$sPKField = MetaModel::DBGetKey($sSomeClass);
+		$iNbIdsDeleted = 0;
+		$bExecuteQuery = true;
+
+		while ($bExecuteQuery) {
+			$oSet = new DBObjectSet($oFilter, array(), array(), null, $iMaxBufferSize);
```
Molkobain: Rather than passing many default values, you can use the `SetLimit()`:

```suggestion
			$oSet = new DBObjectSet($oFilter);
			$oSet->SetLimit($iMaxBufferSize);
```

core/metamodel.class.php
```diff
@@ -7095,32 +7095,41 @@ public static function BulkUpdate(DBObjectSearch $oFilter, array $aValues)
...
+					$sPKField = MetaModel::DBGetKey($sSomeClass);
 
-				$sDeleteSQL = "DELETE FROM `$sTable` WHERE `$sPKField` IN ($sIdList)";
-				CMDBSource::DeleteFrom($sDeleteSQL);
+					$sDeleteSQL = "DELETE FROM `$sTable` WHERE `$sPKField` IN ($sIdList)";
+					CMDBSource::DeleteFrom($sDeleteSQL);
+				}
+				$iNbIdsDeleted += $iNbIds;
+			}
+			if ($iNbIds == 0 || $iNbIds < $iMaxBufferSize) {
```
Molkobain: Not sure to understand this statement correctly, will discuss IRL

Molkobain (changes requested):

Molkobain:
I feel like we should rename the conf parameter to be more explicit about what it does and its usage. It's not a generic buffer size but a "chunk size" or "batch size" for purging data.

Also, why didn't added the conf parameter like the others in the big array instead of a constant?

piRGoif:
Discussed during technical review today.

Romain and Anne-Catherine will do a pre-review together. In particular we need to be sure : 
- where is this \MetaModel::PurgeData mehod used in iTop ?
- why do we have a memory limit error ? Does the GetColumnAsArray() will load the whole objects besides we only asked for 2 fields (id and finalclass) ?

About the config parameter : 
- using an attribute + a getter is the "old way". @accognet please move the declaration to the \Config::$m_aSettings instead
- the parameter shouldn't be visible by default (`show_in_conf_sample` array key)
- the param name could be `purge_data.chunk_size` ? 
- default value could be 1000

core/config.class.inc.php
```diff
@@ -1004,6 +1004,15 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
+		'purge_data.max_chunk_size' => [
+			'type' => 'integer',
+			'description' => 'Maximum item deleted per loop. Used in function MetaModel::purgeData',
```
```suggestion
			'description' => 'Maximum item deleted per loop. Used in function MetaModel::PurgeData',
```

core/config.class.inc.php
```diff
@@ -1004,6 +1004,15 @@ class Config
...
+			// examples... not used
```
```suggestion
```

core/metamodel.class.php
```diff
@@ -7095,32 +7095,42 @@ public static function BulkUpdate(DBObjectSearch $oFilter, array $aValues)
...
 
-				$sDeleteSQL = "DELETE FROM `$sTable` WHERE `$sPKField` IN ($sIdList)";
-				CMDBSource::DeleteFrom($sDeleteSQL);
+					$sDeleteSQL = "DELETE FROM `$sTable` WHERE `$sPKField` IN ($sIdList)";
+					CMDBSource::DeleteFrom($sDeleteSQL);
+				}
+				$iNbIdsDeleted += $iNbIds;
+			}
+			if ($iNbIds == 0 || $iNbIds < $iMaxChunkSize) {
+				$bExecuteQuery = false;
```
Molkobain: Can this be simplify to `$iNbIds < $iMaxChunkSize` ?

Molkobain (changes requested):

favincen:
Hi guys,
I am one of those who asked for a generic bulk deletion solution a while back, in Combodo's ticket R-032835 (which is linked to ER N°4342).
But discovering and reading the content of this PR, while I'm happy something is being done on this front, I realize we might not mean the same : 

As far as I understand the code, PurgeData will do only a brutal SQL purge of the object matching the provided filter. But it will NOT guarantee data integrity because it will not do a proper update or delete of related data (update N:N link tables and related objects' history, purge deleted objects' related CMDBChangeOp, block deletions not respecting mandatory link with manual deletion, etc.). As its comments state, "it bypasses standard objects handlers".
So it seems this current work would be useless for the use case where the data to be purged has datamodel dependencies (so any instance of class inheriting from cmdbAbstractObject).

What I meant when asking for a generic bulk deletion what to be able to mass delete any instance of classes inheriting cmdbAbstractObject, while respecting all iTop data handeling standards. So it would need to perform all iTop's standard data management to respect data integrity and proper historysation (like any deletionPlan would do). 
I saw that last version 1.3 of PersonnalDataAnonymizer extension includes some sort of such a batch mass data processing but at the moment I could not find any documentation on how it could be used outside of personal data anonymation for mass delete.

If that's ok with you, I would be interested to discuss this matter with you in any way you would see fit.

Hipska:
Hi @favincen, some of the developers are available on the [iTop community Slack team](https://join.slack.com/t/itopbycombodo/shared_invite/zt-uoat5wuf-C0g_1tqaeI9~EaqHxV7Q1Q).

Molkobain:
Remaining tasks:
- Add a unit test on the method
- Apply suggestions

core/config.class.inc.php
```diff
@@ -1119,6 +1119,14 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
+		'purge_data.max_chunk_size' => [
+			'type' => 'integer',
+			'description' => 'Maximum item deleted per loop. Used in function MetaModel::PurgeData',
```
```suggestion
			'description' => 'Maximum number of items deleted per loop. Used in function MetaModel::PurgeData',
```

tests/php-unit-tests/unitary-tests/core/MetaModelTest.php
```diff
@@ -456,6 +457,31 @@ public function IsObjectInDBProvider(): array
...
+
+    /**
+     * @return void
+     * @throws CoreException
+     * @throws \OQLException
+     */
+    public function testPurgeData(){
+        //Set max_chunk_size to 2 (default 1000) to test chunk deletion with only 10 items
+        $oConfig = MetaModel::GetConfig();
+        $oConfig->Set('purge_data.max_chunk_size',2);
```
Molkobain:
```suggestion
        $oConfig->Set('purge_data.max_chunk_size', 2);
```

tests/php-unit-tests/unitary-tests/core/MetaModelTest.php
```diff
@@ -456,6 +457,31 @@ public function IsObjectInDBProvider(): array
...
+        MetaModel::SetConfig($oConfig);
+
+        $aPkPerson = [];
+        for ($i=0;$i<10;$i++) {
```
Molkobain:
```suggestion
        for ($i=0; $i < 10; $i++) {
```

tests/php-unit-tests/unitary-tests/core/MetaModelTest.php
```diff
@@ -456,6 +457,31 @@ public function IsObjectInDBProvider(): array
...
+            $oPerson = $this->CreatePerson($i, 1);
+            $sClass = get_class($oPerson);
+            $aPkPerson[] = $oPerson->GetKey();
+        }
+
+        $sDeleteOQL = 'SELECT '.$sClass.' WHERE id IN ('.implode(',',$aPkPerson).')';
```
Molkobain:
```suggestion
        $sDeleteOQL = 'SELECT '.$sClass.' WHERE id IN ('.implode(',', $aPkPerson).')';
```

tests/php-unit-tests/unitary-tests/core/MetaModelTest.php
```diff
@@ -456,6 +457,31 @@ public function IsObjectInDBProvider(): array
...
 		];
 	}
+
+    /**
+     * @return void
+     * @throws CoreException
+     * @throws \OQLException
+     */
+    public function testPurgeData(){
+        //Set max_chunk_size to 2 (default 1000) to test chunk deletion with only 10 items
```
Molkobain:
```suggestion
        // Set max_chunk_size to 2 (default 1000) to test chunk deletion with only 10 items
```

core/metamodel.class.php
Molkobain: This hasn't been fixed or argumented.

Molkobain (commented):
Just the remaining conversation

core/metamodel.class.php
accognet: Done

Molkobain (approved):

Molkobain:
Great, you can "squash & merge" !

PR #322, opened 2022-08-08
N°4792 - memory limit is reached when an AttributeExternalKey is pointing to a class with an AttributeBlob

eespie:
What was done:
  * When requesting values for external keys in the case of combobox (request all possible values) the columns fetched are limitted to the minimum to avoid loading the full objects including blobs
  * __toString() was also added to DBSearch to ease debug

### DIFF
--- a/application/ui.extkeywidget.class.inc.php
+++ b/application/ui.extkeywidget.class.inc.php
@@ -211,7 +211,7 @@ class UIExtKeyWidget
 			$sClassAllowed = $oAllowedValues->GetClass();
 			$bAddingValue = false;
 
-			// #4792 - load only the required fields
+			// N°4792 - load only the required fields
 			$aFieldsToLoad = [];
 
 			$aComplementAttributeSpec = MetaModel::GetNameSpec($oAllowedValues->GetClass(), FriendlyNameType::COMPLEMENTARY);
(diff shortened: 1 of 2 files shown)


application/ui.extkeywidget.class.inc.php
```diff
@@ -211,14 +211,23 @@ public function DisplaySelect(WebPage $oPage, $iMaxComboLength, $bAllowTargetCre
 			$sClassAllowed = $oAllowedValues->GetClass();
 			$bAddingValue = false;
 
+			// #4792 - load only the required fields
```
Molkobain:
```suggestion
			// N°4792 - load only the required fields
```

Just change this as #XXX are usually used for SourceForge ticket reference.

Molkobain (changes requested):
Except from the "typo" seems all good to me ✌

_Edit: You should be able to apply the suggestion directly from the PR UI_

Molkobain:
_Also, for internal PRs can you add the "internal" label? Thanks_

eespie:
changed comment and fix data used from MetaModel::GetNameSpec()

Hipska:
Problem was the whole object from the external key was loaded? And the solution is now only the fields to make the friendlyname? Great improvement!

One question, will a programmer still be able to get any other field (by lazy loading) like this: `$oObject->Get('external_id')->Get('blob_field')`?

eespie:
The use of OptimizeColumnLoad() method is to be used with caution. If another column is requested with Get() then the object will be fully reloaded. So yes, it works, but the performances will be very bad.
In this specific case, the object is not used anymore after building the list.

PR #325, opened 2022-08-10
N°3979 - Handle initial values on dependent fields

accognet:
When a field is hidden (due to display conditions), his initial (or previous) value is kept in metadata and post in a new field named previous_value. 
Some changes for this feature are in templates-base extension.

### DIFF
--- a/js/console_form_handler.js
+++ b/js/console_form_handler.js
@@ -35,8 +35,7 @@ $(function()
         {
             var me = this;
             
-            this.element
-                .append('<div class="last-error"></div>')
+            this.element.append('<div class="last-error"></div>')
                 .addClass('console_form_handler');
 
             this.options.oWizardHelper = window[this.options.wizard_helper_var_name];
@@ -48,13 +47,25 @@ $(function()
         // revert other modifications here
         _destroy: function()
         {
-            this.element
-            .removeClass('console_form_handler');
+            this.element.removeClass('console_form_handler');
             this._super();
         },
+	    _updatePreviousValues: function()
+	    {
+		    let me=this;
+	    	if(this.element.find('[data-attribute-previous-value]').length>0) {
+	    		let aPreviousValues = {};
+			    $(this.element.find('[data-attribute-previous-value]')).each(function (idx, elt) {
+				    aPreviousValues[$(elt).data('field-id')] = $(elt).data('attribute-previous-value');
+			    });
+			    me.element.find('[data-field-id=previous_values]').find('input[type=hidden]').val(JSON.stringify(aPreviousValues));
+		    }
+
+	    },
         _onUpdateFields: function(event, data)
         {
             var me = this;
+	        me._updatePreviousValues();
             var sFormPath = data.form_path;
             var sUpdateUrl = GetAbsoluteUrlAppRoot()+'pages/ajax.render.php';
 
@@ -64,7 +75,6 @@ $(function()
                 {
                     operation: 'custom_fields_update',
                     attcode: this.options.custom_field_attcode,
-                    //current_values: this.getCurrentValues(),
                     requested_fields: data.requested_fields,
                     form_path: sFormPath,
                     json_obj: this.options.oWizardHelper.UpdateWizardToJSON()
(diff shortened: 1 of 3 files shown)


js/console_form_handler.js
```diff
@@ -48,13 +47,25 @@ $(function()
...
         _destroy: function()
         {
-            this.element
-            .removeClass('console_form_handler');
+            this.element.removeClass('console_form_handler');
             this._super();
         },
+	    _updatePreviousValues: function()
+	    {
+		    let me=this;
```
rquetiez: Please respect the spaces (me = this)

js/console_form_handler.js
```diff
@@ -48,13 +47,25 @@ $(function()
...
+	    	if(this.element.find('[data-attribute-previous-value]').length>0) {
```
rquetiez: this.element.find('[data-attribute-previous-value]') is used three times. I suggest to factorize it into a variable previousValuesElements

rquetiez:
Work still to be done

1. Implement the same new behavior in the portal
2. Check that the change on readonly fields -cosmetic but not requested- is ok for every use case.

js/console_form_handler.js
accognet: I think it's not necessary just for 2 times (the third time doesn't work). and for me,  it's easier to undestand like this.

js/console_form_handler.js
```diff
@@ -85,6 +83,7 @@ $(function()
                         me.element.find('.last-error').text(data.error);
                     }
                     me._onUpdateAlways(data, sFormPath);
+	                me.element.find('[data-field-id=previous_values]').find('input[type=hidden]').val('{}');
```
Molkobain: Please put quotes around attribute value in selectors.

```suggestion
	                me.element.find('[data-field-id="previous_values"]').find('input[type="hidden"]').val('{}');
```

accognet:
Done

js/form_handler.js
```diff
@@ -93,11 +93,23 @@ $(function()
...
 
+		_updatePreviousValues: function()
+		{
+			let me = this;
+			if(this.element.find('[data-attribute-previous-value]').length>0) {
+				let aPreviousValues = {};
+				$(this.element.find('[data-attribute-previous-value]')).each(function (idx, elt) {
+					aPreviousValues[$(elt).data('field-id')] = $(elt).data('attribute-previous-value');
+				});
+				me.element.find('[data-field-id=previous_values]').find('input[type=hidden]').val(JSON.stringify(aPreviousValues));
```
Molkobain: Please put quotes in selectors, we discussed about this several times.

```suggestion
				me.element.find('[data-field-id="previous_values"]').find('input[type="hidden"]').val(JSON.stringify(aPreviousValues));
```

sources/Renderer/Console/FieldRenderer/ConsoleSimpleFieldRenderer.php
```diff
@@ -111,15 +112,17 @@ public function Render()
...
 					{
-						$oText->SetIsDisabled(true);
+						$oValue->AddSubBlock(UIContentBlockUIBlockFactory::MakeStandard())->AddSubBlock(HtmlFactory::MakeHtmlContent($this->oField->GetCurrentValue()));
+						$oValue->AddSubBlock(InputUIBlockFactory::MakeForHidden("",$this->oField->GetCurrentValue(), $this->oField->GetGlobalId()));
 					}
 					else
 					{
+						$oText = new TextArea("",$this->oField->GetCurrentValue(),$this->oField->GetGlobalId(),40,8);
+						$oText->AddCSSClass('ibo-input-field-wrapper ibo-input');
+						$oValue->AddSubBlock($oText);
```
Molkobain, line 125: You are adding several CSS classes so use the corresponding method `AddCSSClasses`, otherwise it will mess with the internal representation and possibly with some processing.

js/form_handler.js
```diff
@@ -93,11 +93,23 @@ $(function()
 			return this.options.field_set.triggerHandler('get_current_values');
 		},
 
+		_updatePreviousValues: function()
```
Molkobain, line 96: Add comment with iTop versions in which this is introduced (3.0.2, 3.1.0)

Molkobain (changes requested):

PR #326, opened 2022-08-10
N°5318 - Fix DBObject::CheckValue() messages being HTML encoded when not necessary

accognet:
for 2.7.8 and 3.02

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -4735,11 +4735,15 @@ EOF
 			$sCSSClass = $bResult ? HILIGHT_CLASS_NONE : HILIGHT_CLASS_CRITICAL;
 			$sChecked = $bResult ? 'checked' : '';
 			$sDisabled = $bResult ? '' : 'disabled';
+
+			$aErrorsToDisplay = array_map(function($sError) {
+				return utils::HtmlEntities($sError);
+			}, $aErrors);
 			$aRows[] = array(
 				'form::select' => "<input type=\"checkbox\" class=\"selectList\" $sChecked $sDisabled\"></input>",
 				'object' => $oObj->GetHyperlink(),
 				'status' => $sStatus,
-				'errors' => '<p>'.($bResult ? '' : utils::HtmlEntities(implode('</p><p>', $aErrors))).'</p>',
+				'errors' => '<p>'.($bResult ? '' : utils::HtmlEntities(implode('</p><p>', $aErrorsToDisplay))).'</p>',
 				'@class' => $sCSSClass,
 			);
 			if ($bResult && (!$bPreview))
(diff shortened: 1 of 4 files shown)


application/cmdbabstract.class.inc.php
```diff
@@ -4739,7 +4739,7 @@ public static function DoBulkModify($oP, $sClass, $aSelectedObj, $sCustomOperati
 				'form::select' => "<input type=\"checkbox\" class=\"selectList\" $sChecked $sDisabled\"></input>",
 				'object' => $oObj->GetHyperlink(),
 				'status' => $sStatus,
-				'errors' => '<p>'.($bResult ? '' : implode('</p><p>', $aErrors)).'</p>',
+				'errors' => '<p>'.($bResult ? '' : utils::HtmlEntities(implode('</p><p>', $aErrors))).'</p>',
```
Molkobain: This is most likely to encode the `'</p><p>'` tags, have you tried to go through that line to see how it renders?

Molkobain:
> for 2.7.8 and 3.02

The PR is currently set to be merged in the "develop" branch, can you change it to "support/2.7", otherwise it won't be merged in 2.7.8 / 3.0.2.

application/cmdbabstract.class.inc.php
Molkobain: You need to apply the HtmlEntities() method on elements of `$aErrors` only. To do so, you can use the `array_map()` function (you can use an anonymous function, no need to declare one), check [here](https://www.php.net/manual/en/function.array-map.php) for the function documentation.

For example (don't commit the following as-is, try it out first):
```
$aErrors = array_map(function($sError) {
        return utils::HtmlEntities($sError);
}, $aErrors);
```

application/cmdbabstract.class.inc.php
```diff
@@ -4735,11 +4735,15 @@ public static function DoBulkModify($oP, $sClass, $aSelectedObj, $sCustomOperati
...
+
+			$aErrorsToDisplay = array_map(function($sError) {
+				return utils::HtmlEntities($sError);
+			}, $aErrors);
 			$aRows[] = array(
 				'form::select' => "<input type=\"checkbox\" class=\"selectList\" $sChecked $sDisabled\"></input>",
 				'object' => $oObj->GetHyperlink(),
 				'status' => $sStatus,
-				'errors' => '<p>'.($bResult ? '' : implode('</p><p>', $aErrors)).'</p>',
+				'errors' => '<p>'.($bResult ? '' : utils::HtmlEntities(implode('</p><p>', $aErrorsToDisplay))).'</p>',
```
Molkobain: Now that you entities encoded `$aErrors` on line 4739, you need to remove encoding on this line, otherwise it will be double encoded and the problem will remain.

```suggestion
				'errors' => '<p>'.($bResult ? '' : implode('</p><p>', $aErrorsToDisplay)).'</p>',
```

Molkobain (changes requested):

pages/UI.php
```diff
@@ -1614,7 +1614,7 @@ function DisplayNavigatorGroupTab($oP)
 								}
 								else
 								{
-									$sError = '<p>'.implode('</p></p>',$aErrors)."</p>\n";
+									$sError = '<p>'.utils::HtmlEntities(implode('</p></p>',$aErrors))."</p>\n";
```
Molkobain: You still have the same issue here.

accognet: Yes, I have seen :-)

PR #327, opened 2022-08-11
N°5024 - Missing result in search widget for external key

accognet:


Molkobain:
PR is targeted to branch "develop" but is must be "support/3.0"

eespie (approved):

pages/ajax.render.php
```diff
@@ -364,7 +364,6 @@ function LogErrorMessage($sMsgPrefix, $aContextInfo) {
 			$sContains = utils::ReadParam('q', '', false, 'raw_data');
 			$bSearchMode = (utils::ReadParam('bSearchMode', 'false') == 'true');
 			$sOutputFormat = utils::ReadParam('sOutputFormat', UIExtKeyWidget::ENUM_OUTPUT_FORMAT_CSV, false, 'raw_data');
-            $sAutocompleteOperation = utils::ReadParam('sAutocompleteOperation', null, false, 'raw_data');
```
Molkobain, line 367: I think we should keep this parameter as this endpoint can be use by something else with a specific operation.
(The other change a few lines below needs to be rollbacked as well then)

js/search/search_form_criteria_enum.js
```diff
@@ -441,28 +440,7 @@ $(function()
...
-								.done(function(oResponseContains, sStatus, oXHR){
-                                    //filter duplicates
-                                    $.each(oResponse, function(index, value) {
-                                        delete oResponseContains[index];
-									});
-
-                                    me._onACSearchContainsSuccess(oResponseContains);
-
-
-                                });
```
Molkobain, line 465: Good catch!

pages/ajax.render.php
accognet: This argument is not used below. So for me, it's easier if we don't get the unused variables.

Molkobain: Oh you are right, I didn't check the content of the `UIExtKeyWidget::AutoComplete()` method. Indeed it is not used anymore there, might be due to 3.0 refactoring.

Then it all good!

Molkobain (approved):

pages/ajax.render.php
```diff
@@ -378,7 +377,7 @@ function LogErrorMessage($sMsgPrefix, $aContextInfo) {
 					$oObj = null;
 				}
 				$oWidget = new UIExtKeyWidget($sTargetClass, $iInputId, '', $bSearchMode);
-				$oWidget->AutoComplete($oPage, $sFilter, $oObj, $sContains, $sOutputFormat, $sAutocompleteOperation);
+				$oWidget->AutoComplete($oPage, $sFilter, $oObj, $sContains, $sOutputFormat, '');
```
Molkobain: Default value is not `''` but `null`. Anyway we might just not pass this argument if it is not used, what do you think?

```suggestion
				$oWidget->AutoComplete($oPage, $sFilter, $oObj, $sContains, $sOutputFormat);
```

Molkobain (changes requested):

Molkobain (approved):

PR #328, opened 2022-08-11
N°5389 - Email Notification templates with AttributeLinkedSetIndirect failed

eespie:
What was done in DBObject::DBUpdate()
  * force reload of the object when a linkSet is modified
  * Call the triggers OnObjectUpdate *after* the reload

Risks: the change of calls ordering between triggers and callback AfterUpdate() may change some extension behaviour

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -4589,7 +4589,7 @@ class AttributeCaseLog extends AttributeLongText
 				{
 					//N°5135 - add impersonation information in caselog
 					if (UserRights::IsImpersonated()){
-						$sOnBehalfOf = $sUserString = Dict::Format('UI:Archive_User_OnBehalfOf_User', UserRights::GetRealUserFriendlyName(), UserRights::GetUserFriendlyName());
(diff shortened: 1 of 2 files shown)


accognet (approved):

core/attributedef.class.inc.php
```diff
@@ -4589,7 +4589,7 @@ public function MakeRealValue($proposedValue, $oHostObj)
 				{
 					//N°5135 - add impersonation information in caselog
 					if (UserRights::IsImpersonated()){
-						$sOnBehalfOf = $sUserString = Dict::Format('UI:Archive_User_OnBehalfOf_User', UserRights::GetRealUserFriendlyName(), UserRights::GetUserFriendlyName());
+						$sOnBehalfOf = Dict::Format('UI:Archive_User_OnBehalfOf_User', UserRights::GetRealUserFriendlyName(), UserRights::GetUserFriendlyName());
```
Molkobain, line 4592: I think this is on the wrong PR

Molkobain (changes requested):
Apart from the impersonate line that should not be here, the fix seems ok. As you mentioned there could be side effects on some extensions but it will be easy to fix them.

Please mind to update the CRUD wiki page on both 3.0 and latest namespace in the wiki.

PR #333, opened 2022-08-26
🐛 N°5522 - Fix session storage (breadcrumbs) not cleared on logout

RohacRichard:
Hi everyone,

there is existing cleanup of PHP SESSION variable on logout, but session storage with breadcrumb stays there. This makes the session storage stick between logins and more importantly, between users. 

I am logged in as admin, have a breadcrumb of admin operations, then log out. Another user can theoretically go onto my computer and my session, login as user account and see, what I was doing as admin in the breadcrumb. There are of course proper security checks in place, so the user cannot get anywhere he shouldn't get. 

This is minor security concern, as it can apply only if browser is left open after logout and stays in the same tab/context. Even if this can be exploited in some way, the user can only see friendlynames of objects, that other used accessed.

This is more a matter of proper cleanup, and a minor security concern.

Thanks for consideration,
Richard

### DIFF
--- a/pages/logoff.php
+++ b/pages/logoff.php
@@ -37,6 +37,7 @@ if ($operation == 'do_logoff')
 	// Note the redirection MUST NOT be made via an HTTP "header" since onunload is called only when the actual content of the DOM
 	// is replaced by some other content. So the "bouncing" page must provide some content (in our case a script making the redirection).
 	$oPage = new AjaxPage('');
+	$oPage->add_script("sessionStorage.clear();");
 	$oPage->add_script("window.location.href='{$sUrl}pages/logoff.php?portal=$bPortal'");
 	$oPage->output();
 	exit;
--- a/pages/logoff.php
+++ b/pages/logoff.php
@@ -37,7 +37,6 @@ if ($operation == 'do_logoff')
 	// Note the redirection MUST NOT be made via an HTTP "header" since onunload is called only when the actual content of the DOM
 	// is replaced by some other content. So the "bouncing" page must provide some content (in our case a script making the redirection).
 	$oPage = new AjaxPage('');
-	$oPage->add_script("sessionStorage.clear();");
 	$oPage->add_script("window.location.href='{$sUrl}pages/logoff.php?portal=$bPortal'");
 	$oPage->output();
 	exit;
--- a/templates/pages/login/logout.html.twig
+++ b/templates/pages/login/logout.html.twig
@@ -13,6 +13,8 @@
 {% endblock %}
 
 {% block script %}
+    {{ parent() }}
+    
 	sessionStorage.clear();
 {% endblock %}
 
--- a/templates/pages/login/logout.html.twig
+++ b/templates/pages/login/logout.html.twig
@@ -12,6 +12,10 @@
 	</div>
 {% endblock %}
 
+{% block script %}
+	sessionStorage.clear();
+{% endblock %}
+
 {% block ready_script %}
     {{ parent() }}
 


Hipska (approved):
Looks good, although I think they would want it in a `script` or `ready_script` block in the logout.html.twig template, as it would then only be executed when actually logged out and not in the pre-step. (Or maybe just the same command between line 99 and 100)

Molkobain:
Hello Rohac,

Thanks for the PR! I think we have a internal bug tracked for this, we could hit 2 birds with 1 stone with that PR. @v-dumas do you recall a tracked bug for this?

Molkobain:
@RohacRichard is this working on your instance? We recently made the logout page redirect to the login page but it was on another part of the application, not the file you changed. See [this commit](https://github.com/Combodo/iTop/commit/1655a377f08efb959a31417f4ca58f7da84c33db)

RohacRichard:
@Molkobain Hey Guillaume, I am not sure I understand the question. Is the change I commited working on my instance? If so, yes, it is. I tested it multiple times.

If you believe the session storage emptying should occur elsewhere, I can move it, no problem 😊

Molkobain:
Indeed, the emptying should rather be in [this file](https://github.com/Combodo/iTop/blob/develop/templates/pages/login/logout.html.twig).
Your current fix is working (just tested it) but it might not in the near future as we changed the login/logout mechanism, so moving it to the dedicated file would be great :)

On a side note, you are completely emptying the session storage.
For now it's ok as it only includes data on the breadcrumbs, but if it contains non-user-related data (brought by an extension or the core in the future) it will remove them as well.

Not sure if this is something we should worry about right now, I'll discuss it with the rest of the team.

pages/logoff.php
```diff
@@ -37,6 +37,7 @@
 	// Note the redirection MUST NOT be made via an HTTP "header" since onunload is called only when the actual content of the DOM
 	// is replaced by some other content. So the "bouncing" page must provide some content (in our case a script making the redirection).
 	$oPage = new AjaxPage('');
+	$oPage->add_script("sessionStorage.clear();");
```
Molkobain: Move this to [this file](https://github.com/Combodo/iTop/blob/develop/templates/pages/login/logout.html.twig)

Molkobain (changes requested):

pages/logoff.php
RohacRichard: Ok, it is done 😊

RohacRichard:
IMHO if the iTop session ended, both the browser session storage and the PHP session should be cleared. If one wants to persist some data between logins, it's better to use a cookie or browser local storage.

But this is up to Combodo and it's team to decide 😁

Molkobain:
> [...] it's better to use [...] browser local storage.
> 
> [...]

You are totally right!

templates/pages/login/logout.html.twig
```diff
@@ -12,6 +12,10 @@
 	</div>
 {% endblock %}
 
+{% block script %}
+	sessionStorage.clear();
+{% endblock %}
```
Molkobain:
```suggestion
{% block script %}
    {{ parent() }}
    
	sessionStorage.clear();
{% endblock %}
```

Just this change to avoid loosing scripts from the parent template and it will be ok for me (you can accept the suggestion directly here, no need to change on you computer)

Molkobain (approved):

Molkobain:
All good for me, will be reviewed next week. Thanks!

Hipska (approved):

pages/logoff.php
piRGoif: Closing the conversation as it is now done !

piRGoif:
Hello,

We discussed this PR during the technical review today.

Do we agree this can only happen if the browser is left open after logout, and iTop is opened in the same context (I mean in a Firefox container sense) ? 
If so can you update your PR description to reflect this condition ?

For us the only security issue would be to see the friendlynames of objects you can't normally see.

It is more a functional question than a technical one (and Guillaume already checked the only technical aspect : where to put the reset), it will be discussed in the functional review next week !

RohacRichard:
Hi @piRGoif,

I agree and updated the PR description, is this alright with you?

Thank you,
Richard

piRGoif:
Perfect, many thanks !

Molkobain:
Accepted in functional review, will be part of the upcoming 3.1.0

Molkobain:
@RohacRichard is your postal still the same? (So we can you new stickers) Otherwise send me an email :)

RohacRichard:
@Molkobain Yes, the postal address is still the same :) Can I also get my name on that sweet sweet contributor list? 😁

piRGoif:
> Can I also get my name on that sweet sweet contributor list? 😁

This is done in 7d1893a2 :)

RohacRichard:
Awesome, thank you Pierre. Glad I could contribute to iTop 😁

Cheers,
Richard

Molkobain:
> @Molkobain Yes, the postal address is still the same :) Can I also get my name on that sweet sweet contributor list? 😁

Sorry I thought I already added you with your previous PR!

RohacRichard:
@Molkobain It's alright, no worries :)

PR #335, opened 2022-09-05
N°5523 - Setup wizard : use the ITOP_APPLICATION constant instead of hardcoded "iTop" string

xtophe38:
These changes replace, where it makes sense, the usage of iTop as a product name to the variable ITOP_APPLICATION as it is already done in most of the setup process.

PR #336, opened 2022-09-05
N°5758 - GDPR consent: Apply to all instances containing the Hub connector

xtophe38:
Up to now,  we define that RGPD consent is required if iTop is not an "iTop Product". This is, I think, too restrictive as other products based on iTop (TeemIp std alone today and ESM products tomorrow) may not require RGPD consent even if they are not "iTop Products". Since instance data are collected through iTop Hub only, I suggest to base the test of RGPD consent requirement on the fact that iTop Hub Connector extension is installed.

### DIFF
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -1595,6 +1595,16 @@ JS
 	}
 
 	/**
+	 * @param array $aModules List of available module codes
+	 *
+	 * @return bool true if the Hub connector is installed
+	 */
+	public static function IsItopHubInstance($aModules)
+	{
+		return array_key_exists('itop-hub-connector', $aModules);
+	}
+
+	/**
 	 * @param array $aModules Available modules with code as key and metadata array as values
 	 *    Same structure as the one returned by {@link \RunTimeEnvironment::AnalyzeInstallation}
 	 * @param string $sExtensionsDir In the setup, get value with the 'extensions_dir' parameter
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -1599,7 +1599,7 @@ JS
 	 *
 	 * @return bool true if the Hub connector is installed
 	 */
-	public static function IsItopHubInstance($aModules)
+	public static function IsConnectableToITopHub($aModules)
 	{
 		return array_key_exists('itop-hub-connector', $aModules);
 	}
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -1598,6 +1598,8 @@ JS
 	 * @param array $aModules List of available module codes
 	 *
 	 * @return bool true if the Hub connector is installed
+	 *
+	 * @since 2.7.8 3.0.3 3.1.0 N°5758 method creation
 	 */
 	public static function IsConnectableToITopHub($aModules)
 	{
--- a/setup/wizardsteps.class.inc.php
+++ b/setup/wizardsteps.class.inc.php
@@ -716,7 +716,7 @@ class WizStepLicense extends WizardStep
 		$sMode = $this->oWizard->GetParameter('install_mode');
 		$aModules = SetupUtils::AnalyzeInstallation($this->oWizard);
 
-		return (($sMode === 'install') && SetupUtils::IsItopHubInstance($aModules));
+		return (($sMode === 'install') && SetupUtils::IsConnectableToITopHub($aModules));
 	}
 
     /**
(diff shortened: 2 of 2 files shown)


Hipska:
I also think the ITSM connectors also require this consent.

PS, the correct English term is GDPR.

piRGoif (changes requested):
I like your suggestion !
Just a little change request : you should create a SetupUtils method, as what was done with IsProductVersion. To test for the presence of the Hub Connector, maybe call it IsItopHubInstance ?

piRGoif:
> I also think the ITSM connectors also require this consent.

For now it isn't. I suppose it is because if you have a package including the ITSM Designer Connector then you have a contract with Combodo or one of its partner ? 

@dcoille can you confirm ?

Molkobain:
> I like your suggestion ! Just a little change request : you should create a SetupUtils method, as what was done with IsProductVersion. To test for the presence of the Hub Connector, maybe call it IsItopHubInstance ?

Like Pierre I would rather have the test using a detection method than an "hard-coded" test, so we can factorize this kind of detections.

xtophe38:
Factorization done as required.

piRGoif (approved):
Thanks, perfect now !

Hipska:
For a private function, it is perfectly doable to do a refactoring to correct english naming:
```php
	private function NeedsGdprConsent()
```

piRGoif:
> For a private function, it is perfectly doable to do a refactoring to correct english naming:
> 
> ```
> 	private function NeedsGdprConsent()
> ```

Yeah, we should do that, but it won't be mandatory for this PR to be accepted.
I'll do the rename :)

Hipska:
Sure, it can be accepted as is indeed. I was just mentioning as it could be done in this PR as well, no need for a separate PR.

Or, indeed, just do it directly on the branch without PR, that works as well 😄

piRGoif:
I've done the rename directly in this PR branch as it will lower the possible conflicts :)

Hipska (approved):

setup/setuputils.class.inc.php
```diff
@@ -1587,6 +1587,16 @@ public static function IsProductVersion($aModules)
 		return array_key_exists('itsm-designer-connector', $aModules);
 	}
 
+	/**
+	 * @param array $aModules List of available module codes
+	 *
+	 * @return bool true if the Hub connector is installed
+	 */
+	public static function IsItopHubInstance($aModules)
```
Molkobain: Reminder for after functional review:
  * Add `@since` once version is chosen
  * Discuss about the function name during technical review as "IsItopHubInstance", to me it means that "this is an iTop Hub instance" (which is not possible) rather than "this iTop instance connects to iTop Hub"

Molkobain:
Functional review: R&D still to do some checks but it is accepted. If chosen implementation is very safe plan for 2.7.8, otherwise 3.0.3

@Hipska there is no GDPR disclaimer for products using the ITSM Designer because they can only be used with a Combodo subscription which contains the legal details.

piRGoif:
Rebased (there were conflicts due to #335) and forced pushed !

setup/wizardsteps.class.inc.php
```diff
@@ -707,15 +707,16 @@ public function ProcessParams($bMoveForward = true)
...
 
 	/**
-	 * @return bool
+	 * @return bool true if we need to display a GDPR confirmation
 	 * @throws \Exception
-	 * @since 2.7.7 3.0.2 3.1.0
+	 * @since 2.7.7 3.0.2 3.1.0 N°5037
 	 */
-	private function NeedsRgpdConsent()
+	private function NeedsGdprConsent()
```
piRGoif, line 715: We'll need a `@since` here also so that the method rename would be documented !

setup/setuputils.class.inc.php
Molkobain: Discussed during technical review, still waiting for a better function name, eg. `IsConnectableToITopHub()`.
Once changed, PR will be technically accepted  @xtophe38

Molkobain (changes requested):

Molkobain:
_Moved to functional review to avoid loosing another month while function is renamed_

Molkobain:
@xtophe38 once method changed, it's ok to merge

piRGoif:
Rebased & force pushed

setup/wizardsteps.class.inc.php
piRGoif: Done !

setup/setuputils.class.inc.php
piRGoif: Renaming done !

PR #337, opened 2022-09-14
N°5073 - Implements line actions in a datatable

bdalsass:
Below is a sample of extra param to enable feature:

		$aExtraParams['row_actions'] = [
			[
				'tooltip'       => 'add an element',
				'icon_classes'  => 'fas fa-plus',
				'js_row_action' => 'console.log("Action ID:");console.log(iActionId);',
			],
			[
				'tooltip'       => 'remove an element',
				'icon_classes'  => 'fas fa-minus',
				'js_row_action' => 'console.log("Action row:");console.log(oTrElement);',
			],
			[
				'tooltip'       => 'open in new tab',
				'icon_classes'  => 'fas fa-external-link-square-alt',
				'js_row_action' => 'console.log("Action table:");console.log(oDatatable);',
			],
			[
				'icon_classes'  => 'fas fa-ellipsis-v',
				'js_row_action' => 'console.log("Action data:");console.log(aData);',
			],
		];


Hipska:
Do you have a screenshot of such an action in action?

bdalsass:
> 

2022-09-15_14h11_50

js/dataTables.main.js
```diff
@@ -78,4 +78,29 @@ function getMultipleSelectionParams(listId)
...
+{
+	let aColumn = {
+		type: "html",
+		orderable: false,
+		render: function ( data, type, row, meta ) {
+			return $(`#${sTableId}_actions_buttons_template`).html();
+		}
+	};
+
+	if(iColumnTargetIndex != -1){
```
Molkobain:
```suggestion
	if (iColumnTargetIndex !== -1) {
```

We rather have strict type comparisons unless it is known that the data type can vary (eg. can be either `'1'` or `1` depending on the consumer).
This avoid future regression like the one we had on _template-base_ for instance.

js/dataTables.main.js
```diff
@@ -78,4 +78,29 @@ function getMultipleSelectionParams(listId)
...
\ No newline at end of file
+}
+
+/**
+ * Return column JSON declaration for row actions.
+ * Could be part of column or columnDefs declaration of datatable.js.
+ *
+ * @param sTableId
+ * @param iColumnTargetIndex
+ * @returns {*}
```
Molkobain, line 89:
```suggestion
 * @returns {*}
 * @since 3.1.0
```

sources/Application/UI/Base/Component/DataTable/DataTableUIBlockFactory.php
```diff
@@ -180,6 +182,39 @@ protected static function RenderDataTable(DataTable $oDataTable, string $sStyle,
...
 
+	/**
+	 * Make a row actions toolbar template.
+	 *
+	 * @param $oTable
+	 *
+	 * @return \Combodo\iTop\Application\UI\Base\Component\Template\Template
+	 * @since 3.1.0
+	 */
+	public static function MakeActionRowToolbarTemplate($oTable)
```
Molkobain: If I understand correctly, `$oTable` will be a `UIBlock` using the `tTableRowActions` trait right?

We can't type hint the argument with the trait, nor can we type hint with a union of the table classes as it is only supported by PHP 8.0+.

Maybe we can type hint with a `\Combodo\iTop\Application\UI\Base\iUIBlock` at least and explicit in the PHPDoc that the object needs to use the said trait. What do you think?

sources/Application/UI/Base/Component/DataTable/StaticTable/FormTable/FormTable.php
```diff
@@ -18,11 +19,13 @@
  */
 class FormTable extends StaticTable
 {
+	use tTableRowActions;
+
```
Molkobain: You don't need this as the trait is already inherited from `StaticTable`

sources/Application/UI/Base/Component/DataTable/StaticTable/StaticTable.php
```diff
@@ -81,6 +83,16 @@ public function GetColumns(): array
 		return $this->aColumns;
 	}
 
+	/**
+	 * Return columns count.
+	 *
+	 * @return int
```
Molkobain, line 89:
```suggestion
	 * @return int
	 * @since 3.1.0
```

sources/Application/UI/Base/Component/Template/Template.php
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * Class Template
+ *
+ * @author Benjamin Dalsass<benjamin.dalsass@combodo.com>
+ * @package Combodo\iTop\Application\UI\Base\Component\Template
+ * @since 3.1.0
+ */
+class Template extends UIContentBlock
+{
+	use tUIContentAreas;
```
Molkobain: If your `Template` UIBlock is very simple and only have 1 area where sub blocks will be added, you don't need the whole `tUIContentAreas` trait overhead, it can be removed. You'll just have to use the AddSubBlock / GetSubBlocks methods.

templates/base/components/datatable/row-actions/handler.js.twig
```diff
@@ -0,0 +1,15 @@
...
+{# @license     http://opensource.org/licenses/AGPL-3.0 #}
+
+// for each row action
+{% if oUIBlock.HasRowActions() %}
+    {% for aAction in oUIBlock.GetRowActions() %}
+        $('#{{ oUIBlock.GetId() }} tbody').on('click', 'button[data-action-id="{{ loop.index0 }}"]', function() {
+            let iActionId = $(this).data('action-id');
+            let oDatatable = $('#{{ oUIBlock.GetId() }}').DataTable();
+            let oTrElement = $(this).closest('tr');
+            let aData = $('#{{ oUIBlock.GetId() }}').DataTable().row(oTrElement).data();
```
Molkobain:
```suggestion
            let aData = oDatatable.row(oTrElement).data();
```

Molkobain (dismissed):

sources/Application/UI/Base/Component/DataTable/DataTableUIBlockFactory.php
bdalsass: 👍I update the code

js/dataTables.main.js
```diff
@@ -78,4 +78,33 @@ function getMultipleSelectionParams(listId)
...
+ * @since 3.1.0
+ * @since 3.1.0
+ * @since 3.1.0
+ * @since 3.1.0
```
Molkobain:
```suggestion
 * @since 3.1.0
```

sources/Application/UI/Base/Component/DataTable/DataTableUIBlockFactory.php
```diff
@@ -180,6 +183,46 @@ protected static function RenderDataTable(DataTable $oDataTable, string $sStyle,
...
+	 *
+	 * @return \Combodo\iTop\Application\UI\Base\Component\Template\Template
+	 * @throws \Exception
+	 * @since 3.1.0
+	 */
+	public static function MakeActionRowToolbarTemplate(iUIBlock $oTable)
+	{
+		// test trait
+		$sTableClass = get_class($oTable);
+		if (!Utils::IsClassUsesDeepTrait($sTableClass, tTableRowActions::class)) {
```
Molkobain:
```suggestion
		if (!utils::IsClassUsesDeepTrait($sTableClass, tTableRowActions::class)) {
```

The `Utils` class with a capital "U" is not the class from iTop core, it's from another collector or extension. It might work but it can be misleading when doing "grep" on the codebase.

sources/Application/UI/Base/Component/DataTable/DataTableUIBlockFactory.php
```diff
@@ -970,24 +1039,44 @@ public static function MakeForForm(string $sRef, array $aColumns, array $aData =
...
+			'display_aliases',
+			/** string comma separated list of class aliases to display */
+			'list_id',
+			/** string list outer id */
+			'selection_enabled',
+			/** list of id in witch select is allowed, if not exists all lines are selectable */
+			'id_for_select',
+			/**give definition of id for select checkbox*/
+			'row_actions',
+			/** array of blocks displayed on every row */
```
Molkobain, line 1079: Is that change on purpose? Seems less readable to me (but it would be fine)

sources/Application/UI/Base/Component/DataTable/StaticTable/StaticTable.php
```diff
@@ -81,6 +83,20 @@ public function GetColumns(): array
...
+	 * @since 3.1.0
+	 * @since 3.1.0
+	 * @since 3.1.0
+	 * @since 3.1.0
```
Molkobain:
```suggestion
	 * @return int
	 * @since 3.1.0
```

application/utils.inc.php
```diff
@@ -3143,18 +3143,53 @@ public static function FormatInitialsForMedallion(string $sInitials): string
...
+	 * @param string $class class to check
+	 * @param string $trait trait to search for
+	 *
+	 * @return bool
+	 * @since 3.1.0
+	 */
+	public static function IsClassUsesDeepTrait(string $class, string $trait): bool
+	{
+		return in_array($trait, self::ClassUsesDeep($class, true));
+  }
```
```suggestion
	/**
	 * Return traits array used by a class and by parent classes hierarchy.
	 *
	 * @see https://www.php.net/manual/en/function.class-uses.php#110752
	 *
	 * @param string $sClass Class to scan
	 * @param bool $bAutoload Autoload flag
	 *
	 * @return array traits used
	 * @since 3.1.0
	 */
	public static function ClassUsesDeep(string $sClass, bool $bAutoload = true): array
	{
		$aTraits = [];
		do {
			$aTraits = array_merge(class_uses($sClass, $bAutoload), $aTraits);
		} while ($sClass = get_parent_class($sClass));
		foreach ($aTraits as $sTrait => $same) {
			$aTraits = array_merge(class_uses($sTrait, $bAutoload), $aTraits);
		}

		return array_unique($aTraits);
	}

	/**
	 * Test trait usage by a class or by parent classes hierarchy.
	 *
	 * @param string $sClass Class to check
	 * @param string $sTrait Trait to search for
	 *
	 * @return bool
	 * @since 3.1.0
	 */
	public static function IsClassUsesDeepTrait(string $sClass, string $sTrait): bool
	{
		return in_array($sTrait, self::ClassUsesDeep($sClass, true));
  }
```

application/utils.inc.php
```diff
@@ -3143,18 +3143,53 @@ public static function FormatInitialsForMedallion(string $sInitials): string
...
+	 *
+	 * @see https://www.php.net/manual/en/function.class-uses.php#110752
+	 *
+	 * @param string $class class to scan
+	 * @param bool $autoload autoload flag
+	 *
+	 * @return array traits used
+	 * @since 3.1.0
+	 */
+	public static function ClassUsesDeep(string $class, bool $autoload = true): array
```
Molkobain: The method name is not very intuitive (to me at least) as it does not mention that it will only return traits.
What about:
  * (Enum|List)ClassTraits
  * (Enum|List)TraitsUsedByClass

application/utils.inc.php
```diff
@@ -3143,18 +3143,53 @@ public static function FormatInitialsForMedallion(string $sInitials): string
...
+	/**
+	 * Test trait usage by a class or by parent classes hierarchy.
+	 *
+	 * @param string $class class to check
+	 * @param string $trait trait to search for
+	 *
+	 * @return bool
+	 * @since 3.1.0
+	 */
+	public static function IsClassUsesDeepTrait(string $class, string $trait): bool
```
Molkobain: More understandable (still for my PoV at least), that being said do you think something like "IsClassUsingTrait" would be more understandable or am I the only that it bothers? 😁

Molkobain (changes requested):
_PS: If you want to apply several suggestion at once you can "add them to batch", then apply them._

application/utils.inc.php
bdalsass: done

application/utils.inc.php
bdalsass: done with IsTraitUsedByClass

application/cmdbabstract.class.inc.php
```diff
@@ -1294,6 +1294,24 @@ public static function GetDisplaySetBlock(WebPage $oPage, DBObjectSet $oSet, $aE
...
+			[
+				'tooltip'       => 'open in new tab',
+				'icon_classes'  => 'fas fa-external-link-square-alt',
+				'js_row_action' => 'console.log("Action table:");console.log(oDatatable);',
+			],
+			[
+				'icon_classes'  => 'fas fa-ellipsis-v',
+				'js_row_action' => 'console.log("Action data:");console.log(aData);',
+			],
+		];
```
Molkobain: Is this for debug purposes?

- Tooltip should be a dict entry and start with a capitalized letter
- JS console outputs should be removed

Molkobain (approved):

steffunky (approved):

application/cmdbabstract.class.inc.php
bdalsass: yes, unwanted commit.
I remove it before merging branch.

PR #331, opened 2022-08-17
N°4947 Fix \EMail::LoadConfig (v2)

piRGoif:
This replaces #278, as the Email class was refactored a lot during 2.7.7.
The goal is to now load the config always using the standard `\utils::GetConfig` method.

PR #338, opened 2022-09-14
N°5797  - Email: Fix wrong config load

piRGoif:
Following #331 I see there are calls to \MetaModel::GetConfig in Send methods in Email children classes.
That means we can't use those methods without having loaded the datamodel !!

There is \EMail::LoadConfig though that don't have this pre-requisite and we should use it !

### DIFF
--- a/sources/Core/Email/EmailLaminas.php
+++ b/sources/Core/Email/EmailLaminas.php
@@ -315,7 +315,8 @@ class EMailLaminas extends Email
 		if ($bForceSynchronous) {
 			return $this->SendSynchronous($aIssues, $oLog);
 		} else {
-			$bConfigASYNC = MetaModel::GetConfig()->Get('email_asynchronous');
+			$oConfig = $this->LoadConfig();
+			$bConfigASYNC = $oConfig->Get('email_asynchronous');
 			if ($bConfigASYNC) {
 				return $this->SendAsynchronous($aIssues, $oLog);
 			} else {
--- a/sources/Core/Email/EmailSwiftMailer.php
+++ b/sources/Core/Email/EmailSwiftMailer.php
@@ -270,13 +270,11 @@ class EmailSwiftMailer extends EMail
 		}
 		else
 		{
-			$bConfigASYNC = MetaModel::GetConfig()->Get('email_asynchronous');
-			if ($bConfigASYNC)
-			{
+			$oConfig = $this->LoadConfig();
+			$bConfigASYNC = $oConfig->Get('email_asynchronous');
+			if ($bConfigASYNC) {
 				return $this->SendAsynchronous($aIssues, $oLog);
-			}
-			else
-			{
+			} else {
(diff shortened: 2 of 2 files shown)


sources/Core/Email/EmailLaminas.php
```diff
@@ -315,7 +315,8 @@ public function Send(&$aIssues, $bForceSynchronous = false, $oLog = null)
 		if ($bForceSynchronous) {
 			return $this->SendSynchronous($aIssues, $oLog);
 		} else {
-			$bConfigASYNC = MetaModel::GetConfig()->Get('email_asynchronous');
+			$oConfig = $this->LoadConfig();
+			$bConfigASYNC = $oConfig->Get('email_asynchronous');
```
Hipska, line 319: Keep it on 1 line?
```suggestion
			$bConfigASYNC = $this->LoadConfig()->Get('email_asynchronous');
```

sources/Core/Email/EmailSwiftMailer.php
```diff
@@ -270,13 +270,11 @@ public function Send(&$aIssues, $bForceSynchronous = false, $oLog = null)
 		}
 		else
 		{
-			$bConfigASYNC = MetaModel::GetConfig()->Get('email_asynchronous');
-			if ($bConfigASYNC)
-			{
+			$oConfig = $this->LoadConfig();
+			$bConfigASYNC = $oConfig->Get('email_asynchronous');
```
Hipska, line 274:
```suggestion
			$bConfigASYNC = $this->LoadConfig()->Get('email_asynchronous');
```

Hipska (approved):

sources/Core/Email/EmailLaminas.php
piRGoif: There are no rule for intermediate variables for now in the dev team... I wrote with my own preference... We'll see what the other think on the next review !

Molkobain (approved):

Molkobain:

sources/Core/Email/EmailLaminas.php
Molkobain: Discussed during the technical review, we don't have strict convention for this yet. We leave it to the developer's discretion.

sources/Core/Email/EmailSwiftMailer.php
Molkobain: Discussed during the technical review, we don't have strict convention for this yet. We leave it to the developer's discretion.

Hipska: Yes indeed, it was only a suggestion. It's useless to assign `$oConfig` if you don't need it the rest of the time..

Molkobain:
Accepted during functional review. Will be part of the upcoming 2.7.9.
@piRGoif will create the corresponding bug as N°4947 was for the previous part.

PR #339, opened 2022-09-21
N°6077 - Attachment class: now creation_date and user_id are init by default

piRGoif:
New fields were added to the class in 2.7.0 with N°330
The creation_date isn't initialized by default, that means every consumer needs to fill it. Currently Mail To Ticket extension doesn't...

This PR adds a `DBInsertNoReload` override in the class so that the field is set by default to the current date.
This would solve part of N°4081

### DIFF
--- a/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
+++ b/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
@@ -110,6 +110,8 @@
     public function DBInsertNoReload()
     {
         $this->SetCurrentDateIfNull('creation_date');
+        $this->SetIfNull('user_id', CMDBChange::GetCurrentUserId());
+
         return parent::DBInsertNoReload();
     }
         ]]></code>
--- a/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
+++ b/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
@@ -102,6 +102,17 @@
         </field>
       </fields>
       <methods>
+        <method id="__construct">
+          <comment></comment>
+          <static>false</static>
+          <access>public</access>
+          <type>Overload-ExNihilo</type>
+          <code><![CDATA[	public function __construct()
+	{
+	  parent::__construct();
+		$this->Set('creation_date', new Date());
+	}]]></code>
+        </method>
         <method id="MapContextParam">
           <comment><![CDATA[/**
 	 * Maps the given context parameter name to the appropriate filter/search code for this class
--- a/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
+++ b/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
@@ -110,7 +110,7 @@
     public function DBInsertNoReload()
     {
         $this->SetIfNull('creation_date', time());
-        $iKey = parent::DBInsertNoReload();
+        return parent::DBInsertNoReload();
     }
         ]]></code>
         </method>
--- a/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
+++ b/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
@@ -109,7 +109,7 @@
           <code><![CDATA[
     public function DBInsertNoReload()
     {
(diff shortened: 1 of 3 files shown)


Hipska:
Why not implemented as `$this->SetCurrentDate('creation_date');`?

Or even better, as a lot of other datamodel classes already do, as `$this->SetIfNull('creation_date', time());` in an overload of `OnInsert` or `DBInsertNoReload`.

piRGoif:
Indeed you're right, second solution is what is made everywhere in the datamodel ! 
I will switch to it !

datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
```diff
@@ -101,6 +101,18 @@
...
       <methods>
+        <method id="DBInsertNoReload">
+          <static>false</static>
+          <access>public</access>
+          <type>Overload-DBObject</type>
+          <code><![CDATA[
+    public function DBInsertNoReload()
+    {
+        $this->SetIfNull('creation_date', time());
+        $iKey = parent::DBInsertNoReload();
```
Hipska:
```suggestion
        return parent::DBInsertNoReload();
```

piRGoif: 😱
Thanks, I didn't saw that :(

Hipska (approved):

datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
```diff
@@ -101,6 +101,18 @@
...
       </fields>
       <methods>
+        <method id="DBInsertNoReload">
+          <static>false</static>
+          <access>public</access>
+          <type>Overload-DBObject</type>
+          <code><![CDATA[
+    public function DBInsertNoReload()
+    {
+        $this->SetIfNull('creation_date', time());
```
Molkobain: During support/3.0 merge, use the ~`SetComputedDateIfNull()`~ `SetCurrentDateIfNull()` instead.

Molkobain (approved):

Molkobain:

Molkobain:
Functional review: Pierre will run the audit to list impacted customers. As we tend to not make any datamodel change in a maintenance release, it might be merged in develop only.

piRGoif:
Didn't find any Attachment children in the clients customization (details are in N°4081)
Setting this PR to be reviewed for next functional review !

datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
piRGoif: I didn't find any use of this method and don't understand its phpdoc ? 
What are its benefits ? How should we use it ?

Molkobain: Wrong copy/paste, I meant `SetCurrentDateIfNull()`

Molkobain:
Accepted during functional review but for branch develop instead of support/2.7.
As this is addressing a non blocking bug with a partial fix (other part must be done in the Mail To Ticket Automation extension) we rather stick to our retrofitting guidelines and don't make a datamodel change in a minor version.

To be done before next technical review then merge:
- [x] Add default user_id in the Attachment creation as a fallback in case the consumer didn't pass it.
- [x] Check whereas we need to remove the creation date being set in the portal like it was done for the backoffice
- [x] Rebase PR onto develop branch
- [x] Update PR's title with the bug number in our internal tracker

piRGoif:
Rebased branch + PR on develop

datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
Hipska: I also think that might be a better one to use..

piRGoif: Done in 64b7abf3 (target branch is now develop)

piRGoif:
PR ready to review again

Hipska (approved):

Molkobain:
You can merge Pierre.

jbostoen:
Please also see https://sourceforge.net/p/itop/tickets/2167/

PR #340, opened 2022-09-27
:bug: N°5703 - Fix navigation menu drawer under dashlets on Safari

larhip:
Hi there,

I know officially you are not supporting the Browser Safari and I am also aware of the facts that the following fix seems to be more related to an issue with Safari than with iTop.

But anyhow since Safari is a widely used browser (at least among Mac users), this PR should workaround the issue regarding the hidden menu bar (see also: https://sourceforge.net/p/itop/discussion/itop-hub/thread/fc97ec4a10/?limit=25#1cc0/9930) it seems to be related to an issue with the z-index (https://stackoverflow.com/questions/40895387/z-index-not-working-on-safari-fine-on-firefox-and-chrome)

With this small css snippet in this PR the issue is gone, I should not have an impact on other browsers.

Reference: https://sourceforge.net/p/itop/tickets/2104/


Molkobain:
Hello Lars,

Can you add a screenshot of the visual issue before the fix.
And can you add screenshots of Safari / Chrome / Firefox working correctly after the fix.

As this is a workaround for an unsupported browser, we might consider it because you are a valued contributor, but only if it doesn't affect the others. Also, in that kind of fix we like to add a comment explaining why we did that and where the solution came from. I'll make a suggestion in the PR

Cheers,
Guillaume

Molkobain:
Can you also move the fix to the `css/backoffice/_shame.scss` file please?

css/backoffice/layout/_navigation-menu.scss
```diff
@@ -274,6 +274,7 @@ $ibo-navigation-menu--user-info--height--is-expanded: 100% !default;
 	&.ibo-is-active{
 		.ibo-navigation-menu--drawer{
 			right: calc(-1 * #{$ibo-navigation-menu--drawer--width});
+			-webkit-transform:translate3d(0,0,0);
```
Molkobain:
```suggestion
			/* Workaround for a Safari (unsupported browser) issue: https://sourceforge.net/p/itop/discussion/itop-hub/thread/fc97ec4a10/?limit=25#1cc0/9930 */
			/* Fix found here: https://stackoverflow.com/questions/40895387/z-index-not-working-on-safari-fine-on-firefox-and-chrome */
			-webkit-transform: translate3d(0,0,0);
```

larhip:
Thanks for your fast feedback @Molkobain !

Moved the fix into `_shame.scss` also added your suggested comments.

Behavior in safari **before** the fix:
safari_before

Behavior in safari **after** the fix:
safari_after

Behavior in Firefox after the fix (still the same):
firefox_after

Behavior in Chrome after the fix (still the same):
chrome_after


Enjoy your community Event! :v:

css/backoffice/_shame.scss
```diff
@@ -43,3 +43,9 @@
 .ibo-svg-illustration--container > svg *[fill="#6c63ff"]{
 	fill: $ibo-svg-illustration--fill;
 }
+
+// Workaround for a Safari (unsupported browser) issue: https://sourceforge.net/p/itop/discussion/itop-hub/thread/fc97ec4a10/?limit=25#1cc0/9930
+// Fix found here: https://stackoverflow.com/questions/40895387/z-index-not-working-on-safari-fine-on-firefox-and-chrome
+.ibo-navigation-menu.ibo-is-active .ibo-navigation-menu--drawer{
+    -webkit-transform:translate3d(0,0,0);
```
Molkobain:
```suggestion
    -webkit-transform: translate3d(0,0,0);
```

Molkobain (changes requested):

Molkobain:
Thanks for the update Lars, it will help pushing this forward.
Will keep you updated early october during the next technical review.

> Enjoy your community Event! ✌️

I could not make it this year 😭😭
But I'm sure it was great 😊

Cheers!
Guillaume

Molkobain (approved):

piRGoif:
Just updated the description to add the SF ticket reference.
Many thanks for the fix Lars !

There is also [another ticket on a similar prb with Microsoft Edge](https://sourceforge.net/p/itop/tickets/2114/), I'll have a look to check if your fix address it.

larhip:
> There is also [another ticket on a similar prb with Microsoft Edge](https://sourceforge.net/p/itop/tickets/2114/), I'll have a look to check if your fix address it.

I am afraid my fix will not address that issue as well, since I just added some css for the navigation menu not for any other components. But for sure it might be the MS Edge problem got a similar root cause (something related to z-index...)
To be honest: I am not 100 percent sure whether I understood the concept of z-index vs translateZ in all details 🙈

css/backoffice/_shame.scss
```diff
@@ -43,3 +43,9 @@
 .ibo-svg-illustration--container > svg *[fill="#6c63ff"]{
 	fill: $ibo-svg-illustration--fill;
 }
+
+// Workaround for a Safari (unsupported browser) issue: https://sourceforge.net/p/itop/discussion/itop-hub/thread/fc97ec4a10/?limit=25#1cc0/9930
+// Fix found here: https://stackoverflow.com/questions/40895387/z-index-not-working-on-safari-fine-on-firefox-and-chrome
+.ibo-navigation-menu.ibo-is-active .ibo-navigation-menu--drawer{
+    -webkit-transform: translate3d(0,0,0);
```
Hipska: Not sure if WebKit needs to be addressed specifically, I think this is also good:
```suggestion
    transform: translate3d(0,0,0);
```

larhip: Webkit need not to be addressed specifically (transform works as well), but the idea was to address specifically since there is no need for this change i.e. for firefox

Hipska: No need, but also doesn't harm. Next to that I think the `-webkit` prefix is also relevant for Chrome, Edge and Opera.

larhip: think you are right Chrome and Edge impacted by `-webkit` as well and I am with this change doesn't harm.
But see you any benefit in using `transform` instead of `-webkit-transform`?

Hipska: Using web standards instead of proprietary implementations? Also that would eliminate possible future FF only bug by not having this specific directive for that browser. (So all browsers have the same settings for these classes)

larhip: Is `-webkit-transform` a proprietary implementations?
Anyhow I am absolutely fine with your suggestion but would be interested in @piRGoif or @Molkobain opinion here?

Hipska: Any browser prefixed one is a proprietary implementation that might or might not follow the standard. Mostly that is because they exist before it became a standard..

piRGoif: Hello,
Indeed I'm not sure what would be the right decision here... We already tried to follow web standards as best as we can considering the legacy code, but a bug like this happened anyway...

I'm more inclined with Lars way of thinking (if it works, don't upgrade ;) ), but I understand your point Thomas (same code for all of the browsers)

We'll debate what would be the best option during the technical review !

Molkobain: It's fine by me as long as it still works as intended on Firefox (the only non-webkit based supported browsers).
Also, did you went through most of the backoffice GUI to make sure it did not introduce any glitch on other parts such as modals, datatables etc?

Molkobain: @larhip ? Functional review is next week.

Molkobain: PS: Generic property would be great if no impact on FF / Chrome.

Molkobain:
Accepted during technical review (pending the change / answer on the generic property instead of the webkit one)

css/backoffice/_shame.scss
larhip: So I tested both possibilities webkit and non-webkit implementations, and checke on Safari, Chrome and Firefox not only the menu but also the detail views, list views, ...

I was not able to see any difference using webkit or not, and I was also not able to see any glitches or something similar on any view.

> Accepted during technical review (pending the change / answer on the generic property instead of the webkit one)

When I got your feedback right you prefer the Webstandard way (without webkit)? 
Since this change seems to have no impact to firefox I would be fine with that as well.
Could you give me a brief feedback, so I could adapt it or not?

Molkobain: Yes property `transform` is preferred over `-webkit-transform` :)

_Edit: I might have messed something up in my comments, some of yesterday messages appear as of just now._

Molkobain:
On a side note, @BenGrenoble found this [macOS emulation on Linux](https://github.com/arindas/mac-on-linux-with-qemu.git). We are currently trying to reproduce the bug on it.
Even though we won't do extensive testing on Safari, it could be a good way to reproduce bugs you report to ease the acceptation process.

Hipska (approved):

css/backoffice/_shame.scss
larhip: Done ✅ 

> Edit: I might have messed something up in my comments, some of yesterday messages appear as of just now.

Thanks for the note I was already confused about conversation history :-)

Molkobain:
Will be part of iTop 3.1

I'll create the corresponding bug in our tracker and update the PR's title.

logicsthinkaboutit:
hey,
Why didn't you include the existing and known error as a bug fix in the new 3.0.3 version? Now the error is there again.

piRGoif:
> Why didn't you include the existing and known error as a bug fix in the new 3.0.3 version? 

Hello,
This was the product team decision.
As said widely in this PR and on the various SourceForge discussions, we don't have ways to reproduce the bug and no Apple products are on our QA chain. When we cannot guarantee the quality of a bugfix, we prefer to include it only in next major release, as it allows wider manual testing.

PR #343, opened 2022-10-01
Activity panel: Fix compose button opening log picker menu when only 1 log available

Molkobain:
## Symptom

When on an object with only 1 log in the activity panel, if you click on the "Compose" button when being on the "Activity" tab, the log picker is displayed with only one item, forcing us to an avoidable extra click.

// TODO: Insert GIF of problem

## Proposition

When on an object with only 1 log in the activity panel, open the log directly when "Compose" button is pressed.

// TODO: Insert GIF of solution

## Arguments in favor

  * Better UX as it reduces the number of clicks to get to the log form

## Reproduction

Use the following XML delta which adds a log on the `Person` class.

```
<itop_design xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="3.0">
  <classes>
    <class id="Person">
      <fields>
        <field id="log" xsi:type="AttributeCaseLog" _delta="define">
          <sql>log</sql>
          <default_value/>
        </field>
      </fields>
      <presentation>
        <details>
          <items>
            <item id="log" _delta="define">
              <rank>100</rank>
            </item>
          </items>
        </details>
      </presentation>
    </class>
  </classes>
</itop_design>
```

### DIFF
--- a/dictionaries/fr.dictionary.itop.ui.php
+++ b/dictionaries/fr.dictionary.itop.ui.php
@@ -496,7 +496,7 @@ Nous espérons que vous aimerez cette version autant que nous avons eu du plaisi
 	'UI:Menu:Actions' => 'Actions',
 	'UI:Menu:OtherActions' => 'Autres Actions',
 	'UI:Menu:Transitions' => 'Transitions~~',
-	'UI:Menu:OtherTransitions' => 'Other Transitions~~',
+	'UI:Menu:OtherTransitions' => 'Autres Transitions',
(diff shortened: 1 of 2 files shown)


piRGoif:
I think this is replaced by #344 ? setting the appropriate label for this one

Molkobain:
> I think this is replaced by #344 ? setting the appropriate label for this one

Indeed, thanks. I tried to rebase it on support/3.0 locally but failed to do so.

PR #344, opened 2022-10-01
N°5646 - Activity panel: Fix compose button opening log picker menu when only 1 log available

Molkobain:
## Symptom

When on an object with only 1 log in the activity panel, if you click on the "Compose" button when being on the "Activity" tab, the log picker is displayed with only one item, forcing us to an avoidable extra click.

image

## Proposition

When on an object with only 1 log in the activity panel, open the log directly when "Compose" button is pressed.

## Arguments in favor

  * Better UX as it reduces the number of clicks to get to the log form

## Reproduction

Use the following XML delta which adds a log on the `Person` class.

```
<itop_design xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="3.0">
  <classes>
    <class id="Person">
      <fields>
        <field id="log" xsi:type="AttributeCaseLog" _delta="define">
          <sql>log</sql>
          <default_value/>
        </field>
      </fields>
      <presentation>
        <details>
          <items>
            <item id="log" _delta="define">
              <rank>100</rank>
            </item>
          </items>
        </details>
      </presentation>
    </class>
  </classes>
</itop_design>
```

### DIFF
--- a/js/layouts/activity-panel/activity-panel.js
+++ b/js/layouts/activity-panel/activity-panel.js
@@ -747,7 +747,7 @@ $(function()
 
 			// - Helpers on case logs entry forms
 			/**
-			 * @returns {integer} The number of caselog entry form
+			 * @returns {integer} The number of caselog entry forms
 			 * @private
 			 * @since 3.1.0
 			 */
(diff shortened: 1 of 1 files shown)


js/layouts/activity-panel/activity-panel.js
```diff
@@ -737,6 +746,14 @@ $(function()
 			},
 
 			// - Helpers on case logs entry forms
+			/**
+			 * @returns {integer} The number of caselog entry form
```
Molkobain:
```suggestion
			 * @returns {integer} The number of caselog entry forms
```

steffunky (approved):

PR #345, opened 2022-10-01
N°5681 - Add support for "Ctrl + Enter" and "Meta (Cmd) + Enter" submit on multi-line fields

Molkobain:
## Use case

Add support for "Ctrl + Enter" or "Cmd + Enter" submit on multi-line fields such as `AttributeText` and `AttributeHTML`.

## Arguments in favor

  * To me this is one of the small frustrating things when using iTop on a daily basis, being unable to submit a form after filling a multi-line field; being forced to use the mouse to click on "Submit".
  * Align behavior with single-line fields behavior for which hitting the "Enter" key submits the form.
  * Productivity improvement for "keyboard people". For example try to create a new UserRequest in the backoffice:
    * Fill the organization, hit "Tab"
    * Fill the caller, hit "Tab",
    * Fill the title, hit "Tab",
    * Fill the description and finally now that you want to submit the form, you can't hit "Enter" ou "Ctrl + Enter"
  * This is a common behavior among popular apps:
    * Email clients to send an email (Outlook, Thunderbird, GMail, ...)
    * IDEs to submit dialogs such as commit dialogs (PHPStorm, IntelliJ, GitHub web, ...)
    * Jira to submit edits on a task description
    * ...

## Tested use cases

  * Backoffice
    * Submit of an object `AttributeText` (`FAQ->summary`)
    * Submit of an object `AttributeHTML` (`UserRequest->description`)
    * Submit of an object log in edition (`UserRequest->public_log`)
    * Submit of an object log in read-only (`UserRequest->public_log`)
    * Submit of an object single-line attribute (`UserRequest->title`)
  * Portal
    * Submit of an object log in edition (`UserRequest->public_log`)

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -2278,7 +2278,7 @@ JS
 {$sValidationSpan}{$sReloadSpan}
 HTML;
 					$oPage->add_ready_script(
-						<<<EOF
+						<<<JS
                         $('#$iId').closest('.field_input_text').find('.fullscreen_button').on('click', function(oEvent){
                             var oOriginField = $('#$iId').closest('.field_input_text');
                             var oClonedField = oOriginField.clone();
@@ -2291,7 +2291,14 @@ HTML;
                                 oOriginField.find('textarea').triggerHandler('change');
                             });
                         });
-EOF
+                        
+                        // Submit host form on "Ctrl + Enter" or "Meta (Cmd) + Enter" keyboard shortcut
+                        $('#$iId').on('keyup', function (oEvent) {
+                            if ((oEvent.ctrlKey || oEvent.metaKey) && oEvent.key === 'Enter') {
+                                $(this).closest('form').trigger('submit');
+                            }
+                        });
+JS
 					);
 				break;
 
--- a/datamodels/2.x/itop-portal-base/portal/public/js/portal_form_handler.js
+++ b/datamodels/2.x/itop-portal-base/portal/public/js/portal_form_handler.js
@@ -58,6 +58,8 @@ $(function()
 			if((this.options.cancel_url !== null) && (this.options.cancel_url !== ''))
 				this.options.cancel_rule.url = this.options.cancel_url;
 
+			this._bindEvents();
+
 			this._super();
 		},
    
@@ -79,6 +81,16 @@ $(function()
 		{
 			this._super( key, value );
 		},
+		_bindEvents: function() {
+			const me = this;
+
+			// Submit event from the form should be treated as a click on the submit button
+			// as it processes things before sending the request
+			this.element.on('submit', function(oEvent) {
+				me._onSubmitClick(oEvent);
+			});
+		},
+
 		// - Callback when some fields have been touched
 		_onFieldsTouched: function(oEvent)
 		{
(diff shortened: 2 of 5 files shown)


Molkobain:
Hello @Hipska , could you try on macOS if this works or has any side effect?
Thanks

Molkobain:
@steffunky can you tell me more about the lib you use for hotkeys in the 3.0?

steffunky:
> @steffunky can you tell me more about the lib you use for hotkeys in the 3.0?

We used Mousetrap js in 3.0 to handle hotkeys.
Care as I'm not sure it handles Ctrl value for Mac (as it considers right click on Mac as "Ctrl" value)

Hipska:
This is what that tool says on their reference page:
>  ### Combination of keys
>  ```js
>  Mousetrap.bind('ctrl+s', function(e) {
>      _saveDraft();
>  });
> ```
>  If you want to bind multiple key commands to the same callback you can pass in an array for the first argument:
>  ```js
>  Mousetrap.bind(['ctrl+s', 'command+s'], function(e) {
>      _saveDraft();
>  });
> ```
>  Mousetrap 1.4 introduced a generic `mod` helper which lets you set cross platform shortcuts.
>  ```js
>  Mousetrap.bind('mod+s', _saveDraft);
> ```
>  On Mac this ends up mapping to `command+s` whereas on Windows and Linux it maps to `ctrl+s`.
>  
>  This differs from the array example above because there both `ctrl+s` and `command+s` will trigger save on Mac whereas with the `mod` helper only `command+s` will.
>  

This is anyway not relevant as  you want specifically `ctrl+enter`, so that should work. Also `control` can indeed be used for right click, but only when clicking while pressing that button. You don't click the mouse when pressing `control` and `enter`.

piRGoif:
When we did the review I was thinking about `js/jquery.hotkeys.js` that is here since 684e9e35 : I introduced it to handle... Ctrl+Enter shortcut in the run query screen :)

Honestly I don't remember why I added this lib ? It seems to handle the shortcut directly in the JQuery `on()` binding (third parameter) ? Not clear for me the lib does that reading [its README](https://github.com/tzuryby/jquery.hotkeys) :/ 

@Hipska do you confirm the code is working on your Mac ? 
On this platform, is the Ctrl+Enter shortcut is as popular as it is on PC ?

Hipska:
Did not try this branch (no time/ not setup a easy testing method for this).

For example on Slack `shift+enter` is used to add a newline in a message and none of the other shortcuts use the control key.

GitHub:
image

piRGoif:
Thanks ! Indeed the GitHub shortcut is exactly what we needed.
We'll have to test if the PR JS code triggers when doing Command+Enter on Mac...

Molkobain:
Now supporting "Cmd + Enter" on MacOS.

**What was done:**
Log entry is now submitted on either "Ctrl + Enter" or "Meta + Enter".
"Meta" key means "Windows" key on Windows / Unix, "Command" key on MacOS.

|                       | Windows / Unix | MacOS |
|----------------|------------------|----------|
| Ctrl + Enter    | Submit              | Submit  |
| Meta + Enter | Do nothing       | Submit   |

application/cmdbabstract.class.inc.php
```diff
@@ -2291,6 +2291,13 @@ public static function GetFormElementForField($oPage, $sClass, $sAttCode, $oAttD
                                 oOriginField.find('textarea').triggerHandler('change');
                             });
                         });
+                        
+                        // Submit host form on "Ctrl + Enter" keyboard shortcut
+                        $('#$iId').on('keyup', function (oEvent) {
+                            if (oEvent.ctrlKey && oEvent.key === 'Enter') {
```
Molkobain: Meta key missing

piRGoif:
Can you also update the run query screen to use the same code ? 
And remove the jquery hotkeys lib ?

Molkobain:
> Can you also update the run query screen to use the same code ? And remove the jquery hotkeys lib ?

I won't do this on this PR, it's not part of the object forms. Also it would require deprecating the hotkeys.js file, migrating the possible current usages and so on. So I rather keep it to the PR original purpose.

Molkobain:
Accepted during function review.

PR #347, opened 2022-10-01
N°5598 - Add support for PHP traits in XML datamodel classes

Molkobain:
## Objective

Add support for PHP traits in XML datamodel classes

_Note that this is different than [Jeffrey's ticket](https://sourceforge.net/p/itop/tickets/2022) on "reusable XML snippets". Here the need is to be able to use an existing PHP trait in an XML class, exactly like we already do with the `php_parent` node_

## Use case

In the `molkobain-datacenter-view` ([here](https://github.com/Molkobain/itop-datacenter-view/blob/master/src/molkobain-datacenter-view/datamodel.molkobain-datacenter-view.xml#L138), [here](https://github.com/Molkobain/itop-datacenter-view/blob/master/src/molkobain-datacenter-view/datamodel.molkobain-datacenter-view.xml#L296) and [here](https://github.com/Molkobain/itop-datacenter-view/blob/master/src/molkobain-datacenter-view/datamodel.molkobain-datacenter-view.xml#L597)) and `molkobain-datacenter-view-extended` (repo. not public) modules I add several methods (4) to different classes (6), their codes are the same for all the classes. \
For now I have to duplicate it on each class (or make an indirection to call a factorized method, but still I have to declare the method in the XML) which is not a great DX at all.

With this we could use a unique PHP trait for all these XML classes:
  * No redundant code, it's all factorized in the trait
  * Less XML deltas for the same result
  * => Better / easier maintenance

It would also bring XML DM classes a step closer to what PHP DM classes can do / be.

## TODOs once technical principle accepted

  * [ ] Unit tests on the PHP traits

## TODOs on PR is merged

  * [ ] Documentation: Update XML reference
  * [ ] Documentation: Don't forget to explicit that trait MUST be either:
    * Fully qualified
    * Loadable via autoloader or already loaded by another mean (eg. `require_once`)

## Example

Here is what it would look like in the XML (to be compared with links of what it currently looks like in `molkobain-datacenter-view`)

```
<itop_design xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="3.1">
  <classes>
    <class id="Enclosure">
      <php_traits>
        <php_trait id="example" _delta="define">\Molkobain\iTop\Extension\DatacenterView\Model\tPositioningMethods</php_trait>
      </php_traits>
    </class>
    <class id="DatacenterDevice">
      <php_traits>
        <php_trait id="example" _delta="define">\Molkobain\iTop\Extension\DatacenterView\Model\tPositioningMethods</php_trait>
      </php_traits>
    </class>
    <class id="PDU">
      <php_traits>
        <php_trait id="example" _delta="define">\Molkobain\iTop\Extension\DatacenterView\Model\tPositioningMethods</php_trait>
        <php_trait id="another-example" _delta="define">tSomeCoreTrait</php_trait>
      </php_traits>
    </class>
  </classes>
</itop_design>
```

And in a compiled PHP class

```
class PDU extends PowerConnection
{
	use \Molkobain\iTop\Extension\DatacenterView\Model\tPositioningMethods;
	use tSomeCoreTrait;

	public static function Init()
	{
		[...]
	}

    [...]
}
```

### DIFF
--- a/setup/compiler.class.inc.php
+++ b/setup/compiler.class.inc.php
@@ -2246,8 +2246,10 @@ EOF;
 		//
 		$sClassName = $oClass->getAttribute('id');
 		$bIsAbstractClass = ($oProperties->GetChildText('abstract') == 'true');
-		$oPhpParent = $oClass->GetUniqueElement('php_parent', false);
 		$aRequiredFiles = [];
+
+		// - PHP parent
+		$oPhpParent = $oClass->GetUniqueElement('php_parent', false);
 		if ($oPhpParent)
 		{
 			$sParentClass = $oPhpParent->GetChildText('name', '');
@@ -2271,6 +2273,19 @@ EOF;
 		{
 			$sParentClass = $oClass->GetChildText('parent', 'DBObject');
 		}
+
+		// - PHP traits
+		$aPhpTraits = [];
+		$oPhpTraits = $oClass->GetUniqueElement('php_traits', false);
+		if ($oPhpTraits) {
+			foreach ($oPhpTraits->getElementsByTagName('php_trait') as $oPhpTrait) {
+				$sPhpFQTN = $oPhpTrait->GetText();
+				if (utils::IsNotNullOrEmptyString($sPhpFQTN)) {
+					$aPhpTraits[] = $sPhpFQTN;
+				}
+			}
+		}
+
 		$sInitMethodCalls =
 			<<<EOF
 $sAttributes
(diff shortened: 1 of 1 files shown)


Molkobain:
Accepted during technical review, that being said Romain would like to discuss it a bit more.

I'll create the corresponding bug in the internal bug tracker, as well as the corresponding future ITSM Designer bug to enable the feature in it.

Meanwhile the PR is kept in "Technical review needed" for future discussion with Romain.

favincen:
I also already had to duplicate identical methods into different classes that do not share a parent (nothing below cmdbAbstractObject class). I support the idea 👍 .
If this PR is validated and committed, it would be great to allow usage in ITSM-Designer (creating traits in PHP snippets and allow to reference them if a class). But that would be a separate enhancement request, on an another product right ? (just don't know how to register that)

NB: Not shure this first contribution is properly done. Let me know if somehow I misbehave on the way to contribute to a PR discussion, so I can improve 😃

Molkobain:
Hello @favincen , great to see that you have the use as well! Once this is accepted / merged, you can totally create a request on the support portal for the support of this feature in the ITSM Designer :)

Molkobain:
PR closed, we'll rather follow the "XML templates" lead, main reasons:
- We want to keep an abstraction layer between the XML and the compiled code as much as possible to ease migrations and "under the hood evolutions"
- It's not in tune with "reducing the open bar APIs"

Main drawback of the "XML templates":
- It won't allow to bring class/trait properties for usage in the methods. But as it is a corner case, we feel like it's ok.

PR #351, opened 2022-10-07
N°2013 - Setup: Cannot execute if existing config file contains an inaccessible MySQL server

Hipska:
When running setup to update iTop or install new extensions, but also want to change the DB settings at the same time an error occurs at the module selection wizard steps as it tries to connect to the old database (which might not exist) instead of the new one given a few steps before.

This seems to be like this since at least iTop 2.5.0

### DIFF
--- a/setup/wizardsteps.class.inc.php
+++ b/setup/wizardsteps.class.inc.php
@@ -1348,6 +1348,18 @@ class WizStepModulesChoice extends WizardStep
 		if ($sConfigPath !== null)
 		{
 			$oConfig = new Config($sConfigPath);
+
+			$aParamValues = array(
+				'db_server' => $oWizard->GetParameter('db_server', ''),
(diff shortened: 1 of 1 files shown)


piRGoif:
Hello,
Many thanks for this fix ! 
It was already known (N°2013) but no fix was available for now.

piRGoif (approved):

piRGoif:
Discussed [on the community slack](https://itopbycombodo.slack.com/archives/C02D0C7NNQG/p1665141636698289) : seems like a regression brought by https://github.com/Combodo/iTop/commit/ca28eeb596465ce164526c609bb6e6935195e7be#diff-8c55850df4e71248c227c080e065cf48f4c8bae63ac7e3bcb2864a0a5a20b876L1246-L1260

Molkobain:

Molkobain:

PR #341, opened 2022-10-01
N°4703 - Add "chat" / "in person" as possible "origin" value for tickets

Molkobain:
## Objective

Follow-up of SourceForge [ticket #2019](https://sourceforge.net/p/itop/tickets/2019/), adding the "chat" and "in_person" origin to `Ticket`/`UserRequest`/`Incident` classes to best reflect reality of modern helpdesk.

## TODOs once PR is merged

  * [ ] Add migration audit check

PR #352, opened 2022-10-10
N°4703 - Add Dutch translations to new origins

Hipska:
Adds Dutch translations to #341


piRGoif (approved):

piRGoif:
Note : origin attribute is defined on each children classes

Hipska:
@piRGoif I did exactly the same as original #341.

piRGoif:
> I did exactly the same as original

Yes :) I didn't recall this attribute was defined in children classes, so I just put a note so everyone won't spent time figuring why the translations are defined at the children class level :)

Molkobain (approved):

Molkobain:
@piRGoif as this is just about new dict. entries, maybe we can move it directly to functional review as we already both reviewed it. What do you think?

steffunky (approved):

Molkobain:

PR #354, opened 2022-10-14
N°5622 - Do not verify ca in mysqldump if not configured

larhip:
### Symptom

Using mysql-client (not MariaDB) during the backup the CA is always verified even if no CA Path is configured.

_Current behavior_
image
image

_Expected behavior_
Backup should be successful using mysql client and ssl without configured ca

### Reproduction

- Setup iTop (3.0 or 2.7) with separate web and database host
- Install mysql-client (not mariadb) on web server
- Add the config setting `'db_tls.enabled' => true`
- Do _not_ add `db_tls.ca` or leave it empty

### Cause

iTop adds the mysql option `--ssl-mode=VERIFY_CA` in case of mysql and tls enabled, but do not check whether ca is configured.
If no ca path is configured it should be `--ssl-mode=REQUIRED` instead of `VERIFY_CA`.

### DIFF
--- a/setup/backup.class.inc.php
+++ b/setup/backup.class.inc.php
@@ -464,13 +464,13 @@ EOF;
 	 * @param Config $oConfig
 	 *
 	 * @return string TLS arguments for CLI programs such as mysqldump. Empty string if the config does not use TLS.
+	 * @throws \MySQLException
 	 *
-	 * @uses \CMDBSource::GetDBVendor() so needs a connection opened !
-	 * @uses \CMDBSource::GetDBVersion() so needs a connection opened !
+	 * @uses \CMDBSource::IsSslModeDBVersion() so needs a connection opened !
 	 *
 	 * @since 2.5.0 N°1260
 	 * @since 2.6.2 2.7.0 N°2336 Call DB to get vendor and version (so CMDBSource must be init before calling this method)
-	 * @link https://dev.mysql.com/doc/refman/5.6/en/connection-options.html#encrypted-connection-options "Command Options for Encrypted Connections"
+	 * @link https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#encrypted-connection-options Command Options for Encrypted Connections
 	 */
 	public static function GetMysqlCliTlsOptions($oConfig)
 	{
@@ -480,13 +480,17 @@ EOF;
 			return '';
 		}
 		$sTlsOptions = '';
-
-		$sDBVendor = CMDBSource::GetDBVendor();
-		$sDBVersion = CMDBSource::GetDBVersion();
-		$sMysqlSSLModeVersion = '5.7.0'; //Mysql 5.7.0 and upper deprecated --ssl and uses --ssl-mode instead
-		if ($sDBVendor === CMDBSource::ENUM_DB_VENDOR_MYSQL && version_compare($sDBVersion, $sMysqlSSLModeVersion, '>='))
+		// Mysql 5.7.11 and upper deprecated --ssl and uses --ssl-mode instead
+		if (CMDBSource::IsSslModeDBVersion())
 		{
-			$sTlsOptions .= ' --ssl-mode=VERIFY_CA';
+			if(empty($oConfig->Get('db_tls.ca')))
+			{
+				$sTlsOptions .= ' --ssl-mode=REQUIRED';
+			}
+			else
+			{
+				$sTlsOptions .= ' --ssl-mode=VERIFY_CA';
+			}
 		}
 		else
 		{
--- a/test/setup/DBBackupTest.php
+++ b/test/setup/DBBackupTest.php
@@ -61,8 +61,16 @@ class DBBackupTest extends ItopTestCase
 		$oConfigToTest->Set('db_tls.enabled', true);
 		$sCliArgsMinCfg = DBBackup::GetMysqlCliTlsOptions($oConfigToTest);
 
-		// depending on the MySQL version, we would have `--ssl` or `--ssl-mode=VERIFY_CA`
-		$this->assertStringStartsWith(' --ssl', $sCliArgsMinCfg);
+		// depending on the MySQL vendor, we would have `--ssl` or `--ssl-mode=REQUIRED`
+		if (CMDBSource::IsSslModeDBVersion())
+		{
+			$this->assertStringStartsWith(' --ssl-mode=REQUIRED', $sCliArgsMinCfg);
+		}
+		else
+		{
+			$this->assertStringStartsWith(' --ssl', $sCliArgsMinCfg);
+			$this->assertStringNotContainsString('--ssl-mode', $sCliArgsMinCfg);
+		}
 	}
 
 	/**
@@ -81,7 +89,17 @@ class DBBackupTest extends ItopTestCase
 		$oConfigToTest->Set('db_tls.ca', $sTestCa);
 		$sCliArgsCapathCfg = DBBackup::GetMysqlCliTlsOptions($oConfigToTest);
 
-		$this->assertStringStartsWith(' --ssl', $sCliArgsCapathCfg);
+		// depending on the MySQL vendor, we would have `--ssl` or `--ssl-mode=VERIFY_CA`
+		if (CMDBSource::IsSslModeDBVersion())
+		{
+			$this->assertStringStartsWith(' --ssl-mode=VERIFY_CA', $sCliArgsCapathCfg);
+		}
+		else
+		{
+			$this->assertStringStartsWith(' --ssl', $sCliArgsCapathCfg);
+			$this->assertStringNotContainsString('--ssl-mode', $sCliArgsCapathCfg);
+
+		}
 		$this->assertStringEndsWith('--ssl-ca='.DBBackup::EscapeShellArg($sTestCa), $sCliArgsCapathCfg);
 	}
 }
--- a/core/cmdbsource.class.inc.php
+++ b/core/cmdbsource.class.inc.php
@@ -1611,4 +1611,22 @@ class CMDBSource
 
 		return 'ALTER DATABASE'.CMDBSource::GetSqlStringColumnDefinition().';';
 	}
+
+	/**
+	 * Check which mysql client option (--ssl or --ssl-mode) to be used for encrypted connection
+	 *
+	 * @return bool true if --ssl-mode should be used, false otherwise
+	 * @throws \MySQLException
+	 *
+	 * @link https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#encrypted-connection-options "Command Options for Encrypted Connections"
+	 */
+	public static function IsSslModeDBVersion()
+	{
+		if (static::GetDBVendor() === static::ENUM_DB_VENDOR_MYSQL)
+		{
+			//Mysql 5.7.0 and upper deprecated --ssl and uses --ssl-mode instead
+			return version_compare(static::GetDBVersion(), '5.7.11', '>=');
+		}
+		return false;
+	}
 }


Hipska (approved):

piRGoif (changes requested):
Good catch, thanks ! 

Just a note : this method (\DBBackup::GetMysqldumpCommand) is called from the setup as from the manual backup or the scheduled one.

Can you change accordingly the corresponding test ? 
It is in `test/setup/DBBackupTest.php`
If you feel uncomfortable doing it, just tell me I can do it (if you check the "[Allow edits from maintainers](https://docs.github.com/en/github-ae@latest/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)" option)

Also the doc url might be changed to the 5.7 version : it contains the `--ssl_mode` parameter values, whereas the current 5.6 link doesn't show them
https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#option_general_ssl-mode

piRGoif:
Also this is a good candidate to be merged in the support/2.7 branch, but we'll have to wait the functional review to discuss this.

larhip:
Hi @piRGoif 
thanks a lot for you feedback!

> Can you change accordingly the corresponding test ? It is in `test/setup/DBBackupTest.php`

I already checked the test cases previously, but since the test did just check whether the options starts with `--ssl` I did not see a need for a change. Anyhow I changed the comment in the test now. I have no idea how to implement a test which will check it more in detail, since the current implementation is not able to mock the database client, is it?

> If you feel uncomfortable doing it, just tell me I can do it (if you check the "[Allow edits from maintainers](https://docs.github.com/en/github-ae@latest/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)" option)

I tried to allow you as maintainer, but unfortunately I was not aber to find the corresponding checkbox :-(
image

> Also the doc url might be changed to the 5.7 version : it contains the `--ssl_mode` parameter values, whereas the current 5.6 link doesn't show them https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#option_general_ssl-mode

The link has been changed from 5.6 to 5.7 but I thought it would be better to set the anchor to the top chapter (encrypted-connection-options) since the depreciation info of `--ssl` is not the the chapter you mentioned.

Molkobain (approved):

Molkobain:
@larhip the test cases can be update to check that the right `--ssl-mode` option is passed, I think that would suffice. Once done, the PR will be mature enough for me.

setup/backup.class.inc.php
```diff
@@ -486,7 +486,14 @@ public static function GetMysqlCliTlsOptions($oConfig)
 		$sMysqlSSLModeVersion = '5.7.0'; //Mysql 5.7.0 and upper deprecated --ssl and uses --ssl-mode instead
 		if ($sDBVendor === CMDBSource::ENUM_DB_VENDOR_MYSQL && version_compare($sDBVersion, $sMysqlSSLModeVersion, '>='))
```
Hipska: Also, since iTop 3.x only [supports](https://www.itophub.io/wiki/page?id=3_0_0:install:requirements) MySQL 5.7+ the version check can be removed and this part should always be executed?

Hipska: Oh wait, does MariaDB still use `--ssl`? In that case, it can be this:
```suggestion
		if (CMDBSource::GetDBVendor() === CMDBSource::ENUM_DB_VENDOR_MYSQL)
```

Hipska: Another note on this: It's perfectly possible to have a DB server with MySQL and the `mysqldump` tool from MariaDB and vice versa, so be careful with these checks.

Molkobain: > Also, since iTop 3.x only [supports](https://www.itophub.io/wiki/page?id=3_0_0:install:requirements) MySQL 5.7+ the version check can be removed and this part should always be executed?

Depends if we manage to make this merged in the `support/2.7` branch 🤞

Hipska: Currently it is targeted at `develop`, which means iTop 3.1
My other point still counts..

Molkobain: > Currently it is targeted at develop, which means iTop 3.1

> Also this is a good candidate to be merged in the support/2.7 branch, but we'll have to wait the functional review to discuss this.

Hipska: Correct, but my other point still stands..

larhip: Damn you are right @Hipska `CMDBSource::GetDBVendor` will check the server version / vendor not the client, missed that :-(

I'll try to find a way to check the client version and come back as soon as I got a solution..

larhip: OK I checked it.
Looks like the only other way to figure out the client version, would be by executing `mysqldump --version`
So it might be possible to do something like this:
```
$sMySQLDump = $this->GetMysqldumpCommand() . " --version";
exec($sCommand, $aOutput, $iRetCode);
```
and use a RegEx to get the vendor and the version from the output (not sure how is is such a RegEx since the version string ist not really standardized).

I am really not sure how "stable" such an implementation is...
It does feel a little bit "dirty" to grep for the version in the output of an exec command, doesn't it?

Anyhow the change would be bigger that originally planned from my side..

Another way would be the add a warning in the documentation, that it would be recommended to use that same mysql client and server version in case of an TLS encryption (in fact this warning is true for the current code without this PR as well, since currently you are already using the server version to check which client options to be set). But for sure a comment in the documentation will not address the real technical issue..

What do you think @Molkobain @Hipska ?

larhip:
> @larhip the test cases can be update to check that the right `--ssl-mode` option is passed, I think that would suffice. Once done, the PR will be mature enough for me.

@Molkobain I tried to update the test case to check for the right `--ssl-mode`. Would you mind to review those changes?

setup/backup.class.inc.php
Hipska: IMO as said before, only a check if it is MySQL or MariaDB should be enough. (then it is not possible to backport to 2.7)

larhip: It that case the problem regarding the usage of DB server with MySQL and the mysqldump tool from MariaDB (and vice versa) will still exist an a brief note in the documentation would make sense anyhow, wouldn't it?

Regarding to backport or not, I do not have a strong opinion, so up to the Combodo as the maintainer :-)

test/setup/DBBackupTest.php
```diff
@@ -81,7 +91,18 @@ public function testGetMysqlCliTlsOptionsWithTlsAndCa()
 		$oConfigToTest->Set('db_tls.ca', $sTestCa);
 		$sCliArgsCapathCfg = DBBackup::GetMysqlCliTlsOptions($oConfigToTest);
 
-		$this->assertStringStartsWith(' --ssl', $sCliArgsCapathCfg);
+		// depending on the MySQL version, we would have `--ssl` or `--ssl-mode=VERIFY_CA`
+		$sDBVendor = CMDBSource::GetDBVendor();
+		$sDBVersion = CMDBSource::GetDBVersion();
+		$sMysqlSSLModeVersion = '5.7.0'; //Mysql 5.7.0 and upper deprecated --ssl and uses --ssl-mode instead
+		if ($sDBVendor === CMDBSource::ENUM_DB_VENDOR_MYSQL && version_compare($sDBVersion, $sMysqlSSLModeVersion, '>='))
```
Molkobain: Only if you want to backport to 2.7 anyway: This is duplicated 3 times (2 in unit tests, 1 in the business logic), I would rather have it extracted in a dedicated (both version and condition) method to avoid changes in the business logic to make the tests diverge from reality.

setup/backup.class.inc.php
Molkobain: To me it depends how much we want to backport this fix to the `support/2.7`.

It would be great to do it, but if we are not sure on how to detect the vendor (uncertain RegEx), then I'll rather just fix this in the `support/3.0` to ensure stability.
Also as this is the first time this issue is reported, it might not be so frequent, so the need to backport it is not that big to me.

So long story short I'll go the Hipska way.

larhip: Fine for me, I've already updated the code ✅

test/setup/DBBackupTest.php
larhip: This should be obsolete since there will be no backport to 2.7 :-)

Molkobain (commented):
_All in comments_

Molkobain (approved):

Molkobain:
All good to me, thank you both 🙌

test/setup/DBBackupTest.php
```diff
@@ -61,8 +61,15 @@ public function testGetMysqlCliTlsOptionsWithTlsNoCa()
 		$oConfigToTest->Set('db_tls.enabled', true);
 		$sCliArgsMinCfg = DBBackup::GetMysqlCliTlsOptions($oConfigToTest);
 
-		// depending on the MySQL version, we would have `--ssl` or `--ssl-mode=VERIFY_CA`
-		$this->assertStringStartsWith(' --ssl', $sCliArgsMinCfg);
+		// depending on the MySQL vendor, we would have `--ssl` or `--ssl-mode=REQUIRED`
+		if (CMDBSource::GetDBVendor() === CMDBSource::ENUM_DB_VENDOR_MYSQL)
+		{
+			$this->assertStringStartsWith('--ssl-mode=REQUIRED', $sCliArgsMinCfg);
```
Hipska: Does this assert pass? As the string should start with a space.

Hipska (approved):

test/setup/DBBackupTest.php
Molkobain: Indeed, I wrote all comments at once. It can be discarded :)

test/setup/DBBackupTest.php
Molkobain: +1

larhip: Shame on me :-(
I did not run the tests again, since they are not platform independent and on my MySQL-Test-Env I got no phpunit and no composer and I would like to save that time, my fault :-(

Anyhow it's no fixed and the tests pass.
image

Thanks for your watchful eye @Hipska

test/setup/DBBackupTest.php
```diff
@@ -81,7 +88,15 @@ public function testGetMysqlCliTlsOptionsWithTlsAndCa()
...
 
-		$this->assertStringStartsWith(' --ssl', $sCliArgsCapathCfg);
+		// depending on the MySQL vendor, we would have `--ssl` or `--ssl-mode=VERIFY_CA`
+		if (CMDBSource::GetDBVendor() === CMDBSource::ENUM_DB_VENDOR_MYSQL)
+		{
+			$this->assertStringStartsWith(' --ssl-mode=VERIFY_CA', $sCliArgsCapathCfg);
+		}
+		else
+		{
+			$this->assertStringStartsWith(' --ssl', $sCliArgsCapathCfg);
```
Hipska, line 99: This test also succeeds when the string is ` --ssl-mode=VERIFY_CA`.

test/setup/DBBackupTest.php
```diff
@@ -61,8 +61,15 @@ public function testGetMysqlCliTlsOptionsWithTlsNoCa()
...
-		// depending on the MySQL version, we would have `--ssl` or `--ssl-mode=VERIFY_CA`
-		$this->assertStringStartsWith(' --ssl', $sCliArgsMinCfg);
+		// depending on the MySQL vendor, we would have `--ssl` or `--ssl-mode=REQUIRED`
+		if (CMDBSource::GetDBVendor() === CMDBSource::ENUM_DB_VENDOR_MYSQL)
+		{
+			$this->assertStringStartsWith(' --ssl-mode=REQUIRED', $sCliArgsMinCfg);
+		}
+		else
+		{
+			$this->assertStringStartsWith(' --ssl', $sCliArgsMinCfg);
```
Hipska, line 71: This test also passes when the string is ` --ssl-mode=REQUIRED`

Hipska (changes requested):

setup/backup.class.inc.php
```diff
@@ -466,11 +466,10 @@ protected function EnumerateTables()
 	 * @return string TLS arguments for CLI programs such as mysqldump. Empty string if the config does not use TLS.
 	 *
 	 * @uses \CMDBSource::GetDBVendor() so needs a connection opened !
-	 * @uses \CMDBSource::GetDBVersion() so needs a connection opened !
 	 *
 	 * @since 2.5.0 N°1260
 	 * @since 2.6.2 2.7.0 N°2336 Call DB to get vendor and version (so CMDBSource must be init before calling this method)
-	 * @link https://dev.mysql.com/doc/refman/5.6/en/connection-options.html#encrypted-connection-options "Command Options for Encrypted Connections"
+	 * @link https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#encrypted-connection-options "Command Options for Encrypted Connections"
```
piRGoif: Thanks for the change !
You can remove the quotes around the title, see https://docs.phpdoc.org/guide/references/phpdoc/tags/link.html#link

```suggestion
	 * @link https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#encrypted-connection-options Command Options for Encrypted Connections
```

I didn't get what you were saying in your comment ? 

> I thought it would be better to set the anchor to the top chapter (encrypted-connection-options) since the depreciation info of --ssl is not the the chapter you mentioned.

The old and the new URL are pointing to the same `encrypted-connection-options` anchor ? And in the 5.6 URL there is the `--ssl` parameter documented ?

setup/backup.class.inc.php
```diff
@@ -480,13 +479,17 @@ public static function GetMysqlCliTlsOptions($oConfig)
 			return '';
 		}
 		$sTlsOptions = '';
-
-		$sDBVendor = CMDBSource::GetDBVendor();
-		$sDBVersion = CMDBSource::GetDBVersion();
-		$sMysqlSSLModeVersion = '5.7.0'; //Mysql 5.7.0 and upper deprecated --ssl and uses --ssl-mode instead
-		if ($sDBVendor === CMDBSource::ENUM_DB_VENDOR_MYSQL && version_compare($sDBVersion, $sMysqlSSLModeVersion, '>='))
+		// Mysql 5.7.0 and upper deprecated --ssl and uses --ssl-mode instead
+		if (CMDBSource::GetDBVendor() === CMDBSource::ENUM_DB_VENDOR_MYSQL)
```
piRGoif: Why removing the version_compare call ? Because MySQL min version is 5.7.0 ? (N°2330 in iTop 3.0.0)
I would keep it anyway so that the code snippet won't rely on any assertions other that testing the real thing (and plus it could cause a bug if we decide to rebase on support/2.7 !)

test/setup/DBBackupTest.php
piRGoif: You're right ! 
Maybe add a test to confirm that the string doesn't start with `--ssl-mode=` ?

piRGoif (changes requested):

setup/backup.class.inc.php
Hipska: Do we know if `--ssl-mode` isn't supported on MySQL 5.6?

piRGoif: Well I digged in a little more

In the [param documentation for MySQL 5.7](https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#option_general_ssl-mode) it is said : 

> The [--ssl-mode](https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#option_general_ssl-mode) option was added in MySQL 5.7.11.

And on [the --ssl doc](https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#option_general_ssl) : 

> The client-side [--ssl](https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#option_general_ssl) option is deprecated as of MySQL 5.7.11 and is removed in MySQL 8.0

So indeed the previous test wasn't correct and should be fixed...
To my understanding : 
* `--ssl` should be returned for MySQL if version < 5.7.11, or for MariaDB (?)
* And `--ssl-mode` must be returned for MySQL >= 5.7.11.

setup/backup.class.inc.php
larhip: > You can remove the quotes around the title
Will do that 👍 

> I didn't get what you were saying in your comment ?

OK now I am not sure whether I got what you are saying 😀 
In your comment at the top you suggested to change the link to 5.7 (instead of 5.6). But you suggested a link to the sub-chapter for the `--ssl-mode`-option 
image

Since the option `--ssl-mode` would be only used for mysql >= 5.7.11 and vor older versions `--ssl` IMHO it would make more sense to link to the top chapter (encrypted-connection-options) which includes the description von the sub chapter for `--ssl` and `--ssl-mode`

Does my explanation make sense to you or do you prefer the link to the `--ssl-mode`-sub chapter?

setup/backup.class.inc.php
larhip: So ok, I will
  * Add the version compare again (using version 5.7.11 as "break point")
    * `--ssl` for MariaDB and MySQL version < 5.7.11
    * `--ssl-mode` for MySQL version >= 5.7.11
  * Change the test cases as well (keeping the comment of @Molkobain in mind regarding extracting it in a method and add an assertion in case it should not be `--ssl-mode`)

Did I forgot anything?

setup/backup.class.inc.php
piRGoif: Oh yeah sorry :D

Totally agree to link to https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#encrypted-connection-options ("Command Options for Encrypted Connections" chapter)  !
Instead of https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#option_general_ssl-mode (--ssl_mode param) I suggested earlier, which is indeed a bad idea !

Sorry for the confusion, thanks for the clarification :)

setup/backup.class.inc.php
piRGoif: Seems all good to me, many thanks !

larhip: Sorry for the delay!

So ok that should be done now, so what I did:
  * Add a new method in `CMDBSource` to check which ssl option to be used
    * Do check vendor and version within this method
  * Using this method in `DBBackup::GetMysqlCliTlsOptions`
  * Change the Test cases as well

Could you please check again? :)

piRGoif (approved):

test/setup/DBBackupTest.php
piRGoif: Problem solved, closing conversation

test/setup/DBBackupTest.php
piRGoif: Problem solved, closing conversation

test/setup/DBBackupTest.php
piRGoif: Problem solved, closing conversation

setup/backup.class.inc.php
piRGoif: Problem solved, closing conversation

setup/backup.class.inc.php
piRGoif: Modification done, closing conversation

Hipska (approved):

Molkobain:

Molkobain:
@larhip did you test the patch on your customer's env?

larhip:
Yes we did test on customer's env, but just an early version of this patch (like it was in the initial PR)
I could test the final version beginning of January in this env.

piRGoif:
Hello,

> I could test the final version beginning of January in this env.

Ok thanks, will wait for your feedback (as an extra precaution) before merging.

larhip:
Hi @piRGoif 
we just tested it on test env of customer and the patch works fine.

Cheers
Lars

piRGoif:
Thanks Lars ! 
Marging !

PR #358, opened 2022-11-04
N°4494 - Fix auto-locking on log save and transition

accognet:
It may be better fixes with more code changes. This one works for a quick fix.

### DIFF
--- a/js/layouts/activity-panel/activity-panel.js
+++ b/js/layouts/activity-panel/activity-panel.js
@@ -117,7 +117,7 @@ $(function()
 					locked_by_someone_else: 'locked_by_someone_else',
 				},
 			},
-			action_promise: null,	// N°4494 - Promise used to do the action (following the log entry send) only once the lock is released
+			release_lock_promise_resolve: null,	// N°4494 - Resolve callback of the Promise used for the action following the log entry send, which must be done only once the lock is released
 
 			// the constructor
 			_create: function () {
@@ -978,9 +978,15 @@ $(function()
 						sStimulusCode = ((sStimulusCode ?? '') === '') ? null : sStimulusCode;
 						if (null !== sStimulusCode) {
 							if (me.options.lock_enabled) {
-								me.action_promise = $.Deferred();
-								me.action_promise.promise().then(function () {
+								// Use a Promise to ensure that we redirect to the stimulus page ONLY when the lock is released, otherwise we might lock ourselves
+								const oPromise = new Promise(function(resolve) {
+									// Store the resolve callback so we can call it later from outside
+									me.release_lock_promise_resolve = resolve;
+								});
+								oPromise.then(function () {
 									window.location.href = GetAbsoluteUrlAppRoot()+'pages/UI.php?operation=stimulus&class='+me._GetHostObjectClass()+'&id='+me._GetHostObjectID()+'&stimulus='+sStimulusCode;
+									// Resolve callback is reinitialized in case the redirection fails for any reason and we might need to retry
+									me.release_lock_promise_resolve = null;
 								});
 							} else {
 								window.location.href = GetAbsoluteUrlAppRoot()+'pages/UI.php?operation=stimulus&class='+me._GetHostObjectClass()+'&id='+me._GetHostObjectID()+'&stimulus='+sStimulusCode;
@@ -1202,9 +1208,8 @@ $(function()
 						// Tried to release our lock
 						else if ('release_lock' === oParams.operation) {
 							sNewLockStatus = me.enums.lock_status.unknown;
-							if (me.action_promise !== null) {
-								me.action_promise.resolve();
-								me.action_promise = null;
+							if (me.release_lock_promise_resolve !== null) {
+								me.release_lock_promise_resolve();
 							}
 						}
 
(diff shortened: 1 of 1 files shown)


Molkobain (changes requested):
The analysis of the root cause is good, sometimes the redirection to the transition form is done before the XHR to release the lock is complete.

That being said, we should rather use a promise to solve this than using a `setInterval`

js/layouts/activity-panel/activity-panel.js
```diff
@@ -117,6 +117,7 @@ $(function()
 					locked_by_someone_else: 'locked_by_someone_else',
 				},
 			},
+			action_promise: null,
```
Molkobain:
```suggestion
			action_promise: null,	// N°4494 - Promise used to do the action (following the log entry send) only once the lock is released
```

js/layouts/activity-panel/activity-panel.js
```diff
@@ -973,14 +974,26 @@ $(function()
...
-							window.location.href = GetAbsoluteUrlAppRoot()+'pages/UI.php?operation=stimulus&class='+me._GetHostObjectClass()+'&id='+me._GetHostObjectID()+'&stimulus='+sStimulusCode;
+							// For now, we don't hide the forms as the user may want to add something else
+							me.element.find(me.js_selectors.caselog_entry_form).trigger('clear_entry.caselog_entry_form.itop');
+
+							if (me.options.lock_enabled) {
+								me.action_promise = new Promise(function () {
+										window.location.href = GetAbsoluteUrlAppRoot()+'pages/UI.php?operation=stimulus&class='+me._GetHostObjectClass()+'&id='+me._GetHostObjectID()+'&stimulus='+sStimulusCode;
+									},
+									function () {
+										//todo
```
Molkobain:
```suggestion
										// We do nothing in case of failure for now
```

js/layouts/activity-panel/activity-panel.js
```diff
@@ -973,14 +974,26 @@ $(function()
...
-						// For now, we don't hide the forms as the user may want to add something else
-						me.element.find(me.js_selectors.caselog_entry_form).trigger('clear_entry.caselog_entry_form.itop');
-
 						// Redirect to stimulus
 						// - Convert undefined, null and empty string to null
 						sStimulusCode = ((sStimulusCode ?? '') === '') ? null : sStimulusCode;
 						if (null !== sStimulusCode) {
-							window.location.href = GetAbsoluteUrlAppRoot()+'pages/UI.php?operation=stimulus&class='+me._GetHostObjectClass()+'&id='+me._GetHostObjectID()+'&stimulus='+sStimulusCode;
+							// For now, we don't hide the forms as the user may want to add something else
+							me.element.find(me.js_selectors.caselog_entry_form).trigger('clear_entry.caselog_entry_form.itop');
```
Molkobain: It's no longer necessary to move this right? So we can just revert these lines and 994-997.

js/layouts/activity-panel/activity-panel.js
```diff
@@ -1153,6 +1165,10 @@ $(function()
 						// Tried to release our lock
 						else if ('release_lock' === oParams.operation) {
 							sNewLockStatus = me.enums.lock_status.unknown;
+							if (me.action_promise != null) {
```
Molkobain:
```suggestion
							if (me.action_promise !== null) {
```

Molkobain (changes requested):

js/layouts/activity-panel/activity-panel.js
accognet: Yes, you're right. I just commit the change.

js/layouts/activity-panel/activity-panel.js
```diff
@@ -975,12 +976,18 @@ $(function()
...
 						me.element.find(me.js_selectors.caselog_entry_form).trigger('clear_entry.caselog_entry_form.itop');
-
 						// Redirect to stimulus
 						// - Convert undefined, null and empty string to null
 						sStimulusCode = ((sStimulusCode ?? '') === '') ? null : sStimulusCode;
 						if (null !== sStimulusCode) {
-							window.location.href = GetAbsoluteUrlAppRoot()+'pages/UI.php?operation=stimulus&class='+me._GetHostObjectClass()+'&id='+me._GetHostObjectID()+'&stimulus='+sStimulusCode;
+							if (me.options.lock_enabled) {
+								me.action_promise = $.Deferred();
+								me.action_promise.promise().then(function () {
```
Molkobain: Why did you changed from `new Promise()` to this? I would rather keep the [vanilla JS way](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/Promise) of using promises.

Molkobain (approved):

js/layouts/activity-panel/activity-panel.js
Molkobain: Done in [fb77491](https://github.com/Combodo/iTop/pull/358/commits/fb77491a63834c009cb97cdd1022db9046f2d939)

steffunky (approved):

PR #359, opened 2022-11-08
N°6101 Replace old Ctrl+Enter detection in run_query with the new one

piRGoif:
The Ctrl+Enter shortcut was added in the run query screen with [684e9e3](https://github.com/Combodo/iTop/commit/684e9e35375ec1709f51c0f8692ee1414ce21616) in iTop 2.5.0, introducing the JQuery Hotkeys plugin.

In #345 the Ctrl+Enter shortcut was added (for iTop 3.1.0) in multiline form fields using a more modern technique, without any plugin dependency.

This PR aims to use this new technique.

Note that : 
* I added a `@deprecated` comment in js/jquery.hotkeys.js, but we need to create corresponding tickets in Combodo BD
* test/VerifyOQL.php was using the same code as run_query.php. This is an old page (it is throwing warnings in iTop 3.x) but I modify it anyway as the fix is exactly the same

TODO : 
* [x] tech review
* [x] functional review
* [x] create ticket : JQuery Hotkeys depreciation => N°6102
* [x] create ticket : JQuery Hotkeys removal => N°6103
* [x] create ticket : run_query modification

### DIFF
--- a/js/jquery.hotkeys.js
+++ b/js/jquery.hotkeys.js
@@ -1,6 +1,12 @@
 /*jslint browser: true*/
 /*jslint jquery: true*/
 
+
+/**
+ * @deprecated 3.1.0 use JQuery event keys instead
+ * @link https://github.com/Combodo/iTop/pull/345 Ctrl+Enter shortcut added in iTop object forms
+ */
+
 /*
  * jQuery Hotkeys Plugin
  * Copyright 2010, John Resig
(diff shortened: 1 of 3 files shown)


Molkobain (approved):

Molkobain:
@piRGoif will check with Romain if the test file can be deleted as it is an old thing.

Molkobain:
Pierre will create the corresponding ticket and migration check.

PR #363, opened 2022-11-15
N°5619 - Hide newsroom menu when no provider

dflaven:
The newsroom menu can be hidden by configuration, but is not automatically hidden when there is no newsroom provider installed.
This pull request just hides the menu automatically in such a case.

### DIFF
--- a/sources/application/UI/Base/Component/PopoverMenu/NewsroomMenu/NewsroomMenu.php
+++ b/sources/application/UI/Base/Component/PopoverMenu/NewsroomMenu/NewsroomMenu.php
@@ -21,6 +21,7 @@ namespace Combodo\iTop\Application\UI\Base\Component\PopoverMenu\NewsroomMenu;
 
 
 use Combodo\iTop\Application\UI\Base\Component\PopoverMenu\PopoverMenu;
+use MetaModel;
 
 /**
  * Class NewsroomMenu
@@ -64,5 +65,16 @@ class NewsroomMenu extends PopoverMenu
 	public function GetParamsAsJson(): string
 	{
 		return json_encode($this->aParams);
+	}
+	
+	/**
+	 * Check if there is any Newsroom provider configured
+	 * @since 3.1.0 N°5619
+	 * @return boolean
+	 */
+	public static function HasProviders(): bool
+	{
+		$aProviders = MetaModel::EnumPlugins('iNewsroomProvider');
+		return count($aProviders) > 0;
 	}
 }
--- a/sources/application/UI/Base/Component/PopoverMenu/NewsroomMenu/NewsroomMenu.php
+++ b/sources/application/UI/Base/Component/PopoverMenu/NewsroomMenu/NewsroomMenu.php
@@ -70,7 +70,7 @@ class NewsroomMenu extends PopoverMenu
 	/**
 	 * Check if there is any Newsroom provider configured
 	 * @since 3.1.0 N°5619
-	 * @return boolean
+	 * @return bool
 	 */
 	public static function HasProviders(): bool
 	{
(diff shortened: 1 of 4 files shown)


dflaven:
The second commit fixes a crash when there is actually no provider at all (not reproducible on a Community iTop, since there is the iTop Hub provider...)

sources/application/UI/Base/Component/PopoverMenu/NewsroomMenu/NewsroomMenuFactory.php
```diff
@@ -49,6 +49,16 @@ public static function MakeNewsroomMenuForNavigationMenu()
...
+	/**
+	 * Check if there is any Newsroom provider configured
+	 * @return boolean
+	 */	
+	public static function HasProviders()
+	{
+		$aProviders = MetaModel::EnumPlugins('iNewsroomProvider');
+		return count($aProviders) > 0;
+	}
+
```
Molkobain: Please add the `@since 3.1.0 N°5619` annotation in the PHPDoc as well as the `: bool` return type hint.

Moreover, I feel like this should be a method of the `Combodo\iTop\Application\UI\Base\Component\PopoverMenu\NewsroomMenu`NewsroomMenu` class itself rather than the factory. The factory isn't an helper, it's just there to ease the assembly or a specific use case.

Molkobain (changes requested):

sources/application/UI/Base/Component/PopoverMenu/NewsroomMenu/NewsroomMenuFactory.php
Molkobain: @dflaven ?

sources/application/UI/Base/Layout/NavigationMenu/NavigationMenu.php
```diff
@@ -269,7 +270,7 @@ public function GetSubBlocks(): array
 	 */
 	public function IsNewsroomEnabled(): bool
 	{
-		return MetaModel::GetConfig()->Get('newsroom_enabled');
+		return (MetaModel::GetConfig()->Get('newsroom_enabled') && NewsroomMenuFactory::HasProviders());
```
Molkobain: But it does work on Essential/Professional products though, are you sure this is the root cause?

Molkobain (changes requested):

sources/application/UI/Base/Layout/NavigationMenu/NavigationMenu.php
Hipska: According to https://github.com/Combodo/iTop/pull/363#issuecomment-1318400298 it's "crashing". Don't know what exactly he means with that.

Molkobain: @dflaven

sources/application/UI/Base/Component/PopoverMenu/NewsroomMenu/NewsroomMenu.php
```diff
@@ -65,4 +66,15 @@ public function GetParamsAsJson(): string
 	{
 		return json_encode($this->aParams);
 	}
+	
+	/**
+	 * Check if there is any Newsroom provider configured
+	 * @since 3.1.0 N°5619
+	 * @return boolean
```
Molkobain:
```suggestion
	 * @return bool
```

sources/application/UI/Base/Layout/NavigationMenu/NavigationMenu.php
dflaven: Without the second commit it was crashing (during the construction of the iTop WebPage) on Essential/Professional because the (empty) newsroom menu object was not created (because there is not newroom provider on such products).

Molkobain: Discussed with Denis, I could not reproduce the issue on a Professional package. We would like to be able to reproduce the issue in order to ensure the fix is necessary and correct. Denis will dig further to have a reproduction procedure.

Molkobain:
Closing this PR as it could be reproduced.

PR #362, opened 2022-11-10
New PT-BR Translation

eduardomozart:

PR #367, opened 2022-11-18
Update .gitignore

eduardomozart:
It's a discussion started with @Hipska at PR #362 and I decided to open this PR to bother Hipska (LOL) and because I believe this change should be part of iTop project.
I know it can be added to the global ``.gitignore`` file, but newer contributes may not have these entries in their machine and commit unuseful codes to the iTop branch.
If these entries should be set in every global ``.gitignore`` files of every contributor around the world, following the same logic of Hipska, please remove the specific IDE entries because every developers that uses them must have it added to their global ``.gitignore`` file. Why don't we remove ``.gitignore`` file at all and follow the principle that everyone will have their global ``.gitignore`` file properly set?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -152,3 +152,6 @@ local.properties
 
 # Windows
 Thumbs.db
+
+# vi
+*~
--- a/.gitignore
+++ b/.gitignore
@@ -153,5 +153,3 @@ local.properties
 # Windows
 Thumbs.db
 
-# vi
-*~
--- a/.gitignore
+++ b/.gitignore
@@ -146,3 +146,9 @@ local.properties
 .cache-main
 .scala_dependencies
 .worksheet
+
+# Mac
+.DS_Store
+
+# Windows
+Thumbs.db


Hipska:
You are putting things a bit simple. There are perfect reasonings for having IDE specific ignores in the project `.gitignore` file but no system-wide in them. There are are enough articles written about that over the last 10 years, we don't need to repeat that discussion.

~~Following your logic `thumbs.db` should be added as well?~~
Edit: While I was writing you actually did that..

eduardomozart:
GitHub Desktop starts with the purpose of being simple to commit code changes to any project abstracting the Git layer.
You don't even need to know that Git exists at all - simply install it, click at the bottom of GitHub Desktop into any GitHub project page to sync it locally, make your changes and commit them through the GUI.
What I'm trying to say is that just because it's simple for you, it doesn't mean it will be simple for everyone.
When any user browse the iTop repo locally, the ``.DS_Store`` and ``Thumbs.db`` files are generated on the fly by the File Manager without user intervention. Any user attempting to not commit them ignoring them through the GitHub Desktop software, it will be added to the .gitignore file of the iTop project and users will commit a lot of changes to this file doing so because it adds the full file path to the iTop ``.gitignore`` file (there will be multiple ``.DS_Store`` instances into the iTop ``.gitignore`` file). Of course we can't add every IDE and specific cenarios to the ``.gitignore`` file, but the most common ones that can affect newcomers should also be part of the project itself.
I believe the @Molkobain and @piRGoif are the most active developers at the project and their opinion will decide the future of this PR. If they decided it's not useful to help newcomers to contribute to the project believing they have 10+ years of knowledge and a full knowledge of Git and what the ``.gitignore`` file does, I'll follow their decision.

piRGoif:
Hello,
Thanks Eduardo for raising the question in a dedicated PR !
The .gitignore file has already lots of ignored files... Myself I'm not opposed to this integration.
I'm changing the PR status so that the question could be discussed with the rest of the dev team.

Hipska:
The suggestion is not even complete for [macOS](https://github.com/github/gitignore/blob/main/Global/macOS.gitignore), [Windows](https://github.com/github/gitignore/blob/main/Global/Windows.gitignore) or [Linux](https://github.com/github/gitignore/blob/main/Global/Linux.gitignore).

[Reading](https://augustl.com/blog/2009/global_gitignores/) on adding it to the global gitignore.

PS: also have a look at the location these OS specific gitignore examples are located in that gitignore repository.

eduardomozart:
Those are not exactly necessary because they do not affect the project. The intention isn't add every possibility, but only the files that are create automatically by the File Manager or common editors. My suggestion only includes the Thumbnail files (that are created automatically when the file has previews, like browsing folders of the project that contains images) and temporary files created by vi/nano/pico.

Molkobain:
Accepted during technical review but only for Windows / macOS files. We would rather remove the "vi" part as one should be aware that it has an opened file before committing. Once done, we can merge.

.gitignore
```diff
@@ -146,3 +146,12 @@ local.properties
...
 .worksheet
+
+# Mac
+.DS_Store
+
+# Windows
+Thumbs.db
+
+# vi
+*~
```
Molkobain:
```suggestion
```

Molkobain (changes requested):

.gitignore
eduardomozart: Ok!

Molkobain: PS: You can apply the suggestion directly from GitHub to avoid doing the commit yourself :)

Molkobain:
Thanks, it's merged!

PR #368, opened 2022-11-18
N°5793 - HTML Sanitizer: Allow 'start' attribute in 'ol' tag

jbostoen:
Just an example here. This fork is based on Combodo's Mail to Ticket Automation, which relies on internal methods.

The real contents look like this in MS Outlook (text censored):
image

The contents in iTop look like this:
image

HTML source:

```
<ol style="margin-top:0in" start="1" type="1"><li class="MsoListParagraph" style="margin-left:0in;mso-list:l0 level1 lfo1">xxx.</li></ol>

<p class="MsoNormal"> </p>

<ol style="margin-top:0in" start="2" type="1"><li class="MsoListParagraph" style="margin-left:0in;mso-list:l0 level1 lfo1">xxxx.</li></ol>
```

After stripping: there is indeed a paragraph in between, and iTop will ignore the "start".


Molkobain (approved):

Molkobain:
Thanks for the PR Jeffrey! Makes total sense, will push it forward during next reviews.

Hipska (approved):

piRGoif (approved):
As a reference : https://www.w3.org/TR/html4/struct/lists.html#edef-OL

Molkobain:
The idea was accepted during the technical review.
We would like you to update the corresponding unit test though, it can be found in test/core/sanitizer.

_PR will be reviewed functionally next week anyway to save time for the next iteration._

Molkobain (changes requested):
Update corresponding unit test please.

jbostoen:
Do I just need to edit utf-8_wrong_character_email_truncated.txt ?
Or create a new file? (and if so, do I need to add a reference to it somewhere in another file?)

Molkobain:
I think it should rather be `whitelist_test.html` but its content doesn't seem right. @piRGoif ?

piRGoif:
Hello !

About the test, I would add the new attribute in the \Combodo\iTop\Test\UnitTest\Core\Sanitizer\HTMLDOMSanitizerTest::WhiteListProvider method, to check if the attribute is kept after the sanitize run.

Maybe add also a new data set in \Combodo\iTop\Test\UnitTest\Core\Sanitizer\HTMLDOMSanitizerTest::RemoveBlackListedTagContentProvider, to check for OL with attributes to remove and the new one to keep ?

Molkobain:
Will be merged once unit test is updated.

jbostoen:
Would this qualify?

I extended the PR a bit based on info from W3Schools:
* https://www.w3schools.com/tags/tag_li.asp
* https://www.w3schools.com/tags/tag_ol.asp

Basically:
* li: also allow 'value' (to set a starting value)
* ol: also allow 'reversed' (decreasing list), 'type' (see link above) and 'start' (to set a starting value).

I quickly fiddled around, but didn't see the real difference between li > value and ol > start. ( https://stackoverflow.com/questions/33374212/difference-value-and-start-in-html-lists )

Molkobain (approved):

Molkobain:
Yes, it's closely related :)

test/core/sanitizer/HTMLDOMSanitizerTest.php
```diff
@@ -213,6 +213,12 @@ public function RemoveBlackListedTagContentProvider()
 				'html' => '<html><head></head><body lang="EN-GB" link="#0563C1" vlink="#954F72">bar</body></html>',
 				'expected' => 'bar',
 			),
+			
+			'list test' => array(
```
Molkobain:
```suggestion
			'ordered list with attributes' => array(
```

jbostoen:
Mind to also update https://www.itophub.io/wiki/page?id=latest:admin:rich_text_limitations

Molkobain:
Done but will only appear once 3.1 wiki is published :)

PR #371, opened 2022-11-23
:memo: CONTRIBUTING: Improvements

piRGoif:
Add some info on what the PR description should contain.
Also 2 little modification (header level and text case)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,6 +137,9 @@ When your code is working, please:
 * squash as much as possible your commits,
 * rebase your branch on our repo last commit,
 * create a pull request
+* in the pull request description, mind to add all the information useful to understand why you're suggesting this modification and anything necessary to dive into your work. Especially:
+  - bugfixes: exact steps to reproduce the bug (given/when/then), description of the bug cause and what solution is implemented 
+  - enhancements: use cases, implementation details if needed
 * mind to check the "[Allow edits from maintainers](https://docs.github.com/en/github-ae@latest/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)" option !
 
 Detailed procedure to work on fork and create PR is available [in GitHub help pages](https://help.github.com/articles/creating-a-pull-request-from-a-fork/).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -136,13 +136,11 @@ When your code is working, please:
 
 * Squash as much as possible your commits,
 * Rebase your branch on our repo last commit,
-* Create a pull request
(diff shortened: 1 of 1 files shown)


Hipska (approved):

CONTRIBUTING.md
```diff
@@ -134,15 +134,18 @@ Our tests are located in the `test/` directory, containing a PHPUnit config file
 
 When your code is working, please:
 
-* stash as much as possible your commits,
-* rebase your branch on our repo last commit,
-* create a pull request
-* mind to check the "[Allow edits from maintainers](https://docs.github.com/en/github-ae@latest/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)" option !
+* Stash as much as possible your commits,
```
Molkobain: Do you mean "squash"?

piRGoif: Woops, this typo is here from the beginning :/  Better late than never :o)
Will fix it directly in the current file starting from support/2.7 !

piRGoif: Done in e9c91d98
Closing conversation

Molkobain: It still says "Stash" for me even though you fixed it at some point in the past

Molkobain (changes requested):

CONTRIBUTING.md
Hipska: That commit doesn't seem to be on this branch anymore. But also, what a strange advise: "Squash as much as possible"? What does that even mean?

To me, it makes reviewing easier if the changes are separate commits, compared to how Jeffrey does it by force pushing the branch completely. Then it is not easy to see what has been changed between your last view/review and the latest push..

Stashing as much as possible still looked as a good advise..

Molkobain: > Stashing as much as possible still looked as a good advise..

But what does it mean in case of a PR? You only stash locally what you don't want to commit/push don't you?

Hipska: You are right. Both ways don't make much sense 😄

Molkobain: To me "squash you commits" meant : Clean your branch as much as possible by squashing commits that can be (eg. "Fix 1", "Fix 2", ...) which we strongly suggest. Otherwise in some case we have to make a "Squash and merge" to keep the history clean.

piRGoif: The typo fix was pushed in e9c91d98, and I didn't rebased this branch afterwards. 
Going to fix it ! 

About the meaning of the sentence : we wanted to ask to make the history as much logic as possible before asking for a review. 
During dev, you will often push multiple commits in a row, the history will just represent your coding workflow and can be hard to understand. 
So we are asking to do the appropriate history corrections before asking for a review : regroup same modifications together in particular.

piRGoif:
Branch rebased & force pushed.
Now we get the stash/squash typo fix !

Molkobain (approved):

Molkobain:
Functional review will be skipped as this is purely on the technical side.

PR #374, opened 2022-12-01
N°5765 - Setup: Never cache folder permissions test response

Molkobain:
Avoid false-positive and false-negative when the setup tries to check that folder permissions are set correctly as the result of the request could be cached.

Thanks to @mraenker !

### DIFF
--- a/setup/setup.js
+++ b/setup/setup.js
@@ -54,8 +54,8 @@ function ExecuteStep(sStep)
 }
 
 function CheckDirectoryConfFilesPermissions(sWikiVersion){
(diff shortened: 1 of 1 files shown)


piRGoif (approved):

piRGoif:
As a reference : https://api.jquery.com/jquery.ajax/#jQuery-ajax-url-settings

steffunky (approved):

Molkobain:
PR rebased onto support/2.7 instead of support/3.0 as the test was introduced in 2.7.7.

Molkobain:
Accepted during functional review. Will be part of the upcoming 2.7.9

PR #375, opened 2022-12-02
N°5603 - Autocomplete fails with error for an external key pointing to an abstract class with no friendlyname

accognet:
Autocomplete fails with error for an external key pointing to an abstract class with no friendlyname.

### DIFF
--- a/core/valuesetdef.class.inc.php
+++ b/core/valuesetdef.class.inc.php
@@ -264,36 +264,48 @@ class ValueSetObjects extends ValueSetDefinition
 		{
 			case 'equals':
 				$aAttributes = MetaModel::GetFriendlyNameAttributeCodeList($sClass);
-				$sClassAlias = $oFilter->GetClassAlias();
-				$aFilters = array();
-				$oValueExpr = new ScalarExpression($sContains);
-				foreach($aAttributes as $sAttribute)
-				{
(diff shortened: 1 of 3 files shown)


Molkobain:
I understand that the code was already mostly duplicated but as the code is growing could you factorize the whole switch/case in a dedicated method to ease maintenance? For example `PrepareFilterForLoadValues(DBSearch $oFilter, string $sOperation)`.

Molkobain:
Moreover this will allow to add a unit test on this new method.

Molkobain (approved):

Molkobain:
Discussed with Anne-Catherine, code has been factorized a bit more and branch has been rebased onto support/3.0 to solve conflicts. Ready for technical review.

rquetiez:
This change involves both a "one line fix" and a refactoring of duplicated code. Both are ok.
That said, the one line fix would be suitable for a fix on a minor release, whereas the refactoring makes it more risky : the diff is quite confusing and we do not have relevant unit test on that specific part.
Therefore, I do **not** recommend to merge that on branch 3.0. Rather, we'd better differ the change to **3.1**.

rquetiez (changes requested):
Branch develop is more suitable for such a change

Molkobain:
Discussed IRL, accepted during functional review. Go for 3.0.3 / 3.1.0

PR #376, opened 2022-12-07
N°5559 - Prevent obsolete replica to lock iTop object

v-dumas:
Currently **every replicas** which are mapped to an itop object, are locking the object fields according to the Datasynchro field update policy, regardless of the replica status.

I am proposing to change this. An "obsolete" synchro replica, would no more lock the iTop object. 

- A Person or User could be anonymized for eg. if its associated replicas are **all** "obsolete".
- This PR does not modify on purpose the behavior of "obsolete" replicas preventing the deletion of the iTop object. The reason why I don't want to change it, is related to the fact that data synchronization between applications can have unwanted glitches:
- If the Source data had a glitch and objects were missing from the source, their replicas become "obsolete", the iTop associated objects can be modified by users, but not deleted (_if the Data Synchro was configured that way_). Once the synchronization is restored, the Source objects are back again, the replicas are synchronized again and every fields controlled by the synchro are restored to the values provided by the Source. So no real impact, everything is back to what it was before.
- But in that same scenario, if iTop objects deletion stopped to be protected, when thing come back, also the iTop object is recreated, we have lost all its history, relationship and non datasynchro controlled fields values.


core/dbobject.class.php
```diff
@@ -4937,28 +4937,27 @@ public function GetMasterReplica()
 	 * @throws \MySQLException
 	 * @throws \OQLException
 	 */
-	public function GetSynchroData()
+	public function GetSynchroData($bWithoutObsolete = false)
```
piRGoif: Not sure about the parameter name.
I did a little search, I just came up with the bAllowAllData as an inspiration ? So maybe you should better have `$bAllowObsolete=true` ?

piRGoif (approved):

piRGoif:
Directly set to functional review to avoid loosing a month !

core/dbobject.class.php
Molkobain: Throughout the code we have many occurences of `$bShowObsolete` and `$bAllowAllData` (btw the later is based on the obsolete mechanism, but the status of the replica), so I would rather have some phrased that way too.

Whereas it should be `$bAllowObsolete` or `$bIncludeObsolete` or else, I don't really know. Let's discuss it during the next review.

Moreover, by adding a parameter to the DBObject::GetSynchroData() method you might break all overloadings of this method in sub-classes. Did you run an audit to see there is any in the extensions / customizations?

piRGoif: > Throughout the code we have many occurences of `$bShowObsolete` and `$bAllowAllData` 

Oh I missed the first one yes !
Found this parameter name in \ormLinkSet::ToDBObjectSet, but there are also multiple variables named this way.

> Moreover, by adding a parameter to the DBObject::GetSynchroData() method you might break all overloadings of this method in sub-classes. 

The method is tagged as internal so we shouldn't have to ?

Molkobain: It's tagged as internal yes so third parties should not us it.
But as long as a method is public is can be used by us during the creation of an extension or to answer a support issue, even though it should not. So auditing the possible impact is mandatory to me to anticipate what could break for our customers.

piRGoif: Ok, I was answering on searching for uses in customer customization. 
We did search on a lot of Combodo modules (77 clones I have on my laptop to be precise) and didn't find any caller.

Molkobain:
Accepted during functional review upon the following:
  * Vincent will check bug N°787 that seems similar and update it accordingly
  * Vincent will add a conf. param. to restore the previous behavior if necessary. Conf. param. name TBD.

Molkobain:
@v-dumas any update?

v-dumas:
I have added a global config parameter "synchro_obsolete_replica_locks_object" boolean, default true, so it does not change current behavior.
I have tested with extension combodo-anonymizer 1.3.0, Person with obsolete synchro-replica can now be anonymized when the above parameter is set to false, while without this param, it fails silently.

piRGoif:
Maybe the new parameter should be with show_in_conf_sample=false ?
Also mind you have conflict for now... Can you do a rebase ?

PR #379, opened 2022-12-15
N°5784 - PHP 8.0: Fix mandatory attribute not visible in transition form due to bad emptiness test

Molkobain:
**Issue description**
See ticket [#2136 on SF](https://sourceforge.net/p/itop/tickets/2136/)

**Cause**
In transition form, when an attribute is supposed to be mandatory we check if it already has a value.
The current test is not done correctly as it is testing if the value is an empty string `''`, when in the reported case it could be `0` as it was an external key. This weak test doesn't return true with PHP 8.0+ as you can [see there](https://3v4l.org/j7Oep).

image

**Solution**
_Since the first version of this PR, internal discussion took place to prepare the best solution and introduce the new method in the maintenance version (3.0.3), this PR is a media for the discussion._
  * `AttributeDefinition::HasAValue(mixed $proposedValue): bool` has been introduced. **MIND that implementation for several attribute types are to be discussed**, this PR is here to find the best way to do it.
  * `DBObject::HasAValue(string $sAttCode): bool` is a shortcut to the above method to ease DX.

The main idea behind `HasAValue()` is to determine is a value has been set by the user in a form or by an automation. It is really different from `AttributeDefinition::IsNull()`, 2 examples:
  * On an attribute string `''` should be considered as NOT a set value, whereas it is not null.
  * On an attribute ext. key `0` should be considered as NOT a set value, whereas it is not null.

_PS: I'll add the unit tests once we decide how to do all the implementations_

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -3374,13 +3374,14 @@ EOF
 			// Consider only the "expected" fields for the target state
 			if (array_key_exists($sAttCode, $aExpectedAttributes)) {
 				$iExpectCode = $aExpectedAttributes[$sAttCode];
+
 				// Prompt for an attribute if
 				// - the attribute must be changed or must be displayed to the user for confirmation
 				// - or the field is mandatory and currently empty
 				if (($iExpectCode & (OPT_ATT_MUSTCHANGE | OPT_ATT_MUSTPROMPT)) ||
-					(($iExpectCode & OPT_ATT_MANDATORY) && ($this->Get($sAttCode) == ''))) {
-					$oAttDef = MetaModel::GetAttributeDef($sClass, $sAttCode);
+					(($iExpectCode & OPT_ATT_MANDATORY) && (false === $this->HasAValue($sAttCode)))) {
 					$aArgs = array('this' => $this);
+					$oAttDef = MetaModel::GetAttributeDef($sClass, $sAttCode);
 					// If the field is mandatory, set it to the only possible value
 					if ((!$oAttDef->IsNullAllowed()) || ($iExpectCode & OPT_ATT_MANDATORY)) {
 						if ($oAttDef->IsExternalKey()) {
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -8218,7 +8218,7 @@ class AttributeBlob extends AttributeDefinition
 		}
 
 		// Empty file (no content, just a filename) are supported since PR {@link https://github.com/Combodo/combodo-email-synchro/pull/17}, so we check for both empty content and empty filename to determine that a document has no value
-		return utils::IsNullOrEmptyString($proposedValue->GetData()) && utils::IsNullOrEmptyString($proposedValue->GetFileName());
+		return utils::IsNotNullOrEmptyString($proposedValue->GetData()) && utils::IsNotNullOrEmptyString($proposedValue->GetFileName());
 	}
 
 
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -711,9 +711,10 @@ abstract class AttributeDefinition
 	 * @return bool True if $proposedValue is an actual value set in the attribute, false is the attribute remains "empty"
 	 * @since 3.0.3, 3.1.0 N°5784
 	 */
-	public function HasAValue(mixed $proposedValue): bool
+	public function HasAValue($proposedValue): bool
 	{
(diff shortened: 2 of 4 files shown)


steffunky (approved):
We should indeed push for polymorphism in iTop 3.1 branch, but this is good enough for 3.0 👍

Hipska (approved):

application/cmdbabstract.class.inc.php
```diff
@@ -3374,12 +3374,23 @@ public function DisplayStimulusForm(WebPage $oPage, $sStimulus, $aPrefillFormPar
...
+
+				$oAttDef = MetaModel::GetAttributeDef($sClass, $sAttCode);
+				// Test if value is empty
+				if ($oAttDef instanceof AttributeExternalKey) {
+					$bIsValueEmpty = $this->Get($sAttCode) === 0;
+				} elseif ($oAttDef instanceof AttributeBlob) {
+					$bIsValueEmpty = $this->Get($sAttCode)->IsEmpty();
+				} else {
+					// Weak comparison on purpose to keep existing behavior on attribute types that are not explicitly tested
+					$bIsValueEmpty = $this->Get($sAttCode) == '';
```
Hipska: What about this?
```suggestion
					$bIsValueEmpty = empty($this->Get($sAttCode));
```

Hipska (approved):
Indeed, the `IsEmpty()` method on `AttributeDefinition` is a good proposal.

application/cmdbabstract.class.inc.php
Hipska: Nevermind, this gives false positives. The original is better indeed.

Hipska:
However there is already the `IsNull` method which can already cover this use-case..

Molkobain:
> However there is already the `IsNull` method which can already cover this use-case..

Actually the `AttributeDefinition::IsNull` method returns `true` if the value is null not if it is empty, for instance if you check `AttributeDecimal::IsNull` it will return `is_null($value)` but when $value is either an empty string or a zero decimal it will return `false` which is not what we need in that use case. See [this 3v4l](https://3v4l.org/EJmH3).

Hipska:
The empty string will become `null` because of `MakeRealValue`, and `0` should not be considered 'empty' in the case of a decimal field?

application/cmdbabstract.class.inc.php
Hipska: (can't mark as resolved, please do)

Molkobain: Done, thanks

piRGoif:
What are the downsides of introducing the new AttributeDefinition method in support/3.0 ? I don't see any... To be discussed during technical review.

Molkobain:
Description and code has been updated to reflect the internal discussion. hoping to have some feedback so we can code it the best way 🙏

Hipska:
> The empty string will become `null` because of `MakeRealValue`, and `0` should not be considered 'empty' in the case of a decimal field?

No answer on this yet, and I still don't see the difference between `IsNull` or `HasAValue` meaning..

Molkobain:
> > The empty string will become `null` because of `MakeRealValue`, and `0` should not be considered 'empty' in the case of a decimal field?

Unfortunately `AttributeDefinition::HasAValue()` can be called directly with a `$proposedValue` that has not been prepared by `MakeRealValue()`. So we can't make this assomption.

> No answer on this yet, and I still don't see the difference between `IsNull` or `HasAValue` meaning..

See [this comment](https://github.com/Combodo/iTop/pull/379#issuecomment-1354554067)
`IsNull` should actually be deprecated as its implementation does not reflect is naming. But first we should check each usage of `IsNull` to see if we are testing whether the values is exactly `null` or if it has not value which is not the same.

rquetiez (approved):
Still not sure that the naming/semantic problem has been 100% solved, but for sure we have made some progress: the design has been discussed and is now better understood.

Hipska:
Okay to deprecate `IsNull` in favour of this new method, although I liked the naming of it more.

What will `HasAValue` return for a decimal or integer field having the value `0`?

Molkobain:
> What will `HasAValue` return for a decimal or integer field having the value `0`?

It will return `true`, for that kind of attribute it will return false only if the value is `null` or `''`.
The sole purpose of this method is to know if there is a value in the attribute, no matter if it comes from the user, the default value, a prefill form or else.

Hipska:
Okay, that is perfect. I think it wasn't like that in the beginning, or you communicated it differently..

PR #380, opened 2022-12-20
N°4852 - Navigation menu: Display "+" dict entries as a tooltip on sub-nodes

v-dumas:


application/menunode.class.inc.php
```diff
@@ -320,12 +320,13 @@ public static function GetSubMenuNodes($sMenuGroupIdx, $aExtraParams = array())
 			}
 
 			$aSubMenuNodes[] = [
-				'sId' => $oSubMenuNode->GetMenuId(),
-				'sTitle' => $oSubMenuNode->GetTitle(),
-				'bHasCount' => $oSubMenuNode->HasCount(),
-				'sUrl' => $oSubMenuNode->GetHyperlink($aExtraParams),
+				'sId'              => $oSubMenuNode->GetMenuId(),
+				'sTitle'           => $oSubMenuNode->GetTitle(),
+				'sLabel'           => $oSubMenuNode->GetLabel(),  // Bug #4852
```
piRGoif: No need to add this inline comment : we will get the information from the git history.

However, it would be interesting to understand what is the value returned by \MenuNode::GetLabel, so maybe add more information in this later method PHPDoc block ? 

Also, you can document the \ApplicationMenu::GetSubMenuNodes returned array with the [array shapes syntax](https://www.itophub.io/wiki/page?id=latest:customization:coding_standards#arrays_with_named_keysarray_shapes) ?

piRGoif (changes requested):
Maybe some more efforts on the corresponding documentation...

templates/base/layouts/navigation-menu/menu-node.html.twig
```diff
@@ -1,18 +1,18 @@
...
-			<span class="ibo-navigation-menu--menu-node-label" data-role="ibo-navigation-menu--menu-node-label" title="{{ aMenuNode.sTitle }}">{{ aMenuNode.sTitle }}</span>
+                <span class="ibo-navigation-menu--menu-node-counter ibo-is-hidden" data-menu-id="{{ aMenuNode.sId }}"></span>
+            {% endif %}
+        </a>
+    {% else %}
+        <span class="ibo-navigation-menu--menu-node-title" data-role="ibo-navigation-menu--menu-node-title">
+			<span class="ibo-navigation-menu--menu-node-label" data-role="ibo-navigation-menu--menu-node-label" title="{{ aMenuNode.sLabel }}">{{ aMenuNode.sTitle }}</span>
 		</span>
-	{% endif %}
+    {% endif %}
```
Molkobain: Careful, your IDE replaced the tab indentions with white spaces. Please revert it and adjust your IDE settings.

application/menunode.class.inc.php
Molkobain: I'm not comfortable with the description being retrieved with a method named `GetLabel` as it is very similar to `GetTitle`. Not intuitive at all to me, something like `GetDescription` should have been way better, but that's not your fault, they are already named like this 😕

As for Pierre's comment, indeed the bug number is not mandatory but if you keep it, please use the internal bug tracker convention (eg N°4852) as `#` is for SourceForge tickets.

Molkobain (changes requested):

application/menunode.class.inc.php
v-dumas: Does [this](https://github.com/Combodo/iTop/pull/380/commits/8902033b78dab1cd6acba9b7ec87f47f8be2caad) do the job ?

application/menunode.class.inc.php
```diff
@@ -746,7 +757,7 @@ public function GetEntriesCount()
 	}
 
 	/**
-	 * @return string
+	 * @return string the + dictionary entry for this menu if exists, otherwise the Title
```
Molkobain:
```suggestion
	 * @return string The "+" dictionary entry for this menu if exists, otherwise the Title
```

Molkobain (approved):
All good to me!

Molkobain:
Once it's ok with Pierre as well, poke us so we squash & merge the PR

application/menunode.class.inc.php
```diff
@@ -746,7 +757,7 @@ public function GetEntriesCount()
 	}
 
 	/**
-	 * @return string
+	 * @return string The "+" dictionary entry for this menu if exists, otherwise the Title
```
piRGoif: Actually this isn't totally true as if there aren't any "+" dict key, we won't simply call GetTitle...

```suggestion
	 * @return string The "+" dictionary entry for this menu if exists, otherwise the Title (if we have a parent title, will output parentTitle / currentTitle)
```

piRGoif (changes requested):
Thanks for all those modifications 👌
Just a little remark on a phpdoc

application/menunode.class.inc.php
piRGoif: Thanks for the modifications ! I've just made a new review.
Closing this conversation.

application/menunode.class.inc.php
v-dumas: Do you have an example, where this combination of title is occurring, I have never seen it on any iTop...
sub-menu-title

piRGoif: I described the code I saw in GetLabel(), but currently we are using GetTitle in the menu right ? 
There are multiple callers for this \MenuNode::GetLabel method but might not be very easy to spot where the parent + current string can appear ?

piRGoif (approved):

PR #382, opened 2022-12-22
N°2916 - Import of IPv6 addresses fails when reconciliation is done on the ip

accognet:
Add an interface that should be implemented in order to handle  object type fields on import. It adds a function to transform the object (like ormIPv6) into a string.
This function is used twice during the CVS import:
- to find the corresponding objects in iTop if the field is a search criterion.
- when updating the object if the field is read-only.

### DIFF
--- a/core/bulkchange.class.inc.php
+++ b/core/bulkchange.class.inc.php
@@ -11,7 +11,7 @@ define('UTF8_BOM', chr(239).chr(187).chr(191)); // 0xEF, 0xBB, 0xBF
 
 /**
  * CellChangeSpec
- * A series of classes, keeping the information about a given cell: could it be changed or not (and why)?  
+ * A series of classes, keeping the information about a given cell: could it be changed or not (and why)?
  *
  * @package     iTopORM
  */
@@ -144,7 +144,7 @@ class CellStatus_Ambiguous extends CellStatus_Issue
 
 /**
  * RowStatus
- * A series of classes, keeping the information about a given row: could it be changed or not (and why)?  
+ * A series of classes, keeping the information about a given row: could it be changed or not (and why)?
  *
  * @package     iTopORM
  */
@@ -220,7 +220,7 @@ class RowStatus_Issue extends RowStatus
  */
 class BulkChange
 {
-	protected $m_sClass; 
+	protected $m_sClass;
 	protected $m_aData; // Note: hereafter, iCol maybe actually be any acceptable key (string)
 	// #@# todo: rename the variables to sColIndex
 	protected $m_aAttList; // attcode => iCol
@@ -261,7 +261,7 @@ class BulkChange
 		$this->m_sReportCsvSep = $sSeparator;
 		$this->m_sReportCsvDelimiter = $sDelimiter;
 	}
-   
+
 	protected function ResolveExternalKey($aRowData, $sAttCode, &$aResults)
 	{
 		$oExtKey = MetaModel::GetAttributeDef($this->m_sClass, $sAttCode);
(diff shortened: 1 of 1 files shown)


core/metamodel.class.php
```diff
@@ -4180,18 +4180,18 @@ public static function PrepareQueryArguments($aArgs, $aMoreArgs = array(), $aExp
...
-					{
+					} elseif (is_null($value)) {
 						$aScalarArgs[$sArgName] = null;
-					}
-					elseif (is_array($value))
-					{
+					} elseif (is_array($value)) {
 						$aScalarArgs[$sArgName] = $value;
+					} else {
+						if (is_object($value) && method_exists($value, 'GetValueForQuery')) {
```
Hipska:
```suggestion
						if ($value instanceof iOrmset) {
```
https://3v4l.org/RRMSR

core/iOrmset.php
```diff
@@ -0,0 +1,20 @@
...
+ * @copyright   Copyright (C) 2010-2022 Combodo SARL
+ * @license     http://opensource.org/licenses/AGPL-3.0
+ */
+
+/*
+ * this interface add a function witch allow import on object (like ormIPv6)
+ * @since 3.1.0
+ */
+
+interface iOrmset
```
Hipska: I don't think this is actually a set of ORM objects? If it is, shouldn't the naming more be like this?
```suggestion
interface iOrmSet
```

Hipska (changes requested):

core/metamodel.class.php
accognet: This was my first idea, but this fix is ​​for Teemip first. I waited to meet xtophe and validate that it's ok for him to lose 3.0 compatibility in his extension. I can now make this change.

Molkobain:
Discussed during technical review, Anne-Catherine and Christophe will:
- Revert modifications in the core
- Work on the AttributeIPXX attributes to implement the `MakeValueFromString` methods as it is how the CSV import should build the value for reconcialiation.
- Check where it fails then
- Adjust the PR with the new fix if necessary

Molkobain (changes requested):
Check previous comment

core/metamodel.class.php
Hipska: How will this break 3.0 compatibility?

piRGoif:
> * Work on the AttributeIPXX attributes to implement the `MakeValueFromString` methods as it is how the CSV import should build the value for reconcialiation.

The method PHPDoc might be changed also ? Currently it is : 

`Get the value from a given string (plain text, CSV import)`

Maybe explain "the value" is for example ormSet, etc ?

core/iOrmset.php
Hipska: Hi @accognet,  you resolved, but I didn't see an answer to my question?

Hipska (approved):

Hipska:
Would this also solve the issue mentioned in https://github.com/TeemIp/teemip-core-ip-mgmt/issues/14?

Molkobain:
Anne-Catherine will check with Olivier to add a test for a composite attribute (eg. a linkset)

Molkobain (approved):

Molkobain:
Discussed with Alexandre during iTop 3.0.3 init meeting, if the unit test is made / validated before code freeze (2023/03/10), the PR will be merged into the 3.0.3, otherwise it will be postponed.

xtophe38:
Successfully tested on develop & support/3.0 branches with coming TeemIp 3.1.

PR #383, opened 2023-01-02
N°5841 - Non-admin managing User can't see Administrator by default

v-dumas:


Hipska:
Please add a description for public pull requests

Molkobain:
> Please add a description for public pull requests

Hello Thomas,

We tend to use PRs more and more for internal code review of planned/ongoing development tickets. Such tickets (no "Project" assigned on the PR) are different from other internal PRs as the description is already written in french in our internal bug tracker. We won't necessarily take the time to translate the description and transcript the functional discussion from its case log.

Only PRs which are a "personal initiative" must have a detailed description in our process.
Ideally we would have a public bug tracker and have it sync with the corresponding PR, but that not for any time soon 😥

Cheers,
Guillaume

Molkobain (changes requested):
PR contains commits from another PR (#380), your branch doesn't diverge from the right commit. I'll get in touch with you.

Hipska:
I ask this so the community has some context in order to do a peer review. Doesn’t have to be the complete detailed bug report, user story, internal discussion and so on. Only a few lines are enough. Otherwise you could just keep the PR internal.

Molkobain:
> I ask this so the community has some context in order to do a peer review. Doesn’t have to be the complete detailed bug report, user story, internal discussion and so on. Only a few lines are enough. [...]

Yes and I totally understand, that's why I insist so there is at least the explicit PR title.

> [...] Otherwise you could just keep the PR internal.

Do you know how to make private / internal PRs on a public repo without forking it?

Molkobain (approved):

Hipska:
> Do you know how to make private / internal PRs on a public repo without forking it?

No, a private fork (or internal on a local system) should be created in order to do that. It makes sense as the bug is also on an internal tracker.

PR #386, opened 2023-01-13
Missing org_id in network interface class

dariocodes:
Added org_id to the network interface class so that network interfaces are not shown for every organization, This was especially an issue for TEEMIP extensions. The next step after this pull request is confirmed and tested is to add the org_id to all the relevant interfaces in the TEEMIP extensions. I personally already implemented this in my own itop production environment.

### DIFF
--- a/datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
+++ b/datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
@@ -5385,6 +5385,8 @@
         <naming>
           <attributes>
             <attribute id="name"/>
+            <attribute id="org_id"/>
+						<attribute id="organization_name"/>
           </attributes>
         </naming>
         <style>
@@ -5403,6 +5405,16 @@
           <default_value/>
           <is_null_allowed>false</is_null_allowed>
         </field>
+        <field id="org_id" xsi:type="AttributeExternalKey" _delta="define">
+					<sql>org_id</sql>
+					<target_class>Organization</target_class>
+					<is_null_allowed>false</is_null_allowed>
+					<on_target_delete>DEL_MANUAL</on_target_delete>
+				</field>
+				<field id="organization_name" xsi:type="AttributeExternalField" _delta="define">
+					<extkey_attcode>org_id</extkey_attcode>
+					<target_attcode>name</target_attcode>
+				</field>
       </fields>
       <methods/>
       <presentation>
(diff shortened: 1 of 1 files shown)


Molkobain:
Hello @dariocodes ,

Thanks for the PR!

We usually don't accept PR relative to any datamodel changes as stated in the CONTRIBUTING.md, but I'll present it during the functional review anyway. Will keep you updated. :)

> All datamodel modification should be done in an extension. Beware that such change would impact all existing customers, and could prevent them from upgrading! Combodo has a long experience of datamodel changes: they are very disruptive! This is why we avoid them in iTop core, especially the changes on existing objects/fields.
> If you have an idea you're sure would benefit to all of iTop users, you may [create a corresponding ticket](https://sourceforge.net/p/itop/tickets/new/) to submit it, but be warned that there are lots of good reasons to refuse such changes.

Did you discuss this with @xtophe38 to integrate this change directly in TeemIP instead of iTop?

Guillaume

Hipska:
@Molkobain I also don't like the way it is implemented as discussed in https://github.com/TeemIp/teemip-core-ip-mgmt/issues/18

dariocodes:
@Molkobain Thanks for the reply. I am quite new to git commits but I'm an avid user of both itop and teemip and ran into this issue so I figured I'd try and contribute. As I said I implemented this in this manner on my own instance. I could also create a different implementation using the `AttributeExternalField` as @Hipska suggested. I'll also keep in mind the guidelines next time I'll contribute. 

I haven't tried implementing it directly into TEEMIP yet. this would definitely be a better option if possible.

Anyway, thanks for taking it into consideration.

Molkobain:
Thank you both for the extra info!
FYI, technical review happen the first tuesday of each month, functional review the second tuesday of each month.

piRGoif:
Maybe @v-dumas you can give you advise in the meantime ?

v-dumas:
I don't have a definitive answer yet. It will depends on the impact such change would generate on Combodo's customer.
If this ends up not to be an issue, then the preferred method would be an AttributeExternalField.
As it takes about 5 minutes to add this AttributeExternalField for any Combodo customers using the ITSM Designer, it did not get any priority. Especially as changing it on the Community version, will cost another 10min for each customer, assuming they understood and saw the message telling them to revert their own change...

xtophe38:
Hello,

I tend to agree that it would be better if network interfaces could be "protected' from being visible by any iTop user. As a consequence, I'll integrate in TeemIp an external fields to network interface final classes.

Molkobain:
Fix has been made directly in TeemIP.
  - Issue [TeemIP#18](https://github.com/TeemIp/teemip-core-ip-mgmt/issues/18)
  - Commit https://github.com/TeemIp/teemip-core-ip-mgmt/commit/6f6b31d288630de1d38f4eef74ad5929e5825d60

So we are closing the PR on this side :)

PR #100, opened 2019-11-27
N°2783 - Add support for custom zlists

Hipska:
This is a discussion with @dflaven about the possibility to let extensions add new types of ZList to be used if that extension creates a new type of presentation of the objects. I saw it is supported by `MetaModel` class under [RegisterZList](https://github.com/Combodo/iTop/blob/3a791162c5993f6f1f04ec26879652128a356946/core/metamodel.class.php#L2728) to register new types: [Standard attribute lists](https://github.com/Combodo/iTop/blob/3a791162c5993f6f1f04ec26879652128a356946/core/metamodel.class.php#L7529-L7538)

The only problem is that the setup compiler only compiles some predefined ZLists. I already made some changes to that, but it needs to be tested and reviewed by @Combodo.

PR #389, opened 2023-01-17
N°2783 - Add support for custom zlists

Molkobain:
This is a reboot of #100 as we need to move forward and Thomas lacks time.

**Collection**
  * Custom zlists must be defined under the `/itop_design/classes/class/presentation/custom_presentations` which will contain zlists brought by customizations (extensions, Designer, ...)
  * ~This collection node must be added empty to all standard DM classes~ => Not anymore
  * ~This collection node must be added empty by the compiler on XML datamodel files during the version upgrade (see _iTopDesignFormat::From30To31()_)~ => Not anymore, it's an optional node

**Definition of the zlist**
  * Each custom zlist will be represented by a `<custom_presentation id="ZLIST_ID">[...]</custom>` node
  * The node must contain the same XML structure as other zlists (no particular test to do at compile time)
  * Custom zlist can (optional) expose a description of their purpose so it will be displayed in the Designer as they will be no specific UI display (nothing to there for you Thomas, just to explain why it's in the example below)

**Implementation**
  * Compiler should **append** custom zlist to the standard ones (= same PHP structure)
  * Compiler should throw an exception if a custom zlist has the same name (ID) as a standard one (need for explicit ID will be added in the documentation)
  * `MetaModel::GetZListItems()` will be used to retrieve both standard and custom zlists. If a custom zlist does not exist, it will return an empty array as it already does.

**Example**
```
<itop_design>
    <classes>
        <class id="UserRequest">
            <presentation>
                <details>[...]</details>
                <list>[...</list>
                [...]
                <custom_presentations>
                    <custom_presentation id="webhook-action-slack">
                        <items>
                            <item id="caller_id">
                                <rank>10</rank>
                            </item>
                            <item id="org_id">
                                <rank>20</rank>
                            </item>
                        </items>
                    </custom_presentation>
                </custom_presentations>
            </presentation>
        </class>
        <class id="Incident">
            <presentation>
                <custom_presentations>
                    <custom_presentation id="webhook-action-slack">
                        <items>
                            <item id="caller_id">
                                <rank>10</rank>
                            </item>
                        </items>
                    </custom_presentation>
                </custom_presentations>
            </presentation>
        </class>
    </classes>
    <meta>
        <presentation>
            <custom_presentations>
                <custom_presentation id="webhook-action-slack">
                    <description><![CDATA[Will be used by the "Slack notification" Action to display the corresponding attributes in the Slack message. Refer to the "Webhooks integration" extension documentation for more information.]]></description>
                </custom_presentation>
            </custom_presentations>
        </presentation>
    </meta>
</itop_design>
```

**Needs to be done**
  * [ ] Update XML Reference in the wiki
  * [X] Discuss what we want to do in XML conversion methods (iTopDesignFormat)
    We just need to clean nodes when downgrading. Don't automatically add nodes on upgrade as initially discussed.
  * [X] Add unit tests for XML conversion methods

### DIFF
--- a/setup/compiler.class.inc.php
+++ b/setup/compiler.class.inc.php
@@ -95,6 +95,11 @@ class MFCompiler
 	protected $oFactory;
 
 	protected $aRootClasses;
+	/**
+	 * @var array $aCustomListsCodes Codes of the custom zlists
+	 * @since 3.1.0
+	 */
+	protected $aCustomListsCodes;
 	protected $aLog;
 	protected $sMainPHPCode; // Code that goes into core/main.php
 	protected $aSnippets;
@@ -652,7 +657,21 @@ EOF;
 				$aWebservicesFiles[] = "MetaModel::IncludeModule(MODULESROOT.'/$sRelativeDir/$sRelFileName');";
 			}
 		} // foreach module
-		
+
+		// Register custom zlists
+		$this->sMainPHPCode .= <<<PHP
+/**
+ * Custom zlists 
+ */
+PHP;
+		foreach ($this->aCustomListsCodes as $sCustomListCode) {
+			// Note: HEREDOC used to ease finding of \MetaModel::RegisterZList() method usages
+			$this->sMainPHPCode .= <<<PHP
+MetaModel::RegisterZList('$sCustomListCode', ['description' => 'Custom zlist $sCustomListCode', 'type' => 'attributes']);
+PHP;
+		}
+		$this->sMainPHPCode .= "\n";
+
 		// Compile the dictionaries -out of the modules
 		//
 		$sDictDir = $sTempTargetDir.'/dictionaries';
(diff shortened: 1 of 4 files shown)


Hipska:
Not “customs”. As said that’s a totally different thing. (Douane)

Molkobain:
> Not “customs”. As said that’s a totally different thing. (Douane)

It already "custom_presentations" if you look at the code, just forgot to update the description. Thanks 😁

Hipska (approved):
Looks good, I just don't think it's needed to be added on every class already? Doesn't harm either anyway..

Molkobain:
> Looks good, I just don't think it's needed to be added on every class already? Doesn't harm either anyway..

Yeah, that was reverted. I don't recall why we said it should be in the first place 😅

setup/itopdesignformat.class.inc.php
```diff
@@ -1133,6 +1133,9 @@ protected function From31To30($oFactory)
 		// N°5563 AttributeLinkedSetIndirect
 		// - remove read_only attribute (added in 3.1)
 		$this->RemoveNodeFromXPath("/itop_design/classes/class/fields/field[@xsi:type='AttributeLinkedSetIndirect']/read_only");
+
+		// N°2783 Custom zlists
+		$this->RemoveNodeFromXPath("/itop_design/classes//class/presentation/custom_presentations");
```
Hipska, line 1136: The item above only has 1 / between `classes` and `class`. Typo?

setup/compiler.class.inc.php
```diff
@@ -3843,6 +3870,25 @@ public static function Init()
...
+	 *
+	 * @return string PHP Code to declare a zlist
+	 * @since 3.1.0 N°2783
+	 */
+	private function GeneratePhpCodeForZlist(string $sListCode, DOMNode $oListNode): string
+	{
+		$aAttributes = $oListNode->GetNodeAsArrayOfItems();
+		if(!is_array($aAttributes)) {
+			$aAttributes = array();
+		}
```
Hipska, line 3898: Strange, a method named `GetXXXAsArray` not returning an array?

Hipska (approved):

setup/compiler.class.inc.php
Molkobain: Yeah... Check the method documentation / implementation. it can return a string 😕

setup/itopdesignformat.class.inc.php
Molkobain: The item above should be fixed (and has actually been fixed in the develop branch with 8e3bad11ae7bd6ae580a4cfbebce5c0b801b54a3). The double slash is there because in some contexts the classes can be represented as a tree, not flatten.

setup/compiler.class.inc.php
```diff
@@ -2179,30 +2198,38 @@ protected function RegisterEventListeners()
...
+
+			// Cannot have a custom zlist with an ID that is among the reserved ones {@see $aListRef}
+			if (array_key_exists($sListCode, $aListRef)) {
+				throw new DOMFormatException('Custom zlist "'.$sListCode.'" cannot be compiled as it is using a reserved identifier ('.implode(', ', array_keys($aListRef)).')');
+			}
 
-				$sZAttributes = var_export($aAttributes, true);
-				$sZlists .= "		MetaModel::Init_SetZListItems('$sListCode', $sZAttributes);\n";
+			// Store custom zlist code for further registration
+			if (false === in_array($sListCode, $this->aCustomListsCodes)) {
```
piRGoif, line 2241: Might log a warning if there is a duplicate id ?

piRGoif (changes requested):
Indeed missing tests ;)
You could add easily the code for the XML migration methods as a start ?

setup/compiler.class.inc.php
Molkobain: Why? A same custom list ID can be defined in 2 classes.

Molkobain:
> Indeed missing tests ;) You could add easily the code for the XML migration methods as a start ?

Yes, it's in the checklist at the bottom of the description and why the PR is still in draft.
Why the "requested changes"?

piRGoif:
> Why the "requested changes"?

Because of missing tests :D But yes your pr is in draft, I should have added simple comments instead of a review.

setup/compiler.class.inc.php
piRGoif: Well I was thinking of multiple declarations of the same custom id on the same class... but actually the compiler should already deal with that thanks to the _delta standard processing. 
Closing conversation, sorry for the noise :/

Molkobain:
_As discussed in the original PR (#100), this was already accepted during functional review. Only technical review is necessary._

Hipska (approved):

Molkobain:

PR #390, opened 2023-01-19
N°5893 - Log triggers exception in CRUD stack

piRGoif:
When a trigger throws an exception, blocking the object operation, it's currently very hard to understand what's going on.
This PR adds a log on each trigger exception (insert, update, delete, apply stimulus) that will help to diagnose such error.

In an ideal world, all and every external code call (method override, extensibility interfaces) should be protected by a try/catch block... But it would cost a little bit much, and so another ticket exists to have a greater cover : N°3514.

### DIFF
--- a/core/trigger.class.inc.php
+++ b/core/trigger.class.inc.php
@@ -284,7 +284,7 @@ abstract class TriggerOnObject extends Trigger
 	 */
 	public function LogException($oException, $oObject)
 	{
-		$sObjectKey = ($oObject->IsNew()) ? 'N/A' : $oObject->GetKey();
+		$sObjectKey = $oObject->GetKey(); // if object wasn't persisted yet, then we'll have a negative value
 
 		$aContext = [
 			'exception.class'      => get_class($oException),
--- a/core/trigger.class.inc.php
+++ b/core/trigger.class.inc.php
@@ -259,20 +259,47 @@ abstract class TriggerOnObject extends Trigger
 	public function IsTargetObject($iObjectId, $aChanges = array())
 	{
 		$sFilter = trim($this->Get('filter'));
-		if (strlen($sFilter) > 0)
-		{
+		if (strlen($sFilter) > 0) {
 			$oSearch = DBObjectSearch::FromOQL($sFilter);
 			$oSearch->AddCondition('id', $iObjectId, '=');
 			$oSearch->AllowAllData();
 			$oSet = new DBObjectSet($oSearch);
 			$bRet = ($oSet->Count() > 0);
-		}
-		else
-		{
+		} else {
 			$bRet = true;
 		}
 
 		return $bRet;
+	}
+
+	/**
+	 * @param Exception $oException
+	 * @param \DBObject $oObject
+	 *
+	 * @return void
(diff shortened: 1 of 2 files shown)


core/trigger.class.inc.php
```diff
@@ -259,21 +259,48 @@ public function DoActivate($aContextArgs)
...
+	 *
+	 * @return void
+	 *
+	 * @uses \IssueLog::Error()
+	 *
+	 * @since 2.7.9 3.0.3 3.1.0 N°5893
+	 */
+	public function LogException($oException, $oObject)
+	{
+		$sObjectKey = ($oObject->IsNew()) ? 'N/A' : $oObject->GetKey();
```
Molkobain: Don't you want to log the actual temp key for the object? Maybe it could help to see in some cases that it's -2 or -3 that caused the failure. I don't have any use case in mind, just that it's common knowledge that negative keys are for objects that are not stored yet, so it's a waste to hide this info IMO.

core/trigger.class.inc.php
```diff
@@ -259,21 +259,48 @@ public function DoActivate($aContextArgs)
...
+	 * @param Exception $oException
+	 * @param \DBObject $oObject
+	 *
+	 * @return void
+	 *
+	 * @uses \IssueLog::Error()
+	 *
+	 * @since 2.7.9 3.0.3 3.1.0 N°5893
+	 */
+	public function LogException($oException, $oObject)
```
Molkobain, line 285: The current implementation suppose that a trigger is always activated by an object, but it is not necessarily the case is it?
To me only triggers derived from TriggerOnObject are.

We don't have any yet, but we could totally have triggers activated by something else like a timing (every 2 hours). We should discuss this we Romain as well, but keeping the object as not mandatory to the base trigger class would be best IMO.

piRGoif: > The current implementation suppose that a trigger is always activated by an object, but it is not necessarily the case is it?

That's why this LogException method is defined in the TriggerOnObject class and not on the Trigger root class :o)

core/trigger.class.inc.php
piRGoif: OK
Changed in 72b9c396

core/trigger.class.inc.php
Molkobain: Oh my bad then! I shouldn't have check on the GitHub GUI 😁

piRGoif: No worries, and thanks for the review :)
Closing conversation

Molkobain (approved):

odain-cbd:
nice work. seems good to me. 

maybe you could add a test to show how LogException method works. near the one that covers EnrichRaisedException...

a full test that throws a trigger exception may be hard to write as you need to mock things inside our CRUD framework. it is possible but lot of energy....

piRGoif:
@odain-cbd 
> maybe you could add a test to show how LogException method works. near the one that covers EnrichRaisedException...

Sorry the PR was merged quickly and I wasn't there.
After our discussion this morning, test added in 5ee62234.
Thanks !

PR #391, opened 2023-01-24
:sparkles: N°3190 - Edit n:n LinkedSetIndirect in object details using a tagset-like widget

bdalsass:
- Add generic set block ui component
- Add model link set (direct and indirect) attribute (display style)
- Add model link set direct allowed values
- Create link set viewer block UI (BlockLinksSetDisplayAsProperty)
- Add set block ui factory for linkset
- Add object factory and create new endpoint in object controller (with data binder)
- Add link set model, link set repository and link set data transformer services


Hipska:
What is this change actually doing?

Also, why bothering creating a PR if you immediately merge it?

PR #393, opened 2023-01-25
N°5920 - Add linkset's description as corresponding tab's tooltip in object details

Molkobain:
**Objective**
Display the description of a linkset attribute on its tab toggler in the object details. It would avoid to open all tabs to know their purpose.

The description we are talking about:
image

**More global approach**
Add support for any tab to have a description, not just the linkset attributes of an object. Any tab container benefits of it.
This description will then be used as a tooltip on its "tab toggler", but it could also be used in the "tab content" or elsewhere.

With that in mind a new tab description property has been added to all tabs.

**Result**
For "regular" tab togglers, only the description is displayed as the label is never truncated.
image

For tab togglers in the "extra tab" dropdown, both tab label and description are displayed as the label could be truncated if too long.
_Always having the label in the tooltip can be redundant, but I don't want to use JS to detect if it is truncated or not_
image


templates/base/layouts/tab-container/layout.html.twig
```diff
@@ -41,7 +46,14 @@
                         <div class="ibo-tab-container--extra-tabs-list ibo-is-hidden" data-role="ibo-tab-container--extra-tabs-list">
                             {% for oTab in oUIBlock.GetSubBlocks() %}
                                 <a href="#tab_{{ oTab.GetId()|sanitize(constant('utils::ENUM_SANITIZATION_FILTER_ELEMENT_IDENTIFIER')) }}"
-                                   class="ibo-tab-container--extra-tab-toggler" data-role="ibo-tab-container--extra-tab-toggler">
+                                   class="ibo-tab-container--extra-tab-toggler" data-role="ibo-tab-container--extra-tab-toggler"
+                                    {% if oTab.HasDescription() == true %}
+                                        {# Display both title and description as the title could be truncated if too long #}
+                                        data-tooltip-content="{{ oTab.GetTitle() }}{{ '<br/><br/>' }}{{ oTab.GetDescription() }}"
```
Molkobain: Not really proud of this, but I don't really see another to do it without being overkill. 🤔

steffunky: You're already using `data-tooltip-html-enabled`, you could just add a `<div class="ibo-tab-container--extra-tab-toggler--tooltip--tab-title">` around your `{{ oTab.GetTitle() }}` and write a CSS rule with padding

Hipska (approved):

templates/base/layouts/tab-container/layout.html.twig
Molkobain: Done with [2289d45](https://github.com/Combodo/iTop/commit/2289d45883412d73c95964256c41179a40b05618)
Validated IRL with Stephen

PR #241, opened 2021-10-28
Add uniqueness rule to lnkContactToFunctionalCI

Hipska:
While the UI for both `FunctionalCI` and `Contact` dictate there can be no duplicates for the linked list, it is still possible to create duplicates by using workarounds like for example import.

By adding a rule, this will be prevented and doesn't confuse users.

PR #416, opened 2023-01-26
Feature/5825 -Links friendlynames & uniqueness

v-dumas:
Add on all Standard Datamodel Links classes:
- Friendlyname order aligned to the class Name
- Align details order to friendlyname (& class name logic)
- Uniqueness rule to prevent duplicates


Hipska:
Okay, this one should close #241 when merged.

PR #440, opened 2023-01-30
N°5904 - Use attribute linked set edit mode to enable actions

bdalsass:


Molkobain (changes requested):
Discussed IRL with Benjamin & Stephen, JS selectors with ID will cause collision when multiple linksets in an object.

Benjamin will change this to have a approach similar to what's done on UIBlocks with data-role attributes.

sources/Application/UI/Base/Component/Input/Set/Set.php
```diff
@@ -183,6 +188,30 @@ public function HasRemoveItemButton(): bool
...
 	}
 
+	/**
+	 * SetAddOptionButtonJsOnClick.
+	 *
+	 * @param string $sJsOnClick
+	 *
+	 * @return $this
+	 */
+	public function SetAddOptionButtonJsOnClick(string $sJsOnClick): Set
```
Molkobain, line 198: We don't put type hint on return as it will be an issue with inheritance. Will talk about this further IRL.
```suggestion
	public function SetAddOptionButtonJsOnClick(string $sJsOnClick)
```

Molkobain (commented):
By moving some JS files (`js/linkswidget.js`, `js/linksdirectwidget.js`) to another folder, you are going to break some extensions. If you keep this approach, you must:
  * Create the corresponding ticket in the bug tracker
  * Create the audit rule for these files to anticipate migration of the concerned extensions

For example `js/linksdirectwidget.js` is used on combodo-series, combodo-alarm-console, ...
Same for others.

We can also think about adding these JS files to the iTopWebPage::COMPATIBILITY_MOVED_LINKED_SCRIPTS_REL_PATH array to ease migration. Either way, we should talk about this IRL.

Molkobain (approved):

PR #442, opened 2023-02-06
N°5942 AttributeBlob now displays an image when applicable

piRGoif:
As pointed in [SF #2130](https://sourceforge.net/p/itop/tickets/2130/), when displaying a list of DocumentFile we don't have thumbnails when attached files are images.

This PR pulls up methods GetAsHtml an related methods from AttributeImage to AttributeBlob : now if an AttributeBlob contains an image it will do the same as AttributeImage. See this DocumentFile list after update : 

image

The mime type detection is in the new method \AttributeBlob::IsWebImage

Note that the old \AttributeImage::GetAsHTML is using multiple attribute parameters that are defined in the compiler : 

* display_max_width
* display_max_height
* default_image

I made the bold move of moving those init from AttributeImage to AttributeBlob... But a better way might be to only keep default values ? To be discussed during review.

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -180,20 +180,19 @@ abstract class AttributeDefinition
 	/** @var string */
 	protected $m_sHostClass = '!undefined!';
 
-	public function Get($sParamName)
-	{
-		return $this->m_aParams[$sParamName];
+	/**
+	 * @since 3.1.0 N°5942 new defaultValue param
+	 */
+	public function Get($sParamName, $defaultValue = null) {
+		return $this->m_aParams[$sParamName] ?? $defaultValue;
 	}
 
-	public function GetIndexLength()
-	{
+	public function GetIndexLength() {
 		$iMaxLength = $this->GetMaxSize();
-		if (is_null($iMaxLength))
-		{
+		if (is_null($iMaxLength)) {
 			return null;
 		}
-		if ($iMaxLength > static::INDEX_LENGTH)
-		{
+		if ($iMaxLength > static::INDEX_LENGTH) {
 			return static::INDEX_LENGTH;
 		}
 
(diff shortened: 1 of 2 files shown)


Molkobain:
Technical implementation left apart, from my PoV the current functional implementation raises a lot of functional questions:
  * You loose the file name and size in lists which are handy when searching / looking for something specific (eg. a particular filename, a file bigger than a certain size,...)
  * You can't open / download the image as excepted with a file attribute in the object details.

Having a preview available on hover like it's done in the attachement could be another approach. But that will be discussed during the functional review 😁

piRGoif:
> You can't open / download the image as excepted with a file attribute in the object details.

What do you mean ? 
In the DocumentFile details screen, I can click on the picture and opens the file :
image

piRGoif:
Rebased, solved conflicts and forced pushed

piRGoif:
Rebase and force push
Solved some conflicts with N°6240 / #484

piRGoif:
Technical review :
- Move IsWebImage to the utils method ?
- There is the existing \InlineImage::IsImage method, is it duplicate ? We should use the same method for both needs
- AttDef parameters (max_width etc) : will crash as is in AttributeImage (test on class name, not instance of). We should pick values from methods in AttributeBlob which will be overriden in AttributeImage ?

But we agreed to keep it as is, as we need to discuss functionally : 
- Having file size / name etc in a big list is nice to be able to sort, filter, ...
- On export, can we select distinct blob properties (file name, size, mime type, ...) ?
- A solution could be to display the image in a tooltip (same as hovercard on ext keys)

TO be continued in functional review !

Molkobain:
Declined during functional review, Pierre will give more details.

piRGoif:
There were multiple reasons invoked for rejecting this PR.
Main one is that having an attribute rendering dependent of the attribute value seems like a very bad idea.

Two things we saw during functional review : 
- on export, we have the file URL
- in the details page, a file preview is displayed

A possible solution would be to display the file preview as a tooltip when hovering ?

Anyhow, closing this PR.

Hipska:
The image preview tooltip on hovering sounds like a good idea indeed.

piRGoif:
Yes :) Now just need for a good will to code this behavior :)

PR #444, opened 2023-02-10
N°3190 - Add XML version migration for new AttributeLinkedSet[Indirect] <display_style> tag

bdalsass:


Hipska:
May I ask when this `From31To30` migration takes place? In which scenario does that happen?

Molkobain (approved):

Molkobain:
> May I ask when this `From31To30` migration takes place? In which scenario does that happen?

During MTT / MTP in the ITSM Designer when working on an iTop that is older than 3.1.

PR #445, opened 2023-02-13
N°5906 Launch Ticket impact analysis after lnk modification

piRGoif:
For now analysis impact update is only launched after creating / modifying each Ticket children object (OnInsert / OnUpdate callbacks).
But in 3.1 lnk can be edited directly in the GUI !

This moves analysis impact update in a listener on the EVENT_DB_LINKS_CHANGED event.
On each lnk modification cmdbAbstractObject will record remote objects (\cmdbAbstractObject::$aLinkModificationsStack attribute).
Those records will be used : 
* just before firing \EVENT_DB_UPDATE_DONE, if the current object was recorded, fires EVENT_DB_LINKS_CHANGED 
* at the end of the session (UI.php and ajax.render.php) send one EVENT_DB_LINKS_CHANGED  for each remaining recorded object

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -4559,6 +4559,8 @@ HTML;
 				static::FireEventDbLinksChangedForAllObjects();
 			}
 		}
+
+		return $oDeletionPlan;
 	}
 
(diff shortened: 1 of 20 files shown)


piRGoif:
The PR was merged manually by Eric in 8940051c

PR #447, opened 2023-02-16
N°5993 - Add purge mechanism to log system

bdalsass:
Add purge mechanism:

If `log_purge_enable `is set to true, clean log folder by deleting files with last modification time older than `log_purge_max_keep_daysdays`.

- Add 2 new configuration parameters: log_purge_enable, log_purge_max_keep_days.
- Add purge process to log rotating scheduled task


core/config.class.inc.php
```diff
@@ -113,70 +113,86 @@ class Config
...
-			'value' => '',
-			'source_of_value' => '',
+		'log_purge_enable'                   => [
+			'type'                => 'bool',
+			'description'         => 'Optional purge activation.',
+			'default'             => false,
+			'value'               => '',
+			'source_of_value'     => '',
+			'show_in_conf_sample' => false,
+		],
```
Hipska: Would it make sense to enable this by default? Or at least show the option in the config file, as that will make it easier to enable.

core/log.class.inc.php
```diff
@@ -1341,19 +1355,87 @@ public function Process($iUnixTimeLimit)
 			$oLogFileNameBuilder->ResetLastModifiedDateForFile();
 			$oLogFileNameBuilder->CheckAndRotateLogFile();
 		}
+
+		// Purge logs if purge enabled
+		if (MetaModel::GetConfig()->Get(LogAPI::ENUM_CONFIG_PARAM_PURGE_ENABLE)) {
+			$this->PurgeLogs();
```
Hipska, line 1361: Nothing is done with the returned `$aFilesResult`. Intentional?

core/config.class.inc.php
```diff
@@ -113,70 +113,86 @@ class Config
...
+			'source_of_value'     => '',
 			'show_in_conf_sample' => false,
 		],
-		'event_service.debug.filter_events' => [
-			'type' => 'array',
-			'description' => 'List of events name to filter Event Service debug messages',
-			'default' => [],
-			'value' => '',
-			'source_of_value' => '',
+		'log_purge_enable'                   => [
```
```suggestion
		'log_purge_enabled'                   => [
```

Preterit form is used in all params, I would like to keep the consistency.

core/log.class.inc.php
```diff
@@ -639,7 +639,22 @@ abstract class LogAPI
...
-	public const ENUM_CONFIG_PARAM_DB = 'log_level_min.write_in_db';
+	public const ENUM_CONFIG_PARAM_DB   = 'log_level_min.write_in_db';
+
+
+	/**
+	 * Parameter to enable log purge.
+	 *
+	 * @since 3.1.0
+	 */
+	public const ENUM_CONFIG_PARAM_PURGE_ENABLE = 'log_purge_enable';
```
Molkobain:
```suggestion
	public const ENUM_CONFIG_PARAM_PURGE_ENABLED = 'log_purge_enabled';
```

core/log.class.inc.php
```diff
@@ -1341,19 +1355,87 @@ public function Process($iUnixTimeLimit)
 			$oLogFileNameBuilder->ResetLastModifiedDateForFile();
 			$oLogFileNameBuilder->CheckAndRotateLogFile();
 		}
+
+		// Purge logs if purge enabled
+		if (MetaModel::GetConfig()->Get(LogAPI::ENUM_CONFIG_PARAM_PURGE_ENABLE)) {
```
Molkobain:
```suggestion
		if (MetaModel::GetConfig()->Get(LogAPI::ENUM_CONFIG_PARAM_PURGE_ENABLED)) {
```

core/log.class.inc.php
```diff
@@ -1341,19 +1355,87 @@ public function Process($iUnixTimeLimit)
...
+		}
+	}
+
+	/**
+	 * PurgeLogs.
+	 *
+	 * Purge test last modification time of file in log folder and delete
+	 * files that haven't modifications since ENUM_CONFIG_PARAM_PURGE_MAX_KEEP_DAYS
+	 *
+	 * @return array process feedback
```
Molkobain, line 1371:
```suggestion
	 * @return array process feedback
	 * @since 3.1.0
```

core/log.class.inc.php
```diff
@@ -1341,19 +1355,87 @@ public function Process($iUnixTimeLimit)
...
+		if (MetaModel::GetConfig()->Get(LogAPI::ENUM_CONFIG_PARAM_PURGE_ENABLE)) {
+			$this->PurgeLogs();
+		}
+	}
+
+	/**
+	 * PurgeLogs.
+	 *
+	 * Purge test last modification time of file in log folder and delete
+	 * files that haven't modifications since ENUM_CONFIG_PARAM_PURGE_MAX_KEEP_DAYS
```
Molkobain:
```suggestion
	 * files that haven't modifications since {@see \LogAPI::ENUM_CONFIG_PARAM_PURGE_MAX_KEEP_DAYS}
```

core/log.class.inc.php
Molkobain: +1, seems like the errors are not output anywhere.

Molkobain (changes requested):

core/log.class.inc.php
bdalsass: This method can be accessed by the future log center module controller. (I actually work on it)
It will allow to trigger the purge manually and see the purge result client side.
I check with product team to ensure it makes sense.

core/config.class.inc.php
bdalsass: I check with product team.

core/log.class.inc.php
bdalsass: Also, i will output the JSON representation in a info log entry.

piRGoif:
Hello,

Before doing a review I have some questions on the basics : 

* why adding the purge inside the existing LogFileRotationProcess and not in a dedicated iScheduledProcess impl ?
* isn't a single log_purge_max_keep_days config param would be enough ? If value is anything else than a positive integer > 0 then purge will be done, else nothing happens ?

core/config.class.inc.php
bdalsass: The log_purge_enable initial value will be set to false to keep the system iso functional.
The show_in_conf_sample will be set to false to be homogeneous with other logs settings.

bdalsass:
> 



> 

The process is very closed to the file rotating background process, so we have decided to keep the 2 tasks together.
Having 2 parameters allows to disable the purge without loosing the max_keep_days value.

piRGoif:
No time to do a review :(

Molkobain:
> No time to do a review :(

Recent changes (76f667b) have been reviewed by 3 people already

core/config.class.inc.php
Hipska: Meh 👎

Hipska:
1 with comments, 1 with requesting changes and the last one is a pending review. I don't think that counts 🙂

Molkobain:
> > No time to do a review :(
> 
> Recent changes ([76f667b](https://github.com/Combodo/iTop/commit/76f667b1bd093ac988aaf80de194eb55f036f63a)) have been reviewed by 3 people already

> 1 with comments, 1 with requesting changes and the last one is a pending review. I don't think that counts 🙂

I meant reviewed IRL

Hipska:
Why creating public pull requests if they are reviewed discussed internally/privately?

Molkobain:
We already had that conversation [here](https://github.com/Combodo/iTop/pull/383#issuecomment-1369060827)

piRGoif:
> Recent changes have been reviewed by 3 people already

Next time, please make it implicit by adding a comment before merging O:)

> Why creating public pull requests if they are reviewed discussed internally/privately?

Pull request are a very good tool for review, but there aren't any mean to make them private.

Hipska:
@Molkobain Yes, and I still stand on the same point. The discussion ended in silence.

@piRGoif Yes, there are many ways: https://github.com/Combodo/iTop/pull/383#issuecomment-1375286918

piRGoif:
To my knowledge, private forks on public repo can only be created in a security advisory draft ? 
But anyway, that would be very costly either on merging complexity and team productivity. I wouldn't consider this technique as a valid option.

Molkobain:
> @Molkobain Yes, and I still stand on the same point. The discussion ended in silence.

> To my knowledge, private forks on public repo can only be created in a security advisory draft ? But anyway, that would be very costly either on merging complexity and team productivity. I wouldn't consider this technique as a valid option.

Exactly, that's why for me the discussion was closed on that matter for now. Especially as we stated the criteria that allow you to know if a PR is strictly internal or public.

These criteria could be communicated for widely I agree or we could find a way to flag internal PRs but we won't set a secondary repo just for that it's too much overhead regarding the benefits and our workforce.

Among possible way to flag PRs:
  * Dedicated GiHub project for public PRs (always the case, but we could rename it) and dedicated GitHub project for internal PRs.
  * Prefix on PRs' title
  * Add a "Private" label on these PRs

_But I don't thuink this PR is not the place to have this conversation._

Hipska:
So where can we discuss this? (Preferably in public so other community members can join) Or will we end again in silence?

Molkobain:
I don't know, Slack may be the best place as people on SF don't really come here for PRs.

PR #449, opened 2023-02-23
N°6023 Fix cannot load SVG files in AttributeImage since 3.0.0

piRGoif:
A merge error in ddd6bf2 (support/3.0 as target branch) causes the SVG sanitization to always return an empty string.
In consequence trying to load a SVG file in a AttributeImage will result in an empty content loaded everytime.

### DIFF
--- a/core/htmlsanitizer.class.inc.php
+++ b/core/htmlsanitizer.class.inc.php
@@ -34,32 +34,36 @@ abstract class HTMLSanitizer
 
 	/**
 	 * Sanitize an HTML string with the configured sanitizer, falling back to HTMLDOMSanitizer in case of Exception or invalid configuration
+	 *
 	 * @param string $sHTML
+	 * @param string $sConfigKey
+	 *
 	 * @return string
 	 */
-	public static function Sanitize($sHTML)
+	public static function Sanitize($sHTML, $sConfigKey = 'html_sanitizer')
 	{
-		$sSanitizerClass = MetaModel::GetConfig()->Get('html_sanitizer');
-		if(!class_exists($sSanitizerClass))
-		{
+		$sSanitizerClass = utils::GetConfig()->Get($sConfigKey);
+		if (!class_exists($sSanitizerClass)) {
 			IssueLog::Warning('The configured "html_sanitizer" class "'.$sSanitizerClass.'" is not a valid class. Will use HTMLDOMSanitizer as the default sanitizer.');
 			$sSanitizerClass = 'HTMLDOMSanitizer';
-		}
-		else if(!is_subclass_of($sSanitizerClass, 'HTMLSanitizer'))
-		{
-			IssueLog::Warning('The configured "html_sanitizer" class "'.$sSanitizerClass.'" is not a subclass of HTMLSanitizer. Will use HTMLDOMSanitizer as the default sanitizer.');
-			$sSanitizerClass = 'HTMLDOMSanitizer';
+		} else if (!is_subclass_of($sSanitizerClass, 'HTMLSanitizer')) {
+			if ($sConfigKey === 'html_sanitizer') {
+				IssueLog::Warning('The configured "'.$sConfigKey.'" class "'.$sSanitizerClass.'" is not a subclass of HTMLSanitizer. Will use HTMLDOMSanitizer as the default sanitizer.');
+				$sSanitizerClass = 'HTMLDOMSanitizer';
+			}
+			if ($sConfigKey === 'svg_sanitizer') {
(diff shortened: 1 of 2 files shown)


piRGoif:
Discussed with Alexandre : planned for the upcoming 3.0.3

core/htmlsanitizer.class.inc.php
```diff
@@ -34,32 +34,36 @@ abstract public function DoSanitize($sHTML);
 
 	/**
 	 * Sanitize an HTML string with the configured sanitizer, falling back to HTMLDOMSanitizer in case of Exception or invalid configuration
+	 *
 	 * @param string $sHTML
+	 * @param string $sConfigKey
```
Molkobain: Maybe add examples of conf keys?
```suggestion
	 * @param string $sConfigKey eg. 'html_sanitizer', 'svg_sanitizer'
```

Molkobain (approved):

core/htmlsanitizer.class.inc.php
```diff
@@ -34,32 +34,36 @@ abstract public function DoSanitize($sHTML);
...
 	 */
-	public static function Sanitize($sHTML)
+	public static function Sanitize($sHTML, $sConfigKey = 'html_sanitizer')
 	{
-		$sSanitizerClass = MetaModel::GetConfig()->Get('html_sanitizer');
-		if(!class_exists($sSanitizerClass))
-		{
+		$sSanitizerClass = utils::GetConfig()->Get($sConfigKey);
+		if (!class_exists($sSanitizerClass)) {
 			IssueLog::Warning('The configured "html_sanitizer" class "'.$sSanitizerClass.'" is not a valid class. Will use HTMLDOMSanitizer as the default sanitizer.');
```
Hipska, line 47: Should be `"'.$sConfigKey.'"`

core/htmlsanitizer.class.inc.php
```diff
@@ -34,32 +34,36 @@ abstract public function DoSanitize($sHTML);
...
 		}
-		catch(Exception $e)
-		{
-			if($sSanitizerClass != 'HTMLDOMSanitizer')
-			{
+		catch (Exception $e) {
+			if ($sSanitizerClass != 'HTMLDOMSanitizer') {
 				IssueLog::Warning('Failed to sanitize an HTML string with "'.$sSanitizerClass.'". The following exception occured: '.$e->getMessage());
 				IssueLog::Warning('Will try to sanitize with HTMLDOMSanitizer.');
 				// try again with the HTMLDOMSanitizer
```
Hipska, line 69: This second run won't ever throw exceptions?

core/htmlsanitizer.class.inc.php
```diff
@@ -34,32 +34,36 @@ abstract public function DoSanitize($sHTML);
...
+		} else if (!is_subclass_of($sSanitizerClass, 'HTMLSanitizer')) {
+			if ($sConfigKey === 'html_sanitizer') {
+				IssueLog::Warning('The configured "'.$sConfigKey.'" class "'.$sSanitizerClass.'" is not a subclass of HTMLSanitizer. Will use HTMLDOMSanitizer as the default sanitizer.');
+				$sSanitizerClass = 'HTMLDOMSanitizer';
+			}
+			if ($sConfigKey === 'svg_sanitizer') {
+				IssueLog::Error('The configured "'.$sConfigKey.'" class "'.$sSanitizerClass.'" is not a subclass of '.HTMLSanitizer::class.' ! Won\'t sanitize string.');
+
+				return $sHTML;
+			}
```
Hipska, line 58: The `if` statement can be removed, as the method will then also work for any other type of key.

core/htmlsanitizer.class.inc.php
piRGoif: Thanks, fixed in bfe55183

core/htmlsanitizer.class.inc.php
piRGoif: You're right, but as the if was part of the initial commit we'll have to review the modification.
I'll add another pr

core/htmlsanitizer.class.inc.php
piRGoif: You mean, the DoSanitize call L71 ? It could indeed, but I'm not really sure what we should do then. Certainly throw an exception as if HTMLDOMSanitizer is crashing we are at risk ?

core/htmlsanitizer.class.inc.php
piRGoif: Done : #450

PR #450, opened 2023-02-27
N°6173 \HTMLSanitizer::Sanitize : Fix handling only svg_sanitizer

piRGoif:
As suggested by Hipska in https://github.com/Combodo/iTop/pull/449#discussion_r1118397193

### DIFF
--- a/core/htmlsanitizer.class.inc.php
+++ b/core/htmlsanitizer.class.inc.php
@@ -51,8 +51,7 @@ abstract class HTMLSanitizer
 			if ($sConfigKey === 'html_sanitizer') {
 				IssueLog::Warning('The configured "'.$sConfigKey.'" class "'.$sSanitizerClass.'" is not a subclass of HTMLSanitizer. Will use HTMLDOMSanitizer as the default sanitizer.');
 				$sSanitizerClass = 'HTMLDOMSanitizer';
-			}
-			if ($sConfigKey === 'svg_sanitizer') {
+			} else {
 				IssueLog::Error('The configured "'.$sConfigKey.'" class "'.$sSanitizerClass.'" is not a subclass of '.HTMLSanitizer::class.' ! Won\'t sanitize string.');
 
 				return $sHTML;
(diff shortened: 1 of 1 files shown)


core/htmlsanitizer.class.inc.php
```diff
@@ -51,8 +51,7 @@ public static function Sanitize($sHTML, $sConfigKey = 'html_sanitizer')
 			if ($sConfigKey === 'html_sanitizer') {
 				IssueLog::Warning('The configured "'.$sConfigKey.'" class "'.$sSanitizerClass.'" is not a subclass of HTMLSanitizer. Will use HTMLDOMSanitizer as the default sanitizer.');
```
Hipska:
```suggestion
				IssueLog::Warning('The configured "html_sanitizer" class "'.$sSanitizerClass.'" is not a subclass of '.HTMLSanitizer::class.'. Will use HTMLDOMSanitizer as the default sanitizer.');
```

Hipska (approved):

core/htmlsanitizer.class.inc.php
piRGoif: I like the second suggestion, but prefer not to replace $sConfigKey with its literal value.
Pushed in 0c115c28

Hipska: It can't be any other value and it makes searching for errors in the source code easier.

Hipska (approved):

Molkobain (approved):

core/htmlsanitizer.class.inc.php
piRGoif: We tend to limit as much as possible literals in the code. If 'html_sanitizer' is renamed, we would have to do the rename twice here...

Molkobain:

piRGoif:
SVG sanitizer was introduced in 2.6.5 / 2.7.6 / 3.0.0 so ok to introduce fix in support/2.7

PR #453, opened 2023-03-07
N°6105 - Cleanup unnecessary use of dirname(__FILE__)

Hipska:
The `__DIR__` constant is available [since php 5.3](https://3v4l.org/1lRa2) so the boilerplate can be removed.

I thought of not changing it in `approot.inc.php` or the setup files so they can be compatible with older versions of php to show warnings/errors, but it seems the `setup/index.php` is loading `lib/autoload.php` which is using `__DIR__` as well, so there is no reason to keep the boilerplate at all..

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -1753,7 +1753,6 @@ SQL;
 	 * it at all (losing the direct access to the page) :
 	 *
 	 * ```php
-	 * if (!defined('__DIR__')) define('__DIR__', dirname(__FILE__));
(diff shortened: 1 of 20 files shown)


Molkobain (approved):

Molkobain:

Molkobain:

piRGoif:
Warning, this modification generated a crash in TicketsInstaller::AfterDatabaseCreation, which is called with \call_user_func_array : I guess then the `__DIR__` variable equals the one of the file which did the \call_user_func_array call...
There might be other places where this would cause an issue ?

PR #455, opened 2023-03-10
Fix transaction retries on deadlock not working

piRGoif:
In 2.7.0 we added transactions in CRUD operations, and we also tried to add a retry mechanism in case of a deadlock.
This wasn't working : the `$iTransactionRetry` was always initialized to 1, thus the retry did never happen. 
With this fix we are getting the retries count from the existing `db_core_transactions_retry_count` config parameter. Default value is 3, so we will try one time, and if a deadlock occurs we will retry 2 times max.

### DIFF
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -2744,10 +2744,13 @@ abstract class DBObject implements iDisplay
 		if ($bIsTransactionEnabled)
 		{
 			// TODO Deep clone this object before the transaction (to use it in case of rollback)
-			// $iTransactionRetryCount = MetaModel::GetConfig()->Get('db_core_transactions_retry_count');
-			$iTransactionRetryCount = 1;
-			$iTransactionRetryDelay = MetaModel::GetConfig()->Get('db_core_transactions_retry_delay_ms');
+			$oConfig = MetaModel::GetConfig();
+			if (is_null($oConfig)) {
+				throw new ConfigException('Cannot load config');
+			}
+			$iTransactionRetryCount = $oConfig->Get('db_core_transactions_retry_count');
 			$iTransactionRetry = $iTransactionRetryCount;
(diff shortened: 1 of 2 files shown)


eespie (approved):

core/dbobject.class.php
```diff
@@ -2744,10 +2744,13 @@ public function DBInsertNoReload()
 		if ($bIsTransactionEnabled)
 		{
 			// TODO Deep clone this object before the transaction (to use it in case of rollback)
-			// $iTransactionRetryCount = MetaModel::GetConfig()->Get('db_core_transactions_retry_count');
-			$iTransactionRetryCount = 1;
-			$iTransactionRetryDelay = MetaModel::GetConfig()->Get('db_core_transactions_retry_delay_ms');
+			$oConfig = MetaModel::GetConfig();
+			if (is_null($oConfig)) {
+				throw new ConfigException('Cannot load config');
+			}
```
Molkobain, line 2751: Why not use `\utils::GetConfig()` instead? (Same fpr the other places)

piRGoif: I hesitate to made the change :) But I think making the replacement can bring some confusion : we are in a method that needs datamodel to be loaded, so we should better use the config loaded with it ! The utils method will try different fallbacks, and we don't want that.

Molkobain: Good point. Then maybe we should add a comment explicating that? It would avoid someone to change it later by mistake.

Molkobain (approved):

core/dbobject.class.php
piRGoif: Done in 48eaafca

Molkobain:
  * Retry might have been left to 1 on purpose (see above TODO comment)
  * Add a unit test and mock deadlock situation (already used in TransactionTest?) to check if the object is as excepted during the 2nd+ retries

piRGoif:
Closing PR as I won't do anything more.

PR #456, opened 2023-03-10
N°3795 - Replace JS alert native calls with centralized informative modals

steffunky:
In iTop 3.1 we added a factory to create complex modals and also simple informative modals that can replace native js `alert()` calls.

Replacing these calls will help us:

- Writing visuals tests as it's all in page DOM
- Comply with iTop design
- Add multiple levels of severity to alerts (information, warning, error)
- Common modal API with iTop portal

It was already submitted to us with PR #43 but we rejected it at the time as we were planning iTop backoffice evolution and we were not comfortable with adding a new jQuery library dependency 

This PR replace most `alert()` calls in iTop backoffice with `CombodoModal.OpenInformativeModal()`

modal_error

modal_errord


Molkobain:
Love it 🤩

Discussed with Stephen, "InformativeModal" will have focus on the "Ok" button by default so we can keep the same behavior as the native "alert": Dismissing it quickly by hitting "Enter" key.

steffunky:
> will have focus on the "Ok" button by default so we can keep the same behavior as the native "alert": Dismissing it quickly by hitting "Enter" key.

It seems like jQuery dialogs already have this behavior

js/utils.js
```diff
@@ -1206,6 +1206,12 @@ const CombodoInlineImage = {
  * @since 3.1.0
  */
 let CombodoModal = {
+
+	INFORMATIVE_MODAL_SEVERITY_SUCCESS : 'error',
```
Hipska: Success is an error?

js/extkeywidget.js
```diff
@@ -769,7 +769,7 @@ function ExtKeyWidget(id, sTargetClass, sFilter, sTitle, bSelectMode, oWizHelper
 					$('#fstatus_'+me.id).html('');
 					if (data.id == 0) {
 						$('#label_'+me.id).removeClass('ac_dlg_loading');
-						alert(data.error);
+						CombodoModal.OpenInformativeModal(data.error, 'error');
```
Hipska: Why not using the `CombodoModal.INFORMATIVE_MODAL_SEVERITY_ERROR` here?

Hipska (approved):

js/utils.js
steffunky: Good catch, fixed in [48800c2](https://github.com/Combodo/iTop/pull/456/commits/48800c2cc5d2329e79e63b87f0a2b1bb63bd4dc9)

js/extkeywidget.js
steffunky: Thanks, I overlooked it, fixed in [d3414ba](https://github.com/Combodo/iTop/pull/456/commits/d3414badc476a7be1958a47a9c38a6b75b98d495)

Molkobain:
Discussed with Stephen, we might change how to call modals from

```js
CombodoModal.OpenInformativeModal("Foo", CombodoModal.INFORMATIVE_MODAL_SEVERITY_SUCCESS);
CombodoModal.OpenInformativeModal("Foo", CombodoModal.INFORMATIVE_MODAL_SEVERITY_INFORMATION);
CombodoModal.OpenInformativeModal("Foo", CombodoModal.INFORMATIVE_MODAL_SEVERITY_WARNING);
CombodoModal.OpenInformativeModal("Foo", CombodoModal.INFORMATIVE_MODAL_SEVERITY_ERROR);
```

to

```js
CombodoModal.OpenSucessModal("Foo");
CombodoModal.OpenInformativeModal("Foo");
CombodoModal.OpenWarningModal("Foo");
CombodoModal.OpenErrorModal("Foo");
```

or even

```js
CombodoModal.OpenModalForSucess("Foo");
CombodoModal.OpenModalForInformation("Foo");
CombodoModal.OpenModalForWarning("Foo");
CombodoModal.OpenModalForError("Foo");
```

Thoughts?

js/utils.js
```diff
@@ -1417,5 +1423,41 @@ let CombodoModal = {
...
 		CombodoJSConsole.Debug('CombodoModal.OpenInformativeModal not implemented');
-	}
+	},
+
+	/**
+	 * Open a standard informative modal for success messages.
+	 *
+	 * @param sMessage string Informative success message to be displayed in the modal
+	 */
+	OpenSuccessModal: function(sMessage) {
```
Hipska: Maybe keep support for adding optional `oOptions`? Or should `OpenInformativeModal` be used for that?

Molkobain: And prefix `OpenInformativeModal` with `_` to show that it is a protected method and that we should only use the "functional methods"

Molkobain (approved):

js/utils.js
steffunky: I don't think `OpenInformativeModal` should be private, it can be used to add more parameters with `oOptions` and for extensions to make their own "functional methods" if needed

PR #457, opened 2023-03-13
N°6097 - Enable PHP unit tests on a custom DataModel

Molkobain:
_Internal PR_

## Objective
Enable developers to add unit tests on DataModel elements that are not part of the standard DM (iTop Community), for example:
  * New attribute types that are not used yet
  * An AttributeExternalField pointing on a composite attribute (eg. an AttributeImage)

Being able to add these tests would increase confidence on core changes and their consequences on customizations in extensions / Designer.

## What it does
  * Refactor existing test classes for better extensibility
  * Add a new `ItopCustomDatamodelTestCase` class for tests that need an XML delta

For each test class extending `ItopCustomDatamodelTestCase`, a temporary "php-unit-tests" environment is compiled from the current "production" environment and the XML delta. Then the tests are ran in that environment.

These tests are run in a separate process to avoid tempering with the "regular" tests as we would be switching environments between "php-unit-tests" and "production", which is not supported yet because of the MetaModel not being able to reset its cache and structures.

_Update on 2023-07-19:
Discussed a while back with Romain. We will rather apply ALL deltas together in order to do the compilation only once and speed up these tests. This will require each XML delta to be compatible with each others, but for now we find this acceptable as it will force developers to make a small and meaningful delta for their specific use-cases._

## What needs to be done
  * [X] Rebase on support/2.7 branch
  * [X] Change approach so that we apply ALL XML deltas from `ItopCustomDatamodelTestCase` classes to the `production` environment so we do the compilation only once
  * [X] Finish cleanup after tests are ran
    * Conf. file is cleaned up before first compilation to ensure it is not used
    * Datamodel XML files, cache folder, compiled folders are automatically cleaned during compilation. No need to do anything.
    * Database is dropped / re-created after compilation as we need the MetaModel to be ready in order to drop it
  * [X] Try to refactor base PHP test case classes in a dedicated folder to tidy the root folder a bit
    * Base test case classes are now un `src/BaseTestCase`
  * [X] Decide which annotation to keep in these test classes if they run if separate processes they won't see that the test env. is already compiled.
  => None! Processes separation is done in the `ItopCustomDatamodelTestCase` test class. Running these tests in a separate process ensures that "regular" tests are not impacted by the switch of environment in memory as it is not supported yet.
  * [X] Extend capabilities so we can't have test classes compiled in different environments
  => Override `ItopCustomDatamodelTestCase::GetTestEnvironment()`
  * [X] ~Change namespace of test cases to tend to a PSR-4 notation (could be part of another PR as it is not really related)~
  * [X] Fix autoloader in `TestsRunStartHook`
  => We accept to require approot manually as the file should not move
  * [X] Decide if you we should add `$this->debug()` traces during the preparation / usage of the test env. to help troubleshoot failing test.
  => Removed them as they might not be relevant enough, we'll add them when the need comes.
  * [x] Remove temp. test case classes once PR reviewed

## How to use it
  * Extend `ItopCustomDatamodelTestCase`
  * Overload `ItopCustomDatamodelTestCase::GetDatamodelDeltaAbsPath()` method to specify your XML delta
  * Add your test cases

## Example
_XML delta_
```xml
  <classes>
    <class id="Person">
      <fields>
        <field id="tested_attribute" xsi:type="AttributeString" _delta="define">
          <sql>tested_attribute</sql>
          <default_value/>
        </field>
      </fields>
    </class>
  </classes>
</itop_design>
```

_Test class_
```php
namespace Combodo\iTop\Test\UnitTest\Core;

use Combodo\iTop\Test\UnitTest\ItopCustomDatamodelTestCase;
use DBObjectSet;
use DBSearch;
use MetaModel;

/**
 * No annotation required!
 */
class TestGLATest extends ItopCustomDatamodelTestCase
{
	/**
	 * @inheritDoc
	 */
	public function GetDatamodelDeltaAbsPath(): string
	{
		return APPROOT.'tests/php-unit-tests/unitary-tests/core/TestGLATest.delta.xml';
	}

	public function testFoo()
	{
		static::assertFalse(MetaModel::IsValidAttCode('Person', 'non_existing_attribute'));
		static::assertTrue(MetaModel::IsValidAttCode('Person', 'tested_attribute'));
	}
}
```


Molkobain:
Ready for discussion

Molkobain:
Ready for review @piRGoif @rquetiez ! 🙌

tests/php-unit-tests/src/BaseTestCase/ItopTestCase.php
```diff
@@ -65,6 +56,17 @@ protected function tearDown(): void
...
+	/**
+	 * Overload this method to require necessary files through {@see \Combodo\iTop\Test\UnitTest\ItopTestCase::RequireOnceItopFile()} and {@see \Combodo\iTop\Test\UnitTest\ItopTestCase::RequireOnceUnitTestFile()}
+	 *
+	 * @return void
+	 * @since 2.7.9 3.0.4 3.1.0
+	 */
+	protected function LoadRequiredFiles(): void
+	{
+		// Empty until we actually need to require some files in the class
+	}
```
piRGoif: Why this method, instead of the existing setUp overrides ? 
To provide additional value, maybe have instead 2 methods, LoadRequiredItopFiles / LoadRequiredTestFiles, which would return a simple array ?

tests/php-unit-tests/src/BaseTestCase/ItopDataTestCase.php
```diff
@@ -128,6 +120,51 @@ protected function tearDown(): void
...
+	}
+
+	/**
+	 * @return string Environment the test will run in
+	 * @since 2.7.9 3.0.4 3.1.0
+	 */
+	protected function GetTestEnvironment(): string
+	{
+		return self::DEFAULT_TEST_ENVIRONMENT;
+	}
```
piRGoif, line 140: Having a method allows to create overrides in children classes to have the env name computed
But I'm not sure we need such customization for now ? To keep things simple, I would say just use the constant in PrepareEnvironment() ?

application/utils.inc.php
```diff
@@ -1228,13 +1228,23 @@ public static function GetCurrentEnvironment()
...
 	}
 
+	/**
+	 * @return string A path to the folder into which data can be written
+	 * @internal
+	 * @since N°6097 2.7.10 3.0.4 3.1.1
+	 */
+	public static function GetDataPath(): string
+	{
+		return APPROOT.'data/';
```
piRGoif, line 1238: in approot we already have an APPCONF constant, maybe we can add an APPDATA constant and use it here ?

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
```diff
@@ -0,0 +1,198 @@
...
+	{
+		parent::__construct($name, $data, $dataName);
+
+		// Ensure that a test class derived from this one runs in a dedicated process as it changes the MetaModel / environment on the fly and
+		// for now we have no way of switching environments properly in memory and it will result in other (regular) test classes to fail as they won't be on the expected environment.
+		//
+		// If we don't do this, we would have to add the `@runTestsInSeparateProcesses` on *each* test classes which we want to avoid for obvious possible mistakes.
+		// Note that the `@runClassInSeparateProcess` don't work in PHPUnit yet.
+		$this->setRunClassInSeparateProcess(true);
+	}
```
piRGoif, line 52: We absolutely need to report this modification in ItopDataTestCase, so we can remove any annotations in its children classes !
Of course this is subject of another task, I've added N°6632 in our db.

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
```diff
@@ -0,0 +1,198 @@
...
+		return 'php-unit-tests';
+	}
+
+	/**
+	 * @return string Absolute path to the {@see \Combodo\iTop\Test\UnitTest\ItopCustomDatamodelTestCase::GetTestEnvironment()} folder
+	 */
+	final private function GetTestEnvironmentFolderAbsPath(): string
+	{
+		return APPROOT.'env-'.$this->GetTestEnvironment().'/';
+	}
```
piRGoif, line 93: This is much alike \utils::GetCompiledEnvironmentPath
Maybe create \utils::GetEnvironmentPath with an env parameter, and use it here and in \utils::GetCompiledEnvironmentPath ?

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
```diff
@@ -0,0 +1,198 @@
...
+		$this->RequireOnceItopFile('setup/runtimeenv.class.inc.php');
+	}
+
+	/**
+	 * @return string Environment used as a base (conf. file, modules, DB, ...) to prepare the test environment
+	 */
+	protected function GetSourceEnvironment(): string
+	{
+		return 'production';
+	}
```
piRGoif, line 76: Again, simply overriding the DEFAULT_TEST_ENVIRONMENT would be enough

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
```diff
@@ -0,0 +1,198 @@
...
+	/**
+	 * @return bool True if the {@see \Combodo\iTop\Test\UnitTest\ItopDataTestCase::GetTestEnvironment()} is ready (compiled, but not started)
+	 *
+	 * @details Having the environment ready means that it has been compiled for this global tests run, not that it is a relic from a previous global tests run
+	 */
+	final private function IsEnvironmentReady(): bool
+	{
+		// As these test cases run in separate processes, the best way we found to let know a process if its environment was already prepared for **this run** was to compare the modification times of:
+		// - its own env-<ENV> folder
+		// - a file generated at the beginning of the global test run {@see \Combodo\iTop\Test\UnitTest\Hook\TestsRunStartHook}
```
piRGoif, line 116: Pull up this in the method phpdoc block
Inline comments always get lost and have less visibility :/

tests/php-unit-tests/src/Hook/TestsRunStartHook.php
```diff
@@ -0,0 +1,63 @@
...
+require_once __DIR__ . '/../../../../approot.inc.php';
+
+use PHPUnit\Runner\AfterLastTestHook;
+use PHPUnit\Runner\BeforeFirstTestHook;
+use utils;
+
+/**
+ * Class TestsRunStartHook
+ *
+ * IMPORTANT: This will no longer in PHPUnit 10.0 and there is no alternative for now, so we will have to migrate it when the time comes
```
piRGoif:
```suggestion
 * IMPORTANT: This will no longer work in PHPUnit 10.0 and there is no alternative for now, so we will have to migrate it when the time comes
```

piRGoif (dismissed):
Bravo for this great improvement !

tests/php-unit-tests/src/BaseTestCase/ItopTestCase.php
Molkobain: Discussed IRL on tuesday, finally I kept the methods but separated it in 2.
Having the methods and calling the parent is less error prone than merging arrays IMO.

tests/php-unit-tests/src/BaseTestCase/ItopDataTestCase.php
Molkobain: Discussed IRL on tuesday, we keep the method.

application/utils.inc.php
Molkobain: Discussed IRL on tuesday, we keep the method as it is a utility of the utils class. Also, constants in approot.inc.php should only be what is strictly neccessary for the app as it is both its foundation and included in a lot of pages/scripts.

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
Molkobain: Discussed IRL on tuesday, method doesn't exists in support/2.7.

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
Molkobain: Done

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
Molkobain: Discussed IRL on tuesday, method PHPDoc explains the purpose of the method itself and how it should be used / what to expect from it. This block of doc explains how / why we did it internally, it is not for the consumer but the futur maintainers.

tests/php-unit-tests/src/Hook/TestsRunStartHook.php
Molkobain: Done, thanks

PR #458, opened 2023-03-14
N°6098 updateLicenses : check availability of the required JQ command

piRGoif:
This packaging script requires both bash and the JQ command.
If the later isn't available, it will run without throwing an error...

With this change the script will now check directly at launch for the JQ command availability, and exit in error if it isn't.

### DIFF
--- a/.make/license/updateLicenses.php
+++ b/.make/license/updateLicenses.php
@@ -19,17 +19,25 @@
  * The target license file path is in `$xmlFilePath`
  */
 
-$iTopFolder = __DIR__ . "/../../" ;
-$xmlFilePath = $iTopFolder . "setup/licenses/community-licenses.xml";
+$iTopFolder = __DIR__."/../../";
+$xmlFilePath = $iTopFolder."setup/licenses/community-licenses.xml";
+
+$jqExec = shell_exec("jq --help"); // --help param is mandatory otherwise the script will freeze
+if ((null === $jqExec) || (false === $jqExec)) {
+	echo "/!\ JQ is required but cannot be launched :( \n";
+	echo "Check this script PHPDoc block for instructions\n";
+	die - 1;
+}
+
 
 function get_scope($product_node)
 {
 	$scope = $product_node->getAttribute("scope");
 
-	if ($scope === "")
-	{   //put iTop first
+	if ($scope === "") {   //put iTop first
 		return "aaaaaaaaa";
 	}
+
 	return $scope;
 }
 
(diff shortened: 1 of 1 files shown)


.make/license/updateLicenses.php
```diff
@@ -19,17 +19,25 @@
...
-$iTopFolder = __DIR__ . "/../../" ;
-$xmlFilePath = $iTopFolder . "setup/licenses/community-licenses.xml";
+$iTopFolder = __DIR__."/../../";
+$xmlFilePath = $iTopFolder."setup/licenses/community-licenses.xml";
+
+$jqExec = shell_exec("jq --help"); // --help param is mandatory otherwise the script will freeze
+if ((null === $jqExec) || (false === $jqExec)) {
+	echo "/!\ JQ is required but cannot be launched :( \n";
+	echo "Check this script PHPDoc block for instructions\n";
+	die - 1;
```
Molkobain: Like for `exit()`, return code should be in parenthesis. I don't recall the symptom if we don't but we had similar issues with CLI scripts within the SaaS team if I recall correctly. Poke @odain-cbd

Molkobain (changes requested):

.make/license/updateLicenses.php
```diff
@@ -19,17 +19,25 @@
  * The target license file path is in `$xmlFilePath`
  */
 
-$iTopFolder = __DIR__ . "/../../" ;
-$xmlFilePath = $iTopFolder . "setup/licenses/community-licenses.xml";
+$iTopFolder = __DIR__."/../../";
+$xmlFilePath = $iTopFolder."setup/licenses/community-licenses.xml";
+
+$jqExec = shell_exec("jq --help"); // --help param is mandatory otherwise the script will freeze
```
Hipska: Or just `jq -V` or `which jq`?

.make/license/updateLicenses.php
Hipska: I also don't like this notation..

piRGoif:
Technical review : parenthesis on die call, plus check `jq -V`

.make/license/updateLicenses.php
piRGoif: Woops, thanks ! 
Fixed in 6b52da4d

.make/license/updateLicenses.php
piRGoif: I like the `-V` param ! Implemented in 76450e8e
About which, I'm not sure it is widely available on Windows systems ?

Hipska: On windows it would at least be `jq.exe`

piRGoif: Calling a command without the extension works on Windows

Molkobain (approved):

Molkobain:

PR #459, opened 2023-03-20
🐛 Cleanup deprecated global variable

Hipska:
Variable [$php_errormsg](https://www.php.net/manual/en/reserved.variables.phperrormsg.php) has been deprecated since PHP 7.2 and has been removed in PHP 8.0, so I'm converting this code to make debugging easier.

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -1942,10 +1942,10 @@ SQL;
 			$fp = @fopen($sUrl, 'rb', false, $ctx);
 			if (!$fp)
 			{
-				global $php_errormsg;
-				if (isset($php_errormsg))
+				$aError = error_get_last();
+				if (is_array($aError))
 				{
-					throw new Exception("Wrong URL: $sUrl, $php_errormsg");
+					throw new Exception("Wrong URL: $sUrl, ".$aError['message']);
 				}
 				elseif ((strtolower(substr($sUrl, 0, 5)) == 'https') && !extension_loaded('openssl'))
 				{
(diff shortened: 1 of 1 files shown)


Hipska:
The `Exception`s could maybe also be replaced by `CoreException`?

Molkobain (approved):

Molkobain:
Will be discussed today, should be retrofitted to 3.0.x branch as it was removed with PHP 8.0.

Molkobain:
Starting with iTop 3.1 the cURL PHP extension is mandatory to run the setup, so instead of this we would rather replace the whole else condition content with an exception message explaining the the extension is missing.

Why the message if the extension is mandatory? Because the PHP configuration could be altered after iTop's installation. It would help troubleshooting.

Why not just keep the original fix then? Because the fallback in the else condition has several bugs and we no longer support it.

Hipska:
Great 🎉 Even better!

Molkobain:
@Hipska do you feel like updating the PR to reflect that?

Hipska:
Looks like a total different thing to me.

Molkobain:
So we close this PR?

Hipska:
Sure, or just merge it for the meantime so you don't get any errors until the code is actually removed 😉

Note that this deprecated var is also used on other places..

Molkobain:
I'll ask during the functional review if we can retrofit this into the 3.0 as it is supposed to be compatible with PHP 8.0.
But the answer might be to install cURL anyway. Will see.

piRGoif:
Functional review : PR is rejected. We will simply remove the else in the develop branch : N°6172

PR #464, opened 2023-03-22
exec.php : restore ability to deploy modules using symlinks

piRGoif:
Since N°3863 / #221 in iTop 3.1.0 we are allowing exec.php to access pages in module sub directories, thus a path check was added with \utils::RealPath to protect from path traversal attacks.

But this is blocking module deployements from outside iTop using symlinks : exec.php is throwing a "Wrong module, page name or environment..." error. But that deployement method is really convenient when developping...

With this PR, the check isn't done anymore if \utils::IsDevelopmentEnvironment is true.

### DIFF
--- a/pages/exec.php
+++ b/pages/exec.php
@@ -71,20 +71,26 @@ $oKPI->ComputeAndReport("Session Start");
 $sEnvFullPath = APPROOT.'env-'.$sEnvironment;
 $sPageRelativePath = $sModule.'/'.$sPage;
 $sPageEnvFullPath = $sEnvFullPath.'/'.$sPageRelativePath;
-if (is_link($sPageEnvFullPath)) {
-	$oConfig = utils::GetConfig();
-	$sSourceDir = $oConfig->Get('source_dir'); // generated at compile time, works for legacy build with datamodels/1.x
-	// in case module was compiled to symlink, we need to check against real linked path as symlink is resolved
-	$aPossibleBasePaths = [
-		APPROOT.$sSourceDir,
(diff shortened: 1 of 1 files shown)


Molkobain:
Why don't you just use "deployment" in your IDE? I don't get why we should introduce "complexity" in the code to comply to your own setup.

piRGoif:
We'll have a chat IRL on next review, this would be clearer, but to clear it up for everyone else :

I used PHPStorm deployment few years ago and turned to symlink as it was easier and more reliable. But I'm using deployment again since the original PR was merged.

To me initial configuration is long and a lot error prone (especially about the configuration needed so that debug keeps working), adding a directory is long, and I had some problems with file not synced so I'm never sure and in consequence when something isn't working as expected it adds a step to check.

About initial configuration : 
- don't ever forget to add path mappings
- don't forget to add exclusions depending on your configuration

Adding a directory : 
- 3 fields to fill in the deployment configuration dialog (local path, deployment path, web path)
- cannot simply press tab to go from one field to another
- add the path mapping in PHP / Servers
- don't forget to do a first sync manually after adding the deployment

Steps to deploy with symlinks : 
- fire mklink.exe (or my mklink.php), having autocompletion in the CLI
- no other step, no sync prb, no debug prb

piRGoif:
Technical review : meeting with Guillaume to review problems with PHPStorm deployment

piRGoif:
Closing PR, the team is using PHPStorm deployment feature without needing this improvement.

PR #474, opened 2023-04-04
N°6125 - Issue with GetAttributeFlags and GetInitialStateAttributeFlags within iTop 3.0.2

accognet:
In 2.7, the parameter _status_ in WizardHelper was empty when the object have no life cycle. This is not the case in 3.0.
Consequence : in 3.0, when the fields are refreshed thanks to an ajax call, objet remains in old status even if we have changed it in the page.
This PR fix this issue

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -3058,9 +3058,9 @@ EOF
 
 		$iFieldsCount = count($aFieldsMap);
 		$sJsonFieldsMap = json_encode($aFieldsMap);
-		$sState = '';
+		$sLifecycleStateForWizardHelper = '';
 		if (MetaModel::HasLifecycle($sClass)) {
-			$sState = $this->GetState();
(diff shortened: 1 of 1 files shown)


piRGoif:
Technical review : OK
Just rename `$sState` var to document it is init for WizardHelper, something like `$sLifecycleStateForWizardHelper`

piRGoif:
Discussed with the product team, no need for a functional review.
@accognet just rename the var and then you can merge

Molkobain:
> Discussed with the product team, no need for a functional review. @accognet just rename the var and then you can merge

This is based on support/3.0 not develop, whereas the ticket is planned on 3.1 (develop).
I'm in favor for this to be retrofitted in support/3.0, but please make sure that product team is aware of this *before* merging, and indicate it in the ticket so we don't caught anybody by surprise.

piRGoif:
Good catch Guillaume, I've asked them.

piRGoif:
Product team is ok to merge as is, in the support/3.0 branch, and then up to develop

PR #480, opened 2023-04-19
N°1608 - Organization attachments not visible for some users

eespie:
Change attachments access rights to use the host objects access rights if organization check failed

### DIFF
--- a/core/ormdocument.class.inc.php
+++ b/core/ormdocument.class.inc.php
@@ -256,7 +256,19 @@ class ormDocument
 			$oObj = MetaModel::GetObject($sClass, $id, false, false);
 			if (!is_object($oObj))
 			{
-				throw new Exception("Invalid id ($id) for class '$sClass' - the object does not exist or you are not allowed to view it");
+				// If access to the document is not granted, check if the access to the host object is allowed
+				$oObj = MetaModel::GetObject($sClass, $id, false, true);
+				if ($oObj instanceof Attachment) {
+					$sItemClass = $oObj->Get('item_class');
+					$sItemId = $oObj->Get('item_id');
+					$oHost = MetaModel::GetObject($sItemClass, $sItemId, false, false);
+					if (!is_object($oHost)) {
+						$oObj = null;
+					}
+				}
+				if (!is_object($oObj)) {
+					throw new Exception("Invalid id ($id) for class '$sClass' - the object does not exist or you are not allowed to view it");
+				}
 			}
 			if (($sSecretField != null) && ($oObj->Get($sSecretField) != $sSecretValue))
 			{
@@ -286,6 +298,7 @@ class ormDocument
 					$oObj->Set($sAttCode, $oDocument);
 					// $oObj can be a \DBObject or \cmdbAbstractObject so we ahve to protect it
 					if (method_exists($oObj, 'AllowWrite')) {
+						// AllowWrite method is implemented in cmdbAbstractObject, but $oObject could be a DBObject or CMDBObject
 						$oObj->AllowWrite();
 					}
 					$oObj->DBUpdate();
--- a/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
+++ b/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
@@ -186,36 +186,25 @@
           <type>Overload-ExNihilo</type>
           <code><![CDATA[	public function SetDefaultOrgId()
 	{
-		// First check that the organization CAN be fetched from the target class
-		//
-		$sClass = $this->Get('item_class');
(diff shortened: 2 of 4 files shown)


core/ormdocument.class.inc.php
```diff
@@ -284,7 +296,9 @@ public static function DownloadDocument(WebPage $oPage, $sClass, $id, $sAttCode,
 				if($sContentDisposition === static::ENUM_CONTENT_DISPOSITION_ATTACHMENT) {
 					$oDocument->IncreaseDownloadsCount();
 					$oObj->Set($sAttCode, $oDocument);
-					$oObj->AllowWrite();
+					if (method_exists($oObj, 'AllowWrite')) {
+						$oObj->AllowWrite();
+					}
```
Molkobain: Why is this necessary? Will discuss it with Eric tomorrow.

Molkobain:
@jbostoen this is a follow-up of your previous PR. You might want to have a look :)

core/ormdocument.class.inc.php
Molkobain: Discussed with Eric, sometimes $oObj si a \DBObject which doesn't have the method.

Molkobain (approved):

datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
```diff
@@ -186,36 +186,25 @@
...
-							$oCurrentPerson = MetaModel::GetObject('Person', UserRights::GetContactId(), false);
-							if ($oCurrentPerson)
-							{
-						 		$this->Set('item_org_id', $oCurrentPerson->Get($sAttCode));
-						 	}
-						}
-					}
-				}
-			}
+    // Second: check that the organization CAN be fetched from the current user
```
Hipska: There is no First anymore..

Hipska (commented):
This PR sounds to have similarities with #303

Is it a regression from that one?

datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
eespie: > This PR sounds to have similarities with https://github.com/Combodo/iTop/pull/303
> Is it a regression from that one?

Not at all :) it is the console part to allow non-admin to manage attachments using the access rights of the host object when the organization check fails.

core/ormdocument.class.inc.php
piRGoif:
```suggestion
					if (method_exists($oObj, 'AllowWrite')) {
						// AllowWrite method is implemented in cmdbAbstractObject, but $oObject could be a DBObject or CMDBObject
						$oObj->AllowWrite();
					}
```

jbostoen:
@eespie @molkobain This is IMHO leading to a regression.

Before this PR: the item_org_id could be 0, because it was derived from the host object and that's what was stored in the DB. There was an explicit check to only set it if the host object's class had an org_id.

In the new implementation, this is ignored. So in my use case for which my PR was accepted: if I uploaded an attachment linked to a class with no org_id, the org_id was 0, as I would expect it to be; as there's no org_id either on the host object.

After this PR, it's now the organization of the user who uploaded it.

piRGoif:
Hi Jeffrey, 
After a chat with the product team, I can confirm this is an intended behavior change, asked by multiple clients to avoid having attachments accessible to any user.
This should have been added to the documentation though, and we might have missed that. I'll have a look !

jbostoen:
So what would be the workaround now for existing users who want to attach attachments to host objects without org?

As you'd currently run into an error when using the portal?

piRGoif:
Can you create a topic on the forums so we can have a proper discussion there ?

Hipska:
What error would you get in the portal?

Also, @piRGoif maybe this behaviour should be configurable (old vs new) in the config?

jbostoen:
Done => https://sourceforge.net/p/itop/discussion/general/thread/cf8687cf7c/


@Hipska  before: users would see that an attachment is available in the portal, for an object with no org ID (think FAQs, downloads, ...).
After my (accepted) PR: the rights of the host object are respected. If no org ID, it's available to anyone who has access to the class.

Now, a few commits later, the org_id ALWAYS gets set on the attachment. If the host object has no org ID, the attachment takes the org ID from the user who uploads the attachment. So this differs from the previous behavior, where - also before my PR if I remember correctly - the org ID of the attachment was always 0 in this case.

Hipska:
So it would introduce access problems when the Attachment was uploaded/created by a user with a different org compared to the user in the portal? That looks like a viable problem indeed.

@piRGoif Could you inform product team of this consequence? It could be that they were not aware of this use case.

jbostoen:
@eespie Can you check please if this changed behavior is intended?

PR #482, opened 2023-04-19
N°6208 - Router: Add mechanism to generate complete route URL from its code

Molkobain:
_Internal PR_

## Objective

Following the introduction of the "Router" to automatically redirect a unique endpoint (UI.php) to any controller/operation; 
 this PR enhance the Router service to provide a way to generate URL from its code instead of manually building it.
 
This is very similar to Symfony UrlGenerator service but more basic for now.

_How to build an URL before_

`$sURL = utils::GetAbsoluteUrlAppRoot() . '/pages/UI.php?route=object.modify&class=UserRequest&id=123'`

_How to build an URL now_

```
$oRouter = Router::GetInstance();
$sURL = $oRouter->GenerateUrl('object.modify', ['class' => 'UserRequest', 'id' => 123]);
```

Both will output `https://itop.domain.com/pages/UI.php?route=object.modify&class=UserRequest&id=123`

## Advantages

  * Easier to build URLs without calling the app root
  * Easier to build URLs without url-encoding query string parameters (even for arrays and objects)
  * Easier maintenance over time as the URL form can change but not the consumer

## Limitations

  * For now it's not possible to have placeholders within the URL like the Symfony service, but we can add it in the future without any impact on the consumers which is the most critical point.
  * For now it is not possible to have default values on route params like in the Symfony service, bt we can add it in the future without any impact on consumers as well.


bdalsass (approved):
great !!!

steffunky:
Looks great! :+1:

steffunky (approved):

Hipska:
Is `?route=object.modify` something new? I can't recall it being used before. And does the old method (`?operation=modify`) still work?

piRGoif:
> Is `?route=object.modify` something new? 

Yes, see the UI.php in the develop branch : https://github.com/Combodo/iTop/blob/develop/pages/UI.php#L319

Molkobain:
> Is `?route=object.modify` something new? I can't recall it being used before. And does the old method (`?operation=modify`) still work?

It has been introduced with [this commit](https://github.com/Combodo/iTop/commit/a553616ea2b7049b0639eba3b8fcd9de3e399d30).
  * Old `operation` param is still supported
  * Now the new `route` param can be used on either UI.php or ajax.render.php to call business logic from any controller, may it be in the core, in an extension or else.

This is to end increasing the size of the `switch` statements in these pages and ease maintenability.
Also this avoids developer to add their own endpoints in an extension, their now just have to add the controller.

PR #483, opened 2023-04-21
On link modal creation, make ext key to current object readonly, but still submit its value

steffunky:
_Internal_

In order to avoid linking an object to an unwanted object from links tab, make the external key to current object prefilled and readonly, yet still submit the value in the form.

Example with a new WorkOrder that'll  be linked to a UserRequest:

image


sources/Application/Helper/FormHelper.php
```diff
@@ -0,0 +1,109 @@
...
+	/**
+	 * Get attribute flag for an object allowing to cross-check with extra flags present in a form
+	 * 
+	 * @param $oObject
+	 * @param $sAttCode
+	 * @param array $aExtraFlags
+	 *
+	 * @return int|mixed
+	 */
+	public static function GetAttributeFlagsForObject($oObject, $sAttCode, array $aExtraFlags = []) {
```
Molkobain:
```suggestion
	 * @param \DBObject $oObject
	 * @param string $sAttCode
	 * @param array $aExtraFlags
	 *
	 * @return int
	 */
	public static function GetAttributeFlagsForObject(DBObject $oObject, string $sAttCode, array $aExtraFlags = []): int {
```

Stricter code to avoid unpredicted usages

sources/Application/Helper/FormHelper.php
```diff
@@ -0,0 +1,109 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Get attribute flag for an object allowing to cross-check with extra flags present in a form
+	 * 
+	 * @param $oObject
+	 * @param $sAttCode
+	 * @param array $aExtraFlags
```
Hipska:
```suggestion
	 * @param CMDBObject $oObject
	 * @param string $sAttCode
	 * @param array $aExtraFlags
```

sources/Application/Helper/FormHelper.php
```diff
@@ -0,0 +1,109 @@
...
+	 */
+	public static function GetAttributeFlagsForObject($oObject, $sAttCode, array $aExtraFlags = []) {
+		$iFlags = $oObject->GetFormAttributeFlags($sAttCode);
+		if (array_key_exists($sAttCode, $aExtraFlags)) {
+			// the caller may override some flags if needed
+			$iFlags |= $aExtraFlags[$sAttCode];
+		}
+		return $iFlags;
+	}
+}
```
Hipska, line 110: No newline at end of file

sources/Application/Helper/FormHelper.php
```diff
@@ -0,0 +1,109 @@
+<?php
+/*
+ * @copyright   Copyright (C) 2010-2023 Combodo SARL
+ * @license     http://opensource.org/licenses/AGPL-3.0
+ */
+
+namespace Combodo\iTop\Application\Helper;
+
+use AttributeBlob;
+use Dict;
```
Molkobain:
```suggestion
use AttributeBlob;
use DBObject;
use Dict;
```

sources/Application/Helper/FormHelper.php
```diff
@@ -0,0 +1,109 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Update flags to be sent to form with url parameters
+	 * For now only supports "readonly" param
+	 * 
+	 * @param $oObject
+	 * @param array $aExtraParams
```
Hipska:
```suggestion
	 * @param CMDBObject $oObject
	 * @param array $aExtraParams
```

sources/Application/Helper/FormHelper.php
```diff
@@ -0,0 +1,109 @@
...
+	 *
+	 * @return void
+	 * @throws \Exception
+	 */
+	public static function UpdateFlagsFromContext($oObject, array &$aExtraParams): void
```
Molkobain:
```suggestion
	 * @param \DBObject $oObject
	 * @param array $aExtraParams
	 *
	 * @return void
	 * @throws \Exception
	 */
	public static function UpdateFlagsFromContext(DBObject $oObject, array &$aExtraParams): void
```

Hipska (changes requested):
Some minor doc changes.

Also, I'm not sure if the composer update should be part of this PR. Shouldn't that be a separate one?

Molkobain (changes requested):
Sounds good

sources/Application/Helper/FormHelper.php
Hipska: Even better than my suggestion 😉

sources/Application/Helper/FormHelper.php
steffunky: I'm not sure what u mean :grin: 
Additionally, I need autoloader to be updated so continuous integration correctly works, I'll rebase on develop anyway before merging to be sure it's up to date

Molkobain (approved):

sources/Application/Helper/FormHelper.php
Hipska: See the last line of the file in GitHub: https://github.com/Combodo/iTop/pull/483/files/1cd72937f9ed04cac0bac5c6d8e2ed1684e40fbc#diff-13110662c0473b630688b00147b699a5e6500c3cd21b9fbaa335713a2c9b7c12 It as an ugly red icon, meaning you need to add a return/enter after the last `}`.

PR #484, opened 2023-04-25
N°6240 - Improve display of picture in read or edit mode

accognet:
Internal

### DIFF
--- a/css/backoffice/components/_field.scss
+++ b/css/backoffice/components/_field.scss
@@ -64,9 +64,16 @@ $ibo-field--enable-bulk--checkbox--margin-left: $ibo-spacing-300 !default;
         word-break: break-word;
         white-space: inherit;
       }
+
       & pre {
         white-space: break-spaces;
       }
+    }
+  }
+
+  &[data-attribute-type="AttributeImage"] {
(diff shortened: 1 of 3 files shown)


css/backoffice/components/_field.scss
```diff
@@ -64,11 +64,22 @@ $ibo-field--enable-bulk--checkbox--margin-left: $ibo-spacing-300 !default;
...
         white-space: inherit;
       }
+
       & pre {
         white-space: break-spaces;
       }
     }
   }
+
+  &[data-attribute-type="AttributeImage"] {
```
Molkobain, line 74: Overall fix seems good, I just need to check if the SCSS part it placed corrected.

Molkobain:

_Guillaume will check if the SCSS is in the right place_

css/backoffice/components/_field.scss
Molkobain: Fix is placed correctly. 👍
Just add a comment to explain why this is necessary (not just that it can overflow, explain why we use grid instead of another rule), see how it's done on rules above.

piRGoif:
Functional review : already validated during bug review, ready to merge !

PR #492, opened 2023-05-10
N°1386 - Advanced Search: Navigation in list - Browse this list

accognet:
The back button only returns once.  
The CSS for the navigation is to be validated :
image


Molkobain:
Discussed IRL with Anne-Catherine, I feel like we should discuss the specs with Vincent (at least) first as the expectations might have change since the initial ticket. And as this is a bigger PR than usual, it would be a shame to code it almost completely if we have to redo everything after functional review.

Molkobain:
Support PRs review: This is too big for a PR, it should be handled properly in a dedicated project. Also, current approach uses hacks more than a globally designed approach.

Closing this PR for now, it can be re-opened if this ticket is designed properly.

PR #494, opened 2023-05-12
 N°6047 - Make iTop scripts work by HTTP via token with specific scopes

odain-cbd:
Import/export scripts are run under new specific ContextTag (IMPORT/EXPORT). 

This is an easy way to let our dedicated extension know whether authentication is authorized for specific CLI calls.

### DIFF
--- a/core/contexttag.class.inc.php
+++ b/core/contexttag.class.inc.php
@@ -3,7 +3,7 @@
 //
 //   This file is part of iTop.
 //
-//   iTop is free software; you can redistribute it and/or modify	
+//   iTop is free software; you can redistribute it and/or modify
 //   it under the terms of the GNU Affero General Public License as published by
 //   the Free Software Foundation, either version 3 of the License, or
 //   (at your option) any later version.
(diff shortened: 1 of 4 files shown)


core/contexttag.class.inc.php
```diff
@@ -58,6 +58,13 @@ class ContextTag
 	public const TAG_SETUP   = 'Setup';
 	public const TAG_SYNCHRO = 'Synchro';
 	public const TAG_REST    = 'REST/JSON';
+
+	/**
+	 * @since 3.1.0 N°6047
+	 */
+	public const TAG_IMPORT    = 'Import';
+	public const TAG_EXPORT    = 'Export';
```
Molkobain:
```suggestion
	/**
	 * @since 3.1.0 N°6047
	 */
	public const TAG_IMPORT    = 'Import';
		/**
	 * @since 3.1.0 N°6047
	 */
	public const TAG_EXPORT    = 'Export';
```

Molkobain (approved):

PR #501, opened 2023-06-04
N°7082 Possibility to not only force synchronous sending of e-mails, but also asynchronous sending

jbostoen:
Possibility to not only force synchronous sending of e-mails, but also asynchronous sending.

Currently, the sending of an e-mail in iTop respects whatever is configured (synchronous/asynchronous) in the configuration parameter. iTop already supports an option to forcefully send synchronous e-mails instead, which is mostly useful for sending immediate test e-mails.

Coming to the use case for this PR: extensions such as customer survey, the opposite would be needed. Customer survey sends a lot of individual e-mails, which may trigger a limit with the e-mail provider which starts blocking the e-mails, returning in many failed send attempts. As an iTop admin, I want to send synchronous by default; and asynchronous only for such a specific use case.

This draft is here to open discussion whether the iTop core should already have something on this (customer survey extension still needs a minor update then), or whether the customer survey extension should briefly override the configuration instead.

### DIFF
--- a/core/email.class.inc.php
+++ b/core/email.class.inc.php
@@ -46,6 +46,15 @@ class EMail implements iEMail
 	const ORIGINAL_FORMAT = 1; // Original format, consisting in serializing the whole object, inculding the Swift Mailer's object.
 	// Did not work with attachements since their binary representation cannot be stored as a valid UTF-8 string
 	const FORMAT_V2 = 2; // New format, only the raw data are serialized (base64 encoded if needed)
+	
+	/** @var int ENUM_SEND_DEFAULT This option can be used when sending an e-mail to respect the default configuration parameter. */
+	const ENUM_SEND_DEFAULT = 0;
+	
+	/** @var int ENUM_SEND_FORCE_SYNCHRONOUS This option can be used when sending an e-mail to ignore the default and force synchronous sending instead. Example of a use case: instant e-mail test. */
+	const ENUM_SEND_FORCE_SYNCHRONOUS = 1;
+	
+	/** @var int ENUM_SEND_FORCE_ASYNCHRONOUS This option can be used when sending an e-mail to ignore the default and force synchronous sending instead. Example of a use case: Bulk mails. */
+	const ENUM_SEND_FORCE_ASYNCHRONOUS = 2;
 
 	public function __construct()
 	{
--- a/sources/Core/Email/EmailLaminas.php
+++ b/sources/Core/Email/EmailLaminas.php
@@ -293,22 +293,53 @@ class EMailLaminas extends Email
 		return $aImagesParts;
 	}
 
-	public function Send(&$aIssues, $bForceSynchronous = false, $oLog = null)
+	/**
+	 * Sends an e-mail.
+	 *
+	 * @param string[] $aIssues Array to add any potentially encountered issues to.
+	 * @param int|bool $iSyncAsync Specify whether the e-mail will be sent per default configuration, or whether there will be a forced (a)synchronous sending. One of Email::ENUM_SEND_* constants. To support legacy, it also allows a boolean (true = send synchronous). 
+	 * @param Object|null $oLog Log
+	 *
+	 * @details By default, the send method will respect the preference to send e-mails in an (a)synchronous way as defined in the iTop configuration by the administrator.
+	 * In some use cases, it may be necessary to override this behavior. For example, for some tests it may be best if e-mails are always sent instantly (synchronous).
+	 * Asynchronous may be preferred when sending a lot of bulk e-mails at once, to avoid hitting rate limits of e-mail providers (e.g. customer survey extension).
+	 *
+	 * @since 3.2.0 Previously, $iSyncAsync was a boolean ($bForceSynchronous) and this method only allowed to forcefully send e-mails synchronously even when the default was asynchronous.
+	 *
+	 * @return
+	 */
+	public function Send(&$aIssues, $iSyncAsync = Email::ENUM_SEND_DEFAULT, $oLog = null)
 	{
 		//select a default sender if none is provided.
 		if (empty($this->m_aData['from']['address']) && !empty($this->m_aData['to'])) {
 			$this->SetRecipientFrom($this->m_aData['to']);
 		}
 
-		if ($bForceSynchronous) {
+		// In previous iTop versions, $iSyncAsync was $bForceSynchronous. To retain backward compatibility, this check is in place.
+		if($iSyncAsync === true) {
+			// This legacy mode forces synchronous sending, ignoring whatever default was configured.
 			return $this->SendSynchronous($aIssues, $oLog);
 		} else {
-			$oConfig = $this->LoadConfig();
-			$bConfigASYNC = $oConfig->Get('email_asynchronous');
-			if ($bConfigASYNC) {
-				return $this->SendAsynchronous($aIssues, $oLog);
-			} else {
-				return $this->SendSynchronous($aIssues, $oLog);
+			
+			switch($iSyncAsync) {
+				
+				case Email::ENUM_SEND_FORCE_SYNCHRONOUS:
+					return $this->SendSynchronous($aIssues, $oLog);
+				
+				case Email::ENUM_SEND_FORCE_ASYNCHRONOUS:
+					return $this->SendAsynchronous($aIssues, $oLog);
+					
+				case Email::ENUM_SEND_DEFAULT:
+				default:
+				
+					// Default behavior.
+					$oConfig = $this->LoadConfig();
+					$bConfigASYNC = $oConfig->Get('email_asynchronous');
+					if($bConfigASYNC) {
+						return $this->SendAsynchronous($aIssues, $oLog);
+					} else {
+						return $this->SendSynchronous($aIssues, $oLog);
+					}
 			}
 		}
 	}


sources/Core/Email/EmailLaminas.php
```diff
@@ -293,24 +293,59 @@ protected function EmbedInlineImages(string &$sBody)
...
 	{
 		//select a default sender if none is provided.
 		if (empty($this->m_aData['from']['address']) && !empty($this->m_aData['to'])) {
 			$this->SetRecipientFrom($this->m_aData['to']);
 		}
 
-		if ($bForceSynchronous) {
+		// In previous iTop versions, $iSyncAsync was $bForceSynchronous. To retain backward compatibility, this check is in place.
+		// @todo Decide: log a deprecation warning or something? Or turn $iSyncAsync officially into a mixed value?
+		if(is_bool($iSyncAsync) == true && $iSyncAsync == true) {
```
Hipska:
```suggestion
		if(is_bool($iSyncAsync) && $iSyncAsync == true) {
```
Or maybe just like this (less explicit):
```suggestion
		if($iSyncAsync === true) {
```

sources/Core/Email/EmailLaminas.php
```diff
@@ -293,24 +293,59 @@ protected function EmbedInlineImages(string &$sBody)
...
+					} 
+					else {
+						return $this->SendSynchronous($aIssues, $oLog);
+					}
+				
 			}
+			
 		}
+			
 	}
```
Hipska: Those empty lines can be left out?

sources/Core/Email/EmailLaminas.php
```diff
@@ -293,24 +293,59 @@ protected function EmbedInlineImages(string &$sBody)
...
+				case Email::ENUM_SEND_DEFAULT:
+				default:
+				
+					// Default behavior.
+					$oConfig = $this->LoadConfig();
+					$bConfigASYNC = $oConfig->Get('email_asynchronous');
+					if($bConfigASYNC) {
+						return $this->SendAsynchronous($aIssues, $oLog);
+					} 
+					else {
```
Hipska:
```suggestion
					} else {
```

sources/Core/Email/EmailLaminas.php
```diff
@@ -293,24 +293,59 @@ protected function EmbedInlineImages(string &$sBody)
...
 			return $this->SendSynchronous($aIssues, $oLog);
-		} else {
-			$oConfig = $this->LoadConfig();
-			$bConfigASYNC = $oConfig->Get('email_asynchronous');
-			if ($bConfigASYNC) {
-				return $this->SendAsynchronous($aIssues, $oLog);
-			} else {
-				return $this->SendSynchronous($aIssues, $oLog);
+		}
+		else {
```
Hipska:
```suggestion
		} else {
```

Hipska (commented):
Code logic looks good, some minor remarks

jbostoen:
@piRGoif @Molkobain Can this idea be discussed?

piRGoif:
Corresponding SF ticket : https://sourceforge.net/p/itop/tickets/2157/
Will be discussed on next reviews.

Molkobain:
Technical review: Sounds like a good idea, will check during functional review if ok for product, them will dig more into how we would like it to be implemented.

sources/Core/Email/EmailLaminas.php
Molkobain: I would be in favor of the second proposal :)

jbostoen:
Alright, let me know when to proceed :)

Molkobain:
Functional review:
- Ok on the principle for this PR (Core part. Customer survey part would be part of another PR later)
- Deeper discussion on how we would like it to be done will take place during next technical release
- Target is iTop 3.2, so develop branch is ok
- For future Customer Survey PR: 
  - Sync/Async param. should not be a module param but an enum attribute (`"sync" / "async"`) of the `Survey` object itself. That way it will be easier to see how it will behave in a glance and to manage what is what (same behavior is planned on the `ActionWebhook`)
  - Add a check to write on `Survey` to warn user if it is configured as async and CRON is not running
  - Later, Combodo will try to introduce a standard API to do this on classes

jbostoen:
Alright, adjusted everything according to our discussion above.

Would be nice to still have it in iTop 3.1 (even if it's in a maintenance release, as iTop 3.1 will be LTS I think?). 😁

Molkobain:
> Would be nice to still have it in iTop 3.1 (even if it's in a maintenance release, as iTop 3.1 will be LTS I think?). 😁

LTS won't be iTop 3.1, it is more likely to be a 3.2.x but this is no sure yet.

core/email.class.inc.php
```diff
@@ -46,6 +46,15 @@ class EMail implements iEMail
 	const ORIGINAL_FORMAT = 1; // Original format, consisting in serializing the whole object, inculding the Swift Mailer's object.
 	// Did not work with attachements since their binary representation cannot be stored as a valid UTF-8 string
 	const FORMAT_V2 = 2; // New format, only the raw data are serialized (base64 encoded if needed)
+	
+	/** @var Integer ENUM_SEND_DEFAULT This option can be used when sending an e-mail to respect the default configuration parameter. */
```
Molkobain:
```suggestion
	/** @var int ENUM_SEND_DEFAULT This option can be used when sending an e-mail to respect the default configuration parameter. */
```

core/email.class.inc.php
```diff
@@ -46,6 +46,15 @@ class EMail implements iEMail
...
+	const ENUM_SEND_DEFAULT = 0;
+	
+	/** @var Integer ENUM_SEND_FORCE_SYNCHRONOUS This option can be used when sending an e-mail to ignore the default and force synchronous sending instead. Example of a use case: instant e-mail test. */
```
Molkobain:
```suggestion
	/** @var int ENUM_SEND_FORCE_SYNCHRONOUS This option can be used when sending an e-mail to ignore the default and force synchronous sending instead. Example of a use case: instant e-mail test. */
```

core/email.class.inc.php
```diff
@@ -46,6 +46,15 @@ class EMail implements iEMail
...
+	const ENUM_SEND_FORCE_SYNCHRONOUS = 1;
+	
+	/** @var Integer ENUM_SEND_FORCE_ASYNCHRONOUS This option can be used when sending an e-mail to ignore the default and force synchronous sending instead. Example of a use case: Bulk mails. */
```
Molkobain:
```suggestion
	/** @var int ENUM_SEND_FORCE_ASYNCHRONOUS This option can be used when sending an e-mail to ignore the default and force synchronous sending instead. Example of a use case: Bulk mails. */
```

sources/Core/Email/EmailLaminas.php
```diff
@@ -293,22 +293,51 @@ protected function EmbedInlineImages(string &$sBody)
 		return $aImagesParts;
 	}
 
-	public function Send(&$aIssues, $bForceSynchronous = false, $oLog = null)
+	/**
+	 * Sends an e-mail.
+	 *
+	 * @param String[] $aIssues Array to add any potentially encountered issues to.
+	 * @param Integer|Boolean $iSyncAsync Specify whether the e-mail will be sent per default configuration, or whether there will be a forced (a)synchronous sending. One of Email::ENUM_SEND_* constants. To support legacy, it also allows a boolean (true = send synchronous).
```
Molkobain:
```suggestion
	 * @param string[] $aIssues Array to add any potentially encountered issues to.
	 * @param int|bool $iSyncAsync Specify whether the e-mail will be sent per default configuration, or whether there will be a forced (a)synchronous sending. One of Email::ENUM_SEND_* constants. To support legacy, it also allows a boolean (true = send synchronous). 
```

sources/Core/Email/EmailLaminas.php
```diff
@@ -293,22 +293,51 @@ protected function EmbedInlineImages(string &$sBody)
...
 
-	public function Send(&$aIssues, $bForceSynchronous = false, $oLog = null)
+	/**
+	 * Sends an e-mail.
+	 *
+	 * @param String[] $aIssues Array to add any potentially encountered issues to.
+	 * @param Integer|Boolean $iSyncAsync Specify whether the e-mail will be sent per default configuration, or whether there will be a forced (a)synchronous sending. One of Email::ENUM_SEND_* constants. To support legacy, it also allows a boolean (true = send synchronous). 
+	 * @param Object|null $oLog Log
+	 *
+	 * @details By default, the send method will respect the preference to send e-mails in an (a)synchronous way as defined in the iTop configuration by the administrator.
```
Molkobain, line 303: Add `@since 3.2.0` to explain that $bForceSynchronous changed to $iSyncAsync

Molkobain (changes requested):

Molkobain:
Final functional review next week.

jbostoen:
Updated per discussion above and squashed into one commit.

sources/Core/Email/EmailLaminas.php
```diff
@@ -293,22 +293,53 @@ protected function EmbedInlineImages(string &$sBody)
...
 		//select a default sender if none is provided.
 		if (empty($this->m_aData['from']['address']) && !empty($this->m_aData['to'])) {
 			$this->SetRecipientFrom($this->m_aData['to']);
 		}
 
-		if ($bForceSynchronous) {
+		// In previous iTop versions, $iSyncAsync was $bForceSynchronous. To retain backward compatibility, this check is in place.
+		if($iSyncAsync === true) {
+			// This legacy mode forces synchronous sending, ignoring whatever default was configured.
 			return $this->SendSynchronous($aIssues, $oLog);
```
Hipska, line 321: Is this check even needed? In php a boolean true will evaluate equal to int 1, which should match the case `Email::ENUM_SEND_FORCE_SYNCHRONOUS` and thus execute the same method.

Molkobain: This is a strict comparison (triple `=`) so it will work. The behavior you describe would happen with a weak comparison (double `=`)

Hipska: Yes indeed. I'm just wondering if a switch case is strict or weak, so that this extra if can be removed..

Molkobain: Either way I'd rather keep the code as is, it is more explicit about how the legacy type is handle.

Hipska (approved):

Molkobain:
@jbostoen can you mark the PR as ready?

For future Customer Survey PR:
- Sync/Async param. should not be a module param but an enum attribute ("sync" / "async") of the Survey object itself. That way it will be easier to see how it will behave in a glance and to manage what is what (same behavior is planned on the ActionWebhook)
- Add a check to write on Survey to warn user if it is configured as async and CRON is not running
- Later, Combodo will try to introduce a standard API to do this on classes

Molkobain:
_This PR is not forgotten, it will be handle when the 3.2 development starts (in the coming months)_

sources/Core/Email/EmailLaminas.php
piRGoif: Closing conversation as the code has evolved since then

sources/Core/Email/EmailLaminas.php
piRGoif: Code did change since this comment. Closing conversation.

sources/Core/Email/EmailLaminas.php
piRGoif: Code did change since this comment. Closing conversation.

sources/Core/Email/EmailLaminas.php
piRGoif: Code did change since this comment. Closing conversation.

piRGoif:
Hello,
I'm really sorry I missed that PR, and it took 4 months to merge :/ 
It is done now !

@jbostoen when creating the customer-survey corresponding PR, can you mention this one ?

jbostoen:
No problem Pierre.

I moved web hosts 3 weeks ago, now it's no longer an issue for me personally anymore. But it's good it's part of the iTop core; and perhaps someone else will be happy to see this enhancement and create a PR for it.

PR #503, opened 2023-06-06
N°6358 - Login REST API

eespie:


webservices/rest.php
```diff
@@ -94,7 +96,8 @@ function json_last_error_msg() {
 	}
 	if ($iRet != LoginWebPage::EXIT_CODE_OK)
 	{
-		switch($iRet)
+        header('HTTP/1.0 401 Unauthorized');
```
Hipska: 🙌 A correct http response code should be set for all Exceptions to be more REST compliant.

See for example "Missing parameter" which returns 200 OK.

PR #504, opened 2023-06-06
N°6555 - Add class description in tooltip of Dashlet badge

accognet:
display the class name + the description in the tooltip on the dashlet badge
image


Hipska (approved):

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -185,6 +188,23 @@ public function SetClassLabel(string $sClassLabel)
 		return $this;
 	}
 
+	/**
+	 * @return string
+	 */
+	public function GetClassDescription(): string
+	{
+		return $this->sClassDescription;
```
piRGoif, line 201: To be completly independant of the twig template, I think we should return the escaped value using \utils::EscapeHtml

piRGoif (changes requested):

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
Molkobain: > return the escaped value using \utils::EscapeHtml

No, never! Only the consumer should encode/escape data as it is the only one to know when it is necessary. Otherwise you can end up with HTML characters where you don't want to. We have the issue in many places in iTop, especially with \AttributeDefinition

Molkobain (changes requested):
The class label should remain in the tooltip as it is necessary when it is so long that it is truncated.
You should rather display the class label on the first line, then the description below if there is one.

Check what has been done in an object tabs when there are hidden the extra tabs menu.

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
piRGoif: Indeed make perfect sense to escape in the presentation layer only.
Plus the current template will escape, so double escaping will produce an invalid result :(
What i feared was we may not control which template is used, if it is escaping or not. But here we are in a core dashlet class.
Closing discussion.

accognet:
Done

Hipska (approved):
Great, do you also have a new example screenshot?

accognet:
Yes, I just updated the description

application/displayblock.class.inc.php
```diff
@@ -1233,6 +1233,10 @@ protected function RenderActions(array $aExtraParams): iUIBlock
 		} else {
 			$oBlock = DashletFactory::MakeForDashletBadge($sClassIconUrl, $sHyperlink, $iCount, $sClassLabel, null, null, $aRefreshParams);
 		}
+		$sClassDescription = MetaModel::GetClassDescription($sClass);
+		if ($sClassDescription != '') {
```
Molkobain:
```suggestion
		if (utils::IsNotNullOrEmptyString($sClassDescription)) {
```

I didn't check if it's necessary to add an `@use utils;` statement.

templates/base/components/dashlet/dashlet-badge.html.twig
```diff
@@ -3,7 +3,16 @@
...
+            {% if oUIBlock.HasClassDescription() == true %}
+                {# Display both class name and description as the name could be truncated if too long #}
+                data-tooltip-content="{{ '<div class="ibo-dashlet-badge--body--tooltip-title">'|escape }}{{ oUIBlock.GetClassLabel() }}{{ '</div><div class="ibo-dashlet-badge--body--tooltip-description">'|escape }}{{ oUIBlock.GetClassDescription() }}{{ '</div>'|escape }}"
+            {% else %}
+                {# Display only class name as it could be truncated if too long #}
+                data-tooltip-content="{{ oUIBlock.GetClassLabel() }}"
+            {% endif %}
+         data-tooltip-html-enabled="true"
+            {# Delay display to avoid having all tooltips appearing when mouse is just passing through the tabs #}
+         data-tooltip-show-delay="300">
```
Molkobain:
```suggestion
            {% if oUIBlock.HasClassDescription() %}
                {# Display both class name and description as the name could be truncated if too long #}
                data-tooltip-content="{{ '<div class="ibo-dashlet-badge--body--tooltip-title">'|escape }}{{ oUIBlock.GetClassLabel() }}{{ '</div><div class="ibo-dashlet-badge--body--tooltip-description">'|escape }}{{ oUIBlock.GetClassDescription() }}{{ '</div>'|escape }}"
                data-tooltip-html-enabled="true"
            {% else %}
                {# Display only class name as it could be truncated if too long #}
                data-tooltip-content="{{ oUIBlock.GetClassLabel() }}"
            {% endif %}
         {# Delay display to avoid having all tooltips appearing when mouse is just passing through the tabs #}
         data-tooltip-show-delay="300">
```

`data-tooltip-html-enabled` is not necessary if only the label is display as there is no markup.

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -29,6 +30,8 @@ class DashletBadge extends DashletContainer
 	protected $iCount;
 	/** @var string */
 	protected $sClassLabel;
+	/** @var string */
+	protected $sClassDescription = '';
```
Molkobain: In UIBlock we tend to not initialize property values there, but in the constructor. Even though this could be discuss within the team, I would rather keep consistency.

Please:
  * Initialize the property in the constructor
  * Add a `@since 3.1.0`

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -185,6 +188,31 @@ public function SetClassLabel(string $sClassLabel)
 		return $this;
 	}
 
+	/**
+	 * @return string
```
Molkobain, line 196: `@since 3.1.0`

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -185,6 +188,31 @@ public function SetClassLabel(string $sClassLabel)
...
+	 */
+	public function GetClassDescription(): string
+	{
+		return $this->sClassDescription;
+	}
+
+	/**
+	 * @param string $sClassDescription
```
Molkobain, line 205: `@since 3.1.0`

Molkobain (changes requested):

Molkobain:
A few more changes and it should be ok.
This will be a nice addition 🙌

Molkobain:
Accepted during function review, target is 3.1.1.

Anne-Catherine will:
- Create the corresponding bug
- Rebase branch onto support/3.1 (feel free to ask how to "rebase onto" as it can mess up the branch)

Then Guillaume will do a final review

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -29,6 +30,10 @@ class DashletBadge extends DashletContainer
 	protected $iCount;
 	/** @var string */
 	protected $sClassLabel;
+	/** @var string
+	 * @since 3.1.1
+	 */
```
Molkobain:
```suggestion
	/**
	 * @var string
	 * @since 3.1.1 3.2.0
	 */
```

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -185,6 +191,34 @@ public function SetClassLabel(string $sClassLabel)
 		return $this;
 	}
 
+	/**
+	 * @return string
+	 * @since 3.1.1
```
Molkobain:
```suggestion
	 * @since 3.1.1 3.2.0
```

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -185,6 +191,34 @@ public function SetClassLabel(string $sClassLabel)
...
+	 */
+	public function GetClassDescription(): string
+	{
+		return $this->sClassDescription;
+	}
+
+	/**
+	 * @param string $sClassDescription
+	 *
+	 * @since 3.1.1
```
Molkobain:
```suggestion
	 * @return DashletBadge
	 * @since 3.1.1 3.2.0
```

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -185,6 +191,34 @@ public function SetClassLabel(string $sClassLabel)
...
+	 */
+	public function SetClassDescription(string $sClassDescription): void
```
Molkobain:
```suggestion
	public function SetClassDescription(string $sClassDescription)
```

sources/Application/UI/Base/Component/Dashlet/DashletBadge.php
```diff
@@ -185,6 +191,34 @@ public function SetClassLabel(string $sClassLabel)
...
+	{
+		$this->sClassDescription = $sClassDescription;
```
Molkobain, line 212:
```suggestion
		$this->sClassDescription = $sClassDescription;
		
		return $this;
```

Molkobain (approved):

Hipska (approved):

PR #507, opened 2023-06-14
N°5324 - [ERGO] Avoid to have users with non-standalone power portal profile only

odain-cbd:
### **Context**
A user with only 'Power portal user' cannot access iTop (broken page). we propose this configurable repair mecanism to stop trapping new iTop users and increase product adoption.

### **Fix**
This fix relies on 3.1 event mecanism. it consists in adding 'Portal user' profile to any User that would end up with only 'Power portal user'. either by creating/editing the User. either by editing the associated URP_UserProfiles directly.

By default repair mecanism is enabled.
It could be forced/tuned by adding below configuration (you could choose any URP_UserProfile not only 'Portal user' for repairment): 
```
	'security.single_profile_completion' => array(
		'Power portal user' => 'Portal user'
	),
```

### **Warning only**
Instead of repairing you have the possibility to only display a warning message afterwards (without preventing from creating/updating users). 

Here a configuration example:
```
	'security.single_profile_completion' => array(
		'Power portal user' => null
	),
```

Repair/Warning mecanism will be disabled if
- Fallback profile does not exist. ie 'User portal' by defaut or the one mentioned in 'poweruserportal-repair-profile' configuration
- current iTop has more than 2 portal XML configured (_<portals>_ sections ). our customer may have more than these 2 portals (itop-portal-base and backoffice). we dont want to silently give direct access to a specific portal during user edition.


datamodels/2.x/itop-profiles-itil/src/UserProfilesEventListener.php
```diff
@@ -0,0 +1,152 @@
...
+			$this->bIsRepairmentEnabled = false;
+			return;
+		}
+
+		$this->bIsRepairmentEnabled = true;
+	}
+
+	public function RepairProfiles(\User $oUser) : void
+	{
+		if (!is_null($oUser))
```
Molkobain: Isn't this condition redundant as `$oUser` is typed and mandatory?
Or add a `?` in the function signature.

datamodels/2.x/itop-profiles-itil/src/UserProfilesEventListener.php
```diff
@@ -0,0 +1,152 @@
...
+	 * @return void
+	 * @throws \ConfigException
+	 * @throws \CoreException
+	 */
+	public function Init($aPortalDispatcherData=null) : void {
+		if (is_null($aPortalDispatcherData)){
+			$aPortalDispatcherData = \PortalDispatcherData::GetData();
+		}
+
+		$sRepairmentProfile = \utils::GetConfig()->GetModuleSetting('itop-profiles-itil', 'poweruserportal-repair-profile', null);
```
Molkobain: Use underscores instead of dashes.
What about naming the parameter "default_portal_user_profile"?

datamodels/2.x/itop-profiles-itil/src/UserProfilesEventListener.php
```diff
@@ -0,0 +1,152 @@
...
+
+		if (is_null($sRepairmentProfile) && sizeof($aPortalDispatcherData) > 2){
+			//when there are further portals we dont want to force a specific portal by repairing the associated profiles to a user
```
Molkobain:
```suggestion
		
		if (is_null($sRepairmentProfile) && count($aPortalDispatcherData) > 2){
			// When there are several portals we don't want to force a specific portal by repairing the associated profiles to a user as we don't know which portal instance should be the default one ("Greater than 2" as the "backoffice" is considered as a "portal" here)
```

Molkobain (commented):
Discussed with Olivier IRL

datamodels/2.x/itop-profiles-itil/src/UserProfilesEventListener.php
odain-cbd: il will add ? and leave this function public. that way anyone can reuse this API even with null User.

datamodels/2.x/itop-profiles-itil/datamodel.itop-profiles-itil.xml
```diff
@@ -544,5 +544,23 @@
...
+    </dictionary>
+    <dictionary id="FR FR" _alteration="added">
+      <english_description>French</english_description>
+      <localized_description>Français</localized_description>
+      <entries>
+        <entry id="Core:DeletedObjectLabel" _alteration="added">Le profil %1$s ne peut être seul. Sans rajout d'autres profiles,
+          l'utilisateur peut rencontrer des problèmes dans iTop.</entry>
+      </entries>
+    </dictionary>
+  </dictionaries>
```
Molkobain:
```suggestion
  <dictionaries>
    <dictionary id="EN US">
      <entries>
        <entry id="Class:User/NonStandaloneProfileWarning" _delta="define">User profile %1$s cannot be standalone. You should add
          other prrofiles otherwise you may encounter access issue with this user.</entry>
      </entries>
    </dictionary>
    <dictionary id="FR FR">
      <entries>
        <entry id="Core:DeletedObjectLabel" _delta="define">Le profil %1$s ne peut être seul. Sans rajout d'autres profiles,
          l'utilisateur peut rencontrer des problèmes dans iTop.</entry>
      </entries>
    </dictionary>
  </dictionaries>
```

You don't need the `_alteration` flags, did you do this through the ITSM Designer? That being said you do need `_delta` flags for it to work on any iTop.
Also, you don't the node I removed as they are for the language definition and both english / french are already present.

core/config.class.inc.php
```diff
@@ -1595,6 +1595,14 @@ class Config
 			'source_of_value'     => '',
 			'show_in_conf_sample' => false,
 		],
+		'security.single-profile-completion' => [
```
```suggestion
		'security.single_profile_completion' => [
```

We use underscores instead of dashes.

datamodels/2.x/itop-profiles-itil/src/UserProfilesEventListener.php
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * Class UserProfilesEventListener
+ *
+ * @package Combodo\iTop\Core\EventListener
+ * @since 3.1 N°5324 - Avoid to have users with non-standalone power portal profile only
+ *
+ */
+class UserProfilesEventListener implements iEventServiceSetup
+{
+	const USERPROFILE_REPAIR_ITOP_PARAM_NAME = 'security.single-profile-completion';
```
Molkobain:
```suggestion
	const USERPROFILE_REPAIR_ITOP_PARAM_NAME = 'security.single_profile_completion';
```

Molkobain (changes requested):

datamodels/2.x/itop-profiles-itil/datamodel.itop-profiles-itil.xml
odain-cbd: sorry XML built via a copy/paste from datamodel-production. i will fix.

FR entry id was incorrect as well...

core/config.class.inc.php
odain-cbd: ok i will rename it

addons/userrights/userrightsprofile.class.inc.php
```diff
@@ -220,6 +220,7 @@ public static function Init()
 	{
 		$aParams = array
 		(
+			"is_link" 			  => true, //since 3.1 N°5324
```
Molkobain:
```suggestion
```

Done in [12dbd0e](https://github.com/Combodo/iTop/commit/12dbd0ed3da279e49ae4858055ebc05180c913d8)

addons/userrights/userrightsprofile.db.class.inc.php
```diff
@@ -326,6 +326,7 @@ public static function Init()
 	{
 		$aParams = array
 		(
+			"is_link" 			  => true, //since 3.1 N°5324
```
Molkobain:
```suggestion
```

Done in [12dbd0e](https://github.com/Combodo/iTop/commit/12dbd0ed3da279e49ae4858055ebc05180c913d8)

addons/userrights/userrightsprojection.class.inc.php
```diff
@@ -269,6 +269,7 @@ public static function Init()
 	{
 		$aParams = array
 		(
+			"is_link" 			  => true, //since 3.1 N°5324
```
Molkobain:
```suggestion
```

Done in [12dbd0e](https://github.com/Combodo/iTop/commit/12dbd0ed3da279e49ae4858055ebc05180c913d8)

core/datamodel.core.xml
```diff
@@ -57,6 +57,7 @@
         <!-- Generated by toolkit/export-class-to-meta.php -->
         <parent>cmdbAbstractObject</parent>
         <properties>
+          <is_link>1</is_link> <!-- since 3.1 N°5324 -->
```
Molkobain:
```suggestion
```

Done in [12dbd0e](https://github.com/Combodo/iTop/commit/12dbd0ed3da279e49ae4858055ebc05180c913d8)

Molkobain (changes requested):

piRGoif:
Discussed with Olivier : the functionality was implemented differently and merged.
Closing this PR !

PR #509, opened 2023-06-20
N° 8058 - 🐛 Fix: when editing text, the original Wiki syntax should be visible - using the class and not the class label.

jbostoen:
Fix: when editing text, the original Wiki syntax should be visible - using the class and not the class label.

Try: in a text field of an object, refer to an object using wiki syntax, such as `[[WebServer::1]]`. Save. So far, everything works as expected. In view mode, it will show a link to Web Server (id 1).

However, when modifying the object (with the text field referencing Web Server 1), the text area will show the class label: `[[Web Server::1]]` . Easily overlooked, but it breaks the wiki syntax when saving the object again.

This PR changes the edit value to actually contain the original reference, using the class name instead of the class label.

See https://sourceforge.net/p/itop/tickets/1841/?limit=25#ddc5

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -4518,8 +4518,7 @@ class AttributeText extends AttributeString
 					$sLabel = (!empty($aMatches[4])) ? trim($aMatches[4]) : null;
 
 					if (MetaModel::IsValidClass($sClass)) {
-						$sClassLabel = MetaModel::GetName($sClass);
-						$sReplacement = "[[$sClassLabel:$sName".(!empty($sLabel) ? " | $sLabel" : "")."]]";
+						$sReplacement = "[[$sClass:$sName".(!empty($sLabel) ? " | $sLabel" : "")."]]";
 						$sValue = str_replace($aMatches[0], $sReplacement, $sValue);
 					}
 				}


Hipska (approved):

Molkobain:
Guillaume will try to reproduce through a unit test before functional review

jf-cbd:
Hello @jbostoen, thanks for the PR.
Your fix works well, but during the review of the related ticket, it has been decided that Wiki syntax should not be supported in text fields (the Wiki syntax remains supported in HTML fields) 😔
So when we'll work on the ticket, we'll remove fully the "conversion" that is actually broken.

jbostoen:
It's been a long time - but if I remember correctly, I actually wanted to fix it for AttributeHTLM (so indeed where links are allowed) - which just happens to inherit the logic from AttributeText.

jf-cbd:
Yes, sorry for the delay. I'm retrieving old PR 😅
In that case could you please confirm that now the Wiki syntax works well in AttributeHTLM (without the PR) ?
The related ticket https://sourceforge.net/p/itop/tickets/1841/?limit=25#ddc5 talks about the description field of a NetworkDevice, and this field is an AttributeText.
I reproduce it, but I can't reproduce it with an AttributeHTLM.

PR #514, opened 2023-06-26
N°5945 - Fix "menu" option in dashlet list having no effect since relations refactoring

accognet:
@internal

### DIFF
--- a/application/displayblock.class.inc.php
+++ b/application/displayblock.class.inc.php
@@ -1871,7 +1871,8 @@ class MenuBlock extends DisplayBlock
 		/** @var array $aToolkitActions Any "legacy" toolkit menu item, which are now displayed in the same menu as the $aRegularActions, after them */
 		$aToolkitActions = [];
 
-		if (!isset($aExtraParams['selection_mode']) || ($aExtraParams['selection_mode'] == "")) {
+		if ((!isset($aExtraParams['selection_mode']) || ($aExtraParams['selection_mode'] == ""))
+			&& (!isset($aExtraParams['menu']) || $aExtraParams['menu'])) {
 			$oAppContext = new ApplicationContext();
 			$sContext = $oAppContext->GetForLink();
 			if (utils::IsNotNullOrEmptyString($sContext)) {


application/displayblock.class.inc.php
```diff
@@ -1871,7 +1871,8 @@ public function GetRenderContent(WebPage $oPage, array $aExtraParams, string $sI
 		/** @var array $aToolkitActions Any "legacy" toolkit menu item, which are now displayed in the same menu as the $aRegularActions, after them */
 		$aToolkitActions = [];
 
-		if (!isset($aExtraParams['selection_mode']) || ($aExtraParams['selection_mode'] == "")) {
+		if ((!isset($aExtraParams['selection_mode']) || ($aExtraParams['selection_mode'] == ""))
+			&& (!isset($aExtraParams['menu']) || $aExtraParams['menu'])) {
```
Molkobain, line 1875:
```suggestion
		// Display menu actions only if...
		if (
			// ... NOT in a selection mode
			(!isset($aExtraParams['selection_mode']) || ($aExtraParams['selection_mode'] == ""))
			// ... "menu" parameter is NOT EXPLICITLY disabled
			&& (!isset($aExtraParams['menu']) || $aExtraParams['menu'] === "1" || $aExtraParams['menu'] === true)
		) {
```

  * Being explicit about what kind of value we expect in the "menu" param
  * Adding comments to explain in which functional use cases we should display the menu

Molkobain (changes requested):
As is it breaks the indirect linksets as it doesn't display the menu anymore. You need to include another change to this PR.

In `Combodo\iTop\Application\UI\Links\Indirect\BlockIndirectLinkSetViewTable::GetExtraParam()`, set the `menu` param from `false` to `MetaModel::GetConfig()->Get('allow_menu_on_linkset')`

```php
class BlockIndirectLinkSetViewTable extends AbstractBlockLinkSetViewTable
{
	[...]

	/** @inheritdoc */
	public function GetExtraParam(): array
	{
		$aExtraParams = array(
			[...]
			'menu'          => MetaModel::GetConfig()->Get('allow_menu_on_linkset'),
			[...]
		);
	[...]
```

Molkobain:
Discussed with Anne-Catherine, I'll apply the changes locally.
Commited under [3c8bea8](https://github.com/Combodo/iTop/commit/3c8bea89212e0c8ec63cd19f1ed04f7d4647f1f9)

PR #519, opened 2023-07-03
N°7292 - Improve Clear function in ExtKeyWidget

accognet:
Currently, the function Clear reset the field only if it's an autocomplete (and is used only in this case). 
Add the hability to reset the field in all cases.


Molkobain (approved):
Fix seems ok, Anne-Cath will:
- Create the corresponding ticket
- Add reproduction procedure on ext. key for 
  - Verify that mandatory hint is displayed when emptied
  - Verify that depend fields are reloaded

Molkobain:
Discussion Alexandre, Benjamin & Guillaume: We need to do the things on the previous post and go forward

Molkobain:

PR #521, opened 2023-07-05
N°6543 - Fix display of AttributeText with width parameter

accognet:
If I define width on a AttributeText field by example with this delta :

  ```
<class id="Ticket" _created_in="itop-tickets" _delta="must_exist">
    <fields>
      <field id="description" xsi:type="AttributeText">
        <width _delta="define">1000px</width>
        <height _delta="define">200px</height>
      </field>
    </fields>
  </class>
```

Display of object is not good:
image
And in edit mode:
image
This PR allow to fix the display. The result is:
image
And in edit mode:
image

In search, the change is not big : 
before 
image
after
image

### DIFF
--- a/application/ui.htmleditorwidget.class.inc.php
+++ b/application/ui.htmleditorwidget.class.inc.php
@@ -74,7 +74,7 @@ class UIHTMLEditorWidget
 		$aConfig = utils::GetCkeditorPref();
 		$sWidthSpec = addslashes(trim($this->m_oAttDef->GetWidth()));
 		if ($sWidthSpec != '') {
-			/*the function min allow to keep text inside the column when width is defined*/
+			/*N°6543 - the function min allow to keep text inside the column when width is defined*/
 			$aConfig['width'] = "min($sWidthSpec,100%)";
 		}
 		$sHeightSpec = addslashes(trim($this->m_oAttDef->GetHeight()));
--- a/application/ui.htmleditorwidget.class.inc.php
+++ b/application/ui.htmleditorwidget.class.inc.php
@@ -71,11 +71,13 @@ class UIHTMLEditorWidget
 		// To change the default settings of the editor,
 		// a) edit the file /js/ckeditor/config.js
 		// b) or override some of the configuration settings, using the second parameter of ckeditor()
+		$sJSDefineWidth = '';
 		$aConfig = utils::GetCkeditorPref();
 		$sWidthSpec = addslashes(trim($this->m_oAttDef->GetWidth()));
 		if ($sWidthSpec != '') {
 			/*N°6543 - the function min allow to keep text inside the column when width is defined*/
 			$aConfig['width'] = "min($sWidthSpec,100%)";
+			$sJSDefineWidth = '$("#cke_'.$iId.' iframe").contents().find("body").css("width", "'.$sWidthSpec.'")';
 		}
 		$sHeightSpec = addslashes(trim($this->m_oAttDef->GetHeight()));
 		if ($sHeightSpec != '') {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: application/ui.htmleditorwidget.class.inc.php [hunk 1]
```diff
@@ -73,13 +73,12 @@ public function Display(WebPage $oPage, $aArgs = array())
 		// b) or override some of the configuration settings, using the second parameter of ckeditor()
 		$aConfig = utils::GetCkeditorPref();
 		$sWidthSpec = addslashes(trim($this->m_oAttDef->GetWidth()));
-		if ($sWidthSpec != '')
-		{
-			$aConfig['width'] = $sWidthSpec;
+		if ($sWidthSpec != '') {
+			/*the function min allow to keep text inside the column when width is defined*/
```
Molkobain: Add ticket number in the comment for easier maintenance.

css/backoffice/components/_field.scss
```diff
@@ -80,16 +80,25 @@ $ibo-field--enable-bulk--checkbox--margin-left: $ibo-spacing-300 !default;
       }
     }
   }
+  /*We need the rule to keep text inside the column when width is defined*/
+  &[data-attribute-type="AttributeText"]:not([data-input-type="html_editor"]) {
```
Molkobain: Add AttributeHTML as well and remove unnecessary `:not([data-input-type="html_editor"])`

_Note: In CSS there is no inheritance of classes, so any attribute derived from AttributeText won't have this rule applied, but this is a known limitation and accepted for now. We might better use something like `AttributeDefinition::GetEditClass()` in the future_

Molkobain:

To discuss in functional review: Currently tables are wrapped as much as possible. Texts outside the table wrap as well, which make it easy to read. With this, texts outside the table will no longer wrap and we'll need to scroll to read them, but as it is only for fields with a specified width it should be ok. Will discuss during functional review.

css/backoffice/components/_field.scss
accognet: After test, it doesn't work without _:not([data-input-type="html_editor"])_

Molkobain: Can you detail why?

accognet: because it displays a double horizontal scrollbar in edit mode

application/ui.htmleditorwidget.class.inc.php
accognet: Bug created => I can add the number

Molkobain:
Accepted during functional review for 3.1.1 upon the following condition:
In the screenshots, the text on the first paragraph doesn't wrap the same way in read-only and edit. It would be better to have the same behavior.

Anne-Catherine will check how to fix this.
Guillaume will then do a final review.

Molkobain:
Support PRs review: SCSS and JS seem good! We just need to check where the change in AttributeDef could have an impact (parts of the app where we use `GetAsHTML` directly)

css/backoffice/components/datatable/_datatable.scss
```diff
@@ -123,6 +123,13 @@ $ibo-fieldsorter--selected--background-color: $ibo-color-blue-200 !default;
     .ibo-datatable--row-actions-toolbar{
       justify-content: end;
     }
+    /*N°6543 - We need the rule to keep text inside the column when width is defined*/
+    >[data-attribute-type="AttributeHtml"],
+    >[data-attribute-type="AttributeText"] {
```
Molkobain:
```suggestion
    /* N°6543 - We need the rule to keep text inside the column when width is defined */
    > [data-attribute-type="AttributeHtml"],
    > [data-attribute-type="AttributeText"] {
```

Molkobain (approved):

Molkobain:

Molkobain:
_Anne-Catherine will update the screenshots to show how it displays in a list_

Molkobain:

Molkobain:
@accognet you can "squash and merge", please mind to keep the PR number in the commit

PR #525, opened 2023-07-27
N°6600 Portal download attachment : don't display anymore SQL query on attachment not found error

piRGoif:
Before PR : 
image

After PR : 
image

Also, after PR a log is made on the CMDBSource channel, info level, containing the SQL query, and object class and id.

Note that on the admin console the behavior is to display an error page, and a simple message (no SQL query).
image

### DIFF
--- a/core/metamodel.class.php
+++ b/core/metamodel.class.php
@@ -6723,7 +6723,13 @@ abstract class MetaModel
 
 		if ($bMustBeFound && empty($aRow))
 		{
-			throw new CoreException("No result for the single row query");
+			$sNotFoundErrorMessage = "No result for the single row query";
+			IssueLog::Info($sNotFoundErrorMessage, LogChannels::CMDB_SOURCE, [
+				'class' => $sClass,
+				'key' => $iKey,
+				'sql_query' => $sSQL,
+				]);
+			throw new CoreException($sNotFoundErrorMessage);
 		}
 
 		return $aRow;
--- a/core/metamodel.class.php
+++ b/core/metamodel.class.php
@@ -6723,7 +6723,7 @@ abstract class MetaModel
 
 		if ($bMustBeFound && empty($aRow))
 		{
-			throw new CoreException("No result for the single row query: '$sSQL'");
+			throw new CoreException("No result for the single row query");
 		}
 
 		return $aRow;
--- a/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
@@ -1038,15 +1038,7 @@ class ObjectController extends BrickController
 			
 			$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, false, true);
 			if ($oAttachment === null) {
-				utils::PushArchiveMode(true);
-				$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, false);
-				utils::PopArchiveMode();
-				if (is_null($oAttachment)) {
-					IssueLog::Info(__METHOD__ . ' at line ' . __LINE__ . ' : Could not load object ' . $sObjectClass . '::' . $sObjectId . '.');
-					throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
-				}
-
-				IssueLog::Info(__METHOD__ . ' at line ' . __LINE__ . ' : access to archived object ' . $sObjectClass . '::' . $sObjectId . '.');
+				throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
 			}
 			$sHostClass = $oAttachment->Get('item_class');
 			$sHostId = $oAttachment->Get('item_id');
--- a/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
@@ -20,7 +20,6 @@
 
 namespace Combodo\iTop\Portal\Controller;
 
-use ArchivedObjectException;
 use AttributeEnum;
 use AttributeFinalClass;
 use AttributeFriendlyName;
@@ -29,7 +28,6 @@ use BinaryExpression;
 use Combodo\iTop\Portal\Brick\CreateBrick;
 use Combodo\iTop\Portal\Helper\ApplicationHelper;
 use Combodo\iTop\Portal\Helper\ContextManipulatorHelper;
-use CoreException;
 use DBObject;
 use DBObjectSearch;
 use DBObjectSet;
--- a/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
@@ -1038,8 +1038,15 @@ class ObjectController extends BrickController
 			
 			$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, false, true);
 			if ($oAttachment === null) {
-				IssueLog::Info(__METHOD__.' at line '.__LINE__.' : Could not load object '.$sObjectClass.'::'.$sObjectId.'.');
-				throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
+				utils::PushArchiveMode(true);
+				$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, false);
+				utils::PopArchiveMode();
+				if (is_null($oAttachment)) {
+					IssueLog::Info(__METHOD__ . ' at line ' . __LINE__ . ' : Could not load object ' . $sObjectClass . '::' . $sObjectId . '.');
+					throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
+				}
+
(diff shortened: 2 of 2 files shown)


piRGoif:
Behavior was submitted on the SourceForge forum : https://sourceforge.net/p/itop/discussion/customizing-itop/thread/2cc7b74178/?limit=25
Thanks Vincenzo K  !

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
```diff
@@ -1035,7 +1037,11 @@ public function DocumentAction(Request $oRequest, $sOperation = null)
 		// When reaching to an Attachment, we have to check security on its host object instead of the Attachment itself
 		if ($sObjectClass === 'Attachment')
 		{
-			$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, true, true);
+			try {
+				$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, true, true);
+			} catch (ArchivedObjectException|CoreException $e) {
+				throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
+			}
```
Molkobain: I would rather use the same pattern as a few lines below (set must be found to false and test for null value) to keep things homogeneous and to avoid missing a future exception type change.

Molkobain (changes requested):

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
piRGoif: As the test to display the error message is outside of the if where the GetObject is done, that would make the code rather complicated ? Something like : 

```php
		// When reaching to an Attachment, we have to check security on its host object instead of the Attachment itself
		if ($sObjectClass === 'Attachment')
		{
			$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, false, true);
			if (false === is_null($oAttachment)) {
				$sHostClass = $oAttachment->Get('item_class');
				$sHostId = $oAttachment->Get('item_id');
			}
		}
		else
		{
			$sHostClass = $sObjectClass;
			$sHostId = $sObjectId;
		}

		// Checking security layers
		// Note: Checking if host object already exists as we can try to download document from an object that is being created
		if (
				(($sObjectClass === 'Attachment') && \is_null($oAttachment))
				|| ((\is_null($oAttachment) || (($sHostId > 0) && !$oSecurityHelper->IsActionAllowed(UR_ACTION_READ, $sHostClass, $sHostId))))
			)
		{
			IssueLog::Warning(__METHOD__.' at line '.__LINE__.' : User #'.UserRights::GetUserId().' not allowed to retrieve document from attribute '.$sObjectField.' as it not allowed to read '.$sHostClass.'::'.$sHostId.' object.');
		}
```

Plus \MetaModel::GetObject is part of the API, we won't change its signature without thinking of the consequences right ?

Molkobain:
```suggestion
			
			$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, false, true);
			if ($oAttachment === null) {
				IssueLog::Info(__METHOD__.' at line '.__LINE__.' : Could not load object '.$sObjectClass.'::'.$sObjectId.'.');
				throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
			}
```

piRGoif: Ok, I though you wanted to reuse the same if for error message below :) 
Thanks !

Molkobain (approved):

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
```diff
@@ -1035,7 +1037,12 @@ public function DocumentAction(Request $oRequest, $sOperation = null)
 		// When reaching to an Attachment, we have to check security on its host object instead of the Attachment itself
 		if ($sObjectClass === 'Attachment')
 		{
-			$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, true, true);
+			
+			$oAttachment = MetaModel::GetObject($sObjectClass, $sObjectId, false, true);
+			if ($oAttachment === null) {
```
Hipska, line 1040: Why not keep `$bMustBeFound` true and catch on exception?

piRGoif: That was asked by Guillaume in the above conversation : https://github.com/Combodo/iTop/pull/525#discussion_r1276788966
But we didn't explain exactly why we prefer this :)

There aren't a precise exception thrown by MetaModel::GetObject when object isn't found, so a CoreException throwned by this method might be caused by something completely different.
Checking for null really means only this.

Hipska (approved):

piRGoif:
Discussed during technical review : Romain proposed to simply do not throw the SQL anymore :) It was added when the main cause of non found objects was bad queries, but the ORM has grown and it shouldn't happen anymore. And if it does, we can still write down quickly a PHPUnit to reproduce and get everything.
Plus of course we don't want to display big queries to the final user...

I'll have a look at if this can be done, and we'll discuss on the modification on next technical review.

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
Hipska: It feels like a step back. I agree that a CoreException isn't a good indicator as well, there should be something like ObjectNotFoundException..

piRGoif:
Done, ready for next technical review !

core/metamodel.class.php
```diff
@@ -6707,7 +6707,7 @@ public static function MakeSingleRow($sClass, $iKey, $bMustBeFound = true, $bAll
 
 		if ($bMustBeFound && empty($aRow))
 		{
-			throw new CoreException("No result for the single row query: '$sSQL'");
+			throw new CoreException("No result for the single row query");
```
Molkobain: I find the SQL rather helpful for debugging, for example when it crashes because nothing is returned, it helps -by checking the joins / wheres- to find which condition make it fail.

What about putting the complete message (with SQL) in the logs (ERROR level) as well (I thought that was the approach you would do when discussed yesterday)

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
```diff
@@ -20,6 +20,7 @@
 
 namespace Combodo\iTop\Portal\Controller;
 
+use ArchivedObjectException;
```
Molkobain:
```suggestion
```

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
```diff
@@ -28,6 +29,7 @@
 use Combodo\iTop\Portal\Brick\CreateBrick;
 use Combodo\iTop\Portal\Helper\ApplicationHelper;
 use Combodo\iTop\Portal\Helper\ContextManipulatorHelper;
+use CoreException;
```
Molkobain:
```suggestion
```

Molkobain (changes requested):

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
piRGoif: Done with 32706270

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
piRGoif: Done with 32706270

core/metamodel.class.php
piRGoif: Yeah, adding a log is a good idea. But as there was no log before (yeah there was on the upper layers of the call stack, but not in this low level method), I don't think using the error level is a good idea.
I've just pushed 7fbe251b with a log on CMDBSource channel and info level.
Ok for you ?

Molkobain: Yep

Molkobain (approved):

core/metamodel.class.php
Hipska: Could maybe be even debug level.. Since you want to use it for debugging ;-)

piRGoif: I have mixed feelings, as before the developer was always getting the SQL query : enabling debug level would add possibly lots of logs you don't want.
We'll debate on the proper log level on the next technical review.

Hipska (approved):

piRGoif:
Technical review : to check with Romain before next functional review

piRGoif:
Review done with Romain !
The same behavior is already implemented in UI.php (case details). In this page, we also handle archived objects to display another error message.
Check if the code can be factorized, and if so add a test.

piRGoif:
Functional review : OK
Romain's feedback to be developed, and review to be done

core/metamodel.class.php
piRGoif: Was validated as is during reviews

piRGoif:
The portal object visualization is handled in \Combodo\iTop\Portal\Controller\ObjectController::ViewAction and doesn't handle archive mode.
Actually, the keywork "archive" doesn't appear in the itop-portal-base and itop-portal modules code (just in phpdoc blocks for ArchivedObjectException)

Discussed with Romain : ok, let's get back to previous version so the attachment behavior will be the same as object behavior.
About the log : might not be very useful, just get rid of it.

PR #526, opened 2023-07-28
N°3677 - Fix AttributeImage.default_image URLs not up to date after app_root_url change

piRGoif:
## Original work
This is a cherry pick of commits made in #176 (original author @sg-gcouronne , many thanks to him !)
The original PR was looked at far too late, so we had no choice but merging manually the changes to a new branch, based on a more recent code.

The original PR aimed at fixing the mandatory run of a setup after changing the `app_root_url` config parameter. There were 3 issues addressed, changing absolute URLs generated during setup on : 
* class icon
* display template
* AttributeImage default_image

## Two concerns already fixed
Since iTop 3.0.0 : 
* the icon compilation is now done in MFCompiler::GenerateStyleDataFromNode, which creates a call to ormStyle passing the relative path to the icon (in other words : absolute url computation is done at runtime)
* display template were removed (N°3233)

## Remaining problem on AttributeImage default_image
Examples in the default datamodel : Person.picture (images/silhouette.png) and admin Person instance.

### To reproduce (on Windows) : 

1. create a new lmhosts entry
2. run setup using an URL with this entry
3. remove the entry (comment lmhosts declaration + run `nbtstat -R`)
4. change app_root_url config param to localhost (DO NOT run setup again !)
5. access the previously installed iTop instance with localhost

Before change : 
Broken default image in the admin Person instance

After change : 
Default image in admin Person instance is OK

## TODO

- [x] Change Test annotations
- [x] Rebase on support/3.2
- ~~Change since annotations~~
- [x] Document AttrimageImage.default_image path is relative to the env root (note that an AttributeImage was added in Action core class : N°2039)

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -8651,7 +8651,34 @@ class AttributeImage extends AttributeBlob
 	 */
 	public function __construct($sCode, $aParams)
 	{
+		//Remove AbsoluteUrlModulesRoot hardcoded to the default_image
+		if (!empty($aParams) && array_key_exists('default_image', $aParams) && !empty($aParams['default_image'])) {
+			$sAbsoluteUrlModulesRoot = utils::GetAbsoluteUrlModulesRoot();
+			if (strpos($aParams['default_image'], $sAbsoluteUrlModulesRoot) !== false) {
+				$aParams['default_image'] = str_replace($sAbsoluteUrlModulesRoot, '', $aParams['default_image']);
+			}
+		}
+
 		parent::__construct($sCode, $aParams);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 * @param string $sParamName
+	 * @return type
+	 */
+	public function Get($sParamName)
+	{
+		$oParamValue = parent::Get($sParamName);
+
+		if ($sParamName == 'default_image') {
+			if (!empty($oParamValue)) // check if not null
+			{
+				return utils::GetAbsoluteUrlModulesRoot() . $oParamValue;
+			}
+		}
+
+		return $oParamValue;
 	}
 
 	public function GetEditClass()
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -8651,14 +8651,6 @@ class AttributeImage extends AttributeBlob
 	 */
 	public function __construct($sCode, $aParams)
 	{
-		//Remove AbsoluteUrlModulesRoot hardcoded to the default_image
-		if (!empty($aParams) && array_key_exists('default_image', $aParams) && !empty($aParams['default_image'])) {
-			$sAbsoluteUrlModulesRoot = utils::GetAbsoluteUrlModulesRoot();
-			if (strpos($aParams['default_image'], $sAbsoluteUrlModulesRoot) !== false) {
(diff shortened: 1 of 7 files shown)


Molkobain:
Technical review:
- Remove AttributeBlob::__construct() as we don't know why it's there.
- Then it should be ok

core/attributedef.class.inc.php
```diff
@@ -8626,9 +8626,36 @@ class AttributeImage extends AttributeBlob
...
+			if (strpos($aParams['default_image'], $sAbsoluteUrlModulesRoot) !== false) {
+				$aParams['default_image'] = str_replace($sAbsoluteUrlModulesRoot, '', $aParams['default_image']);
+			}
+		}
+
 		parent::__construct($sCode, $aParams);
 	}
 
+	/**
+	 * {@inheritDoc}
```
```suggestion
	 * @inheritDoc
```

piRGoif: Actually the phpdoc block doesn't give useful information at all. I'll remove it.

piRGoif: Done in ee2f3765

piRGoif:
> * Remove AttributeBlob::__construct() as we don't know why it's there.

Done in 6ba7c450 (it was AttributeImage constructor by the way)

Molkobain (approved):

Hipska (approved):

Molkobain (changes requested):
Can you add a unit test to ensure that `\AttributeImage::Get()` return the correct URL for the default image even when we change the `app_root_url`?

There is the `Person->picture` attribute in the standard DM that you can use.

piRGoif:
> Can you add a unit test to ensure that \AttributeImage::Get() return the correct URL

Very good idea ! 
Done in bb621f9f

Note that there was a cache in \utils::GetAbsoluteUrlAppRoot (called from \AttributeImage::Get) that cannot be cleared from the test :( I removed it completely, I don't think there should be any perceptible drawback as everything is done in memory ? To be discussed during technical review...

piRGoif:
Well, as removing the cache made utilsTest::testGetAbsoluteUrlAppRootPersistency to crash, I restored it but partially : only for the use case of the test (having app_root_url empty)

Molkobain:
Accepted during functional review in 3.2 because not much requestors, low impact and workaround (run setup again).
We just have to check the utils method, then merge.

tests/php-unit-tests/unitary-tests/core/AttributeImageTest.php
```diff
@@ -0,0 +1,48 @@
...
+use Config;
+use MetaModel;
+use Person;
+use utils;
+
+/**
+ * @runTestsInSeparateProcesses
+ * @preserveGlobalState disabled
+ * @backupGlobals disabled
+ */
```
piRGoif: To remove before merge !

Molkobain:
Support PRs review:
- Ok with the refactoring
- Since iTop 3.2 and N°2039, we have an AttributeImage in the datamodel.core.xml, ok to keep path relative to env-xxx. Pierre will just update the documentation to make it clear that these paths must be relative to tne current environment and not the app root.

tests/php-unit-tests/unitary-tests/core/AttributeImageTest.php
piRGoif: Done !

PR #527, opened 2023-07-31
N°5786 - Problem with text color in public log and in AttributeHTML

accognet:
Internal

### DIFF
--- a/css/backoffice/vendors/_bulma-variables-overload.scss
+++ b/css/backoffice/vendors/_bulma-variables-overload.scss
@@ -13,8 +13,8 @@ $body-overflow-x: hidden !default;
 $body-overflow-y: auto !default;
 
 
-  /*N°5786 - Problem with text color in public log and in AttributeHTML - override bulma css */
-$text-strong: inherit !default;;
+/*N°5786 - Avoid strong text to always be grey (default Bulma color for this var.) so strong text can keep its color. This  is mostly for text within the .ibo-is-html-content. */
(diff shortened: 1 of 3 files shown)


css/backoffice/vendors/_bulma-variables-overload.scss
```diff
@@ -13,6 +13,9 @@ $body-overflow-x: hidden !default;
 $body-overflow-y: auto !default;
 
 
+  /*N°5786 - Problem with text color in public log and in AttributeHTML - override bulma css */
+$text-strong: inherit !default;;
```
Molkobain:
```suggestion
/*N°5786 - Avoid strong text to always be grey (default Bulma color for this var.) so strong text can keep its color. This  is mostly for text within the .ibo-is-html-content. */
$text-strong: inherit !default;
```

Molkobain (changes requested):

Molkobain (approved):

Molkobain:
@steffunky the drawback is that it applies to more than the .ibo-is-html-content, but as we don't use bulma classes much, I felt it was okay. What do you think?

steffunky:
Doest it affect Alerts block ?
Or Backoffice communications' alerts ?

accognet:
Yes, it solves the same problem

piRGoif:
Technical review : OK

piRGoif:
Functional review : OK

PR #531, opened 2023-08-01
N°6618 - Fix crash due to router's cache containing an integer instead of an array

Molkobain:
## Symptom

Critical issue seen on our internal iTop instance, the application is no longer accessible:

```
Uncaught TypeError: Return value of Combodo\iTop\Service\Router\Router::GetRoutes() must be of the type array, int returned in <ITOP>/sources/Service/Router/Router.php:190
Stack trace:
#0 <ITOP>/sources/Service/Router/Router.php(276): Combodo\iTop\Service\Router\Router->GetRoutes()
#1 <ITOP>m/sources/Service/Router/Router.php(205): Combodo\iTop\Service\Router\Router->FindHandlerFromRoute()
#2 <ITOP>/sources/Service/Router/Router.php(98): Combodo\iTop\Service\Router\Router->GetDispatchSpecsForRoute()
#3 <ITOP>m/pages/ajax.render.php(77): Combodo\iTop\Service\Router\Router->CanDispatchRoute()
#4 {main}
thrown
```

## Reproduction

We don't know how it happened nor how to reproduce all the steps to the crash state.
We don't know what triggers the corruption. It worked fine for a couple months, then it happened twice in a day.
We just know how to reproduce the crash itself:

  * On a community iTop, NOT in "dev" mode (either by using a package, or if on a Git clone by setting `'developer_mode.enabled'` to `false`)
  * Open an object in the backoffice to generate the router's cache
  * Open `<ITOP>data/cache-production/router/available-routes.php` and replace content with:
```
<?php

return 1;
```
  * Re-open an object
  * It should crash
 
## Cause

We don't know what caused the corrupted cache file with content `return 1;` instead of `return [.......];`
We see that the cache file is actually re-generated at each call (not much of a cache 💩)

In the [PHP documentation](https://www.php.net/manual/en/function.include.php#example-141) of the `include` statement, we see that _"if an included file as no return statement, then it will return 1"_.

This seems to be consistent with our issue, we just don't know how / why it would include a file that doesn't have the routes as an array.
An hypethesis would be that in the case of multiple concurrent accesses, a first process could have emptied the file but written in it yet, while a second process is reading it before its content is flushed. But we have no way of verifying it. 

## Solution

  * Do NOT re-generate the file at each call
  * Check that cache file actually returns an array and not `1` 

## Workaround

  * Delete the corrupted cache file (<ITOP>data/cache-production/router/availables-route.php)
  * Re-generate it by opening an object in the backoffice
  * Change cache file permissions on the file system to read-only for everybody

This solution works until the next setup / MTT / MTP is done, as it will delete the file.

### DIFF
--- a/sources/Service/Router/Router.php
+++ b/sources/Service/Router/Router.php
@@ -144,7 +144,16 @@ class Router
 		// Try to read from cache
 		if ($bUseCache) {
 			if (is_file($sCacheFilePath)) {
-				$aRoutes = include $sCacheFilePath;
+				$aCachedRoutes = include $sCacheFilePath;
+
+				// N°6618 - Protection against corrupted cache returning `1` instead of an array of routes
+				if (is_array($aCachedRoutes)) {
+					$aRoutes = $aCachedRoutes;
+				} else {
+					// Invalid cache force re-generation
+					// Note that even if it is re-generated corrupted again, this protection should prevent crashes
+					$bMustWriteCache = true;
+				}
 			} else {
 				$bMustWriteCache = true;
 			}
--- a/sources/Service/Router/Router.php
+++ b/sources/Service/Router/Router.php
@@ -138,12 +138,15 @@ class Router
 	{
 		$aRoutes = [];
 		$bUseCache = false === utils::IsDevelopmentEnvironment();
+		$bMustWriteCache = false;
 		$sCacheFilePath = $this->GetCacheFileAbsPath();
 
 		// Try to read from cache
 		if ($bUseCache) {
 			if (is_file($sCacheFilePath)) {
 				$aRoutes = include $sCacheFilePath;
+			} else {
+				$bMustWriteCache = true;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


sources/Service/Router/Router.php
```diff
@@ -144,7 +144,15 @@ public function GetRoutes(): array
...
 		if ($bUseCache) {
 			if (is_file($sCacheFilePath)) {
-				$aRoutes = include $sCacheFilePath;
+				$aCachedRoutes = include $sCacheFilePath;
+
+				// N°6618 - Protection against corrupted cache returning `1` instead of an array of routes
+				if (is_array($aCachedRoutes)) {
+					$aRoutes = $aCachedRoutes;
+				} else {
+					// Invalidate cache but do not re-generate it or we could fall in the same issue we are trying to fix
```
bdalsass: Re-generate it would be nice

bdalsass (approved):

sources/Service/Router/Router.php
Molkobain: Done.

As discussed with Benjamin IRL, I was affraid (,I was petrified 🎵) the cache could be re-generated corrupted again but as we have a protection against the usage of a corrupted file on line 150 it should be ok.

piRGoif (changes requested):
Great job ! 
But you're testing the cache file isn't regenerated, but not that the code isn't crashing anymore if its content is invalid ?

Molkobain:
Good idea, I just added it and forced-push the branch.

PR #532, opened 2023-08-02
N°6619 - Attachment : Make contact_id an AttributeExternalKey instead of AttributeExternalField

accognet:
Description here : https://sourceforge.net/p/itop/tickets/2167/

### DIFF
--- a/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
+++ b/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
@@ -110,9 +110,14 @@
     public function DBInsertNoReload()
     {
         $this->SetCurrentDateIfNull('creation_date');
-        $this->SetIfNull('user_id', CMDBChange::GetCurrentUserId());
-        $this->SetIfNull('contact_id', UserRights::GetContactId());
+        $iUserId = CMDBChange::GetCurrentUserId();
+        if(utils::IsNotNullOrEmptyString($iUserId)){
+               $this->SetIfNull('user_id', $iUserId);
 
+              // Get Contact from user
+              $oUser = MetaModel::GetObject('User', $iUserId);
+              $this->SetIfNull('contact_id', $oUser->Get('contactid'));
+         }
         return parent::DBInsertNoReload();
     }
         ]]></code>
--- a/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
+++ b/datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
@@ -93,12 +93,12 @@
           <target_class>User</target_class>
           <allow_target_creation>false</allow_target_creation>
         </field>
-        <field id="contact_id" xsi:type="AttributeExternalField">
-          <extkey_attcode>user_id</extkey_attcode>
-          <target_attcode>contactid</target_attcode>
-          <dependencies>
-            <attribute id="user_id"/>
-          </dependencies>
+        <field id="contact_id" xsi:type="AttributeExternalKey">
+          <sql>contact_id</sql>
+          <is_null_allowed>true</is_null_allowed>
+          <on_target_delete>DEL_MANUAL</on_target_delete>
+          <target_class>Person</target_class>
+          <allow_target_creation>false</allow_target_creation>
         </field>
       </fields>
       <methods>
@@ -111,6 +111,7 @@
     {
         $this->SetCurrentDateIfNull('creation_date');
         $this->SetIfNull('user_id', CMDBChange::GetCurrentUserId());
+        $this->SetIfNull('contact_id', UserRights::GetContactId());
 
         return parent::DBInsertNoReload();
     }
--- a/datamodels/2.x/itop-attachments/module.itop-attachments.php
+++ b/datamodels/2.x/itop-attachments/module.itop-attachments.php
@@ -204,7 +204,10 @@ SQL;
 			if (version_compare($sPreviousVersion, '3.2.0', '<')) {
 				SetupLog::Info("Upgrading itop-attachment from '$sPreviousVersion' to '$sCurrentVersion'. Starting with 3.2.0, contact_id will be added into the DB...");
 				$sUserTableName = MetaModel::DBGetTable('User');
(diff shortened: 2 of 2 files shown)


datamodels/2.x/itop-attachments/module.itop-attachments.php
```diff
@@ -194,14 +194,22 @@ public static function AfterDatabaseCreation(Config $oConfiguration, $sPreviousV
...
-				{
+				if ($oContainer) {
 					$oAttachment->SetItem($oContainer, true /*updateonchange*/);
 					$iUpdated++;
 				}
 			}
-
 			SetupLog::Info("Initializing attachment/item_org_id - $iUpdated records have been adjusted");
+
+			if (version_compare($sPreviousVersion, '3.2.0', '<')) {
```
Molkobain: Test whether the attribute exists in the datamodel rather than the version number.

datamodels/2.x/itop-attachments/module.itop-attachments.php
```diff
@@ -194,14 +194,22 @@ public static function AfterDatabaseCreation(Config $oConfiguration, $sPreviousV
...
+				SetupLog::Info("Upgrading itop-attachment from '$sPreviousVersion' to '$sCurrentVersion'. Starting with 3.2.0, contact_id will be added into the DB...");
+				$sUserTableName = MetaModel::DBGetTable('User');
+				$sAddContactId = "UPDATE `$sTableName` att, `$sUserTableName` us SET att.contact_id = us.contactid WHERE att.user_id = us.id AND att.contact_id = 0";
```
Molkobain: Get column name from the attribute definition instead of hardcoding it as they can differ between the datamodel and the database.

We need to check with others if `\MetaModel::GetAttributeDefinition()->Get('sql')` is the right method to do so.

Molkobain (changes requested):

Molkobain (approved):
Good job!

datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
```diff
@@ -110,8 +110,14 @@
...
     {
         $this->SetCurrentDateIfNull('creation_date');
-        $this->SetIfNull('user_id', CMDBChange::GetCurrentUserId());
+        $iUserId = CMDBChange::GetCurrentUserId();
+        if(utils::IsNotNullOrEmptyString($iUserId)){
+               $this->SetIfNull('user_id', $iUserId);
 
+              // Get Contact from user
+              $oUser = MetaModel::GetObject('User', $iUserId);
+              $this->SetIfNull('contact_id', $oUser->Get('contactid'));
```
Hipska, line 119: Isn't there a better (already cached) way to retrieve the current user, or the current contact (id)?

piRGoif: Mind this code can be called from CLI context where it isn't the connected User we want to get...

Molkobain: > Isn't there a better (already cached) way to retrieve the current user, or the current contact (id)?

In that particular case (`CMDBChange::GetCurrentUserId()`) we retrieve the ID of the "real" user (in case of impersonation).
We can't use `UserRights::GetContactId()` as it will retrieve the wrong Contact ID in case of impersonation.

We might need to add the necessary APIs to manipulate the "real" contact as well.
I'm not an expert of that part, it would be good to discuss with @rquetiez as well.

Personally I don't think this is a blocking issue. That pattern is used amongst the app, we can choose to improve its performances but it can be part of another PR with a broader scope.

Molkobain:
Moved the PR to "technical review" as Hispka's point will discussed there.

datamodels/2.x/itop-attachments/module.itop-attachments.php
```diff
@@ -194,14 +194,25 @@ public static function AfterDatabaseCreation(Config $oConfiguration, $sPreviousV
...
-
 			SetupLog::Info("Initializing attachment/item_org_id - $iUpdated records have been adjusted");
+
+			if (MetaModel::GetAttributeDef('Attachment', 'contact_id') instanceof AttributeExternalKey) {
+				SetupLog::Info("Upgrading itop-attachment from '$sPreviousVersion' to '$sCurrentVersion'. Starting with 3.2.0, contact_id will be added into the DB...");
+				$sUserTableName = MetaModel::DBGetTable('User');
+				$sUserFieldContactId = MetaModel::GetAttributeDef('User', 'contactid')->Get('sql');
+				$sAttachmentFieldUserId = MetaModel::GetAttributeDef('Attachment', 'user_id')->Get('sql');
+				$sAttachmentFieldContactId = MetaModel::GetAttributeDef('Attachment', 'contact_id')->Get('sql');
+				$sAddContactId = "UPDATE `$sTableName` att, `$sUserTableName` us SET att.$sAttachmentFieldContactId = us.$sUserFieldContactId WHERE att.$sAttachmentFieldUserId = us.id AND att.$sAttachmentFieldContactId = 0";
```
Molkobain:
```suggestion
				$sAddContactId = "UPDATE `$sTableName` att, `$sUserTableName` us SET att.`$sAttachmentFieldContactId` = us.`$sUserFieldContactId` WHERE att.`$sAttachmentFieldUserId` = us.id AND att.`$sAttachmentFieldContactId` = 0";
```

Molkobain:
Accepted during technical review, Anne-Cath you can merge.

datamodels/2.x/itop-attachments/datamodel.itop-attachments.xml
piRGoif: Discussed during technical review, where Romain was present.

We do have a cached user with \UserRights::GetUser, but this represent the logged in user. 

But with attachment, we can create instances with a different user than the logged one... For example in the Mail To Ticket background process, on each mail processed we are getting the corresponding user and setting it to be used by the history API. So we must use the history API to get the correct user (that's what we are doing with CMDBChange::GetCurrentUserId())

For now only the user id is saved by the history API. 
We think that in Attachment::DBInsertNoReload, getting the User object from the id saved in the history API won't be that bad ?

Molkobain:
For the record, we missed a missing "allow all data" when setting the user_id which causes a regression in the portal. Fixed with [0c3e839](https://github.com/Combodo/iTop/commit/0c3e839381fb70ad7856576de4fc1da17278dfa2)

PR #535, opened 2023-08-04
FAF diff for long text attributes

piRGoif:
Today when browsing an iTop object history (activity panel, "activity" tab), on all text attributes we are displaying the previous value. On long texts this is far from perfect as seeing what was modified is complicated ! It would be better to have a diff instead of just the previous value ! 

Here is an example with a long text : 

image

With this PR the changeop is displayed like this : 
image

=> on any CMDBChangeOpSetAttributeLongText and children you now have a binocular link, that will open a modal window displaying the diff

On the same changeop as the example above the modal will display: 

image

The diff itself is made by the [php-diff](https://github.com/jfcherng/php-diff) lib (827ko with all its dependencies, see /lib/jfcherng)
It is generated by the new AttributeTextDiffController, with this endpoint url: `<itop>pages/UI.php?route=attributetext_diff.display_diff&changeop=49945`

TODO

- [X] Enpoint & diff page
- [x] Add link in activity panel 
  - ~~with a proper UiBlock (neutral button with icon ? (use \CMDBChangeOpSetAttributeLongText::GetDescription + BlockRenderer or EditsEntry directly in its twig file ?)~~ => GetDescription only returns some HTML, so to be discussed during technical review
- [X] Popup (with new popup tooling added in 3.0.0 js/pages/backoffice/toolbox.js ??)
  - [X] Factorize link opening (only the cmdbchangeop id in a data attribute and a global JS as event listener ?)
- [X] make diff colums equal sizes
- [X] Solve error on modal opening : "redeclaration of const aLoadedJsFilesRegister"
- [X] Solve no output of inline CSS in AjaxPage
- ~~prev/next browsing~~ => might not be a good idea as the modal is opened in a content dependent size (and I want to keep it that way !). If the first content opened is small, navigating between changeop won't be very easy :/ 
- [X] Update PR description


piRGoif:
Rebased & force pushed

piRGoif:
Closing PR as I won't lead it to a consensus.

PR #537, opened 2023-08-07
N°938 - Improve print of portal object page and portal dashboard page

accognet:
Fix use of browser print action in portal.
Before PR : 
image
After : 
image
image
image


piRGoif:
Maybe try to avoid page breaks inside some fields like description or a caselog comment ? 
Seems like the [break-inside](https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside) CSS property can do that ?

piRGoif (approved):

Molkobain:
_**Important: During functional review, it was agreed that this PR does not answer to the whole specifications of N°938, it just makes the printable page readable. N°938 is about hiding unnecessary parts of the UI (buttons, pagination, ...) and selecting what to print like in the backoffice.**_

I'm not comfortable with the current implementation for the following reasons, see point 3 for an alternative suggestion.

**Point 1**
The print stylesheet includes a lot of code when the real issues are mostly 2 things:
  * Modal windows displayed stacked, making texts overlapping
    => When printing, we can just choose to unstack them and display them one after another
  * URL of hyperlinks being displayed
    => Hide them

All other things seems (eg. navigation menu) to already be handled by the BS native print rules.

**Point 2**
Adding the print stylesheet as a standalone stylesheet could be a good idea as it allows overloading it by a customizations.
But from my POV it should be included in the "portal" stylesheet:
  * The "portal" stylesheet is about the layout of the portal template, so it makes sense to centralize it here
  * The BS stylesheet already includes its own print rules, so we don't need a dedicated stylesheet

Also, if an overload ever comes necessary, it could always be done by adding it in a misc. stylesheet.

**Point 3**
So my suggestion would be to add the following in the "portal" stylesheet (before the "Shameful area" in iTop 3.1)
```scss
@media print {
	/* Prevent URLs from being displayed */
	a[href], img[src] {
		&::after {
			content: none !important;
		}
	}

	/* Force modals to be displayed one after another instead of stacked */
	.modal {
		&.in {
			position: relative;
			top: unset;
			z-index: unset;
			overflow-y: unset;
		}
		
	}

	#drag_overlay {
		display: none;
	}
}
```

_Example of a page with 3 stacked modals_
image

_Result_
image
image
image

_Example of an aggregate page brick_
image

_Result_
image

Molkobain:
Discussed IRL with Alexandre:
- Ok to retrofit this in support/2.7

Molkobain (changes requested):
See previous comment

PR #539, opened 2023-08-08
N°5170 - In a transition DoCheckToWrite returned error, generes a Fatal error

accognet:
Internal

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -3592,7 +3592,17 @@ EOF
 			$oPage->add_ready_script(InlineImage::EnableCKEditorImageUpload($this, $sTempId));
 		} else {
 			//we can directly apply the stimuli
-			$bApplyStimulus = $this->ApplyStimulus($sStimulus); // will write the object in the DB
+			try {
+				$bApplyStimulus = $this->ApplyStimulus($sStimulus); // will write the object in the DB
+			}
+			catch (Exception $e) {
+				if ($sOwnershipToken !== null) {
+					// Release the concurrent lock, if any
+					iTopOwnershipLock::ReleaseLock($sClass, $iKey, $sOwnershipToken);
+				}
+				throw $e;
+			}
+
 			if (!$bApplyStimulus) {
 				throw new ApplicationException(Dict::S('UI:FailedToApplyStimuli'));
 			} else {
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -3592,26 +3592,26 @@ EOF
 			$oPage->add_ready_script(InlineImage::EnableCKEditorImageUpload($this, $sTempId));
 		} else {
 			//we can directly apply the stimuli
+			$sExceptionMessage = null;
 			try {
-				$bApplyStimulus = $this->ApplyStimulus($sStimulus); // will write the object in the DB
+			$bApplyStimulus = $this->ApplyStimulus($sStimulus); // will write the object in the DB
 			}
-			catch (Exception $e) {
+			catch (Exception $oException) {
+				// Catch any exception happening during the stimulus
+				$bApplyStimulus = false;
+				$sExceptionMessage =   ($oException instanceof CoreCannotSaveObjectException) ? $oException->getHtmlMessage() : $oException->getMessage();
+			}
+			finally {
 				if ($sOwnershipToken !== null) {
 					// Release the concurrent lock, if any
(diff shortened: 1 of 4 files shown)


piRGoif:
Functional review : OK for 3.1.1

pages/UI.php
```diff
@@ -1203,10 +1203,19 @@ function DisplayNavigatorGroupTab($oP)
...
 				}
 				catch (ApplicationException $e) {
+					$bApplyTransition = false;
 					$sMessage = $e->getMessage();
 					$sSeverity = 'info';
 					ReloadAndDisplay($oP, $oObj, 'stimulus', $sMessage, $sSeverity);
 				}
+				catch (CoreCannotSaveObjectException $e) {
+					$bApplyTransition = false;
+					$aIssues = $e->getIssues();
```
Molkobain, line 1216: Why retrieve this? It doesn't seem to be used.

datamodels/2.x/itop-portal-base/portal/src/Form/ObjectFormManager.php
```diff
@@ -1215,6 +1215,13 @@ public function OnSubmit($aArgs = null)
 				}
 			}
 		}
+		catch (CoreCannotSaveObjectException $e) {
+			$aData['valid'] = false;
+			$aData['messages']['error'] += array('_main' => array($e->getHtmlMessage()));
```
Molkobain, line 1227: Are you sure you want the HTML message? It could lead to XSS.

pages/UI.php
```diff
@@ -1203,10 +1203,19 @@ function DisplayNavigatorGroupTab($oP)
...
+					$sMessage = $e->getHtmlMessage();
```
Molkobain, line 1217: Are you sure you want the HTML message? It could lead to XSS.

application/cmdbabstract.class.inc.php
```diff
@@ -3580,7 +3580,17 @@ public function DisplayStimulusForm(WebPage $oPage, $sStimulus, $aPrefillFormPar
...
+				$bApplyStimulus = $this->ApplyStimulus($sStimulus); // will write the object in the DB
+			}
+			catch (Exception $e) {
+				if ($sOwnershipToken !== null) {
+					// Release the concurrent lock, if any
+					iTopOwnershipLock::ReleaseLock($sClass, $iKey, $sOwnershipToken);
+				}
+				throw $e;
+			}
+
```
Molkobain: This duplicates the lock release logic, I would rather try something like the following. But it needs another look to it. (Mind that it factorize lines 3594-3603 as well but I can't modify them in the PR)

```suggestion
			try {
				$bApplyStimulus = $this->ApplyStimulus($sStimulus); // Will write the object in the DB
			}
			catch (Exception $oException) {
				// Catch any exception happening during the stimulus
				$bApplyStimulus = false;
				throw $oException;
			}
			finally {
				if (!$bApplyStimulus) {
					// Throw an application oriented exception if necessary
					throw new ApplicationException(Dict::S('UI:FailedToApplyStimuli'));
				} else {
					if ($sOwnershipToken !== null) {
						// Release the concurrent lock, if any
						iTopOwnershipLock::ReleaseLock($sClass, $iKey, $sOwnershipToken);
					}

				return true;
			}

```

Molkobain (changes requested):
Did you discussed it with Vincent? I'm not sure it doesn't exactly what the spec says.

application/cmdbabstract.class.inc.php
accognet: This code does not have the same behavior as the one I wrote....

datamodels/2.x/itop-portal-base/portal/src/Form/ObjectFormManager.php
accognet: Yes, for CoreCannotSaveObjectException, this is the message that is usually displayed

pages/UI.php
accognet: Yes, for CoreCannotSaveObjectException, this is the message that is usually displayed

pages/UI.php
accognet: CoreCannotSaveObjectException was not catch. Without this catch, we don't see the error message

datamodels/2.x/itop-portal-base/portal/src/Form/ObjectFormManager.php
Molkobain: It is a matter of where the message will consumed / displayed, not how it is usually done for this type of exception.

PR #540, opened 2023-08-17
N°7067 - Add setting to change default "password change" URL

peckpeck:
When we use an LDAP authentication, iTop still displays its own change password URL whereas it doesn't work since it cannot change the users' password.

This PR adds a setting to allow the administrator to specify this URL so that users can click on it in the login form and user the proper form.


jbostoen:
Just wondering, is this a generic service URL or something built-in of your LDAP server? If it's not something standard: wouldn't it make more sense to hide/disable iTop's native "change password" URL altogether in case of certain user type accounts? I suppose it could be something like a static function on the User child classes?

I'm support engineer for a product where the changing of the LDAP password can actually be done in the application as well. Makes for a more integrated solution, but I think it's Kerberos based.

peckpeck:
In my case, it is the URL of change password service configured for our LDAP server. It may also be interesting to allow removing the URL in your case.

Hipska:
This implementation changes the URL for all types of users, so also for `UserLocal` who will then not be able to change their password anymore.

peckpeck:
It indeed does not work for all cases, but it at least works for the case when all users are in LDAP.
We are in the login form, so we don't know yet if the user is in LDAP yet or not since the login is inknown.

Hipska:
Oh, I see.

I would at least also change the URL for the menu item when logged in (if applicable to that User class)
image

application/loginform.class.inc.php
```diff
@@ -140,7 +140,11 @@ public function GetTwigContext()
 		$oLoginContext->AddBlockExtension('login_form_footer', new LoginBlockExtension('extensionblock/loginformfooter.html.twig'));
 
 		$bEnableResetPassword = MetaModel::GetConfig()->Get('forgot_password');
-		$sResetPasswordUrl = utils::GetAbsoluteUrlAppRoot() . 'pages/UI.php?loginop=forgot_pwd';
+		$sResetPasswordUrl = MetaModel::GetConfig()->Get('change_password_url');
+		if ($sResetPasswordUrl == '')
+		{
+			$sResetPasswordUrl = utils::GetAbsoluteUrlAppRoot() . 'pages/UI.php?loginop=forgot_pwd';
+		}
```
Molkobain:
```suggestion
		$sResetPasswordUrl = MetaModel::GetConfig()->Get('forgot_password.url');
		if ($sResetPasswordUrl == '')
		{
			$sResetPasswordUrl = utils::GetAbsoluteUrlAppRoot() . 'pages/UI.php?loginop=forgot_pwd';
		}
```

core/config.class.inc.php
```diff
@@ -1683,6 +1683,14 @@ class Config
...
 			'show_in_conf_sample' => false,
 		],
+		'change_password_url' => [
+			'type'                => 'string',
+			'description'         => 'Set this value to your "change password" service URL if you have an LDAP authentication',
+			'default'             => '',
+			'value'               => '',
+			'source_of_value'     => '',
+			'show_in_conf_sample' => false,
+		],
```
```suggestion
		'forgot_password.url' => [
			'type'                => 'string',
			'description'         => 'Set this value to your "forgot password" service URL if it should be handled out of '.ITOP_APPLICATION_SHORT.'. Note that it will apply to all users (iTop users, LDAP users, ...)',
			'default'             => '',
			'value'               => '',
			'source_of_value'     => '',
			'show_in_conf_sample' => false,
		],
```

Molkobain (changes requested):

core/config.class.inc.php
```diff
@@ -1683,6 +1683,14 @@ class Config
...
 			'show_in_conf_sample' => false,
 		],
+		'forgot_password.url' => [
+			'type'                => 'string',
+			'description'         => 'Set this value to your "forgot password" service URL if it should be handled out of '.ITOP_APPLICATION_SHORT.'. Note that it will apply to all users (iTop users, LDAP users, ...)',
+			'default'             => '',
+			'value'               => '',
+			'source_of_value'     => '',
+			'show_in_conf_sample' => false,
+		],
```
Molkobain: @peckpeck can you move this block of code just below the existing `forgot_password` declaration?

Molkobain:
Technical review:
- Ok with the idea of overloading the URL with a conf parameter, we just renamed the conf. param. name as it is only for the "forgot/reset password" URL (login page), not the "change password" (menu for logged users).
- Guillaume will document in the Wiki that it is not recommended to change it when several types of users are available
- @Hipska , regarding the "change password" url, the link is already displayed or not depending on `\User::CanChangePassword()`, ideally we could have another `\User::ChangePasswordURL()` method, but this is for a dedicated PR.

Hipska (approved):

Molkobain:
Functional review:
  - We have an internal use case: ITSM Designer users which need to be redirected to our support portal for password change
  - Guillaume will check with SaaS teams if they have a similar use case
  - As stated before, this PR only makes sense when there are only LDAP users, we need to discuss how it could be handle for iTops with a mix of users and external OAuth providers like (Google, ...), we'll think about it until next review
  - @peckpeck in the meantime you can set the `forgot_password` parameter to false in order to hide the link (if you haven't done it yet 😁)

Hipska (approved):

Molkobain:
@peckpeck you reverted my suggestions to your original proposition, was that intended? We would like this suggestions to be applied in order to proceed with the PR.

peckpeck:
@Molkobain sorry, I made a mistake, I thought I was pushing to a different branch.
I apologize for the mess

Molkobain:
> @Molkobain sorry, I made a mistake, I thought I was pushing to a different branch. I apologize for the mess

No problem, do you want me to apply the suggestions directly? I was able to restore them on the PR.

peckpeck:
Yes you can apply the suggestions, directly

Molkobain:
Guillaume will add the documentation.

Molkobain:
Hello @peckpeck ,

Thanks again for the contribution, as a thank you, we would like to
- Add your name on the [contributors list](https://github.com/Combodo/iTop?tab=readme-ov-file#contributors), would you like to appear with your fullname or just your nickname?
- Send [you stickers](https://github.com/Combodo/iTop/raw/develop/.doc/contributing-guide/2023.contributing-stickers-side-by-side.png), could you send me your fullname and postal address to guillaume.lajarige at combodo.com ?

PR #541, opened 2023-09-01
N°8442 - Document SynchroAttribute columns

Hipska:
This adds possibility to let data collectors know which column names are available to be synchronised.

Problem:
- Sometimes the column name is not the same as iTop attribute code
- Sometimes an attribute has multiple columns (Blob, ...)

See also counterpart PR https://github.com/Combodo/itop-data-collector-base/pull/37.


Hipska:
Can we have some feedback please?

Hipska:
Anyone?

xtophe38 (approved):

jf-cbd:
Hello @Hipska, thanks for the PR, and sorry for the delay.
Answering here for the current PR and https://github.com/Combodo/itop-data-collector-base/pull/37.
Functionally, your PR is very interesting.
Technically, we'll perform a check and come back to you soon because we have a few concerns about performances (e.g.memory limit used by collector) and interface.

jf-cbd:
Hello Hipska,
In your PR, what is the expected behavior with the existing data ?

Hipska:
As you can see, current items aren't updated yet, but I guess that would be needed of course. First question is if @Combodo would accept this addition at all..

jf-cbd:
If there is no performance issue nor technical problem (we'll continue investigating on that), then yes Combodo is interested in this enhancement :) But we also need to think about what happens with the current data.

Hipska:
Current data should be updated during setup.

Hipska:
@jf-cbd Any update? Should I continue to support this?

jf-cbd:
Hello @Hipska, when I'm trying to create a Synchro Data Source (for example on a document file), I have the following error : 
"The object cannot be saved : Unexpected value for attribute 'Import Columns' (import_columns) : Null not allowed", but I don't have any field to set this attribute. Does it work for with an iTop on your branch ?

Hipska:
Strange, doesn't the `CreateSynchroAtt` method get called?

jf-cbd:
Doesn't look like it

Hipska:
You're right, it seems the code has many duplicates. Fixed with 9dd25ca by de-duplicating them.

Hipska:
Is there any update regarding this? I would start to be able to use this feature..

jf-cbd:
Unfortunately, it has been decided in the technical review (this afternoon) because the choice has been made not to do that in the data (datasynchro). Sorry for telling that so late, after requiring changes from you :/
I created a ticket internally so Combodo will create a webservices that will return the metadata of classes.

Hipska:
@jf-cbd do you have the number of that internal ticket? So I can reference it.

PR #542, opened 2023-09-01
N°6695 - Handle multiline dict entries in portal tooltips

piRGoif:
This bug was submitted in a SourceForge ticket : [#2171](https://sourceforge.net/p/itop/tickets/2171/)

To reproduce add this XML to your iTop instance : it will change the UserRequest.status field tooltip.

```xml
    <dictionaries>
        <dictionary id="EN US">
            <entries>
                <entry id="Class:UserRequest/Attribute:status+" _delta="redefine"><![CDATA[This is the first line.
This is the second line.]]></entry>
            </entries>
        </dictionary>
    </dictionaries>
</itop_design>
```
 
On the admin console you'll get this tooltip :
image

But in the user portal : 
image

With this modification both tooltips will keep the line feed.


piRGoif:
Technical review : we have in combodo-approval-extended dict key containing br, check if they are handled ok !
In iTop there are already some dict keys with line feeds.

piRGoif:
For dict keys with `<br>`, the result is the same in both portal and console : the `<br>` is displayed as is.
Example : 
image

piRGoif:
Functional review : 
- Search which key(s) contains a `<br>`
- Multiple multiline dict keys are currently added by Vincent, so yes should be fixed !
- Check which behavior if CrLf or Cr instead of Lf ?

Technical review to be done with Guillaume and Stephen.

piRGoif:
Of all the dict keys containing `<br>`, none are "+" keys which are displayed in a tooltip, so I think we are OK.

piRGoif:
Please @steffunky and/or @Molkobain can you have a look and check if the portal hack implementation is OK for you ?

datamodels/2.x/itop-portal-base/portal/public/css/portal.scss
```diff
@@ -1803,6 +1803,12 @@ table .group-actions .item-action-wrapper .panel-body > p:last-child{
 	text-decoration: line-through;
 }
 
+
+/* Tippy customizations */
```
Molkobain:
```suggestion
/* Tippy: Handle multi-line content */
```

Molkobain (changes requested):
Fix seems good.
Where did you see a lot of changes in the compiled file? I only see what you added.

piRGoif:
> Where did you see a lot of changes in the compiled file? I only see what you added.

Indeed in the PR there is only the added scss code as a difference. But in rocket chat I called for help, the message is still there ? Maybe I did report the modification manually ?

Molkobain:
> Indeed in the PR there is only the added scss code as a difference. But in rocket chat I called for help, the message is still there ? Maybe I did report the modification manually ?

Discussed on IM, issue not reproduced. We let it go for now as it have been a particular dev env.

PR #543, opened 2023-09-07
N°4511 - CKEditor : links are made on all the leading text in Firefox

accognet:
On Firefox, the text after @mention is the wrong color.


piRGoif:
Seems like you did a CK Editor update, which version and why ?

accognet:
Yes, it's an update to version 4.23.0 in order to have the param "followingSpace"

Molkobain:
This has been fixed with CKEditor 5 in the `support/3.2` branch.

@v-dumas @BenGrenoble should we consider to fix it in the `support/3.1` branch or do we close this PR and the corresponding ticket?

BenGrenoble:
This is fixed with ckeditor5 in 3.2.
We won't migrate ckeditor4 in 3.1 to avoid any regression.

piRGoif:
Hello @Molkobain :) 
There is a "rejected" label that would better fit with the decision made than the "invalid" label you did set ?

Molkobain:
You are right, just changed it

PR #544, opened 2023-09-08
N°6660 - Fix define_if_not_exists flag not working on class nodes

accognet:
_delta=define_if_not_existsnow works on the class in the delta.xml file

### DIFF
--- a/setup/modelfactory.class.inc.php
+++ b/setup/modelfactory.class.inc.php
@@ -747,7 +747,7 @@ class ModelFactory
 				$oTargetNode = $oTarget->ImportNode($oSourceNode, true);
 				$sSearchId = $oSourceNode->hasAttribute('_rename_from') ? $oSourceNode->getAttribute('_rename_from') : $oSourceNode->getAttribute('id');
 				if ($oSourceNode->tagName === 'class') {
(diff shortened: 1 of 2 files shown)


rquetiez:
I am clearly opposed to that proposal:
1) There is no test case corresponding to the failing case (I've just written them anyhow)
2) There is no test case ensuring that other types of flag will still work fine for "class" node (and it will be costly to do so)
3) I have written a much more simple change in LoadDelta (one line affected), which corresponds to the mindset in which that part of the code has been designed (unfortunately, it has already been ignored)

Hipska:
@rquetiez Can you point to the branch or commit hash with your much more simple change?

Hipska:
@rquetiez? In the meantime, this bug is still not fixed 😞

Molkobain:
Rejected during support PRs review:
  * This is being reworked within the "technical debt" project on the compiler, this issue won't exist anymore as classes will be like any other XML node.
  * Approach wasn't following the technical standards

Molkobain:
Done with https://github.com/Combodo/iTop/commit/90458f604833fc97ea016460a4d38a089d8afe6f

PR #545, opened 2023-09-12
5491 brokenui dictformat

odain-cbd:
**Issue:**
iTop UI can explode anywhore because of unprotected Dict::Format API. it depends on language used and dictionaries label to display...

in details such a case can occur at any time translation is not safe to vsprint PHP API. for example if there are not enough placeholders provided in dictionary:

**Example**: labelkey = "this label breaks UI without current fix: %1$"

**Fix**
in this PR a try/catch has been added to avoid any UI white page (worst UX effect ever). in that case an error log is printed as well to identify which label/language has to be fixed. in UI the label key will be displayed.

we also added a testcase to identify all broken dictionaries to fix them as soon as possible...

### DIFF
--- a/core/dict.class.inc.php
+++ b/core/dict.class.inc.php
@@ -3,7 +3,7 @@
 //
 //   This file is part of iTop.
 //
-//   iTop is free software; you can redistribute it and/or modify	
+//   iTop is free software; you can redistribute it and/or modify
 //   it under the terms of the GNU Affero General Public License as published by
 //   the Free Software Foundation, either version 3 of the License, or
 //   (at your option) any later version.
@@ -18,7 +18,7 @@
 
 /**
  * Class Dict
- * Management of localizable strings 
+ * Management of localizable strings
  *
  * @copyright   Copyright (C) 2010-2018 Combodo SARL
  * @license     http://opensource.org/licenses/AGPL-3.0
@@ -145,6 +145,24 @@ class Dict
 	 */
 	public static function S($sStringCode, $sDefault = null, $bUserLanguageOnly = false)
 	{
+		$aInfo = self::GetLabelAndLangCode($sStringCode, $sDefault, $bUserLanguageOnly);
+		if (array_key_exists('label', $aInfo)){
+			return $aInfo['label'];
+		}
+		return $sDefault;
+	}
+
+	/**
+	 * Returns a localised string from the dictonary with its associated lang code
+	 *
+	 * @param string $sStringCode The code identifying the dictionary entry
+	 * @param string $sDefault Default value if there is no match in the dictionary
+	 * @param bool $bUserLanguageOnly True to allow the use of the default language as a fallback, false otherwise
+	 *
+	 * @return array with localized label string and used lang code
+	 */
+	private static function GetLabelAndLangCode($sStringCode, $sDefault = null, $bUserLanguageOnly = false)
+	{
 		// Attempt to find the string in the user language
 		//
 		$sLangCode = self::GetUserLanguage();
@@ -154,23 +172,23 @@ class Dict
 		{
 			IssueLog::Warning("Cannot find $sLangCode in dictionnaries. default labels displayed");
 			// It may happen, when something happens before the dictionaries get loaded
-			return $sStringCode;
+			return [ 'label' => $sStringCode, 'lang' => $sLangCode ];
 		}
 		$aCurrentDictionary = self::$m_aData[$sLangCode];
 		if (is_array($aCurrentDictionary) && array_key_exists($sStringCode, $aCurrentDictionary))
 		{
-			return $aCurrentDictionary[$sStringCode];
+			return [ 'label' => $aCurrentDictionary[$sStringCode], 'lang' => $sLangCode ];
 		}
 		if (!$bUserLanguageOnly)
 		{
 			// Attempt to find the string in the default language
 			//
 			self::InitLangIfNeeded(self::$m_sDefaultLanguage);
-			
+
 			$aDefaultDictionary = self::$m_aData[self::$m_sDefaultLanguage];
 			if (is_array($aDefaultDictionary) && array_key_exists($sStringCode, $aDefaultDictionary))
 			{
-				return $aDefaultDictionary[$sStringCode];
+				return [ 'label' => $aDefaultDictionary[$sStringCode], 'lang' => self::$m_sDefaultLanguage ];
 			}
 			// Attempt to find the string in english
 			//
@@ -179,17 +197,17 @@ class Dict
 			$aDefaultDictionary = self::$m_aData['EN US'];
 			if (is_array($aDefaultDictionary) && array_key_exists($sStringCode, $aDefaultDictionary))
 			{
-				return $aDefaultDictionary[$sStringCode];
+				return [ 'label' => $aDefaultDictionary[$sStringCode], 'lang' => 'EN US' ];
 			}
 		}
 		// Could not find the string...
 		//
 		if (is_null($sDefault))
 		{
-			return $sStringCode;
+			return [ 'label' => $sStringCode, 'lang' => null ];
 		}
 
-		return $sDefault;
+		return [ 'label' => $sDefault, 'lang' => null ];
 	}
 
 
@@ -205,19 +223,36 @@ class Dict
 	 */
 	public static function Format($sFormatCode /*, ... arguments ....*/)
 	{
-		$sLocalizedFormat = self::S($sFormatCode);
+		$aInfo = self::GetLabelAndLangCode($sFormatCode);
+		if (array_key_exists('label', $aInfo)){
+			$sLocalizedFormat = $aInfo['label'];
+		} else {
+			$sLocalizedFormat = $sFormatCode;
+		}
+
+		if (array_key_exists('lang', $aInfo)){
+			$sLangCode = $aInfo['lang'];
+		} else {
+			$sLangCode = null;
+		}
+
 		$aArguments = func_get_args();
 		array_shift($aArguments);
-		
+
 		if ($sLocalizedFormat == $sFormatCode)
 		{
 			// Make sure the information will be displayed (ex: an error occuring before the dictionary gets loaded)
 			return $sFormatCode.' - '.implode(', ', $aArguments);
 		}
 
-		return vsprintf($sLocalizedFormat, $aArguments);
+		try{
+			return vsprintf($sLocalizedFormat, $aArguments);
+		} catch(\Exception $e){
+			\IssueLog::Error("Broken label", null, ["sFormatCode" => $sFormatCode, "sLangCode" => $sLangCode ]);
+			return $sFormatCode.' - '.implode(', ', $aArguments);
+		}
 	}
-	
+
 	/**
 	 * Initialize a the entries for a given language (replaces the former Add() method)
 	 * @param string $sLanguageCode Code identifying the language i.e. 'FR-FR', 'EN-US'
@@ -227,7 +262,7 @@ class Dict
 	{
 		self::$m_aData[$sLanguageCode] = $aEntries;
 	}
-	
+
 	/**
 	 * Set the list of available languages
 	 * @param hash $aLanguagesList
@@ -288,7 +323,7 @@ class Dict
 		{
 			$sDictFile = APPROOT.'env-'.utils::GetCurrentEnvironment().'/dictionaries/'.str_replace(' ', '-', strtolower($sLangCode)).'.dict.php';
 			require_once($sDictFile);
-			
+
 			if (self::GetApcService()->function_exists('apc_store')
 				&& (self::$m_sApplicationPrefix !== null))
 			{
@@ -298,7 +333,7 @@ class Dict
 		}
 		return $bResult;
 	}
-	
+
 	/**
 	 * Enable caching (cached using APC)
 	 * @param string $sApplicationPrefix The prefix for uniquely identiying this iTop instance
@@ -342,14 +377,14 @@ class Dict
 			}
 		}
 	}
-	
+
 	public static function MakeStats($sLanguageCode, $sLanguageRef = 'EN US')
 	{
 		$aMissing = array(); // Strings missing for the target language
 		$aUnexpected = array(); // Strings defined for the target language, but not found in the reference dictionary
 		$aNotTranslated = array(); // Strings having the same value in both dictionaries
 		$aOK = array(); // Strings having different values in both dictionaries
-	
+
 		foreach (self::$m_aData[$sLanguageRef] as $sStringCode => $sValue)
 		{
 			if (!array_key_exists($sStringCode, self::$m_aData[$sLanguageCode]))
@@ -357,7 +392,7 @@ class Dict
 				$aMissing[$sStringCode] = $sValue;
 			}
 		}
-	
+
 		foreach (self::$m_aData[$sLanguageCode] as $sStringCode => $sValue)
 		{
 			if (!array_key_exists($sStringCode, self::$m_aData[$sLanguageRef]))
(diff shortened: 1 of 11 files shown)


core/dict.class.inc.php
```diff
@@ -144,6 +144,24 @@ public static function Exists($sStringCode)
 	 * @return string
 	 */
 	public static function S($sStringCode, $sDefault = null, $bUserLanguageOnly = false)
+	{
+		$aInfo = self::GetLabelAndLangCode($sStringCode, $sDefault, $bUserLanguageOnly);
+		if (array_key_exists('label', $aInfo)){
+			return $aInfo['label'];
+		}
```
Hipska: Looking at the code of `GetLabelAndLangCode`, the return will always contain `label`..

core/dict.class.inc.php
```diff
@@ -205,19 +223,36 @@ public static function S($sStringCode, $sDefault = null, $bUserLanguageOnly = fa
...
 		{
 			// Make sure the information will be displayed (ex: an error occuring before the dictionary gets loaded)
 			return $sFormatCode.' - '.implode(', ', $aArguments);
 		}
 
-		return vsprintf($sLocalizedFormat, $aArguments);
+		try{
+			return vsprintf($sLocalizedFormat, $aArguments);
+		} catch(\Throwable $e){
+			\IssueLog::Error("Broken label", null, ["sFormatCode" => $sFormatCode, "sLangCode" => $sLangCode ]);
```
Hipska: Maybe also include the actual error message to help debugging/fixing?

core/dict.class.inc.php
```diff
@@ -205,19 +223,36 @@ public static function S($sStringCode, $sDefault = null, $bUserLanguageOnly = fa
...
+			$sLocalizedFormat = $aInfo['label'];
+		} else {
+			$sLocalizedFormat = $sFormatCode;
+		}
+
+		if (array_key_exists('lang', $aInfo)){
+			$sLangCode = $aInfo['lang'];
+		} else {
+			$sLangCode = null;
+		}
```
Hipska: The keys are always existing..
```suggestion
		['label' => $sLocalizedFormat, 'lang' => $sLangCode] = self::GetLabelAndLangCode($sFormatCode);
```

core/dict.class.inc.php
```diff
@@ -144,6 +144,24 @@ public static function Exists($sStringCode)
...
+	}
+
+	/**
+	 * Returns a localised string from the dictonary with its associated lang code
+	 *
+	 * @param string $sStringCode The code identifying the dictionary entry
+	 * @param string $sDefault Default value if there is no match in the dictionary
+	 * @param bool $bUserLanguageOnly True to allow the use of the default language as a fallback, false otherwise
+	 *
+	 * @return array with localized label string and used lang code
```
piRGoif: Please document using the array shapes syntax (https://wiki.combodo.com/doku.php?id=latest:customization:coding_standards#arrays_with_named_keysarray_shapes)

core/dict.class.inc.php
piRGoif: +1 ! This method content should be rewritten ? Maybe return another key than label for error cases ?

core/dict.class.inc.php
```diff
@@ -154,23 +172,23 @@ public static function S($sStringCode, $sDefault = null, $bUserLanguageOnly = fa
 		{
 			IssueLog::Warning("Cannot find $sLangCode in dictionnaries. default labels displayed");
```
piRGoif: This should say that we cannot find this dict key in any language

core/dict.class.inc.php
piRGoif: +1

core/dict.class.inc.php
piRGoif: I would change the error message for something like "Cannot format dict key with existing placeholders"
Also I agree with Thomas about the error message, as we saw different PHP versions can have different behaving. You may use ExceptionLog ? (only exists starting with support/3.0 for now, but it may be added in support/2.7)

tests/php-unit-tests/src/BaseTestCase/ItopTestCase.php
```diff
@@ -222,6 +222,21 @@ public function InvokeNonPublicMethod($sObjectClass, $sMethodName, $oObject, $aA
...
+	 * @param string $sProperty
+	 *
+	 * @return mixed property
+	 *
+	 * @throws \ReflectionException
+	 * @since 2.7.10 3.1.0
+	 */
+	public function GetNonPublicStaticProperty(string $sClass, string $sProperty)
+	{
+		$oProperty = $this->GetProperty($sClass, $sProperty);
```
piRGoif, line 236: What will be the behavior if the attribute isn't static ? I'm afraid it won't crash and have unpredictable results (depends on PHP version) ? Maybe test that before doing anything ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+				"aPlaceHolders" => ['1'],
+			],
+		];
+	}*/
+
+	/**
+	 * @dataProvider VsprintfProvider
+	 */
+	/*public function testVsprintf($sLabelTemplate, $aPlaceHolders){
+		try{
```
piRGoif, line 233: instead of commenting out all this code, add a markTestSkipped with a clear message please (it's nice to keep this method for evaluating results with different PHP versions, and document uses cases causing errors)

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * iTop is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ * You should have received a copy of the GNU Affero General Public License
+ */
+
```
piRGoif: Check our internal wiki to have the proper copyright header

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+			'traduction ok' => [
+				'sTemplate' => 'toto %1$s titi',
+				'sExpectedTraduction' => 'toto 1 titi',
+			],
+		];
+	}
+
+	/**
+	 * @since 2.7.10 N°5491 - Inconsistent dictionary entries regarding arguments to pass to Dict::Format
+	 * Dict::Format
```
piRGoif: Is this line a typo ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+	 * @dataProvider FormatProvider
+	 */
+	public function testFormat($sTemplate, $sExpectedTraduction){
```
piRGoif: $sExpectedTranslation instead :)

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+		$sLangCode = \Dict::GetUserLanguage();
```
piRGoif, line 46: In this context there are no real user connected, so I found this confusing ? Maybe set clearly the default value (\Dict::$m_sDefaultLanguage) ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+		$aDictByLang = $this->GetNonPublicStaticProperty(\Dict::class, 'm_aData');
+		$sDictKey = 'ITOP::DICT:FORMAT:BROKEN:KEY';
+
+		if (! is_null($sTemplate)){
+			if (array_key_exists($sLangCode, $aDictByLang)){
+				$aDictByLang[$sLangCode][$sDictKey] = $sTemplate;
+			} else {
+				$aDictByLang[$sLangCode] = [$sDictKey => $sTemplate];
```
piRGoif: Can't the assignment be always done the same way ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
piRGoif: Document (with @param ?) that if null is set to $sTemplate that means we won't set the dict key value

piRGoif: I would rename the method name to a less generic name ? Something like testFormatWithOneArgumentAndCustomKey ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+		} else if ((false !== strpos($sLabel, "%s"))
+			|| (false !== strpos($sLabel, "%d"))
+		){
+			$iMaxIndex = 1;
+		}
+
+		return $iMaxIndex;
+	}
+
+	public function LangCodeProvider(){
```
piRGoif, line 111: See DictionariesConsistencyAfterSetupTest::testDictEntryValues
Between the var in this method and the language lists done during setup, we should pick one of those ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+		$this->SetNonPublicStaticProperty(\Dict::class, 'm_aData', $aDictByLang);
+
+		$this->assertEquals($sExpectedTraduction, \Dict::Format($sDictKey, "1"));
+	}
+
+	private function ReadDictKeys($sLangCode) : array {
+		\Dict::InitLangIfNeeded($sLangCode);
+
+		$aDictEntries = $this->GetNonPublicStaticProperty(\Dict::class, 'm_aData');
+		\Dict::InitLangIfNeeded($sLangCode);
```
piRGoif: Why do we make the call twice ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+			}
+		}
+
+		$this->SetNonPublicStaticProperty(\Dict::class, 'm_aData', $aDictByLang);
+
+		$this->assertEquals($sExpectedTraduction, \Dict::Format($sDictKey, "1"));
+	}
+
+	private function ReadDictKeys($sLangCode) : array {
+		\Dict::InitLangIfNeeded($sLangCode);
```
piRGoif, line 62: This method will read compiled dict keys in the current env. We are in a ItopTestCase and not an ItopDataTestCase so I doubt we will have enough data ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+
+		$aKeyArgsCountMap = [];
+		$aKeyArgsCountMap[$sReferenceLangCode] = $this->GetKeyArgCountMap($aReferenceLangDictEntry);
+		//$aKeyArgsCountMap[$sCode] = $this->GetKeyArgCountMap($aDictEntry);
+
+		//set user language
+		$this->SetNonPublicStaticProperty(\Dict::class, 'm_sCurrentLanguage', $sCode);
+
+		$aMismatchedKeys = [];
+		$aLabelCodeNotToCheck = [
```
piRGoif: To ease maintenance, maybe extract this list to a new class ? Like what was done with \Combodo\iTop\Composer\iTopComposer::ListAllowedTestDir / ListDeniedTestDir ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+			if (in_array($sKey, $aLabelCodeNotToCheck)){
+				//false positive: do not test
+				continue;
+			}
+
+				/*if ($iExpectedNbOfArgs === 0){
+				//there is a good chance Dict:S is called, not Dict::Format
+				//avoid to raise false positive broken traductions
+				continue;
+				}*/
```
piRGoif: Why is this commented code kept ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+			'zh' => [ 'ZH CN' ],
+		];
+	}
+
+	/**
+	 * compare en and other dictionaries and check that for all labels there is the same number of arguments
+	 * if not Dict::Format could raise an exception for some languages. translation should be done again...
+	 * @dataProvider LangCodeProvider
+	 */
+	public function testDictEntryValues($sCode)
```
piRGoif: $sCode is quite generic. Maybe rename to something like $sLanguageCodeToTest ?

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
```diff
@@ -0,0 +1,255 @@
...
+						'label key' => $sKey,
+						'expected nb of args' => $iExpectedNbOfArgs,
+						$sCode => $aDictEntry[$sKey],
+						"label value in $sReferenceLangCode" => $aReferenceLangDictEntry[$sKey],
+					]);
+				}
+			}
+		}
+
+		$sErrorMsg = sprintf("%s broken propertie(s) on $sCode dictionaries!", $iCount);
```
piRGoif: Why doing var_dump of the details instead of adding the details at the end of the error message ? I fear the var_dump might be hard to spot in a Jenkins failed test report ?

piRGoif (changes requested):
Will be discussed with Olivier tomorrow !

tests/php-unit-tests/src/BaseTestCase/ItopTestCase.php
odain-cbd: this PR starts from support/2.7. there were missing test APIs that exist already in support/3.x.

to avoid painfull merges I will keep these changes as is

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
odain-cbd: added a comment

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
odain-cbd: added a comment

core/dict.class.inc.php
Hipska: This conversation is marked as resolved, but nothing has changed. So not resolved yet..

tests/php-unit-tests/integration-tests/DictionariesConsistencyAfterSetupTest.php
odain-cbd: topic discussed directly and feedback enhanced

core/dict.class.inc.php
odain-cbd: sorry I repaired it.

odain-cbd: this should be fixed now....

odain-cbd:
Thank you to Hispka and Pierre. I took all (or almost all) your comments into account and fixed

PR #547, opened 2023-09-19
N°6964 - Add API to allow modules to register files to include in the backup

larhip:
In case a custom extension write some files into data directory, it might be necessary to include those files in the backup.

Currently it is just possible to add extra files via config entry, but this will always depend on the local config file. It might be interesting to define extra files via an extension as well.

What do you think?


application/applicationextension.inc.php
```diff
@@ -2269,4 +2269,21 @@ public function InitStats();
...
+ * @api
+ */
+interface iBackupExtraFilesExtension
+{
+	/**
+	 * Returns an array of files and directories to be included in the backup
+	 *
+	 * @api
+	 *
+	 * @return array
```
Hipska:
```suggestion
	 * @return string[]
```

Hipska (approved):

application/applicationextension.inc.php
larhip: yep, thank one make sense :-)

Hipska (approved):

setup/backup.class.inc.php
```diff
@@ -268,6 +268,13 @@ protected function PrepareFilesToBackup($sSourceConfigFile, $sTmpFolder, $bSkipS
 		if (MetaModel::GetConfig() !== null) // During unattended install config file may be absent
 		{
 			$aExtraFiles = MetaModel::GetModuleSetting('itop-backup', 'extra_files', []);
+
+			/** @var iBackupExtraFilesExtension $oExtensionInstance */
+			foreach (MetaModel::EnumPlugins('iBackupExtraFilesExtension') as $oExtensionInstance)
+			{
+				$aExtraFiles = array_merge($aExtraFiles, $oExtensionInstance->GetExtraFiles());
+			}
```
Molkobain: You might want to move this outside of the if condition as the file can be saved even though there is no configuration yet no?

Molkobain (commented):
Is it working on your instance? It shouldn't until the interface is declared in `\MetaModel::InitExtensions()`.
When doing so, please add the interface to the unit test `tests/php-unit-tests/unitary-tests/application/applicationextension/ApplicationExtensionTest.php::ExtensionAPIRegisteredAndCalledProvider()` as well.

setup/backup.class.inc.php
larhip: Changed with Commit d16c37b

larhip:
Damn you are right I forgot to add my changes in `metamodel.class.php` 
I just added this one. (see 3851042)

I have also updated the PHP unit test as requested (9ce61eb)

Beside that there was a little issue, in case you would like to include a subdirectory into the backup (fixed with (234c1b5)

One more issue: Could it be possible that you did not run your php unit tests with PHP 8.1? 
Running those test with 8.1 will throw a php warning: ` Private methods cannot be final as they are never overridden by other classes in /var/www/html/tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php on line 90`
So I removed the `Final`flag for those private methods (see 36483fb)
This last commit might be better to do not via this PR since it is not directly related to this topic, what do you think?

Molkobain:
> One more issue: Could it be possible that you did not run your php unit tests with PHP 8.1? Running those test with 8.1 will throw a php warning: ` Private methods cannot be final as they are never overridden by other classes in /var/www/html/tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php on line 90` So I removed the `Final`flag for those private methods (see [36483fb](https://github.com/Combodo/iTop/commit/36483fb00f45d1abdbed43310f14ffb4ccde6951)) This last commit might be better to do not via this PR since it is not directly related to this topic, what do you think?

It's possible yes, I might have ran them on 7.4 at the time. Thanks for pointing it out! I'll handle it :)

application/applicationextension.inc.php
```diff
@@ -2269,4 +2269,21 @@ public function InitStats();
...
      */
     public function LogOperation($oKpiLogData);
-}
\ No newline at end of file
+}
+
+/**
+ * Implement this interface to add files to the backup
+ *
+ * @api
```
Molkobain, line 2289:
```suggestion
 * @api
 * @since 3.2.0
```

Molkobain: Damn, I thought I would be able to apply the suggestion myself, sorry. I forgot the "edit maintainers" was not active on itomig-de fork.

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
Molkobain, line 1: Remove these changes please, it's not part of the PR and I have to fix it from support/2.7 to develop :)

Molkobain (changes requested):

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
larhip: Corresponding commit has been dropped 👍

Molkobain (approved):

tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
Molkobain: Perfect thanks! I fixed the unit test class with [2a6e5b2](https://github.com/Combodo/iTop/commit/2a6e5b21a945538498e2fda2e5ba7e54ffab5a12)

Hipska (approved):

application/applicationextension.inc.php
```diff
@@ -2269,4 +2269,22 @@ public function InitStats();
...
+interface iBackupExtraFilesExtension
+{
+	/**
+	 * Returns an array of files and directories to be included in the backup
+	 *
+	 * @api
+	 *
+	 * @return string[]
+	 */
+	public function GetExtraFiles();
```
Molkobain:
```suggestion
	public function GetExtraFiles(): array;
```

core/metamodel.class.php
```diff
@@ -7658,6 +7658,7 @@ public static function InitExtensions()
 			'iKPILoggerExtension',
 			'ModuleHandlerApiInterface',
 			'iNewsroomProvider',
+			'iBackupExtraFilesExtension',
```
Molkobain:
```suggestion
```

See next comment

Molkobain:
Accepted during technical review:
- Guillaume creates a modernization ticket to split the `applicationextension.inc.php` file into separate files
- Guillaume creates a modernization ticket to migrate usages of `\MetaModel::EnumPlugins()` to `\utils::GetClassesForInterface()`

setup/backup.class.inc.php
```diff
@@ -265,35 +265,44 @@ protected function PrepareFilesToBackup($sSourceConfigFile, $sTmpFolder, $bSkipS
...
 		{
 			$aExtraFiles = MetaModel::GetModuleSetting('itop-backup', 'extra_files', []);
-			foreach($aExtraFiles as $sExtraFileOrDir)
+		}
+
+		/** @var iBackupExtraFilesExtension $oExtensionInstance */
+		foreach (MetaModel::EnumPlugins('iBackupExtraFilesExtension') as $oExtensionInstance)
+		{
+			$aExtraFiles = array_merge($aExtraFiles, $oExtensionInstance->GetExtraFiles());
+		}
```
Molkobain:
```suggestion
		foreach (utils::GetClassesForInterface('iBackupExtraFilesExtension', '', ['[\\\\/]lib[\\\\/]', '[\\\\/]node_modules[\\\\/]', '[\\\\/]tests[\\\\/]']) as $sExtensionClass)
		{
			/** @var iBackupExtraFilesExtension $oExtensionInstance */
			$oExtensionInstance = new $sExtensionClass();
			$aExtraFiles = array_merge($aExtraFiles, $oExtensionInstance->GetExtraFiles());
		}
```

Rather use this method to retrieve classes implementing the API, it has 2 main advantages (and 1 drawback):
- No need to declare the API interface in `\MetaModel::InitExtensions()`
- No need to explicit require class in the `module.xxx.php` file, it can be autoloaded

_The drawback is that you get FQCN of classes implementing the interface instead of an instance of these classes_

tests/php-unit-tests/unitary-tests/application/applicationextension/ApplicationExtensionTest.php
```diff
@@ -163,6 +163,10 @@ public function ExtensionAPIRegisteredAndCalledProvider(): array
 				\iNewsroomProvider::class,
 				static::ENUM_API_CALL_METHOD_ENUMPLUGINS,
 			],
+			\iBackupExtraFilesExtension::class => [
+				\iBackupExtraFilesExtension::class,
+				static::ENUM_API_CALL_METHOD_ENUMPLUGINS,
```
Molkobain:
```suggestion
				static::ENUM_API_CALL_METHOD_GETCLASSESFORINTERFACE,
```

Molkobain (changes requested):

Molkobain:
Unfortunately I can't apply these changes, could you apply them Lars?

Molkobain:
> * Guillaume creates a modernization ticket to split the `applicationextension.inc.php` file into separate files

N°6932

> * Guillaume creates a modernization ticket to migrate usages of `\MetaModel::EnumPlugins()` to `\utils::GetClassesForInterface()`

N°6936

larhip:
> Unfortunately I can't apply these changes, could you apply them Lars?

@Molkobain should be done, could you please check?

setup/backup.class.inc.php
```diff
@@ -265,35 +265,45 @@ protected function PrepareFilesToBackup($sSourceConfigFile, $sTmpFolder, $bSkipS
...
 		}
+
+		$aExtraFiles = [];
 		if (MetaModel::GetConfig() !== null) // During unattended install config file may be absent
 		{
 			$aExtraFiles = MetaModel::GetModuleSetting('itop-backup', 'extra_files', []);
-			foreach($aExtraFiles as $sExtraFileOrDir)
+		}
+
+		foreach (utils::GetClassesForInterface('iBackupExtraFilesExtension', '', ['[\\\\/]lib[\\\\/]', '[\\\\/]node_modules[\\\\/]', '[\\\\/]tests[\\\\/]']) as $sExtensionClass)
```
Molkobain:
```suggestion
		foreach (utils::GetClassesForInterface('iBackupExtraFilesExtension', '', ['[\\\\/]lib[\\\\/]', '[\\\\/]node_modules[\\\\/]', '[\\\\/]test[\\\\/]', '[\\\\/]tests[\\\\/]']) as $sExtensionClass)
```

Molkobain (commented):
Sorry I forgot this change. Then it's ok, it's works like a charm locally.

Molkobain (approved):

Molkobain:
Functional review:
  - Globally ok for the PR, but curious about the concrete use case you have in mind @larhip ?
  - Including huge files could break the backup archive, making it un-restorable. Guillaume will discuss with Stéphane about the checks/protections we would like to be done.

Molkobain:
> * Including huge files could break the backup archive, making it un-restorable. Guillaume will discuss with Stéphane about the checks/protections we would like to be done.

Discussed with Stephane:
  - There is no size limitation of the tar archive itself
  - The limitation is that each file included in the tar cannot exceed 68GB

As the main issue is the SQL dump, we consider that there is no need to check file size. If something needs to be addressed, it would be outside of that PR to handle the SQL dump size correctly.

Molkobain:
Just waiting for Lars use case before merging

larhip:
We are currently working on an extension, which is writing some files into the filesystem those files are related to the itop data in the database. 
Due to that relation it would be pretty clever to include those file into the backup (and restore).

When I got you right the "only" limitations might be that files bigger than 68GB become problematic?

Molkobain:
> We are currently working on an extension, which is writing some files into the filesystem those files are related to the itop data in the database. Due to that relation it would be pretty clever to include those file into the backup (and restore).

Thanks for the feedback

> When I got you right the "only" limitations might be that files bigger than 68GB become problematic?

Yes, we currently have the problem with instances on which the SQL dump (which is included in the backup) exceeds 68GB in which case the archive cannot be opened. 
That can happen on SQL dumps, but it should be very very rare on files.

Hipska:
According to the specs of tar, there is no real file size limitation. Unless you use a version from before 2004?

application/applicationextension.inc.php
```diff
@@ -2269,4 +2269,22 @@ public function InitStats();
...
+/**
+ * Implement this interface to add files to the backup
+ *
+ * @api
+ * @since 3.2.0
+ */
+interface iBackupExtraFilesExtension
+{
+	/**
+	 * Returns an array of files and directories to be included in the backup
```
Molkobain, line 2283:
```suggestion
	 * Returns an array of files and directories relative paths (from iTop root) to be included in the backup
```

Molkobain:
Accepted during functional review:
  - Guillaume will add a protection for file exceeding 68GB, and a conf. param to disable this if necessary
  - Guillaume will ensure that in case of a too large file, the backup fails in the same way as the usual so the admin notices it

setup/setuputils.class.inc.php
larhip, line 1: @Molkobain since this PR is planned for iTop 3.2, could you maybe have a look at that tiny change here again? Could you imagine to include only this line already with iTop 3.1.2? Without this line I am also not able to use the config param extra_files in a proper way

Molkobain: Hello Lars,

I'll check if we can add this line in the 3.1.2 sure :)
The rest of it will only be in 3.2.0 though

Molkobain: Hello Lars,

After looking a bit more into this, I'm not comfortable setiing `777` on a folder.
Eventually we can set 755, but it should not be necessary if your permissions are ok on the current folder. Have you checked this?

larhip: Hi Guillaume,

to be honest I have busted used '0777' since that is the default, which currently means without that change it is also '0777' (https://www.php.net/manual/en/function.mkdir.php). The only difference is that from now the possibility of the recursive creation is missing.

So my intention was to keep the change as small as possible (lower risk of breaking change).
From my point of view I would be fine with '0755' as well, but especially to introduce it into a minor version (3.1.2) it might be smarter to do it in the same way as before?

Maybe it would be also a good idea for a future version (for 3.2.0 not for 3.1.2) to adapt the code and copy the original permissions? So Something like that:

```PHP
mkdir($sDest, 0777, true);
chmod($sDest, fileperms($sSource));
```

Molkobain: 100% agree with all of your answer :))

Molkobain: Done

Molkobain:
Hello Lars,

Could you please rebase your branch "onto" support/3.2 please so I can merge it? I can't do it myself as this fork doesn't allow maintainers to edit.

Thanks

Molkobain:
(If you need a hand, poke me)

larhip:
I recently did so.
(I did it via GitHub UI so changing the base branch leads to a merge commit instead of a rebase. But since commits of this PR will be squashed into one during the final merge is should be fine anyhow)

Molkobain:
> I recently did so. [...]

Oh my bad sorry :)

PR #549, opened 2023-09-22
N°5472 Notification Action objects : add a last executions tab

piRGoif:
🧐 For now we have a "Notifications" (`UI:NotificationsTab` dict key) tab on each object that is used in triggers.
This is done in `\cmdbAbstractObject::DisplayBareRelations` by querying on `EventNotification` classes.

🤔 This is great for end users, to see which notifications were fired on their `UserRequest` for example.
But it would be nice to have the counterpart in `Action` objects for admin users !

✨ This PR adds a new "Last executions" (`Action:last_executions_tab` and `Action:last_executions_tab+` dict keys) ajax tab in `Action` objects.
This will display all the `EventNotification` saved for the current action in the past days (the limit is set in the `notifications.last_executions_days` config parameter, default is 61 days).

Having an ajax tab (not loaded by default) and being able to change the limit in the config are precaution for perf issues.
Note the list can still be be opened in a new window and filtered as wanted thanks to #222. 

image

⚙ Those `EventNotification` instances are created in each `Action::DoExecute` impl. On existing email sending actions impl we have : 
* ActionEmail::DoExecution : `new EventNotificationEmail()`
  - ActionEmailApprovalRequest : parent call
  - ActionEmailUnauthenticatedForm : parent call
* _ActionWebhook::DoExecute : `new EventWebhook();`

The created classes (`EventNotificationEmail`, `EventWebhook`) are children of `EventNotification`.
For now there are no attribute in this class, parent nor children, to tell whereas we have a success, warning or error :/

-----

🚧 TODO
- [x] replace DisplayBlock by `DataTableUIBlockFactory::MakeForResult`
- [x] ensure every notification events (success, errors, warnings) are traced as EventNotification objects => 24/11 description updated
- [x] replace the DisplayBareRelation by a dasboard attribute (asynchronous by default, customizable) => 24/11 `Action` class is defined in PHP, so having a DisplayBareRelation override would be clearer than a new attribute defined in the Init method I think ?
- [x] use an AjaxTab so that we won't have data load every time we open an Action
- [x] use iTop generic router for the endpoint (see ObjectController)
- [x] in the query, use a config parameter to set number of days (replace 2 months hardcoded value)
- [x] tech review
- [x] functional review
- [x] create bug
- [x] merge
- [x] document (config param, what's new)

### DIFF
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -177,6 +177,14 @@ abstract class Action extends cmdbAbstractObject
 	{
 		parent::DisplayBareRelations($oPage, false);
 
+		$this->GenerateLastExecutionsTab($oPage, $bEditMode);
+	}
+
+	/**
+	 * @since 3.2.0 N°5472 method creation
+	 */
+	protected function GenerateLastExecutionsTab(WebPage $oPage, $bEditMode)
+	{
 		if ($bEditMode) {
 			return;
 		}
@@ -184,12 +192,22 @@ abstract class Action extends cmdbAbstractObject
 		$oPage->SetCurrentTab('action_errors', Dict::S('Action:last_executions_tab'), Dict::S('Action:last_executions_tab+'));
 
 		$oFilter = DBObjectSearch::FromOQL(
-			"SELECT EventNotification WHERE action_id = :action_id AND date > DATE_SUB(NOW(), INTERVAL 2 MONTH)",
+			$this->GetLasExecutionsTabQuery(),
 			['action_id' => $this->GetKey()]
 		);
 		$oSet = new DBObjectSet($oFilter, ['date' => false]);
 		$oExecutionsListBlock = DataTableUIBlockFactory::MakeForResult($oPage, 'action_executions_list', $oSet);
 		$oPage->AddUiBlock($oExecutionsListBlock);
+	}
+
+	/**
+	 * @since 3.2.0 N°5472 method creation
+	 */
+	protected function GetLasExecutionsTabQuery(): string
+	{
+		return <<<OQL
+SELECT EventNotification WHERE action_id = :action_id AND date > DATE_SUB(NOW(), INTERVAL 2 MONTH)
+OQL;
 	}
 }
 
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -203,19 +203,26 @@ abstract class Action extends cmdbAbstractObject
 		$sLastExecutionDaysConfigParamName = 'notifications.last_executions_days';
 		$iLastExecutionDays = $oConfig->Get($sLastExecutionDaysConfigParamName);
 
-		if ($iLastExecutionDays <= 0) {
+		if ($iLastExecutionDays < 0) {
 			throw new InvalidConfigParamException("Invalid value for {$sLastExecutionDaysConfigParamName} config parameter. Param desc: " . $oConfig->GetDescription($sLastExecutionDaysConfigParamName));
 		}
 
-		$oFilter = DBObjectSearch::FromOQL(
-			'SELECT EventNotification WHERE action_id = :action_id AND date > DATE_SUB(NOW(), INTERVAL :days DAY)',
-			[
-				'action_id' => $this->GetKey(),
-				'days' => $iLastExecutionDays,
-			]
-		);
-		$oSet = new DBObjectSet($oFilter, ['date' => false]);
-		$oExecutionsListBlock = DataTableUIBlockFactory::MakeForResult($oPage, 'action_executions_list', $oSet, ['panel_title' => Dict::Format('Action:last_executions_tab_panel_title', $iLastExecutionDays)]);
+		$sActionQueryOql = 'SELECT EventNotification WHERE action_id = :action_id';
+		$aActionQueryParams = ['action_id' => $this->GetKey()];
+		if ($iLastExecutionDays > 0) {
+			$sActionQueryOql .= ' AND date > DATE_SUB(NOW(), INTERVAL :days DAY)';
+			$aActionQueryParams['days'] = $iLastExecutionDays;
+			$sActionQueryLimit = Dict::Format('Action:last_executions_tab_limit_days', $iLastExecutionDays);
+		} else {
+			$sActionQueryLimit = Dict::S('Action:last_executions_tab_limit_none');
+		}
+
+		$oActionFilter = DBObjectSearch::FromOQL($sActionQueryOql, $aActionQueryParams);
+		$oSet = new DBObjectSet($oActionFilter, ['date' => false]);
+
+		$sPanelTitle = Dict::Format('Action:last_executions_tab_panel_title', $sActionQueryLimit);
+		$oExecutionsListBlock = DataTableUIBlockFactory::MakeForResult($oPage, 'action_executions_list', $oSet, ['panel_title' => $sPanelTitle]);
+
 		$oPage->AddUiBlock($oExecutionsListBlock);
 	}
 }
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -187,10 +187,6 @@ abstract class Action extends cmdbAbstractObject
 	 */
 	protected function GenerateLastExecutionsTab(iTopWebPage $oPage, $bEditMode)
 	{
-		if ($bEditMode) {
-			return;
-		}
-
 		$sActionLastExecutionsPageUrl = utils::GetAbsoluteUrlAppRoot()
 			. 'pages/ajax.render.php'
 			. '?operation=notification.action.last_execution_content'
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -180,7 +180,7 @@ abstract class Action extends cmdbAbstractObject
 			return;
 		}
 
-		$oPage->SetCurrentTab('action_errors', Dict::S('Action:activity_tab'), Dict::S('Action:activity_tab+'));
+		$oPage->SetCurrentTab('action_errors', Dict::S('Action:last_executions_tab'), Dict::S('Action:last_executions_tab+'));
 
 		$oFilter = DBObjectSearch::FromOQL(
 			"SELECT EventNotification WHERE action_id = :action_id AND date > DATE_SUB(NOW(), INTERVAL 2 MONTH)",
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -204,7 +204,7 @@ abstract class Action extends cmdbAbstractObject
 		$iLastExecutionDays = $oConfig->Get($sLastExecutionDaysConfigParamName);
 
 		if ($iLastExecutionDays <= 0) {
-			throw new InvalidConfigParamException("Invalid value in the {$sLastExecutionDaysConfigParamName} config parameter");
+			throw new InvalidConfigParamException("Invalid value for {$sLastExecutionDaysConfigParamName} config parameter. Param desc: " . $oConfig->GetDescription($sLastExecutionDaysConfigParamName));
 		}
 
 		$oFilter = DBObjectSearch::FromOQL(
--- a/core/action.class.inc.php
+++ b/core/action.class.inc.php
@@ -45,16 +45,16 @@ abstract class Action extends cmdbAbstractObject
 	{
 		$aParams = array
 		(
-			"category" => "grant_by_profile,core/cmdb",
-			"key_type" => "autoincrement",
-			"name_attcode" => "name",
+			"category"                   => "grant_by_profile,core/cmdb",
+			"key_type"                   => "autoincrement",
+			"name_attcode"               => "name",
 			"complementary_name_attcode" => array('finalclass', 'description'),
-			"state_attcode" => "status",
-			"reconc_keys" => array('name'),
-			"db_table" => "priv_action",
-			"db_key_field" => "id",
-			"db_finalclass_field" => "realclass",
-			"style" => new ormStyle("ibo-dm-class--Action", "ibo-dm-class-alt--Action", "var(--ibo-dm-class--Action--main-color)", "var(--ibo-dm-class--Action--complementary-color)", null, '../images/icons/icons8-in-transit.svg'),
+			"state_attcode"              => "status",
+			"reconc_keys"                => array('name'),
+			"db_table"                   => "priv_action",
+			"db_key_field"               => "id",
+			"db_finalclass_field"        => "realclass",
+			"style"                      => new ormStyle("ibo-dm-class--Action", "ibo-dm-class-alt--Action", "var(--ibo-dm-class--Action--main-color)", "var(--ibo-dm-class--Action--complementary-color)", null, '../images/icons/icons8-in-transit.svg'),
 		);
 		MetaModel::Init_Params($aParams);
 		//MetaModel::Init_InheritAttributes();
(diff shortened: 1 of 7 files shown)


Molkobain:
I'm not sure "Activity" tab is relevant, it's confusing, at first I thought it was something in the activity panel.
Why not name the tab "last_executions_tab" like you already did in the dictionary entries?

piRGoif:
Thanks for the feedback, renaming done !

Hipska:
May I ask why it is put as PHP code and not an `AttributeDashboard`? Looks doable for the current use-case and is more extendable.

piRGoif:
> May I ask why it is put as PHP code and not an `AttributeDashboard`? Looks doable for the current use-case and is more extendable.

Woo I missed that, thanks for the tip, I'll try !

piRGoif:
The Action class is defined in PHP, having a DisplayBareRelation override would be clearer than a new attribute defined in the Init method I think ? 
If the class was defined in XML of course the AttributeDashboard syntax would have been much better.

Hipska:
Oh, right. Another case why a lot of the standard classes should be defined in XML instead of PHP 😉

Hipska (approved):

piRGoif:
Documentation asked on 2023-10-27 FAF meeting done
Ready to be reviewed !

piRGoif:
Rebased and force pushed

Molkobain:
Questions for technical review:
- Shoudn't it be an AJAX tab? Amount of notifications sent in 2 months can be huge (take the public log on Bug ou UserRequest in our iTop)
- Should we use a time interval (eg. 1 month) ou a count limit (eg. 100) ?

piRGoif:
There is a time interval (2 months)
On previous discussions we thought having a simple query shouldn't impact perf.
There was an idea to use a dashboard attribute instead,n but as the class is defined in php it should be easier to override methods than an attribute.

Molkobain:
Technical review: Accepted if tab is switched to an ajax tab to limit performance issues.

core/action.class.inc.php
```diff
@@ -167,6 +168,46 @@ protected function DoCheckIfHasTrigger()
...
+	/**
+	 * @since 3.2.0 N°5472 method creation
+	 */
+	protected function GenerateLastExecutionsTab(WebPage $oPage, $bEditMode)
+	{
+		if ($bEditMode) {
+			return;
+		}
+
+		$oPage->SetCurrentTab('action_errors', Dict::S('Action:last_executions_tab'), Dict::S('Action:last_executions_tab+'));
```
Molkobain:
```suggestion
		$oPage->AddAjaxTab('Action:last_executions_tab', XXX, true /* bCache */, Dict::S('Action:last_executions_tab'), AjaxTab::ENUM_TAB_PLACEHOLDER_LIST);
```

For the ajax tab you can try something like this

Molkobain (changes requested):

Molkobain:
Accepted during functional review:
- Ok for 3.2.0
- Change interval to a config. param. and change it to days instead of months. (eg. `"notifications.last_executions_days" = 60`)

piRGoif:
Rebased branch

core/action.class.inc.php
piRGoif: TODO reported in description
Closing conversation

core/action.class.inc.php
```diff
@@ -168,6 +169,60 @@ protected function DoCheckIfHasTrigger()
...
 	}
+
+	/**
+	 * @since 3.2.0 N°5472 method creation
+	 */
+	public function DisplayBareRelations(WebPage $oPage, $bEditMode = false)
+	{
+		parent::DisplayBareRelations($oPage, false);
+
+		if ($oPage instanceof iTopWebPage) {
```
Molkobain, line 181: Can you remind me why this check is necessary? DisplayBareRelations is only called in iTopWebPage isn't it?

core/action.class.inc.php
```diff
@@ -168,6 +169,60 @@ protected function DoCheckIfHasTrigger()
...
+	}
+
+	/**
+	 * @since 3.2.0 N°5472 method creation
+	 */
+	protected function GenerateLastExecutionsTab(iTopWebPage $oPage, $bEditMode)
+	{
+		if ($bEditMode) {
+			return;
+		}
```
Molkobain: I think it would be better to leave the tab in read only when editing the object.
It's less confusing for the user to always have the same tabs, and it can be use by the user to help fix the current action.

core/action.class.inc.php
```diff
@@ -168,6 +169,60 @@ protected function DoCheckIfHasTrigger()
...
+
+		$sActionLastExecutionsPageUrl = utils::GetAbsoluteUrlAppRoot()
```
Molkobain: We should not use these endpoints anymore, you should either use the deprecated iTop Router or the Symfony Router (come see me for the draft conventions)

core/action.class.inc.php
```diff
@@ -168,6 +169,60 @@ protected function DoCheckIfHasTrigger()
...
+	public function GetLastExecutionsTabContent(WebPage $oPage): void
+	{
+		$oConfig = utils::GetConfig();
+		$sLastExecutionDaysConfigParamName = 'notifications.last_executions_days';
+		$iLastExecutionDays = $oConfig->Get($sLastExecutionDaysConfigParamName);
+
+		if (false === is_int($iLastExecutionDays)) {
+			throw new InvalidConfigParamException("Invalid value in the {$sLastExecutionDaysConfigParamName} config parameter");
+		}
+
```
Molkobain: This is unnecessary, the cast is already done by the \Config class
```suggestion
```

dictionaries/en.dictionary.itop.core.php
```diff
@@ -539,6 +539,8 @@
 	'Class:Action/Attribute:finalclass' => 'Action sub-class',
 	'Class:Action/Attribute:finalclass+' => 'Name of the final class',
 	'Action:WarningNoTriggerLinked' => 'Warning, no trigger is linked to the action. It will not be active until it has at least 1.',
+	'Action:last_executions_tab' => 'Last executions',
+	'Action:last_executions_tab_panel_title' => 'Executions of this action in the past %1$s days',
```
Molkobain: Can you put a screenshot where this will be used? This looks more like a "+" entry or a table's subtitle.

pages/ajax.render.php
```diff
@@ -2569,6 +2569,23 @@ function returnFileUrl(iAttId, sAltText, sSecret) {
 				$oPage = $oController->OperationModify();
 				break;
 
+
+			case 'notification.action.last_execution_content':
```
Molkobain: See comment above about routing

Molkobain (changes requested):

core/action.class.inc.php
piRGoif: cmdbAbstractObject::DisplayBareRelations method signature has WebPage as argument type, whereas AddAjaxTab method is defined in iTopWebPage

core/action.class.inc.php
piRGoif: Fixed in 12a7d95e

core/action.class.inc.php
piRGoif: Indeed, in the \Config::Set method
Test changed (actually we need to reject values <= 0) in d06a3d4e

dictionaries/en.dictionary.itop.core.php
piRGoif: Before implementing the AjaxTab, the second key was indeed a "+" key
But you can't specify "+" key to iTopWebPage::AddAjaxTab :(
So I changed the key to be displayed as a datatable panel title, see \Action::GetLastExecutionsTabContent

piRGoif: I've updated the description screenshot

pages/ajax.render.php
piRGoif: closing conversation to keep only one open on the same subject

core/config.class.inc.php
```diff
@@ -1619,6 +1619,14 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
+		'notifications.last_executions_days' => [
+			'type' => 'integer',
+			'description' => 'Number of days to display in the Action\'s last executions tab',
+			'default' => 61,
+			'value' => 61,
```
Hipska: What a strange default value. Why not `60`?

Hipska (approved):
I'm still under the impression it better would have been a dashboard attribute. That would have solved also a few of the conversations in here..

core/config.class.inc.php
Molkobain: It triggered me also, I think it's because 2 months is most of the time 30+31 days, but it's odd yes 😁

core/action.class.inc.php
piRGoif: Done in 494a376b
Can you review this commit ? 
I'm closing this conversation, don't hesitate to do a new review

core/config.class.inc.php
piRGoif: Apart of February, we have months that last 30 or 31 days, so I did 30 + 31 as a default value to reflect the previous "2 months" value

core/action.class.inc.php
```diff
@@ -168,6 +170,54 @@ protected function DoCheckIfHasTrigger()
...
+	 */
+	public function GetLastExecutionsTabContent(WebPage $oPage): void
+	{
+		$oConfig = utils::GetConfig();
+		$sLastExecutionDaysConfigParamName = 'notifications.last_executions_days';
+		$iLastExecutionDays = $oConfig->Get($sLastExecutionDaysConfigParamName);
+
+		if ($iLastExecutionDays <= 0) {
+			throw new InvalidConfigParamException("Invalid value in the {$sLastExecutionDaysConfigParamName} config parameter");
+		}
```
Molkobain:
```suggestion
		if ($iLastExecutionDays < 0) {
			throw new InvalidConfigParamException("{$sLastExecutionDaysConfigParamName} config. parameter must be a positive integer");
		}
```

In functional review, we discussed that 0 would mean no time limit.
You'll need to adjust the OQL as well.

sources/Controller/Notifications/ActionController.php
```diff
@@ -0,0 +1,48 @@
...
+		if (is_null($sActionId)) {
+			throw new CoreUnexpectedValue($sCannotLoadActionErrorMessage);
+		}
+
+		$oAction = MetaModel::GetObject(Action::class, $sActionId, false);
+		if (is_null($oAction)) {
+			throw new CoreException($sCannotLoadActionErrorMessage);
+		}
+
+		$oPage = new NiceWebPage(Dict::S('UI:BrowseInlineImages'));
```
Molkobain: Why `NiceWebPage` instead of `AjaxWebPage`, and why this dict entry?!

core/action.class.inc.php
```diff
@@ -168,6 +170,54 @@ protected function DoCheckIfHasTrigger()
...
+	}
+
+	/**
+	 * @since 3.2.0 N°5472 method creation
+	 */
+	protected function GenerateLastExecutionsTab(iTopWebPage $oPage, $bEditMode)
+	{
+		$oRouter = Router::GetInstance();
+		$sActionLastExecutionsPageUrl = $oRouter->GenerateUrl('notifications.action.last_executions_tab', ['actionid' => $this->GetKey()]);
```
Molkobain:
```suggestion
		$sActionLastExecutionsPageUrl = $oRouter->GenerateUrl('notifications.action.last_executions_tab', ['action_id' => $this->GetKey()]);
```

The convention is to put an underscore.

sources/Controller/Notifications/ActionController.php
```diff
@@ -0,0 +1,48 @@
...
+	/**
+	 * @throws CoreException if cannot load the Action object
+	 * @throws CoreUnexpectedValue if `actionid` parameter is invalid
+	 * @since 3.2.0 N°5472 creation
+	 */
+	public function OperationLastExecutionsTab()
+	{
+		$sActionId = utils::ReadParam('actionid', null, false);
+		$sCannotLoadActionErrorMessage = __METHOD__ . ': invalid actionid parameter';
+		if (is_null($sActionId)) {
```
Molkobain:
```suggestion
		$sActionId = utils::ReadParam('action_id', null, false);
		$sCannotLoadActionErrorMessage = __METHOD__ . ': invalid actionid parameter';
		if (utils::IsNullOrEmptyString($sActionId)) {
```

You need to handle `""` as well

dictionaries/en.dictionary.itop.core.php
Molkobain: Thanks, makes sense now

Molkobain (changes requested):

core/action.class.inc.php
piRGoif: Done in 4862977d

sources/Controller/Notifications/ActionController.php
piRGoif: Done in 4862977d

core/config.class.inc.php
piRGoif: Documented in d24ea5f0

Hipska: Yeah okay, but we are talking about days here, not months. 61 is such an odd default number. In days I would have chosen 30, 50 or 60.

core/action.class.inc.php
piRGoif: I've changed the exception message with 5b1ff2c3
I used the param description to avoid duplicate. 
We might create a dedicated method in the Config class to check value... But it's out of scope of this PR.

piRGoif: Woops sorry correct commit is c5f6c47c

sources/Controller/Notifications/ActionController.php
piRGoif: Fixed in f18c2e1e

core/action.class.inc.php
piRGoif: 0 as config param value : cdb5556a

core/config.class.inc.php
```diff
@@ -1619,6 +1619,14 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
+		'notifications.last_executions_days' => [
+			'type' => 'integer',
+			'description' => 'Number of days to display in the Action\'s last executions tab',
```
```suggestion
			'description' => 'Number of days to display in the Action\'s last executions tab (0 means no limit)',
```

Maybe we could precise in the description the behavior of `0`?

Molkobain (approved):

piRGoif:
Tech review OK, to check one last time in functional review

piRGoif:
Functional review ok
ready to merge !

core/config.class.inc.php
Hipska: @piRGoif You are forgetting December/January and July/August. So this calculation is wrong 4 months a year.

Hipska:
Please have a look https://github.com/Combodo/iTop/pull/549#discussion_r1489110960

core/config.class.inc.php
piRGoif: Yes you're perfectly right, 2 months isn't always the same as 61 days.
During the functional review this point was raised and everybody was ok with the default value anyway.
Closing conversation

Hipska: Okay then 😂

PR #555, opened 2023-10-05
N°6648 - Use \utils::GetDataPath() instead of hard-coded paths

Molkobain:
**Objective**
Stop duplicating an hard-coded path to the "<ITOP>/data" folder (the "writable" folder for caches, extensions needs, ...) by using a helper function `\utils::GetDataCache()` similar to what as been done for other folders.

Note that the helper was introduced a while back, this PR just makes a bit of cleanup by using it everywhere it can.

### DIFF
--- a/core/simplegraph.class.inc.php
+++ b/core/simplegraph.class.inc.php
@@ -500,17 +500,17 @@ EOF
 		if (file_exists($sDotExecutable))
 		{
 			// create the file with Graphviz
-			if (!is_dir(APPROOT."data"))
+			if (!is_dir(utils::GetDataPath()))
 			{
-				@mkdir(APPROOT."data");
+				@mkdir(utils::GetDataPath());
 			}
-			if (!is_dir(APPROOT."data/tmp"))
+			if (!is_dir(utils::GetDataPath()."tmp"))
 			{
-				@mkdir(APPROOT."data/tmp");
+				@mkdir(utils::GetDataPath()."tmp");
 			}
-			$sImageFilePath = tempnam(APPROOT."data/tmp", 'png-');
+			$sImageFilePath = tempnam(utils::GetDataPath()."tmp", 'png-');
 			$sDotDescription = $this->GetDotDescription();
-			$sDotFilePath = tempnam(APPROOT."data/tmp", 'dot-');
+			$sDotFilePath = tempnam(utils::GetDataPath()."tmp", 'dot-');
 		
 			$rFile = @fopen($sDotFilePath, "w");
 			@fwrite($rFile, $sDotDescription);
@@ -556,17 +556,17 @@ EOF
 		if (file_exists($sDotExecutable))
 		{
 			// create the file with Graphviz
-			if (!is_dir(APPROOT."data"))
+			if (!is_dir(utils::GetDataPath()))
 			{
-				@mkdir(APPROOT."data");
+				@mkdir(utils::GetDataPath());
 			}
-			if (!is_dir(APPROOT."data/tmp"))
+			if (!is_dir(utils::GetDataPath()."tmp"))
 			{
-				@mkdir(APPROOT."data/tmp");
+				@mkdir(utils::GetDataPath()."tmp");
 			}
-			$sXdotFilePath = tempnam(APPROOT."data/tmp", 'xdot-');
+			$sXdotFilePath = tempnam(utils::GetDataPath()."tmp", 'xdot-');
 			$sDotDescription = $this->GetDotDescription(true); // true => don't put (localized) labels in the file, since it makes it harder to parse
-			$sDotFilePath = tempnam(APPROOT."data/tmp", 'dot-');
+			$sDotFilePath = tempnam(utils::GetDataPath()."tmp", 'dot-');
 	
 			$rFile = @fopen($sDotFilePath, "w");
 			@fwrite($rFile, $sDotDescription);
--- a/pages/graphviz.php
+++ b/pages/graphviz.php
@@ -111,17 +111,17 @@ $sDotExecutable = MetaModel::GetConfig()->Get('graphviz_path');
 if (file_exists($sDotExecutable))
 {
 	// create the file with Graphviz
-	$sImageFilePath = APPROOT."data/lifecycle/".$sClass.".svg";
-	if (!is_dir(APPROOT."data"))
+	$sImageFilePath = utils::GetDataPath()."lifecycle/".$sClass.".svg";
+	if (!is_dir(utils::GetDataPath()))
 	{
-		@mkdir(APPROOT."data");
+		@mkdir(utils::GetDataPath());
 	}
-	if (!is_dir(APPROOT."data/lifecycle"))
+	if (!is_dir(utils::GetDataPath()."lifecycle"))
 	{
-		@mkdir(APPROOT."data/lifecycle");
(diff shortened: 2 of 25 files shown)


pages/graphviz.php
```diff
@@ -111,17 +111,17 @@ function GraphvizLifecycle($sClass)
...
 {
 	// create the file with Graphviz
-	$sImageFilePath = APPROOT."data/lifecycle/".$sClass.".svg";
-	if (!is_dir(APPROOT."data"))
+	$sImageFilePath = utils::GetDataPath()."lifecycle/".$sClass.".svg";
+	if (!is_dir(utils::GetDataPath()))
 	{
-		@mkdir(APPROOT."data");
+		@mkdir(utils::GetDataPath());
 	}
```
Hipska, line 118: Why would data path not exist? Looks like a useless check to me.

core/simplegraph.class.inc.php
```diff
@@ -500,17 +500,17 @@ public function DumpAsHtmlImage()
 		if (file_exists($sDotExecutable))
 		{
 			// create the file with Graphviz
-			if (!is_dir(APPROOT."data"))
+			if (!is_dir(utils::GetDataPath()))
 			{
-				@mkdir(APPROOT."data");
+				@mkdir(utils::GetDataPath());
 			}
```
Hipska, line 506: Same here, this directory should always exist?

core/simplegraph.class.inc.php
```diff
@@ -556,17 +556,17 @@ public function DumpAsXDot()
...
```
Hipska, line 562: Same..

setup/applicationinstaller.class.inc.php
```diff
@@ -658,7 +658,7 @@ protected static function DoCompile($aSelectedModules, $sSourceDir, $sExtensionD
 		}
 		
 		// Set an "Instance UUID" identifying this machine based on a file located in the data directory
-		$sInstanceUUIDFile = APPROOT.'data/instance.txt';
+		$sInstanceUUIDFile = utils::GetDataPath().'instance.txt';
 		SetupUtils::builddir(APPROOT.'data');
```
Hipska:
```suggestion
		SetupUtils::builddir(utils::GetDataPath());
```

Hipska (approved):

core/simplegraph.class.inc.php
piRGoif: Depending on the way the package is made, it may not. We should have a more robust way of handling "iTop system" dirs, but it's out of scope of this PR I think ?

pages/graphviz.php
Molkobain: I think it should always be there, I don't know about the check. But I kept the PR focus around the usage of the API, otherwise there is a lot to do and it will never pass 😁

core/simplegraph.class.inc.php
Molkobain: > [...], but it's out of scope of this PR I think ?

Same opinion, I kept the PR focus on using the new API.

core/simplegraph.class.inc.php
Molkobain: _See reply above_

piRGoif:
Thanks for the proposition ! 

Reading all of those string concat, I would suggest having the data subpath as a function parameter ? That would improve readability, as string concats are hard to understand ? 
So, something like `utils::GetDataPath('bulk_export/'.$sToken.'.status')` ?

Also as Thomas pointed out there are multiple tests on data dir existing. We may add one in GetDataPath() ?

setup/applicationinstaller.class.inc.php
piRGoif: Good catch !

core/simplegraph.class.inc.php
Hipska: Could you elaborate on in which situation the data dir might not exist? If I understand I will agree, otherwise there is no point in updating useless code but rather just remove it as cleanup of APPROOT code is exactly the point of this PR 😉

piRGoif: We had a similar bug on the /log dir multiple times in the last year. Either because working on a branch or a fork where all the file contents were removed from the index (an empty dir in git is considered as just no dir at all), or because we had to remove the dir in the build tool recipe. It doesn't occurs very often, but when it does it is VERY frustrating ;o)
It's a kind of defensive programming : iTop shouldn't crash just because some dirs it needs aren't present on the FS... 

Having a centralized way to get those "iTop system" dirs would allow to add such checks... But I share Guillaume fear that having multiple problems handled in the same PR is the best way for the PR to never be merged :/

Hipska: Okay, but then you would expect the check to be on every occurrence, no?

As you are suggesting this in https://github.com/Combodo/iTop/pull/555#issuecomment-1750177986

Hipska:
> Reading all of those string concat, I would suggest having the data subpath as a function parameter ? That would improve readability, as string concats are hard to understand ? So, something like `utils::GetDataPath('bulk_export/'.$sToken.'.status')` ?

That is still a string concat, and the name of the method isn't correct anymore, as in this case you are not getting the data path, but a path to a file.
 
> Also as Thomas pointed out there are multiple tests on data dir existing. We may add one in GetDataPath() ?

That would be fine for me 👍

Hipska (approved):

piRGoif:
Right, another method could be added, like GetPathToDataResource ?
Biggest advantage of having the parameter is that it is enclosed in brackets and the IDE clearly shows that...

Molkobain:
> Reading all of those string concat, I would suggest having the data subpath as a function parameter ? That would improve readability, as string concats are hard to understand ? So, something like `utils::GetDataPath('bulk_export/'.$sToken.'.status')` ?

This PR is just about using the API, not creating a new one / modifying an existing.

Regarding your example I don't think it's a good idea, it would just duplicate `'bulk_export/'`, instead we should have dedicated helpers for each "working" directory. That way any feature / extension can use them without knowing where it actually is on the FS.

core/simplegraph.class.inc.php
piRGoif: I have mixed feelings. Having proper checks would be better handled in another PR...

Hipska: By the way, `data/` is actually part of the repository. So somebody should already had removed it explicitly in order for it to be not existing.. That way you could also check for existence of `core/`, `application/` and so on.

pages/graphviz.php
Molkobain: Can we agree that it is out of this PR's scope?

core/simplegraph.class.inc.php
Molkobain: Can we agree that it is out of this PR's scope?

Hipska: Sure, if you want to keep updating useless code.. It would have been a good opportunity to get rid of it.

(I already approved the PR btw)

piRGoif (approved):
Discussed with Guillaume : we will keep this pr as simple as possible. Some iterations to be done, but later...

Molkobain:

Molkobain:
Accepted functional review, for 3.3.0

piRGoif:
Ready for merge now that the support/3.2 branch exists

PR #355, opened 2022-10-18
N°6016 MissingDependencyException : better text message

piRGoif:
With #280 we improved the message displayed to the user in the setup screens when a module on disk hasn't got all of its dependencies present. The displayed message was based on the MissingDependencyException attributes.

This PR aims to improve the MissingDependencyException message as well : it will benefit to unattended install for example !

Before we were getting this in the log (direct MissingDependencyException message printing) : 
```
Error: The following modules have unmet dependencies: Database maintenance tools (id: combodo-db-tools/3.1.0) depends on itop-structure/3.0.0, Links between virtualization and storage (id: itop-bridge-virtualization-storage/3.1.0) depends on itop-storage-mgmt/2.2.0 + itop-virtualization-mgmt/2.2.0
```

After : 
```
Error: The following modules have unmet dependencies:
Database maintenance tools (id: combodo-db-tools/3.1.0) depends on itop-structure/3.0.0,
Links between virtualization and storage (id: itop-bridge-virtualization-storage/3.1.0) depends on itop-storage-mgmt/2.2.0 + itop-virtualization-mgmt/2.2.0
```

PR #557, opened 2023-10-17
N°6849 - Enhance setup error message in case of unmet module dependen…

odain-cbd:
​When installing an iTop package you may have following error message by CLI setup (almost the same via UI wizard):

```
 Error: The following modules have unmet dependencies:
 iTop XXX  (id: itop-www/1.0.8) depends on: itop-xxx/0.0.1 + itop-yyy/3.0.0 + itop-zzz/1.0.1,
 Combodo YYY (id: combodo-www/1.0.5) depends on: combodo-xxx/1.0.0 + combodo-yyy/2.3.3 
 Encountered stopper issues. Aborting...
```

In that case we could help administrators/developers if iTop output describes which modules are really broken among all dependencies. For that purpose current PR would enhance same output by adding ✅ / ❌ icons :

- in iTop setup wizard screen

image

- in iTop CLI setup output (unattended_install.php)

```
Error: The following modules have unmet dependencies:
iTop XXX  (id: itop-www/1.0.8) depends on: ❌ itop-xxx/0.0.1+ ✅itop-yyy/3.0.0 + ✅ itop-zzz/1.0.1,
Combodo YYY (id: combodo-www/1.0.5) depends on: ❌ combodo-xxx/1.0.0 + ❌ combodo-yyy/2.3.3
...
```

### DIFF
--- a/setup/modulediscovery.class.inc.php
+++ b/setup/modulediscovery.class.inc.php
@@ -220,8 +220,8 @@ class ModuleDiscovery
 	public static function OrderModulesByDependencies($aModules, $bAbortOnMissingDependency = false, $aModulesToLoad = null)
 	{
 		// Order the modules to take into account their inter-dependencies
-		$aDependencies = array();
-		$aSelectedModules = array();
+		$aDependencies = [];
+		$aSelectedModules = [];
 		foreach($aModules as $sId => $aModule)
 		{
 			list($sModuleName, ) = self::GetModuleName($sId);
@@ -232,7 +232,7 @@ class ModuleDiscovery
 			}
 		}
 		ksort($aDependencies);
-		$aOrderedModules = array();
+		$aOrderedModules = [];
 		$iLoopCount = 1;
 		while(($iLoopCount < count($aModules)) && (count($aDependencies) > 0) )
 		{
@@ -256,23 +256,24 @@ class ModuleDiscovery
 		}
 		if ($bAbortOnMissingDependency && count($aDependencies) > 0)
 		{
-			$aModulesInfo = array();
-			$aModuleDeps = array();
+			$aModulesInfo = [];
+			$aModuleDeps = [];
 			foreach($aDependencies as $sId => $aDeps)
 			{
 				$aModule = $aModules[$sId];
+				$aDepsWithIcons = [];
 				foreach($aDeps as $sIndex => $sDepId)
 				{
 					if (self::DependencyIsResolved($sDepId, $aOrderedModules, $aSelectedModules))
 					{
-						$aDeps[$sIndex] = '✅ ' . $sDepId;
+						$aDepsWithIcons[$sIndex] = '✅ ' . $sDepId;
 					} else
 					{
-						$aDeps[$sIndex] = '❌ ' .  $sDepId;
+						$aDepsWithIcons[$sIndex] = '❌ ' .  $sDepId;
 					}
 				}
-				$aModuleDeps[] = "{$aModule['label']} (id: $sId) depends on: ".implode(' + ', $aDeps);
-				$aModulesInfo[$sId] = array('module' => $aModule, 'dependencies' => $aDeps);
+				$aModuleDeps[] = "{$aModule['label']} (id: $sId) depends on: ".implode(' + ', $aDepsWithIcons);
+				$aModulesInfo[$sId] = array('module' => $aModule, 'dependencies' => $aDepsWithIcons);
 			}
 			$sMessage = "The following modules have unmet dependencies:\n".implode(",\n", $aModuleDeps);
 			$oException = new MissingDependencyException($sMessage);
--- a/setup/modulediscovery.class.inc.php
+++ b/setup/modulediscovery.class.inc.php
@@ -76,7 +76,7 @@ class ModuleDiscovery
 		'doc.manual_setup' => 'url',
 		'doc.more_information' => 'url',
 	);
-	
+
 
 	// Cache the results and the source directories
 	protected static $m_aSearchDirs = null;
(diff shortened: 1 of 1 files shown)


piRGoif:
Thanks, this is a very good complement to #355 !

May I suggest having a message like : 

```
Error: The following modules have unmet dependencies:
    iTop www (id: itop-www/1.0.8) dependencies: 
        ❌ UNMET: itop-xxx/0.0.1
        ✔ OK: itop-yyy/3.0.0 + itop-zzz/1.0.1
    Combodo YYY (id: combodo-www/1.0.5) dependencies: 
        ❌ UNMET: combodo-xxx/1.0.0, combodo-yyy/2.3.3
```

Hipska:
Having it on multiple lines makes it already a lot easier to understand/find the issue! I was also thinking of using "missing" instead of "unmet", but maybe that can cause confusion when there is an older version of the required module installed.

odain-cbd:
> Thanks, this is a very complement to #355 !
> 
> May I suggest having a message like :
> 
> ```
> Error: The following modules have unmet dependencies:
>     iTop www (id: itop-www/1.0.8) dependencies: 
>         ❌ UNMET: itop-xxx/0.0.1
>         ✔ OK: itop-yyy/3.0.0 + itop-zzz/1.0.1
>     Combodo YYY (id: combodo-www/1.0.5) dependencies: 
>         ❌ UNMET: combodo-xxx/1.0.0, combodo-yyy/2.3.3
> ```

good idea. 

in CLI mode I propose to have only UNMET word (not OK) before broken dependencies. to reduce the amount of stuff to read (users hate to read things)

in UI mode i propose to replace UNMET/OK by the cross/checkmark icon. but without any label to avoid any further translation. which will add process and interactions/inertia with our international community for small gain.

odain-cbd:
> Having it on multiple lines makes it already a lot easier to understand/find the issue! I was also thinking of using "missing" instead of "unmet", but maybe that can cause confusion when there is an older version of the required module installed.

what about BROKEN or KO for both unmet/missing dependencies?

Hipska:
No, I did like your idea of only using the Emoji instead of the words "unmet" and "ok".

piRGoif:
Actually, do we even need to display the whole dependencies list ? Only the missing ones should be enough ??

odain-cbd:
> Actually, do we even need to display the whole dependencies list ? Only the missing ones should be enough ??

this is a bigger change that should be discussed internally as we remove some useful information. personally I like to have these information.

it could even be ambiguous. for a module that has only one broken dependency among 3. does it mean the broken displayed one is lonely (all others are ok)? or our iTop installation stopped at the very first missing one?

Molkobain:
Very good idea indeed!

  * I'm also in favor of using just the emojis in the UI, adding "OK / UNMEET" just makes the UI more difficult to read.
  * Having no emojis and only the "UNMET" work in CLI is fine by me as well.
  * I'm not sure of what the actual output is, but keeping multilines makes it a lot easier to read, so we should keep that as well.

Could you add a screenshot of both CLI / UI rendering? Thanks

piRGoif:
>> Actually, do we even need to display the whole dependencies list ? Only the missing ones should be enough ??
> 
> for a module that has only one broken dependency among 3. does it mean the broken displayed one is lonely (all others are ok)? or our iTop installation stopped at the very first missing one?

Oh yes, I can imagine why you're saying that as we saw weird behaviors in the past (like the XML errors during aggregation, I suppose you have this in mind ;) )

But the obvious behavior would be to check all the dependency graph right ? Maybe it is just a matter of word choosing ?
What do you think of : 

```
ERROR: there are unmet modules in the dependencies graph
    iTop www (id: itop-www/1.0.8) is missing: ❌itop-xxx/0.0.1
    Combodo YYY (id: combodo-www/1.0.5) is missing: ❌combodo-xxx/1.0.0, ❌combodo-yyy/2.3.3
```

> Having no emojis and only the "UNMET" work in CLI is fine by me as well.

I think it would be fine yes, as most consoles / terminals are unicode now right ? Do you fear any compatibility issues Olivier ?

> Could you add a screenshot of both CLI / UI rendering? Thanks

Yes please update the description, for the old message (yours is on one line whereas it isn't true since 8 months and the merge of #355) , as of the new message

odain-cbd:
> > > Actually, do we even need to display the whole dependencies list ? Only the missing ones should be enough ??
> > 
> > 
> > for a module that has only one broken dependency among 3. does it mean the broken displayed one is lonely (all others are ok)? or our iTop installation stopped at the very first missing one?
> 
> Oh yes, I can imagine why you're saying that as we saw weird behaviors in the past (like the XML errors during aggregation, I suppose you have this in mind ;) )
> 
> But the obvious behavior would be to check all the dependency graph right ? Maybe it is just a matter of word choosing ? What do you think of :
> 
> ```
> ERROR: there are unmet modules in the dependencies graph
>     iTop www (id: itop-www/1.0.8) is missing: ❌itop-xxx/0.0.1
>     Combodo YYY (id: combodo-www/1.0.5) is missing: ❌combodo-xxx/1.0.0, ❌combodo-yyy/2.3.3
> ```
> 
> > Having no emojis and only the "UNMET" work in CLI is fine by me as well.
> 
> I think it would be fine yes, as most consoles / terminals are unicode now right ? Do you fear any compatibility issues Olivier ?
> 
> > Could you add a screenshot of both CLI / UI rendering? Thanks
> 
> Yes please update the description, for the old message (yours is on one line whereas it isn't true since 8 months and the merge of #355) , as of the new message

PR description has been modified accordingly. including also a UI wizard screenshot with the emoj

Hipska (approved):

Molkobain (approved):

piRGoif (changes requested):
Instead of modifying the `aDeps` array, maybe create a new variable ? This is always a risk to change an existing variable :/

Also I still believe showing only missing dependencies is better, but this will be discussed during reviews.

Molkobain (approved):

PR #558, opened 2023-10-20
N°6808 - Rank management (order) in iTop actions

accognet:
Internal

### DIFF
--- a/core/trigger.class.inc.php
+++ b/core/trigger.class.inc.php
@@ -114,11 +114,15 @@ abstract class Trigger extends cmdbAbstractObject
 
 		// Find the related actions
 		$oLinkedActions = $this->Get('action_list');
+
+		// Order actions as expected
 		$aActionListOrdered = [];
 		while ($oLink = $oLinkedActions->Fetch()) {
 			$aActionListOrdered[(int) $oLink->Get('order')][] = $oLink;
 		}
 		ksort($aActionListOrdered);
+		
+		// Execute actions
 		foreach ($aActionListOrdered as $aActionSubList) {
 			foreach ($aActionSubList as $oLink) /** @var \DBObject $oLink */ {
 				/** @var \DBObject $oLink */
--- a/core/trigger.class.inc.php
+++ b/core/trigger.class.inc.php
@@ -116,7 +116,7 @@ abstract class Trigger extends cmdbAbstractObject
 		$oLinkedActions = $this->Get('action_list');
 		$aActionListOrdered = [];
 		while ($oLink = $oLinkedActions->Fetch()) {
-			$aActionListOrdered[$oLink->Get('order')][] = $oLink;
+			$aActionListOrdered[(int) $oLink->Get('order')][] = $oLink;
 		}
 		ksort($aActionListOrdered);
 		foreach ($aActionListOrdered as $aActionSubList) {
--- a/core/trigger.class.inc.php
+++ b/core/trigger.class.inc.php
@@ -114,17 +114,22 @@ abstract class Trigger extends cmdbAbstractObject
 
 		// Find the related actions
 		$oLinkedActions = $this->Get('action_list');
-		while ($oLink = $oLinkedActions->Fetch())
-		{
-			/** @var \DBObject $oLink */
-			$iActionId = $oLink->Get('action_id');
-			/** @var \Action $oAction */
-			$oAction = MetaModel::GetObject('Action', $iActionId);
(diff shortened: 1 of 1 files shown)


Molkobain:
Instead of sorting element in PHP, did you check with @rquetiez if we can specify the sort order on the `lnkTriggerAction` class (`/itop_design/classes/class/properties/order`) so elements are sorted when retrieved from the DB?

accognet:
I have tested by adding the "order" attribute to the lnk and the sorting is not used.

Molkobain:
Accepted during functional review, but we need to check whether the different actions are executed in sequence (wait for the previous one to be done before the next one is executed) or if they are in //. It should be in sequence but there was a confusion during the meeting.

Implementation will be discussed during next technical review.

Molkobain:
Accepted during technical review once sort is done naturally to avoid falling in the usual pitfall "1, 10, 2" order.

> Accepted during functional review, but we need to check whether the different actions are executed in sequence (wait for the previous one to be done before the next one is executed) or if they are in //. It should be in sequence but there was a confusion during the meeting.

Also, actions are executed sequentially.

core/trigger.class.inc.php
```diff
@@ -113,17 +113,22 @@ public function DoActivate($aContextArgs)
...
-			if ($oAction->IsActive())
-			{
-                $oKPI = new ExecutionKPI();
-				$oAction->DoExecute($this, $aContextArgs);
-                $oKPI->ComputeStatsForExtension($oAction, 'DoExecute');
+		$aActionListOrdered = [];
+		while ($oLink = $oLinkedActions->Fetch()) {
+			$aActionListOrdered[$oLink->Get('order')][] = $oLink;
+		}
+		ksort($aActionListOrdered);
```
Molkobain, line 120: Use `natsort()` to avoid actions being executed like "1, 10, 2"

core/trigger.class.inc.php
```diff
@@ -113,17 +113,22 @@ public function DoActivate($aContextArgs)
...
-			/** @var \Action $oAction */
-			$oAction = MetaModel::GetObject('Action', $iActionId);
-			if ($oAction->IsActive())
-			{
-                $oKPI = new ExecutionKPI();
-				$oAction->DoExecute($this, $aContextArgs);
-                $oKPI->ComputeStatsForExtension($oAction, 'DoExecute');
+		$aActionListOrdered = [];
+		while ($oLink = $oLinkedActions->Fetch()) {
+			$aActionListOrdered[$oLink->Get('order')][] = $oLink;
```
Molkobain:
```suggestion
			$aActionListOrdered[(int) $oLink->Get('order')][] = $oLink;
```

To ensure that order is always returned as an int, so ksort work even if Get() starts returning strings.

Molkobain (approved):

core/trigger.class.inc.php
```diff
@@ -113,17 +113,22 @@ public function DoActivate($aContextArgs)
...
-			/** @var \Action $oAction */
-			$oAction = MetaModel::GetObject('Action', $iActionId);
-			if ($oAction->IsActive())
-			{
-                $oKPI = new ExecutionKPI();
-				$oAction->DoExecute($this, $aContextArgs);
-                $oKPI->ComputeStatsForExtension($oAction, 'DoExecute');
+		$aActionListOrdered = [];
+		while ($oLink = $oLinkedActions->Fetch()) {
+			$aActionListOrdered[(int) $oLink->Get('order')][] = $oLink;
```
Hipska, line 121: ~~Erm, I always thought `order` could be any numeric, so also float?~~

Nevermind, I was confusing with `<rank>`.

core/trigger.class.inc.php
```diff
@@ -113,17 +113,22 @@ public function DoActivate($aContextArgs)
...
+		}
+		ksort($aActionListOrdered);
+		foreach ($aActionListOrdered as $aActionSubList) {
```
Molkobain:
```suggestion

		// Order actions as expected
		$aActionListOrdered = [];
		while ($oLink = $oLinkedActions->Fetch()) {
			$aActionListOrdered[(int) $oLink->Get('order')][] = $oLink;
		}
		ksort($aActionListOrdered);
		
		// Execute actions
		foreach ($aActionListOrdered as $aActionSubList) {
```

Molkobain:
All good! You can:
- Rebase branch on support/3.2
- Make sure target branch is support/3.2
- Squash and merge (mind to keep the PR number in the title)

PR #564, opened 2023-10-31
N°6887 - Fix excessive OQL requests to display user's grant matrix

accognet:
1200 times the same request for display the grant matrix... replaced by only 1 time, it's better :-)


Hipska:
Sounds great, but what about just add caching to the `UserRights::ListProfiles` method?

PR #565, opened 2023-10-31
N°6890 - Update request template data though the REST/JSON API

accognet:
internal
tests added in https://github.com/Combodo/itop-request-template/tree/feature/6890-Update_request_template_REST


rquetiez:
Reviewed today with Anne-Cath:

1. Missing new test cases (see request-template)
2. The implementation in MakeRealValue implies a change in the expected behavior (at least for the REST API) : it is not possible to reset the values. A PO decision has to be made, Then tests cases have to be added to cover these use cases.
3. The implementation in CompareValues needs a rework : first, it will issue a warning when some fields are missing on the right value, and the meaning of the change is really unclear. CompareValues should be responsible for comparing PHP arrays, not for compensating the tricky behavior of the above layers.

rquetiez (changes requested):
See my comment on June, the 10th

PR #568, opened 2023-11-07
N°6901 - Enable tracking of iTop active sessions (monitoring)

odain-cbd:
### Goal

Current PR aims at being able at any time to know the amount of user currently active on iTop.
Design

For that purpose a dedicated iTop session handler extends default one. it keeps default implementation by calling each parent method first.
and then a specific iTop implementation is triggered to have local session files in data/sessions folder.

**session file content**
each session file is either empty when iTop session is made without any user authentified (status.php for example).
otherwise it contains below data:

-  login_mode: authentication mode (among UI classical 'form', 'token', 'ldap', sso called 'hybrid-auth', .....)
- user_id: logged in user ID
- creation_time: timestamp of the PHP session creation (used to compute session elapsed active time in seconds)
- context: iTop session context tags are aggregated with ':' separators (empty when simple login or pick among ContextTag values like REST/Synchro/....)

Examples:

- user UI access:

`{"login_mode":"form","user_id":"1","creation_time":1699371169,"context":""}`

- Rest API call (via token)

`{"login_mode":"token","user_id":"4","creation_time":1699373876,"context":"REST\/JSON"}`

- Synchro API call (via token as well)
`{"login_mode":"token","user_id":"6","creation_time":1699373876,"context":"Synchro"}`

**Cleanup**
Garbage collecting of all local session files (in data/sessions folder) is made via a dedicated Background task.


sources/Application/Helper/iTopSessionHandler.php
Molkobain, line 1: Move this to `sources/SessionTracker/SessionHandler.php`

core/cmdbsource.class.inc.php
```diff
@@ -518,7 +518,12 @@ public static function Quote($value, $bAlways = false, $cQuoteStyle = "'")
 		if ($bAlways || is_string($value))
 		{
 			/** @noinspection NullPointerExceptionInspection this shouldn't be called with un-init DB */
-			$value = $cQuoteStyle.DbConnectionWrapper::GetDbConnection()->real_escape_string($value).$cQuoteStyle;
+			$mysqli = DbConnectionWrapper::GetDbConnection();
```
Molkobain: As this change is not really necessary for the PR, but fixed a test case, we'll see if we keep it during PR review.

sources/Application/Helper/iTopSessionGarbageCollector.php
Molkobain, line 1: Move this to `sources/SessionTracker/SessionTrackerGC.php`

- No need to prefix with iTop
- Other GC task are suffixed with "GC" instead of "GarbageCollector"

lib/composer/autoload_classmap.php
Molkobain, line 1: Mind to redump autoloaders after changes below.

sources/Application/Helper/iTopSessionHandler.php
```diff
@@ -0,0 +1,181 @@
...
+use utils;
+
+/**
+ * Class iTopSessionGarbageCollector
+ *
+ * @author Olivier Dain <olivier.dain@combodo.com>
+ * @package Combodo\iTop\Application\Helper
+ * @since 3.1.1 3.2.0
+ */
+class iTopSessionHandler extends \SessionHandler
```
Molkobain: Put overrided methods first, then custom methods.

sources/Application/Helper/iTopSessionHandler.php
```diff
@@ -0,0 +1,181 @@
...
+					$iCreationTime = $aJson['creation_time'];
+				}
+			}
+
+			return json_encode (
+				[
+					'login_mode' => Session::Get('login_mode'),
+					'user_id' => $sUserId,
+					'creation_time' => $iCreationTime,
+					'context' => implode(":", ContextTag::GetStack())
```
Molkobain: Don't use ":" as it is already used in some context tags such as cron tasks, portal instances, synchro DS, ...
Discussed this morning, using "|" seems safe.

sources/Application/Helper/iTopSessionHandler.php
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	public function list_session_files() : array
+	{
+		clearstatcache();
+		if (! is_dir(utils::GetDataPath() . "sessions")) {
+			@mkdir(utils::GetDataPath() . "sessions");
+		}
+
+		return glob(utils::GetDataPath() . "sessions/session_**");
```
Molkobain:
```suggestion
		return glob(utils::GetDataPath() . "sessions/session_*");
```

webservices/cron.php
```diff
@@ -250,7 +250,7 @@ function CronExec($oP, $bVerbose, $bDebug=false)
 					$sStatus = $oTask->Get('status');
 					$sLastRunDate = $oTask->Get('latest_run_date');
 					$sNextRunDate = $oTask->Get('next_run_date');
-					$oP->p(sprintf('| %1$-25.25s | %2$-7s | %3$-19s | %4$-19s |', $sTaskName, $sStatus, $sLastRunDate, $sNextRunDate));
+					$oP->p(sprintf('| %1$-50.50s | %2$-7s | %3$-19s | %4$-19s |', $sTaskName, $sStatus, $sLastRunDate, $sNextRunDate));
```
Molkobain:
```suggestion
					$oP->p(sprintf('| %1$-25.25s | %2$-7s | %3$-19s | %4$-19s |', $sTaskName, $sStatus, $sLastRunDate, $sNextRunDate));
```

My bad

Molkobain (changes requested):
Discussed this morning with Olivier and Eric:
- I commited some changes
- Olivier will do the others

sources/Application/Helper/iTopSessionHandler.php
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	public function open($save_path, $session_name)
+	{
+		$bRes = parent::open($save_path, $session_name);
+
+		$session_id = session_id();
+		IssueLog::Debug("open($session_id)");
```
Molkobain: All logs need to be more verbose: At least tell that you are doing something about the sessions. Better, add a channel and context data.

Molkobain:
_Looks like my notes for this morning review got sent with the ones I just added now. You might have received them by mail, I deleted the old ones, just mind the ones still on the PR_

core/cmdbsource.class.inc.php
odain-cbd: just reverted

odain-cbd:
> Discussed this morning with Olivier and Eric:
> 
>     * I commited some changes
> 
>     * Olivier will do the others

just made all corrections based on your feedbacks. thanks

core/log.class.inc.php
```diff
@@ -576,6 +576,7 @@ class LogChannels
 	public const DATATABLE = 'Datatable';
 
 	public const DEADLOCK = 'DeadLock';
+	public const SESSIONTRACKER = 'SessionTracker';
```
Molkobain, line 583:
```suggestion
	/**
	 * @var string Everything related to PHP sessions tracking
	 * @since 3.1.1 3.2.0 N°6901
	 */
	public const SESSIONTRACKER = 'SessionTracker';
```

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+
+use Combodo\iTop\Application\Helper\Session;
+use ContextTag;
+use Exception;
+use IssueLog;
+use UserRights;
+use utils;
+
+/**
+ * Class iTopSessionGarbageCollector
```
Molkobain:
```suggestion
 * Class SessionHandler
```

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @author Olivier Dain <olivier.dain@combodo.com>
+ * @package Combodo\iTop\Application\Helper
```
Molkobain:
```suggestion
 * @package Combodo\iTop\SessionTracker
```

sources/SessionTracker/SessionGC.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Combodo\iTop\SessionTracker;
+
+/**
+ * Class SessionGC
+ *
+ * @author Olivier Dain <olivier.dain@combodo.com>
+ * @package Combodo\iTop\Application\Helper
```
Molkobain:
```suggestion
 * @package Combodo\iTop\SessionTracker
```

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		return $bRes;
+	}
+
+	public function write($session_id, $data)
+	{
+		$bRes = parent::write($session_id, $data);
+
+		IssueLog::Debug("gc", \LogChannels::SESSIONTRACKER, [
```
Molkobain:
```suggestion
		IssueLog::Debug("write", \LogChannels::SESSIONTRACKER, [
```

Molkobain:
Great thanks! I'll set up a meeting within engineering for the follow-up.

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @author Olivier Dain <olivier.dain@combodo.com>
+ * @package Combodo\iTop\SessionTracker
+ * @since 3.1.1 3.2.0 N°6901
+ */
+class SessionHandler extends \SessionHandler
+{
+	public function destroy($session_id)
+	{
+		IssueLog::Debug("destroy", \LogChannels::SESSIONTRACKER, [
```
Molkobain:
```suggestion
		IssueLog::Debug("Destroy PHP session", \LogChannels::SESSIONTRACKER, [
```

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+		if ($bRes) {
+			$this->unlink_session_file($session_id);
+		}
+
+		return $bRes;
+	}
+
+	public function gc($max_lifetime)
+	{
+		IssueLog::Debug("gc", \LogChannels::SESSIONTRACKER, [
```
Molkobain:
```suggestion
		IssueLog::Debug("Run PHP sessions garbage collector", \LogChannels::SESSIONTRACKER, [
```

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+		$this->gc_with_time_limit($max_lifetime);
+		return $iRes;
+	}
+
+	public function open($save_path, $session_name)
+	{
+		$bRes = parent::open($save_path, $session_name);
+
+		$session_id = session_id();
+		IssueLog::Debug("open", \LogChannels::SESSIONTRACKER, [
```
Molkobain:
```suggestion
		IssueLog::Debug("Open PHP session", \LogChannels::SESSIONTRACKER, [
```

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		return $bRes;
+	}
+
+	public function write($session_id, $data)
+	{
+		$bRes = parent::write($session_id, $data);
+
+		IssueLog::Debug("write", \LogChannels::SESSIONTRACKER, [
```
Molkobain:
```suggestion
		IssueLog::Debug("Write PHP session", \LogChannels::SESSIONTRACKER, [
```

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
```diff
@@ -0,0 +1,299 @@
...
+		parent::tearDown();
+		static::SetNonPublicStaticProperty(ContextTag::class, 'aStack', []);
+
+		if (! is_null($this->sFile) && is_file($this->sFile)){
+			@unlink($this->sFile);
+		}
+	}
+
+	private function CreateUserAndLogIn() : ? string {
+		$_SESSION = [];
```
Molkobain, line 35: This might not be necessary anymore. We should discuss with @rquetiez

Molkobain (commented):
Specifying the log channel is great step forward as it enables us to set the log level for these specific logs.

That being said, I think it would ease reading logs (from all channels) for Support / R&D teams if the messages where a little bit more explicit and say that they are about PHP session.

What do you think @odain-cbd @eespie ?

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+					$iCreationTime = $aJson['creation_time'];
+				}
+			}
+
+			return json_encode (
+				[
+					'login_mode' => Session::Get('login_mode'),
+					'user_id' => $sUserId,
+					'creation_time' => $iCreationTime,
+					'context' => implode("|", ContextTag::GetStack())
```
Molkobain:
```suggestion
					'contexts' => implode("|", ContextTag::GetStack())
```

Making this plural as it contains all context tags currently set.
@odain-cbd you might need to adjust your monitoring module.

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,193 @@
...
+					$iCreationTime = $aJson['creation_time'];
+				}
+			}
+
+			return json_encode (
+				[
+					'login_mode' => Session::Get('login_mode'),
+					'user_id' => $sUserId,
+					'creation_time' => $iCreationTime,
+					'contexts' => implode("|", ContextTag::GetStack())
```
Molkobain:
```suggestion
					'context' => implode("|", ContextTag::GetStack())
```

Previous change was actually stupid as we already use singular everywhere else.

Molkobain:
~@odain-cbd they is an issue with the PR, there now are 29 commits and some changed files that should not be there.
Let's check together when you have a moment.~

Molkobain:
Discussed IRL with Alexandre, Olivier and Romain:
- Planned for iTop 3.1.1 / 3.2.0
- Add random session GC trigger on session opening to ensure files are cleared even without a CRON running. See what is done in `\privUITransactionFile::CleanupOldTransactions()`
- Session handler override will be conditioned to a config. param. disabled by default.
```php
		'session.enable_custom_session_handler' => [
			'type' => 'bool',
			'description' => 'Enable custom handler to have PHP sessions logs (opened, destroyed, ...) and generate additional session files for monitoring purposes',
			'default' => false,
			'value' => false,
			'source_of_value' => '',
		],
```

core/config.class.inc.php
```diff
@@ -1193,6 +1193,30 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
+		'sessions_tracking_enabled' => [
```
rquetiez: It is more common to use a dot between a feature name and the various settings.
=> 'sessions_tracking.enabled'

core/config.class.inc.php
```diff
@@ -1193,6 +1193,30 @@ class Config
...
+			'type' => 'bool',
+			'description' => 'Whether or not the whole mechanism to track active sessions is enabled.',
+			'default' => false,
+			'value' => '',
+			'source_of_value' => '',
+			'show_in_conf_sample' => false,
+		],
+		'sessions_tracking_gc_threshold' => [
+			'type' => 'integer',
+			'description'         => 'probability in percent for the garbage collector to be triggered (100 mean always)',
```
rquetiez: What does that mean? What happens when the garbage collector is triggered?

sources/SessionTracker/SessionHandler.php
```diff
@@ -0,0 +1,234 @@
...
+use utils;
+
+/**
+ * Class SessionHandler
+ *
+ * @author Olivier Dain <olivier.dain@combodo.com>
+ * @package Combodo\iTop\SessionTracker
+ * @since 3.1.1 3.2.0 N°6901
+ */
+class SessionHandler extends \SessionHandler
```
rquetiez, line 27: This is a bit mysterious. I had to focus on the namespace to figure out what was the purpose of such a class having the same name as its parent class.
I suggest adding a comment like "Tracks active user sessions in files"
Possibly also name the class like "FileTrackedSession"

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
```diff
@@ -0,0 +1,303 @@
...
+
+		$oSessionHandler = new SessionHandler();
+		Session::Set('login_mode', 'toto');
+
+		$sContent = $this->GenerateSessionContent($oSessionHandler, $sPreviousFileVersionContent);
+
+		$this->assertNotNull($sContent);
+		$aJson = json_decode($sContent, true);
+		$this->assertNotEquals(false, $aJson, $sContent);
+		$this->assertEquals($sUserId, $aJson['user_id'] ?? '', $sContent);
```
rquetiez: In case the test fails, the message will be cryptic => $sContent
I would rather have something like "Session content should have the user_id entry, found $sContent"

rquetiez: Same comment for most of the assertions above and below.

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
```diff
@@ -0,0 +1,303 @@
...
+		$sCreationTime = $aJson['creation_time'] ?? '';
+		$this->assertNotEquals('', $sCreationTime, $sContent);
+		$this->assertEquals('toto', $aJson['login_mode'] ?? '', $sContent);
+
+		$this->touchSessionFile($oSessionHandler, $session_id);
+		$sNewContent = file_get_contents($this->sFile);
+		$this->assertEquals($sContent, $sNewContent, $sNewContent);
+	}
+
+	public function testTouchSessionFile_UserLoggedInWithImpersonation(){
```
rquetiez: The first part (until UserRights::Impersonate) seems 100% redundant with the preceding test. Either keep only this test (prefered for speed of tests), or remove all the assertions of the first part.

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
```diff
@@ -0,0 +1,303 @@
...
+		$this->assertEquals($sUserId, $aJson['user_id'] ?? '', $sFirstContent);
+		$this->assertEquals(ContextTag::TAG_REST, $aJson['context'] ?? '', $sFirstContent);
+		$this->assertNotEquals('', $aJson['creation_time'] ?? '', $sFirstContent);
+		$this->assertEquals('toto', $aJson['login_mode'] ?? '', $sFirstContent);
+
+		$sNewContent = $this->GenerateSessionContent($oSessionHandler, $sFirstContent);
+		$this->assertEquals($sFirstContent, $sNewContent, $sNewContent);
+	}
+
+	public function testGenerateSessionContentWithSessionContextChange(){
```
rquetiez: The first part is 100% identical to the previous tests.
Please Keep the complete version and rename it to testGenerateSessionContent, for the sake of clarity.

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
```diff
@@ -0,0 +1,303 @@
...
+		$this->assertEquals([], $oSessionHandler->list_session_files());
+
+		for($i=0; $i<=1; $i++) {
+			$this->sFile = $this->GetFilePath($oSessionHandler, uniqid());
+			file_put_contents($this->sFile, "fakedata");
+		}
+
+		$aFoundSessionFiles = $oSessionHandler->list_session_files();
+		foreach ($aFoundSessionFiles as $sFile){
+			$this->assertTrue(is_file($sFile));
```
rquetiez: $this->assertTrue(is_file($sFile), 'list_session_files is should return only existing files. Not found: '.$sFile);

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
```diff
@@ -0,0 +1,303 @@
...
+	/**
+	 * @dataProvider GgcWithTimeLimit_FileWithData
+	 */
+	public function testGgcWithTimeLimit_FileWithData($iTimeLimit, $max_lifetime, $iExpectedProcessed) {
+		$oSessionHandler = new SessionHandler();
+		//remove all
+		$oSessionHandler->gc_with_time_limit(-1);
+		$this->assertEquals([], $oSessionHandler->list_session_files());
+
+		for($i=0; $i<=1; $i++) {
```
rquetiez: Please use a constant or variable to help the reader understanding the relation between this loop and the assertion on line 264.
Suggestion :
for ($i = 0 ; $i < FAKE_FILE_COUNT ; $i++)

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
```diff
@@ -0,0 +1,303 @@
...
+	/**
+	 * @dataProvider GgcWithTimeLimit_EmptyFile
+	 */
+	public function testGgcWithTimeLimit_EmptyFile($iTimeLimit, $iExpectedProcessed) {
+		$oSessionHandler = new SessionHandler();
+		//remove all
+		$oSessionHandler->gc_with_time_limit(-1);
+		$this->assertEquals([], $oSessionHandler->list_session_files());
+
+		for($i=0; $i<=1; $i++) {
```
rquetiez: for ($i = 0 ; $i < FAKE_FILE_COUNT ; $i++)

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
```diff
@@ -0,0 +1,303 @@
...
+				'iTimeLimit' => time() - 1,
+				'iExpectedProcessed' => 1
+			],
+		];
+	}
+
+	/**
+	 * @dataProvider GgcWithTimeLimit_EmptyFile
+	 */
+	public function testGgcWithTimeLimit_EmptyFile($iTimeLimit, $iExpectedProcessed) {
```
rquetiez: What is the difference between an empty file and a file containing the word "fake"?
I wonder if this test could be merged with the previous one.

core/config.class.inc.php
odain-cbd: yes it could be nicer. but this is not compliant with other existing variables like _transactions_gc_threshold_...

i could also have added an array instead (my first choice before seeing transaction* vars):
```
'sessions_tracking' =>
[ 'enabled' => true, 'gc_collector' => 'xx', 'gc_timelimit' => 'YY']
```

core/config.class.inc.php
odain-cbd: it is the probability the gc is triggered in fact at each Session.start(). nothing new as I relied on transaction existing cleanup logic

sources/SessionTracker/SessionHandler.php
odain-cbd: good point. still name should end with SessionHandler at least.

defaut PHP SessionHandler already relies on files. this new iTop SessionHandler keeps on calling his parent first. and afterwards creates files apart in itop folders (/data/...) with very few data.

what about FileActiveSessionTrackerSessionHandler? or ActiveSessionTrackerSessionHandler? seems quite long, no?

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
odain-cbd: one test consists in checking empty files are cleaned up. 

the other is a mix where I test cleanup works on a list of existing non empty files and files that do not existing (fake should be renaming: non-existing-file).

odain-cbd: empty files are for unauthentified active sessions (when we call status.php). in most cases files are not empty and there are authentication data inside (login_mode/context/user_id).

sources/SessionTracker/SessionHandler.php
odain-cbd: add phpdoc only but no renaming

tests/php-unit-tests/unitary-tests/sources/SessionTracker/SessionHandlerTest.php
odain-cbd: merged some tests and add comments in case of failures

sources/SessionTracker/SessionHandler.php
Hipska: What about `TrackingSessionHandler`?

Molkobain: I asked Olivier to named it like that.

For information, it is very common (and a good practice) to have a class with the same name as the one it is extending, the namespace is here to differenciate them.

To ease the understanding, the usual convention is to do something like this (I'm not asking Olivier to do it, just we spread awereness):

```php
namespace A\B;

use SessionHandler as BaseSessionHandler; // This makes an alias for this class to be used in this file

class SessionHandler extends BaseSessionHandler
{
    [...]
}
```

rquetiez (approved):
Reviewed the tests

PR #569, opened 2023-11-08
N°6765 Avoid behat scenario loading issues on portal modal

piRGoif:
Factorize the existing ready script workaround to use it also in field_set.js (used to display forms in portal modal)

### DIFF
--- a/js/field_set.js
+++ b/js/field_set.js
@@ -48,7 +48,7 @@ $(function()
 		_create: function()
 		{
 			var me = this;
-
+			
 			this.element
 			.addClass('field_set');
 			
--- a/js/field_set.js
+++ b/js/field_set.js
@@ -49,8 +49,6 @@ $(function()
 		{
 			var me = this;
 
-			CombodoJsActivity.SetScriptOngoing();
-			
 			this.element
 			.addClass('field_set');
 			
@@ -90,8 +88,6 @@ $(function()
 			{
 				this.buildForm();
 			}
-
-			CombodoJsActivity.SetScriptFinished();
 		},
    
 		// called when created, and later when changing options
--- a/js/field_set.js
+++ b/js/field_set.js
@@ -48,6 +48,8 @@ $(function()
 		_create: function()
 		{
 			var me = this;
+
+			CombodoJsActivity.SetScriptOngoing();
 			
 			this.element
 			.addClass('field_set');
@@ -88,6 +90,8 @@ $(function()
 			{
 				this.buildForm();
 			}
+
+			CombodoJsActivity.SetScriptFinished();
 		},
    
 		// called when created, and later when changing options
--- a/js/utils.js
+++ b/js/utils.js
@@ -1000,12 +1000,25 @@ const CombodoJSConsole = {
  * @since 3.0.4 3.1.1 3.2.0 N°6765
  */
 const CombodoJsActivity = {
+	BODY_DATAATTR_NAME_READY: "data-ready-scripts",
+
+	/**
+	 * Counter so that we set the flag as done only on the last call
+	 * @type number
+	 */
+	OngoingScriptsCount: 0,
+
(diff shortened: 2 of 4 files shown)


piRGoif:
@odain-cbd  and @Molkobain can you do a review before I can merge ?

piRGoif:
Review with Olivier : we should avoid setting done too soon... 
For example : 

1. process A does start
2. process B does start
3. process B does end
4. process A does end

The body attr should only be set to end on 4, not on 3...

Maybe we could simply add a counter as another body attr ? There could be issues with concurrency, but as we just have few consumers now and considering the need (not doing end too soon) it might be not worse it to handle it.

piRGoif:
> we should avoid setting done too soon...

Done with 927d985a

@odain-cbd can you have a look ?

odain-cbd (approved):
muy bien!

Molkobain:
Will discuss IRL tomorrow with Pierre

piRGoif:
As seen with Guillaume, the process is : 

```
ajaxSend
ajaxComplete
object create ready START
field_set START
field_set END
object create ready END
```

As a consequence I moved the start/finish to the object template.

@Molkobain can you do a final review ?

js/field_set.js
Molkobain, line 1: Can you revert this file from the PR so we don't get mislead when overlooking at the history?

js/utils.js
```diff
@@ -994,6 +994,34 @@ const CombodoJSConsole = {
...
+	},
+
+	SetScriptFinished: function() {
+		this.OngoingScriptsCount--;
+
+		if (this.OngoingScriptsCount < 1) {
+			$("body").attr(this.BODY_DATAATTR_NAME_READY, "done");
+		}
+	}
+}
```
Molkobain:
```suggestion
	BODY_DATA_ATTR_NAME_READY: "data-ready-scripts",

	/**
	 * Counter so that we set the flag as done only on the last call
	 * @type number
	 */
	iOngoingScriptsCount: 0,

	SetScriptOngoing: function() {
		this.iOngoingScriptsCount++;
		$("body").attr(this.BODY_DATA_ATTR_NAME_READY, "start");
	},

	SetScriptFinished: function() {
		this.iOngoingScriptsCount--;

		if (this.iOngoingScriptsCount < 1) {
			$("body").attr(this.BODY_DATAATTR_NABODY_DATA_ATTR_NAME_READYE_READY, "done");
		}
	}
}
```

Molkobain (changes requested):

js/utils.js
```diff
@@ -994,6 +994,34 @@ const CombodoJSConsole = {
...
+
+	AddOngoingScript: function() {
+		this.OngoingScriptsCount++;
+		$("body").attr(this.BODY_DATAATTR_NAME_READY, "start");
+	},
+
+	RemoveOngoingScript: function() {
+		this.OngoingScriptsCount--;
+
+		if (this.OngoingScriptsCount < 1) {
```
Molkobain:
```suggestion
	iOngoingScriptsCount: 0,

	AddOngoingScript: function() {
		this.iOngoingScriptsCount++;
		$("body").attr(this.BODY_DATAATTR_NAME_READY, "start");
	},

	RemoveOngoingScript: function() {
		this.iOngoingScriptsCount--;

		if (this.iOngoingScriptsCount < 1) {
```

To be consistent with other helpers, then ok to merge.

Molkobain (changes requested):

js/utils.js
piRGoif: Done in 9ba4cf0d

js/utils.js
piRGoif: Done

js/field_set.js
piRGoif: Done in d3e2dd4a !

PR #570, opened 2023-11-10
Small optimization in the search screen when there is only one result.

accognet:
This avoids loading the whole object, only checking the existence of the object.

### DIFF
--- a/application/displayblock.class.inc.php
+++ b/application/displayblock.class.inc.php
@@ -1911,6 +1911,7 @@ class MenuBlock extends DisplayBlock
 			// Check concurrent lock (can only be lock if we are handling a single object
 			$bLocked = false;
 			if ($iSetCount === 1) {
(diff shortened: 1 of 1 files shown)


Molkobain:
Can you provide the symptom that made you do this fix as well as a reproduction procedure?

accognet:
It's only visible in the log kpi. 
I was looking for a client issue on document file when I saw this little optimization. This avoids loading the blob fields.

accognet:
You can see the request in search screen, when there is only one result.

Molkobain:
Worked with Anne-Catherine on the PR, as is it as an impact on the rest of the code that will be using the `$oSet` variable.
To fix avoid that, we have to refactor several piece of code to use a clone of the set when necessary. But even then there will be a reload deeper in the callstack.

So we agreed that it would add too much complexity to the already complex existing code. Closing this PR.

PR #574, opened 2023-11-17
N°6975 ItopCustomDatamodelTestCase VS symlinks flag

piRGoif:
ItopCustomDatamodelTestCase was added in 2.7.9 / 3.0.3 / 3.1.0
When preparing the test env the compilation is made with the symlink param always set to false. 
Since iTop 3.0.0 we have a symlink flag persisted on disk (N°4092), that is present by default in the iTop repo.
When working with the repo, executing a ItopCustomDatamodelTestCase  leads to removing this flag :(

This is now fixed.

### DIFF
--- a/tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
+++ b/tests/php-unit-tests/src/BaseTestCase/ItopCustomDatamodelTestCase.php
@@ -11,8 +11,8 @@ use Combodo\iTop\Test\UnitTest\Hook\TestsRunStartHook;
 use Combodo\iTop\Test\UnitTest\Service\UnitTestRunTimeEnvironment;
 use Config;
 use Exception;
-use IssueLog;
 use MetaModel;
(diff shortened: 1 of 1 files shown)


piRGoif:
@Molkobain I found this bug while executing the whole tests suite.
Do you agree with the fix ?

Molkobain:
> @Molkobain I found this bug while executing the whole tests suite. Do you agree with the fix ?

- This environment is re-build at each tests run, what is your use case for having the symlinks?
- If we accept this, we could pull it up to support/2.7 ?

piRGoif:
Discussed with Guillaume : 
- the purpose of this fix is to stop removing `/data/.compilation-symlinks` on each ItopCustomDatamodelTestCase execution (and we have one impl in the default test suite : ApplicationExtensionTest)
- the symlink flag was introduced in 3.0.0 (#190) so there are no visible changes in iTop 2.7

Molkobain (approved):

PR #575, opened 2023-11-20
N°6968 - Audit duration : add of a rule multiplie by 4 the time of response

accognet:
internal

### DIFF
--- a/core/dbsearch.class.php
+++ b/core/dbsearch.class.php
@@ -851,11 +851,11 @@ abstract class DBSearch
 			return;
 		}
 
-		if (count($aColumns) == 0)
-		{
-			$aColumns = array_keys(MetaModel::ListAttributeDefs($this->GetClass()));
-			// Add the standard id (as first column)
-			array_unshift($aColumns, 'id');
+        if (count($aColumns) == 0)
+        {
+            $aColumns = array_keys(MetaModel::ListAttributeDefs($this->GetClass()));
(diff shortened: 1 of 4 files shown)


Molkobain:
Support PRs review:
- Rename existing tet to something saying that it only test that the behavior is the same as for the old method
- Add a new unit test for the method itself with following use cases:
  - 1 class with alias
  - 2 classes with aliases, retrieve field from joined class
  - 1 class, 2 attributes (this one should fail for now as we'll have to refactor the method to hadnle this case)
- Refactor method to manage several attributes (eg. templates base)

Molkobain:
Support PRs review:

  * We adjust the new method to handle class alias in addition to the attribute code as we could want to get an attribute from any joined class.
  * We adjust the test cases accordingly
  * If the need to have a method for several attributes ever comes up, we will add it and refactor the code

> * Refactor method to manage several attributes (eg. templates base)

We don't recall what use case we had that needed to handle several attributes, so we forget it for now.

Molkobain:
Questions for next support PRs review:
  * New method returns a hash array with the same of the attribute
    * We don't need it as it always returns only 1 attribute
    * Refactor it and the consumers
    * It will ease test case comparison

Molkobain:
This will be included in iTop 2.7.11 / 3.1.2 / 3.2.0. The PR was merged i support/3.2 by mistake and the commit was cherry-picked to support/2.7.

PR #576, opened 2023-11-21
N°6976 Fix DeprecatedCallsLog error handler never set

piRGoif:
With N°6274 we added a constant to be able to remove the DeprecatedCallsLog error handler when running tests, because this error handler was removing PHPUnit's one.
The code was disfunctional : DeprecatedCallsLog error handler was never set anymore :(

This PR restores the expected behavior : 
- when running PHPUnit, DeprecatedCallsLog won't set an error handler
- but it will on any other case

### DIFF
--- a/approot.inc.php
+++ b/approot.inc.php
@@ -26,8 +26,10 @@ define('ITOP_DESIGN_LATEST_VERSION', '3.0');
 define('ITOP_CORE_VERSION', '3.0.3');
 
 /**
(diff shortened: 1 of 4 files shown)


piRGoif:
@bdalsass and @Molkobain : this is blocking Symfony depreciations logging
Can you do a review so we can merge asap ? 
Many thanks !

bdalsass (approved):

Molkobain (approved):

piRGoif:
Thanks Guillaume and Benjamin for your reviews, merging !

PR #580, opened 2023-12-04
N°6934 - Symfony 6.4 - upgrade Symfony bundles to 6.4

bdalsass:
- Update Symfony lib to version ~6.4.0
- Update code missing return type
- Add a default value to APP_SECRET mandatory parameter
-  Use dependency injection in ExceptionListener & UserProvider classes


Molkobain (approved):
Discussed IRL with Benjamn, I'll trace in the ticket that app_secret should be unique and generated during the setup. while leaving the possibility to the admin to set it in the configuration file.

piRGoif (approved):
Good, not so much to change on our code ! We'll have to see what behat tests will give...

Even if Symfony 6.4 was just released, as an extra precaution I've launched a `composer audit` command and nothing found \o/

PR #581, opened 2023-12-05
N°7034 - Symfony 6.4 - Not passing FILTER_REQUIRE_ARRAY or FILTER_FOR…

bdalsass:
…CE_ARRAY flags to InputBag::filter() when filtering an array will throw BadRequestException

Not passing FILTER_REQUIRE_ARRAY or FILTER_FORCE_ARRAY flags to InputBag::filter() when filtering an array will throw BadRequestException
 
We need to use FILTER_REQUIRE_ARRAY flag where we expected array values.
No change when the valu is not set. (default value si respected default value of null for exemple)

source:
https://github.com/symfony/symfony/blob/6.4/UPGRADE-6.0.md#httpfoundation

### DIFF
--- a/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
@@ -795,7 +795,7 @@ class ObjectController extends BrickController
 
 			// Updating host object
 			$oFormManager->OnUpdate(array(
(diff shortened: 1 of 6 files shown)


Molkobain:
About our doubts with the `null` default value and the filter, this seems like a good time to try to make a unit test on the `RequestManipulator` service to test the `ReadParam` method. This would also be a first step to making a unit test in the Symfony universe.

_I know you hate me_ 😬

PR #582, opened 2023-12-08
 N°7055 - Apply better default value for portal copy object link

Hipska:
The default text added to the clipboard when clicking "Copy object link" on the portal is `<object friendlyname>: <object portal URL>`. This seems not useful in most cases, so this is my proposal to change that to simply `<object portal URL>`.

In the scenario users still want the old text, they can adapt the dictionary item in the designer or with an extension.


Molkobain (approved):

Molkobain:
Good idea yes 👍

Hipska:
This was also discussed on [Slack](https://itopbycombodo.slack.com/archives/C02CAG5MLH0/p1700813661874819) with @steffunky.

steffunky (approved):

steffunky:
Approved by @v-dumas

PR #583, opened 2023-12-08
N°7052 synchro_import.php: fix undefined offset notices

piRGoif:
With #269 was added in 3.0.1 iTop version, in the `/synchro/synchro_import.php` script the `$aIsBinaryToTransform` variable. But :
- when reading the header columns, the array is set only for certain columns
- whereas when reading data the array is accessed on all columns indexes

The fix is to set the array for every columns indexes.

### DIFF
--- a/synchro/synchro_import.php
+++ b/synchro/synchro_import.php
@@ -469,6 +469,8 @@ try
 	$aIsBinaryToTransform = array();
 	foreach ($aInputColumns as $iFieldId => $sInputColumn)
 	{
+		$aIsBinaryToTransform[$iFieldId] = false;
+
 		if (array_key_exists($sInputColumn, $aDateColumns))
 		{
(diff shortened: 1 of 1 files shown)


piRGoif:
Hello @Hipska can you take a look ?

Hipska (approved):

Hipska:
Is the notice for the primary key field? Can you show such a notice?

Molkobain (approved):

piRGoif:
Hello,
No sorry I don't have logs samples.
A client had such notices and made its own analysis of their cause (indeed for each primary keys the aIsBinaryToTransform array index isn't set), and I agreed with his conclusions and made this PR.
Thanks for your review !

Hipska:
Okay, ~~could also have moved the assignment/check before the one for primary_key or before the dates check..~~
Nevermind, that would also introduce issues.

piRGoif:
Support PR review : ok 
Add in the bug reproduction method

piRGoif:
Functional review OK
ready to merge !

PR #597, opened 2024-01-02
N°7130 - Allows to ignore existing column field in setup's data migration method

steffunky:
When migrating data from a table to another, it might happen (when moving a column from children classes to their parent) that multiple classes want to update the same column. Still, it's not possible as we check if the destination table exists.

### DIFF
--- a/setup/moduleinstaller.class.inc.php
+++ b/setup/moduleinstaller.class.inc.php
@@ -248,7 +248,7 @@ abstract class ModuleInstallerAPI
 	 * @throws \MySQLException
 	 * @throws \MySQLHasGoneAwayException
 	 *
-	 * @since 3.2.0 Add parameter $bIgnoreExistingDstColumn
+	 * @since 3.2.0 N°7130 Add parameter $bIgnoreExistingDstColumn
 	 */
 	public static function MoveColumnInDB($sOrigTable, $sOrigColumn, $sDstTable, $sDstColumn, $bIgnoreExistingDstColumn = false)
 	{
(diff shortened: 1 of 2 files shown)


setup/moduleinstaller.class.inc.php
```diff
@@ -242,11 +242,13 @@ public static function RenameEnumValueInDB($sClass, $sAttCode, $sFrom, $sTo)
 	 * @param $sOrigColumn
 	 * @param $sDstTable
 	 * @param $sDstColumn
+	 * @param bool $bIgnoreExistingDstColumn
 	 *
-	 * @throws \MySQLException
 	 * @throws \CoreException
+	 * @throws \MySQLException
+	 * @throws \MySQLHasGoneAwayException
```
Molkobain, line 249:
```suggestion
	 * @throws \MySQLHasGoneAwayException
	 *
	 * @since 3.2.0 Add parameter $bIgnoreExistingDstColumn
```

Molkobain:
Will discuss with Stephen about a unit test on that method to ensure we don't regress on that. Especially as it wouldn't be seen easily on development environments.

setup/moduleinstaller.class.inc.php
```diff
@@ -242,11 +242,15 @@ public static function RenameEnumValueInDB($sClass, $sAttCode, $sFrom, $sTo)
...
+	 *
+	 * @since 3.2.0 Add parameter $bIgnoreExistingDstColumn
```
Molkobain:
```suggestion
	 * @since 3.2.0 N°7130 Add parameter $bIgnoreExistingDstColumn
```

Molkobain (approved):

Molkobain:
_This is part of the "newsroom" ticket_

PR #598, opened 2024-01-04
N°4368 Fix CORB blocking regression

piRGoif:
We added the X-Content-Type-Options HTTP header, in consequence CORB blocking may now occurs. 
With those modifications : 
- sending the header is disabled for certain WebPage impl
- a new `security_header_xcontent_type_options` config parameter can be used to disabled header sending

### DIFF
--- a/application/ajaxwebpage.class.inc.php
+++ b/application/ajaxwebpage.class.inc.php
@@ -54,6 +54,7 @@ class ajax_page extends WebPage implements iTabbedPage
 	/**
 	 * Disabling sending the header so that resource won't be blocked by CORB. See parent method documentation.
 	 * @return void
+	 * @since 2.7.10 3.0.4 3.1.2 3.2.0 N°4368 method creation
 	 */
 	public function add_xcontent_type_options()
 	{
--- a/application/ajaxwebpage.class.inc.php
+++ b/application/ajaxwebpage.class.inc.php
@@ -52,6 +52,15 @@ class ajax_page extends WebPage implements iTabbedPage
 	}
 
 	/**
+	 * Disabling sending the header so that resource won't be blocked by CORB. See parent method documentation.
+	 * @return void
+	 */
+	public function add_xcontent_type_options()
+	{
+		// Nothing to do !
+	}
+
+	/**
 	 * @inheritDoc
 	 * @throws \Exception
 	 */
--- a/application/webpage.class.inc.php
+++ b/application/webpage.class.inc.php
@@ -515,7 +515,7 @@ class WebPage implements Page
 	 * Warning : this header will trigger the Cross-Origin Read Blocking (CORB) protection for some mime types (HTML, XML except SVG, JSON, text/plain)
 	 * In consequence some children pages will override this method.
 	 *
-	 * Sending header can be disabled globally using the `security.header_xcontent_type_options` optional config parameter.
+	 * Sending header can be disabled globally using the `security.enable_header_xcontent_type_options` optional config parameter.
 	 *
 	 * @return void
 	 * @since 2.7.10 3.0.4 3.1.2 3.2.0 N°4368 method creation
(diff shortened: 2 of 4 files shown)


application/ajaxwebpage.class.inc.php
```diff
@@ -51,6 +51,15 @@ function __construct($s_title) {
 		utils::InitArchiveMode();
 	}
 
+	/**
+	 * Disabling sending the header so that resource won't be blocked by CORB. See parent method documentation.
+	 * @return void
```
Molkobain, line 56: Maybe add an `@since` with the ticket number so we can trace it meaning back,

core/config.class.inc.php
```diff
@@ -1320,6 +1320,14 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
+		'security_header_xcontent_type_options' => [
+			'type' => 'bool',
+			'description' => 'If set to false, iTop will stop sending the X-Content-Type-Options HTTP header.',
+			'default' => 'true',
```
```suggestion
		'security.enable_header_xcontent_type_options' => [
			'type' => 'bool',
			'description' => 'If set to false, iTop will stop sending the X-Content-Type-Options HTTP header.',
			'default' => true,
```

- Namespace the conf. param. like for all new params.
- Fix default value to bool instead of string.
- Maybe we could explain in the description for which reasons it could be good to switch it to false?

application/webpage.class.inc.php
```diff
@@ -513,13 +512,34 @@ public function add_xframe_options($sHeaderValue = null)
...
 	{
+		try {
+			$oConfig = utils::GetConfig();
+		} catch (ConfigException|CoreException $e) {
+			$oConfig = null;
+		}
+		if (is_null($oConfig)) {
+			$bSendXContentTypeOptionsHttpHeader = true;
+		} else {
+			$bSendXContentTypeOptionsHttpHeader = $oConfig->Get('security_header_xcontent_type_options');
```
Molkobain:
```suggestion
			$bSendXContentTypeOptionsHttpHeader = $oConfig->Get('security.enable_header_xcontent_type_options');
```

See comment below

application/webpage.class.inc.php
```diff
@@ -513,13 +512,34 @@ public function add_xframe_options($sHeaderValue = null)
 	}
 
 	/**
+	 * Warning : this header will trigger the Cross-Origin Read Blocking (CORB) protection for some mime types (HTML, XML except SVG, JSON, text/plain)
+	 * In consequence some children pages will override this method.
+	 *
+	 * Sending header can be disabled globally using the `security_header_xcontent_type_options` optional config parameter.
```
Molkobain:
```suggestion
	 * Warning: this header will trigger the Cross-Origin Read Blocking (CORB) protection for some mime types (HTML, XML except SVG, JSON, text/plain)
	 * In consequence some children pages will override this method.
	 *
	 * Sending header can be disabled globally using the `security.enable_header_xcontent_type_options` optional config parameter.
```

See comment below

application/xmlpage.class.inc.php
```diff
@@ -48,6 +48,15 @@ function __construct($s_title, $bPassThrough = false)
 		$this->add_header("Content-location: export.xml");
 	}
 
+	/**
+	 * Disabling sending the header so that resource won't be blocked by CORB. See parent method documentation.
+	 * @return void
```
Molkobain, line 53: Maybe add an `@since` with the ticket number so we can trace it meaning back,

Molkobain (changes requested):

core/config.class.inc.php
piRGoif: Argll good catch for the default value :(
I changed also the description
Done in 5a1dd9ff

We'll have a chat on renaming

application/ajaxwebpage.class.inc.php
piRGoif: Done in 5a1dd9ff

application/xmlpage.class.inc.php
piRGoif: Done in 5a1dd9ff

core/config.class.inc.php
piRGoif: Renaming done in c630a3f1

application/webpage.class.inc.php
piRGoif: Done in c630a3f1

application/webpage.class.inc.php
piRGoif: Done in c630a3f1

core/config.class.inc.php
piRGoif: Woops renamed done in da941fca actually (missed the enable prefix)

Molkobain (approved):

PR #599, opened 2024-01-05
N°7122 - Portal: Hide log off button when user can't actually log off (eg. SSO using SAML or other providers)

Molkobain:
**Symptom**
For some users using SSO (eg. SAML), they can't log off through iTop as it will show them an error page.

**Proposed solution**
Do exactly as in the backoffice, only show the "log off" button is the user can actually log off, which is based on a session variable that can be retrieved through `\utils::CanLogOff()`

_**Important:** Careful when merging in support/3.1 branch as the `UserProvider` part changed a lot when migrating to Symfony 5.4_

_Desktop_
image

_Mobile_
image

### DIFF
--- a/datamodels/2.x/itop-portal-base/portal/src/EventListener/UserProvider.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/EventListener/UserProvider.php
@@ -93,7 +93,7 @@ class UserProvider implements ContainerAwareInterface
 		$this->oContainer->set('combodo.current_user', $oUser);
 
         // User allowed to log off or not
-        $this->oContainer->set('combodo.current_user.can_logff', utils::CanLogOff());
+        $this->oContainer->set('combodo.current_user.can_logoff', utils::CanLogOff());
 
 		// Allowed portals
 		$aAllowedPortals = UserRights::GetAllowedPortals();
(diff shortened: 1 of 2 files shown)


bdalsass (approved):

datamodels/2.x/itop-portal-base/portal/src/EventListener/UserProvider.php
```diff
@@ -91,6 +92,9 @@ public function onKernelRequest(GetResponseEvent $oGetResponseEvent)
 		}
 		$this->oContainer->set('combodo.current_user', $oUser);
 
+        // User allowed to log off or not
+        $this->oContainer->set('combodo.current_user.can_logff', utils::CanLogOff());
```
Hipska:
```suggestion
        $this->oContainer->set('combodo.current_user.can_logoff', utils::CanLogOff());
```

datamodels/2.x/itop-portal-base/portal/templates/layout.html.twig
```diff
@@ -3,11 +3,13 @@
 
 {% if app['combodo.current_user'] is defined and app['combodo.current_user'] is not null %}
 	{% set bUserConnected = true %}
+    {% set bUserCanLogOff = app['combodo.current_user.can_logff'] %}
```
Hipska:
```suggestion
    {% set bUserCanLogOff = app['combodo.current_user.can_logoff'] %}
```

Hipska (changes requested):

datamodels/2.x/itop-portal-base/portal/src/EventListener/UserProvider.php
Molkobain: Good catch, thanks!

Hipska (approved):

Molkobain:

PR #600, opened 2024-01-08
N°7645 - PHP 8.1: Fix usage of strpos() & str_replace() with null value when compiling empty dictionary

Hipska:
## To reproduce

Having : 
- an iTop 3.1 clone
- PHP 8.1+ 
- php notices enabled

Deploy this XML, that is redefining a dict key to an empty value : 

```xml
<itop_design version="1.7">
  <dictionaries>
    <dictionary id="EN US" _delta="must_exist">
      <entries>
        <entry id="Class:UserRequest/Attribute:urgency" _delta="force"></entry>
      </entries>
    </dictionary>
  </dictionaries>
</itop_design>
```

Upon compilation (either by the setup or the toolkit) you'll get the following notices : 

```
Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in C:\Dev\wamp64\www\itop-31\setup\compiler.class.inc.php on line 3272

Deprecated: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in C:\Dev\wamp64\www\itop-31\setup\compiler.class.inc.php on line 1170
```

They are blocking during setup.

## Analysis

The methods `FilterDictString` and `QuoteForPHP` both use `strpos` or `str_replace` who give a deprecation notice when the argument is `null`:

```
Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/html/setup/compiler.class.inc.php on line 3231
Deprecated: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /var/www/html/setup/compiler.class.inc.php on line 1146
```

This message is repeated for each empty dictionary item, which seems to happen on a Professional instance connected to the Designer.

Logs taken from a iTop Professional 3.1.0-3. The null argument gives a deprecation notice since [PHP 8.1](https://3v4l.org/W5mRs), which is supported for this iTop version.

### DIFF
--- a/setup/compiler.class.inc.php
+++ b/setup/compiler.class.inc.php
@@ -1165,6 +1165,7 @@ EOF
 	 */	 	
 	protected function QuoteForPHP($sStr, $bSimpleQuotes = false)
 	{
+		$sStr = $sStr ?? '';
 		if ($bSimpleQuotes)
 		{
 			$sEscaped = str_replace(array('\\', "'"), array('\\\\', "\\'"), $sStr);
@@ -3260,7 +3261,7 @@ EOF;
 		file_put_contents($sLanguagesFile, $sLanguagesFileContent);
 	}
 
-	protected static function FilterDictString($s)
+	protected static function FilterDictString(string $s): string
 	{
 		if (strpos($s, '~') !== false)
 		{
--- a/setup/compiler.class.inc.php
+++ b/setup/compiler.class.inc.php
@@ -3221,10 +3221,11 @@ EOF;
 
 			$aEntriesPHP = array();
 			$oEntries = $oDictionaryNode->GetUniqueElement('entries');
+			/** @var MFElement $oEntry */
 			foreach ($oEntries->getElementsByTagName('entry') as $oEntry)
 			{
 				$sStringCode = $oEntry->getAttribute('id');
-				$sValue = $oEntry->GetText();
+				$sValue = $oEntry->GetText('');
 				$aEntriesPHP[] = "\t'$sStringCode' => ".self::QuoteForPHP(self::FilterDictString($sValue), true).",";
 			}
 			$sEntriesPHP = implode("\n", $aEntriesPHP);


piRGoif:
Hello Thomas,
Can you provide some sample dict keys to reproduce the issue ?

Hipska:
I guess it is just `<entry id="dict:code"/>`.

Hipska:
Other PHP 8.1 findings:

```
Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/core/ormcaselog.class.inc.php on line 172
Deprecated: md5(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/core/ormdocument.class.inc.php on line 346
```

piRGoif:
Hello,
Thanks for the 2 other deprecated notices on PHP 8.1, I've created N°7212
Can you detail how to reproduce them, especially the one on ormDocument ?

Hipska:
Already discussed with @Molkobain on Slack.

Displaying a Person on the portal (e.g. click on manager on profile page) causes the `ormDocument` notice.

The other one was when going to a User's details page (might be related to having the activity panel open)

piRGoif:
ormCaselog error is handled in #607

Hipska:
Having this issue as well with 3.1.1:

```
Deprecated:  strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/html/setup/compiler.class.inc.php on line 3264
Deprecated:  str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /var/www/html/setup/compiler.class.inc.php on line 1170
```

This is becoming a major issue since it blocks the setup while upgrading to 3.1.1 when for example the [sv-display-mgmt](https://github.com/Super-Visions/sv-display-mgmt/blob/db7b5d64d18a360d065e4fdfa8004f7029c67f57/datamodel.sv-display-mgmt.xml#L277) is installed.

Hipska:
Even replacing this line with the following isn't a suitable workaround:
```xml
<entry id="Class:Display/Attribute:type_id+" _delta="define"></entry>
```

Looks like there is no way anymore to define an empty dict in xml !!!

piRGoif:
Woops absolutely sorry I forgot this PR :(
I've just tried to reproduce on a standard support/3.1 clone, and I got it ! 
I'll rewrite the description a little, and send the PR back for review.

piRGoif:
@Molkobain will be the new assignee for the reviews next month.

Molkobain (changes requested):
Technical review: We would rather fix the root cause than the symptom:
  * `FilterDictString()`
    * Method type hinting can be enforce as it is only used within the compiler: `FilterDictString(string $s): string`
    * Fix te one usage in the line below your proposal: `self::FilterDictString($sValue ?? '')`
  * `QuoteForPHP()`
    * As it is used in many places, fix its internal usages with `$s = $s ?? '';` at the begining of the method.

setup/compiler.class.inc.php
```diff
@@ -3221,10 +3222,11 @@ protected function CompileDictionaries($oDictionaries, $sTempTargetDir, $sFinalT
 
 			$aEntriesPHP = array();
 			$oEntries = $oDictionaryNode->GetUniqueElement('entries');
+			/** @var MFElement $oEntry */
 			foreach ($oEntries->getElementsByTagName('entry') as $oEntry)
 			{
 				$sStringCode = $oEntry->getAttribute('id');
-				$sValue = $oEntry->GetText();
+				$sValue = $oEntry->GetText('');
 				$aEntriesPHP[] = "\t'$sStringCode' => ".self::QuoteForPHP(self::FilterDictString($sValue), true).",";
```
Molkobain, line 3230:
```suggestion
				$sValue = $oEntry->GetText();
				$aEntriesPHP[] = "\t'$sStringCode' => ".self::QuoteForPHP(self::FilterDictString($sValue ?? ''), true).",";
```

See last comment in the conversation

Molkobain (changes requested):

setup/compiler.class.inc.php
Hipska: Hmm, I thought "We would rather fix the root cause than the symptom" would mean you wanted the root cause fixed instead of adding a workaround?

Molkobain: Yes, what we meant by that is that the `FilterDictString` should be stricter to avoid this symptom. But I overlooked your proposition, it's actually ok. Thanks.

Molkobain (approved):

Molkobain:
Functional review: Accepted.

PR #603, opened 2024-01-25
N°6974 - Flatten classes in datamodel

eespie:
See Bug N°6974

ModelFactory updated to store data models classes in a flat way (before it was hierarchical).
Main update is LoadDelta not storing the classes under their parent class in the XML tree.
The behavior must not change so:

- The compiler should generate the same code
- The assumptions like deleting a class suppress also all the children should be kept
- ModelFactory API should behave the same as before


piRGoif:
There is obviously still some work going on
@eespie can you please change the pr to draft, and revert when it will be ready ?

Hipska:
Also, https://github.com/Combodo/iTop/blob/4e53deec9d96093df87b233249a9c071ae458d40/.github/pull_request_template.md?plain=1#L3-L5

core/designdocument.class.inc.php
```diff
@@ -41,6 +49,10 @@
  */
 class DesignDocument extends DOMDocument
 {
+
+	public const XML_PARSE_BIG_LINES = 4194304;
```
Molkobain, line 54: Can you add a comment on what this is for, and what unit it is?

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
 		}
 		return $sRet;
 	}
+
+	/**
+	 * Check that the current node is actually a class node, under classes
```
Molkobain, line 327:
```suggestion
	 * Check that the current node is actually a class node, under classes
	 * @since 3.1.2 3.2.0 N°6974
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
+	 * UNSAFE: may return nodes marked as _alteration="removed"
+	 * A method with the same signature MUST exist in MFDocument for the recursion to work fine
+	 *
+	 * @param DesignElement $oRefNode The node to search for
+	 * @param string $sSearchId substitutes to the value of the 'id' attribute
+	 *
+	 * @return DesignElement|null
+	 * @throws \Exception
+	 */
+	public function _FindChildNode(DesignElement $oRefNode, $sSearchId = null)
```
Molkobain:
```suggestion
	public function _FindChildNode(DesignElement $oRefNode, $sSearchId = null): ?DesignElement
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
+	 * UNSAFE: may return nodes marked as _alteration="removed"
+	 * A method with the same signature MUST exist in MFDocument for the recursion to work fine
+	 *
+	 * @param DesignElement $oRefNode The node to search for
+	 * @param string $sSearchId substitutes to the value of the 'id' attribute
+	 *
+	 * @return DesignElement|null
+	 * @throws \Exception
+	 */
+	public function _FindChildNodes(DesignElement $oRefNode)
```
Molkobain:
```suggestion
	public function _FindChildNodes(DesignElement $oRefNode): ?DesignElement
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
+	/**
+	 * Find the child node matching the given node.
+	 * UNSAFE: may return nodes marked as _alteration="removed"
+	 * A method with the same signature MUST exist in MFDocument for the recursion to work fine
+	 *
+	 * @param DesignElement $oRefNode The node to search for
+	 * @param string $sSearchId substitutes to the value of the 'id' attribute
+	 *
+	 * @return DesignElement|null
+	 * @throws \Exception
```
Molkobain, line 351:
```suggestion
	 * @throws \Exception
	 * @since 3.1.2 3.2.0 N°6974
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
```
Molkobain, line 368:
```suggestion
	 * @throws \Exception
	 * @since 3.1.2 3.2.0 N°6974
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
+	/**
+	 * Find the child node matching the given node under the specified parent.
+	 * UNSAFE: may return nodes marked as _alteration="removed"
+	 *
+	 * @param \DOMNode $oParent
+	 * @param DesignElement $oRefNode
+	 * @param string|null $sSearchId
+	 *
+	 * @return DesignElement|null
+	 * @throws Exception
```
Molkobain, line 385:
```suggestion
	 * @throws Exception
	 * @since 3.1.2 3.2.0 N°6974
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
+	 */
+	public static function _FindNode(DOMNode $oParent, DesignElement $oRefNode, string $sSearchId = null)
```
Molkobain:
```suggestion
	public static function _FindNode(DOMNode $oParent, DesignElement $oRefNode, string $sSearchId = null): ?DesignElement
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
+		return false;
+	}
+
+	/**
+	 * Find the child node matching the given node.
+	 * UNSAFE: may return nodes marked as _alteration="removed"
+	 * A method with the same signature MUST exist in MFDocument for the recursion to work fine
+	 *
+	 * @param DesignElement $oRefNode The node to search for
+	 * @param string $sSearchId substitutes to the value of the 'id' attribute
```
Molkobain:
```suggestion
	 * @param null|string $sSearchId substitutes to the value of the 'id' attribute
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
+	}
+
+
+	/**
+	 * Find the child node matching the given node.
+	 * UNSAFE: may return nodes marked as _alteration="removed"
+	 * A method with the same signature MUST exist in MFDocument for the recursion to work fine
+	 *
+	 * @param DesignElement $oRefNode The node to search for
+	 * @param string $sSearchId substitutes to the value of the 'id' attribute
```
Molkobain:
```suggestion
```

core/designdocument.class.inc.php
```diff
@@ -307,4 +321,122 @@ public function GetChildText($sTagName, $sDefault = null)
...
+
+	/**
+	 * Find the child node matching the given node under the specified parent.
+	 * UNSAFE: may return nodes marked as _alteration="removed"
+	 *
+	 * @param \DOMNode $oParent
+	 * @param DesignElement $oRefNode
+	 * @param string|null $sSearchId
+	 *
+	 * @return \DOMNodeList|false|mixed
```
Molkobain, line 409:
```suggestion
	 * @return \DOMNodeList|false|mixed
	 * @since 3.1.2 3.2.0 N°6974
```

js/icon_select.js
Molkobain, line 1: I think this should not be here, it's related to another ticket.

Molkobain, line 1: I think this was from another ticket. Either name its commit accordingly or remove it from the PR

core/designdocument.class.inc.php
eespie, line 54:
```suggestion
	/** To fix DOMNode::getLineNo() ref https://www.php.net/manual/en/domnode.getlineno.php */
	public const XML_PARSE_BIG_LINES = 4194304;
```

PR #604, opened 2024-01-25
N°7172 HTML markup for n:n set properties

Timmy38:

- [ ] Would a unit test be relevant and have I added it?


Molkobain:
Discussed with Timothée, constant will be renamed to something like `ENUM_INPUT_TYPE_TAGSET_XXX` to reflect that the input is a tagset-like but with a different approach.

Molkobain (approved):

PR #609, opened 2024-02-05
N°7244 Fix setup ContextTag init

piRGoif:
Initial use case was an implementation of iEventServiceSetup::RegisterEventsAndListeners where we wanted to exclude based on \ContextTag::TAG_SETUP (EventService::RegisterListener has a context parameter for inclusion, but don't have one for exclusions)
We saw that the setup context wasn't present, especially during the call made in \ApplicationInstaller::DoUpdateDBSchema (call is made when initializing the data model)

With this PR the context init is moved on top of the methods.

Also, a context init is now done in the \WizardController::Run method.

Sample impl to reproduce : 

```php
class LoggingEventServiceSetup implements \Combodo\iTop\Service\Events\iEventServiceSetup
{
    public function RegisterEventsAndListeners()
    {
      IssueLog::Info(__CLASS__.'::'.__METHOD__.' call !', null, [
        'ContextTag_stack' => ContextTag::GetStack(),
        'stack' => debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS),
      ]);
    }
}
```

### DIFF
--- a/setup/applicationinstaller.class.inc.php
+++ b/setup/applicationinstaller.class.inc.php
@@ -685,6 +685,8 @@ class ApplicationInstaller
 	 */
 	protected static function DoUpdateDBSchema($aSelectedModules, $sModulesDir, $aParamValues, $sTargetEnvironment = '', $bOldAddon = false, $sAppRootUrl = '')
 	{
+		/** @noinspection PhpUnusedLocalVariableInspection */
+		$oContextTag = new ContextTag(ContextTag::TAG_SETUP);
 		SetupLog::Info("Update Database Schema for environment '$sTargetEnvironment'.");
 		$sMode = $aParamValues['mode'];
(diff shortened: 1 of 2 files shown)


Molkobain (approved):

Molkobain:
Accepted during technical review:
  * Pierre will update the PHPDoc to help developer know when the tag was available in module installer

Molkobain (approved):

jf-cbd (approved):

piRGoif:
Functional review : ok
Will create a ticket and then merge

piRGoif:
Thanks for the test Jimmy ! 
Side note : we saw with Jimmy there is a call remaining at the end of the setup (url /pages/ajax.document.php?operation=dic)
I'll have a look in the coming days, but this shouldn't stop this pr from merging

PR #610, opened 2024-02-08
N°7246 New dict phpunit test to check that there are no "~~" suffix on translated keys + avoid key duplications

piRGoif:
The test method uses the `<fullpath>:<linenumber>` syntax so that output can be clickable in PHPStorm, so fixing files is super easy : 
Clipboard - February 9, 2024 9 29

Many files had problems, I fixed them all.

Note that the base branch is support/3.0 : adding this test in support/2.7 was complicated as the test class was lacking lots of methods, and dict files have many inconsistencies in this branch. Plus the dict files merge from support/2.7 to support/3.0 is almost impossible :/


piRGoif:
TODO add a duplicate keys in same file test !
TODO change constants names so that it will be easier to spot in values print, something like '##ITOP_APPLICATION_SHORT##' instead of 'itop'

Hipska (approved):

piRGoif:
Modifications done, PR is ready !

Hipska (approved):

piRGoif:
Support PR review : some small changes made
Ready for functional review !

piRGoif:
@Molkobain and @rquetiez I've just pushed a modification of the test README file about markTestAsSkipped uses. Can you check ?

Molkobain (approved):

piRGoif:
Functional review : OK !
Will create a ticket and then merge.

PR #611, opened 2024-02-12
N°7157 - Allow users to unsubscribe from notification channels

steffunky:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | N°7157

## Objective (enhancement)
This feature allows users who receive multiple notifications and want to unsubscribe from these notifications or change how they receive it. They want to opt out of a specific notification, which should apply to all future notifications of the same type while keeping the existing ones unchanged. Later, if they wish to re-subscribe, they can browse their opted-out notifications and undo the unsubscribe action, affecting only future notifications and not those generated while unsubscribed. The concept of "news of the same type" encompasses objects of the same class, triggering events, or actions.

ezgif-6-0371f8a9af

This PR adds a UI to subscribe/unsubscribe from notifications that a user has already received.

- [ ] Would a unit test be relevant and have I added it? :shrug: 

- [x] Add a way to unsubscribe from Actions individually


core/action.class.inc.php
```diff
@@ -570,7 +588,7 @@ protected function _DoExecute($oTrigger, $aContextArgs, &$oLog)
 	 * @throws \Exception
 	 * @since 3.1.0 N°918
 	 */
-	protected function PrepareMessageContent($aContextArgs, &$oLog): array
+	protected function PrepareMessageContent($aContextArgs, &$oLog, $oTrigger = null): array
```
Molkobain: We need to add this in the migration notes, as it would break any custom Action derived from this one and overloading this function.

core/action.class.inc.php
```diff
@@ -355,18 +358,20 @@ public static function Init()
...
 	 *
 	 * @return string
+	 * @since 3.2.0 $oTrigger parameter added
 	 * @throws \ArchivedObjectException
 	 * @throws \CoreException
 	 * @throws \CoreUnexpectedValue
 	 * @throws \MySQLException
 	 */
-	protected function FindRecipients($sRecipAttCode, $aArgs)
+	protected function FindRecipients($sRecipAttCode, $aArgs, $oTrigger = null)
```
Molkobain: Same comment as for ProcessMessageContent

Molkobain (approved):
Code is alright for me, you can merge so QA team can test.
But I would like us to add some integration tests to ensure subscription mechanism stability. Will discuss it with Stephen on monday.

Molkobain:
_Just forced-push to rebase branch and resolve conflicts_

core/action.class.inc.php
steffunky: Yes it indeed broke approval Actions for example

PR #616, opened 2024-02-15
N°7163 - Avoid having an empty list when "items per page" is set to 0 as it makes no sense

accognet:
New protections.
If the "items per page" value is not correct, it reverts to the default value

### DIFF
--- a/css/backoffice/components/datatable/_datatableconfig.scss
+++ b/css/backoffice/components/datatable/_datatableconfig.scss
@@ -23,7 +23,8 @@ $ibo-datatableconfig--settings-panel--option--margin-right: $ibo-spacing-200 !de
 .ibo-datatableconfig--attributes-panel--per-page--input{
   margin: $ibo-datatableconfig--attributes-panel--per-page--input--margin-y $ibo-datatableconfig--attributes-panel--per-page--input--margin-x;
   max-width: 4em;
-  border: 1px solid $ibo-color-grey-500;
+  @extend .ibo-input;
+  display: unset;
 }
 
 .ibo-datatableconfig--settings-panel .ibo-panel--body{
--- a/css/backoffice/components/datatable/_datatableconfig.scss
+++ b/css/backoffice/components/datatable/_datatableconfig.scss
@@ -22,6 +22,8 @@ $ibo-datatableconfig--settings-panel--option--margin-right: $ibo-spacing-200 !de
 }
 .ibo-datatableconfig--attributes-panel--per-page--input{
   margin: $ibo-datatableconfig--attributes-panel--per-page--input--margin-y $ibo-datatableconfig--attributes-panel--per-page--input--margin-x;
+  max-width: 4em;
+  border: 1px solid $ibo-color-grey-500;
 }
 
 .ibo-datatableconfig--settings-panel .ibo-panel--body{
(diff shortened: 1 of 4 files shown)


Hipska (approved):

js/dataTables.settings.js
```diff
@@ -186,6 +186,10 @@ $(function () {
 					var sSortDirection = 'asc';
 					var oColumns = $('#datatable_dlg_'+this.options.sListId).find(':itop-fieldsorter').fieldsorter('get_params');
 					var iPageSize = parseInt($('#datatable_dlg_'+this.options.sListId+' input[name="page_size"]').val(), 10);
+					if (isNaN(iPageSize) || iPageSize <= 0) {
```
Molkobain, line 190:
```suggestion
					// Fallback to default page size in case of invalid number
					if (isNaN(iPageSize) || iPageSize <= 0) {
```

Molkobain (approved):

Molkobain (changes requested):
We should also protect the "configure this list" input so we can't pass invalid value. To do so, just change the input type from `type="text"` to `type="number"` and add a `min="1"` attribute to avoid setting 0 and below.

Protection / fallback as already implemented by the PR remain mandatory though, as users can have an invalid value in their preferences.

Hipska (approved):

Hipska:
0 could be interpreted as "no limit"?

css/backoffice/components/datatable/_datatableconfig.scss
```diff
@@ -22,6 +22,8 @@ $ibo-datatableconfig--settings-panel--option--margin-right: $ibo-spacing-200 !de
 }
 .ibo-datatableconfig--attributes-panel--per-page--input{
   margin: $ibo-datatableconfig--attributes-panel--per-page--input--margin-y $ibo-datatableconfig--attributes-panel--per-page--input--margin-x;
+  max-width: 4em;
+  border: 1px solid $ibo-color-grey-500;
```
Molkobain: Try instead, it will bring everything an input is supposed to have.

```suggestion
  @extend .ibo-input;
```

If it doesn't work as expected, use this instead. I think partials dependencies will be ok.

```suggestion
  border: 1px solid $ibo-input--border-color;
```

Molkobain (changes requested):

css/backoffice/components/datatable/_datatableconfig.scss
accognet: with a slightly different patch, I have this display
image

css/backoffice/components/datatable/_datatableconfig.scss
```diff
@@ -22,6 +22,9 @@ $ibo-datatableconfig--settings-panel--option--margin-right: $ibo-spacing-200 !de
 }
 .ibo-datatableconfig--attributes-panel--per-page--input{
   margin: $ibo-datatableconfig--attributes-panel--per-page--input--margin-y $ibo-datatableconfig--attributes-panel--per-page--input--margin-x;
+  max-width: 4em;
+  @extend .ibo-input;
+  display: unset;
```
Molkobain:
```suggestion
  display: initial;
```

We usually put `initial` instead to restore the default behavior.

Molkobain (changes requested):

css/backoffice/components/datatable/_datatableconfig.scss
Molkobain: Once this is fixed, you can:
- Rebase on support/3.2
- Change target branch to support/3.2
- Sqaush and merge (mind to keep the PR number)

PR #620, opened 2024-02-26
N°7288 - Fix page crash due to unescaped characters in relations row actions

Molkobain:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | N°7288

The page hangs, javascript related actions can't be performed anymore (e.g.: changing an object's tab)
The following error message appears in the JS console:
```
Uncaught SyntaxError: Invalid Unicode escape sequence (at
```

1. On iTop 3.1.1
2. With PHP 8.1.0
3. Edit a Person and append "\utils" to their name or surname (it's very important that the backslash and 'u' come consecutively)
4. Save the object
5. See the error message in the JS console and the inability to change tabs

image

## Cause (bug)

This is a continuation of the parent bug (N°6560), when "\u" is put in a string, JavaScript expects it to be a unicode character (e.g., "\u0020" for a space), but that might not always be the case.

- Add the TWIG filter | escape ('js') to strings so that all characters are correctly escaped
- Optionally, use backticks (`) instead of apostrophes (') to delimit strings. This allows for multi-line dictionary entries.

### DIFF
--- a/templates/base/components/datatable/row-actions/handler.js.twig
+++ b/templates/base/components/datatable/row-actions/handler.js.twig
@@ -23,16 +23,16 @@
         {% if aAction.confirmation is defined %}
 
             // Prepare confirmation title
-            let sTitle = '{{ 'UI:Datatables:RowActions:ConfirmationDialog'|dict_s  }}';
+            let sTitle = `{{ 'UI:Datatables:RowActions:ConfirmationDialog'|dict_s|escape('js') }}`;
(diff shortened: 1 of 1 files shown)


bdalsass (approved):

Hipska:
Thanks for the very good description of the PR! 👍

Molkobain:
> Thanks for the very good description of the PR! 👍

You're welcome!
BTW, @steffunky goes the extra mile as well :)

PR #622, opened 2024-02-27
N°7344 rest.php core/get : add try/catch around query execution

piRGoif:
For example with this json_data containing an invalid OQL (single quote after class name) : 

```json
{
   "operation": "core/get",
   "class": "Person",
   "key": "SELECT Person'", "limit": "10", "page": "1",
   "output_fields":"friendlyname"
}
```

Before PR you would get as a result : 

```json
{
    "code": 100,
    "message": "Error: Unexpected input at line 1: '"
}
```

And after PR : 

```json
{
    "code": 100,
    "message": "Error: Query failed to execute: query = SELECT Person', exception_class = Exception, exception_message = Unexpected input at line 1: '"
}
```

### DIFF
--- a/application/applicationextension.inc.php
+++ b/application/applicationextension.inc.php
@@ -1712,7 +1712,7 @@ class RestUtils
             try {
                 $oSearch = DBObjectSearch::FromOQL($key);
             } catch (Exception $e) {
-                throw new CoreOqlException('json_data key parameter: query failed to execute', [
+                throw new CoreOqlException('Query failed to execute', [
                         'query' => $key,
                         'exception_class' => get_class($e),
                         'exception_message' => $e->getMessage(),
--- a/application/applicationextension.inc.php
+++ b/application/applicationextension.inc.php
@@ -1709,8 +1709,16 @@ class RestUtils
 		elseif (is_string($key))
 		{
 			// OQL
-			$oSearch = DBObjectSearch::FromOQL($key);
-		}
+            try {
+                $oSearch = DBObjectSearch::FromOQL($key);
+            } catch (Exception $e) {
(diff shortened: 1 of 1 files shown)


application/applicationextension.inc.php
```diff
@@ -1709,8 +1709,16 @@ public static function GetObjectSetFromKey($sClass, $key, $iLimit = 0, $iOffset
 		elseif (is_string($key))
 		{
 			// OQL
-			$oSearch = DBObjectSearch::FromOQL($key);
-		}
+            try {
+                $oSearch = DBObjectSearch::FromOQL($key);
+            } catch (Exception $e) {
+                throw new CoreOqlException('json_data key parameter: query failed to execute', [
```
Hipska: I think a bit shorter message is fine as well:
```suggestion
                throw new CoreOqlException('Query failed to execute', [
```

piRGoif: This is an option indeed
Now you can be really confused about the message (it is an iTop error, does one of my parameters triggers it ?)
I chose a detailed error message so that rest.php user will now exactly what to fix, but stating "the query inside key param in json_data" might be a bit much.
We'll see during the reviews.

Molkobain (approved):

Molkobain:

Hipska (approved):

application/applicationextension.inc.php
piRGoif: Hello,
During review we decided to go on with your proposition , as the throw is in a method that doesn't know about the json_data structure, and this message is simple and gives the information needed by the consumer.

Molkobain:

PR #623, opened 2024-02-28
N°7697 - Add method to rename DB table during setup

Hipska:
## Base information

## Objective

To have a helper method that can rename a table in DB during setup.

~~The helper method can silently ignore the renaming if the origin table doesn't exist or the destination table already exists. This to avoid errors when running the setup again after the migration has ben done before.~~
EDIT: This part has been rejected by Combodo technical review..

## Proposed solution

Creation of `ModuleInstallerAPI::RenameTableInDB`.

- [x] Is the PR clear and detailled enough so anyone can understand digging in the code?

### DIFF
--- a/setup/moduleinstaller.class.inc.php
+++ b/setup/moduleinstaller.class.inc.php
@@ -291,4 +291,47 @@ abstract class ModuleInstallerAPI
 		CMDBSource::CacheReset($sDstTable);
 	}
 
+	/**
+	 * Rename a table providing:
+	 * - The original name exists
+	 * - The destination name does not exist
+	 *
+	 * @param string $sOrigTable
+	 * @param string $sDstTable
+	 *
+	 * @return void
+	 * @throws CoreException
+	 * @throws CoreUnexpectedValue
+	 * @throws MySQLException
+	 */
+	public static function RenameTableInDB(string $sOrigTable, string $sDstTable)
+	{
+		if ($sOrigTable == $sDstTable)
+		{
+			throw new CoreUnexpectedValue("Origin table and destination table are the same");
+		}
+
+		if (!MetaModel::DBExists(false))
+		{
+			// Install from scratch, no migration
+			return;
+		}
+
+		if (!CMDBSource::IsTable($sOrigTable))
+		{
+			SetupLog::Warning(sprintf('Rename table in DB - Origin table %s doesn\'t exist', $sOrigTable));
+			return;
+		}
+
+		if (CMDBSource::IsTable($sDstTable))
+		{
+			SetupLog::Warning(sprintf('Rename table in DB - Destination table %s already exists', $sDstTable));
+			return;
+		}
+
+		$sQueryRename = sprintf(/** @lang MariaDB */ "RENAME TABLE `%s` TO `%s`;", $sOrigTable, $sDstTable);
+		CMDBSource::Query($sQueryRename);
+
+		CMDBSource::CacheReset($sOrigTable);
+	}
 }
--- a/tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php
+++ b/tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php
@@ -241,4 +241,30 @@ SQL
 		$this->assertEquals('from table 2', $sFromTable2Data, "Data was not moved as expected");
 	}
 
+	/**
+	 * Test that the table has been renamed
+	 *
+	 * @covers ModuleInstallerAPI::RenameTableInDB
+	 *
+	 * @return void
+	 * @throws \CoreException
+	 * @throws \MySQLException
+	 */
+	public function testRenameTableInDB()
+	{
+		$sOrigTable = MetaModel::DBGetTable('Person');
+		$aOrigTableInfo = CMDBSource::GetTableInfo($sOrigTable);
+		$this->assertNotEmpty($aOrigTableInfo, 'Origin table does not exist');
+
+		$sDstTable = static::$sWorkTable;
+		$this->assertFalse(CMDBSource::IsTable($sDstTable), 'Work table already exists');
+
+		ModuleInstallerAPI::RenameTableInDB($sOrigTable, $sDstTable);
+
+		$this->assertEquals($aOrigTableInfo, CMDBSource::GetTableInfo($sDstTable), 'Table was not renamed');
+
+		// Revert
+		ModuleInstallerAPI::RenameTableInDB($sDstTable, $sOrigTable);
+		$this->assertEquals($aOrigTableInfo, CMDBSource::GetTableInfo($sOrigTable));
+	}
 }


setup/moduleinstaller.class.inc.php
```diff
@@ -290,4 +290,41 @@ public static function MoveColumnInDB($sOrigTable, $sOrigColumn, $sDstTable, $sD
...
+		if (!CMDBSource::IsTable($sOrigTable))
+		{
+			// Origin table is not present
+			return;
+		}
+
+		if (CMDBSource::IsTable($sDstTable))
+		{
+			// Destination table is already present
+			return;
```
Molkobain: Maybe in case the destination table already exists we should put something in the log? It would help troubleshooting why a table remains in the DB whilst we expected it to be renamed.

tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php
```diff
@@ -128,4 +128,26 @@ public function MoveColumnInDB_IgnoreExistingDstColumnParamProvider(): array
...
+		$sOrigTable = MetaModel::DBGetTable('Person');
+		$aOrigTableInfo = CMDBSource::GetTableInfo($sOrigTable);
+		$this->assertNotEmpty($aOrigTableInfo, 'Origin table does not exist');
+
+		$sDstTable = static::$sWorkTable;
+		$this->assertFalse(CMDBSource::IsTable($sDstTable), 'Work table already exists');
+
+		ModuleInstallerAPI::RenameTableInDB($sOrigTable, $sDstTable);
+
+		$this->assertEquals($aOrigTableInfo, CMDBSource::GetTableInfo($sDstTable), 'Table was not renamed');
```
Molkobain, line 264: After checking that everything went as expected, you should rename the table back to its original name, otherwise you might pollute other (futures) tests.

Molkobain (changes requested):
See comment on test

setup/moduleinstaller.class.inc.php
Hipska: I took example from the `MoveColumnInDB` helper which also doesn't do this: https://github.com/Combodo/iTop/blob/187a895265eaa90f1a600a614653a92b2dc907ae/setup/moduleinstaller.class.inc.php#L261-L272


Also see what I wrote in the objective about this.

tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php
Hipska: Hmm, aren't tests in php run in their own isolated environment and teardown afterwards? https://github.com/Combodo/iTop/blob/187a895265eaa90f1a600a614653a92b2dc907ae/tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php#L28-L35

Molkobain: * PHP Unit tests are run in the "iTop" production environment (99%) unless specified otherwise (see `\Combodo\iTop\Test\UnitTest\ItopCustomDatamodelTestCase`).
  * PHP Unit tests are run in the same process, there is no isolation for performance reasons. Previously, most tests were isolated but their execution took way too much time due to the MetaModel being started everytime.
  * Also, DDL statements can't be rollbacked during a transaction as they are committed instantly.

So it's best to revert the table to its original state in order to keep it clean for other tests.
Moving the revert in the teardown method would not pollute other test classes, but it would impact other tests from the current class.

setup/moduleinstaller.class.inc.php
Molkobain: > Also see what I wrote in the objective about this.

What do you mean?

Hipska: This exactly:

>The helper method can silently ignore the renaming if the origin table doesn't exist or the destination table already exists. This to avoid errors when running the setup again after the migration has ben done before.

tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php
Hipska: Okay, added a revert.

Molkobain: Thanks

Molkobain (approved):

setup/moduleinstaller.class.inc.php
```diff
@@ -290,4 +290,41 @@ public static function MoveColumnInDB($sOrigTable, $sOrigColumn, $sDstTable, $sD
 		CMDBSource::CacheReset($sDstTable);
 	}
 
+	/**
+	 * Rename a table providing:
+	 * - The original name exists
+	 * - The destination name does not exist
+	 *
+	 * @param $sOrigTable string
+	 * @param $sDstTable string
```
piRGoif:
```suggestion
	 * @param string $sOrigTable
	 * @param string $sDstTable
```

setup/moduleinstaller.class.inc.php
```diff
@@ -290,4 +290,41 @@ public static function MoveColumnInDB($sOrigTable, $sOrigColumn, $sDstTable, $sD
...
+	 *
+	 * @return void
+	 * @throws CoreException
+	 * @throws MySQLException
+	 */
+	public static function RenameTableInDB(string $sOrigTable, string $sDstTable)
+	{
```
piRGoif, line 308: Maybe check that orig and dest aren't the same ?

setup/moduleinstaller.class.inc.php
```diff
@@ -290,4 +290,41 @@ public static function MoveColumnInDB($sOrigTable, $sOrigColumn, $sDstTable, $sD
...
+		{
+			// Install from scratch, no migration
+			return;
+		}
+
+		if (!CMDBSource::IsTable($sOrigTable))
+		{
+			// Origin table is not present
+			return;
+		}
```
piRGoif: Shouldn't an exception being thrown ? Like a `CoreUnexpectedValue` ?

setup/moduleinstaller.class.inc.php
```diff
@@ -290,4 +290,41 @@ public static function MoveColumnInDB($sOrigTable, $sOrigColumn, $sDstTable, $sD
...
+
+		if (CMDBSource::IsTable($sDstTable))
+		{
+			// Destination table is already present
+			return;
+		}
+
+		$sQueryRename = sprintf(/** @lang MariaDB */ "RENAME TABLE `%s` TO `%s`;", $sOrigTable, $sDstTable);
+		CMDBSource::Query($sQueryRename);
```
piRGoif, line 333: I think the @throws in the method phpdoc are because of this call ? 
Maybe we should surround with a try/catch and throw only one exception like a `MySQLException` ?

setup/moduleinstaller.class.inc.php
```diff
@@ -258,7 +258,7 @@ public static function MoveColumnInDB($sOrigTable, $sOrigColumn, $sDstTable, $sD
 			return;
 		}
 
-		if (!CMDBSource::IsTable($sOrigTable) || !CMDBSource::IsField($sOrigTable, $sOrigColumn))
```
piRGoif: Ok, we already have the table existing test in IsField()

tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php
```diff
@@ -128,4 +128,30 @@ public function MoveColumnInDB_IgnoreExistingDstColumnParamProvider(): array
...
+	/**
+	 * Test that the table has been renamed
+	 *
+	 * @covers \ModuleInstallerAPI::MoveColumnInDB
+	 *
+	 * @return void
+	 * @throws \CoreException
+	 * @throws \MySQLException
+	 */
+	public function testRenameTableInDB()
```
piRGoif, line 253: I would also test : 
- orig and dest on the same existing table name
- orig and dest on the same non existing table name
- orig on a non existing table name
- dest on an existing table name

This may be easier to code using a dataprovider returning orig and dest names, and expected result ?

piRGoif (changes requested):
Great work, thanks !
My remarks are mostly questions, and also we may add more uses cases in the phpunit

setup/moduleinstaller.class.inc.php
Hipska: Strange, PHPStorm generated it this way I think.

setup/moduleinstaller.class.inc.php
Hipska: See the `MoveColumnInDB` method where I copied the behaviour from.

setup/moduleinstaller.class.inc.php
Hipska: What should be done with the catched `CoreException` then? IMHO those should not be trowed by `CMDBSource::Query` in the first place.

setup/moduleinstaller.class.inc.php
Hipska: Well actually, they would also not error out since the check for existing destination table would return the method..

setup/moduleinstaller.class.inc.php
piRGoif: Ok right, indeed there are only return in \ModuleInstallerAPI::MoveColumnInDB
This method was added in 2018 so not to far away, it is quite a strange choice :/ 
To be discussed during next review !

setup/moduleinstaller.class.inc.php
piRGoif: What I see is that the throws annotations in the `CMDBSource::Query` phpdoc block are wrong starting in support/2.7 : neither me or PHPStorm are finding what would throw a CoreException in this method.
I've fixed the block with da490739 !

Considering `CMDBSource::Query` will only returns a `MySQLException` (or its `MySQLHasGoneAwayException` child) no need anymore to have a try/catch. 
Can you rebase, and adapt the throws annotations in your method to have either only one `@throws MySQLException`, or both throws for `MySQLException` and `MySQLHasGoneAwayException` ?

Molkobain (changes requested):
Technical review:
  * Unit test:
    * Move table revert in tearDown() method for better robustness in case there is an issue within the test method
  * New method:
    * Add `SetupLog` message in exit conditions (except the `!MetaModel::DBExists(false)`). See `RenameEnumValueInDB()` for a real example, `MoveColumnInDB()` is a bad one.

setup/moduleinstaller.class.inc.php
piRGoif: Closing conversation, we discussed during tech review and this is no longer valid.

setup/moduleinstaller.class.inc.php
Hipska: It turns out the `CoreException` comes from `MetaModel::DBExists`. So it needs to stay even after merge with current develop branch.

Hipska:
Okay @Molkobain, but earlier you said this:
> So it's best to revert the table to its original state in order to keep it clean for other tests. Moving the revert in the teardown method would not pollute other test classes, but it would impact other tests from the current class.

So I understand that it might impact other tests and thus it needed to be added within the test?

About the exit conditions, I thought this was also discussed and agreed before?
> Ok right, indeed there are only return in \ModuleInstallerAPI::MoveColumnInDB This method was added in 2018 so not to far away, it is quite a strange choice :/ To be discussed during next review !

Followed by
> Closing conversation, we discussed during tech review and this is no longer valid.

Do I understand this wrong?

Hipska:
@Molkobain?

Molkobain:
Hello Thomas, nevermidn about the revert table comment, leave it as it is.
Just missing the `SetupLog` messages, then alright for me. :)

Molkobain (approved):

Molkobain:
Thanks!

Molkobain (approved):

Molkobain:
Done, thanks a lot Thomas!

PR #624, opened 2024-02-28
N°5120 - Bring unattended-install CLI script in iTop repository

odain-cbd:
This script is useful to install iTop via CLI. this is already used in our CI and SaaS tools. 

We decided to provided it directly in iTop github repository. as long as it is protected from any HTTP access.


setup/unattended-install/README.md
```diff
@@ -0,0 +1,3 @@
+Unattended-install : this scripts allows to script iTop install and / or updates.
```
Hipska:
```suggestion
# Unattended-install

This scripts allows to script iTop install and / or updates.
```

jbostoen:
Could indeed be good to think about https://github.com/Combodo/iTop/pull/223 as well

setup/unattended-install/README.md
Molkobain: +1 would look nicer :)

setup/unattended-install/unattended-install.php
```diff
@@ -0,0 +1,314 @@
+<?php
+$bBypassMaintenance = true;
+require_once(dirname(__FILE__, 3) . '/approot.inc.php');
```
Molkobain, line 3:
```suggestion
require_once(dirname(__FILE__, 2) . '/approot.inc.php');
```

Shouldn't it be 2 instead?

setup/unattended-install/unattended-install.php
```diff
@@ -0,0 +1,314 @@
...
+	$aDBXmlSettings['server'] = $oConfig->Get('db_host');
+	$aDBXmlSettings['user'] = $oConfig->Get('db_user');
+	$aDBXmlSettings['pwd'] = $oConfig->Get('db_pwd');
+	$aDBXmlSettings['name'] = $oConfig->Get('db_name');
+	$aDBXmlSettings['prefix'] = $oConfig->Get('db_subname');
+	$oParams->Set('database', $aDBXmlSettings);
+	$oParams->Set('url', $oConfig->Get('app_root_url'));
+} else {
+	$aDBXmlSettings = $oParams->Get('database', array());
+}
```
Molkobain:
```suggestion
if ($bUseItopConfig){
	$oConfig = new Config(APPROOT . "/conf/$sTargetEnvironment/config-itop.php");
	$aDBXmlSettings['server'] = $oConfig->Get('db_host');
	$aDBXmlSettings['user'] = $oConfig->Get('db_user');
	$aDBXmlSettings['pwd'] = $oConfig->Get('db_pwd');
	$aDBXmlSettings['name'] = $oConfig->Get('db_name');
	$aDBXmlSettings['prefix'] = $oConfig->Get('db_subname');
	$oParams->Set('database', $aDBXmlSettings);
	$oParams->Set('url', $oConfig->Get('app_root_url'));
}
$aDBXmlSettings = $oParams->Get('database', array());

```

setup/unattended-install/unattended-install.php
```diff
@@ -0,0 +1,314 @@
...
+	$bClean = utils::ReadParam('clean', false, true /* CLI allowed */);
+	if ($bClean)
+	{
+		echo "Cleanup mode detected.\n";
+
+		// Configuration file
+		$sConfigFile = APPCONF.$sTargetEnvironment.'/'.ITOP_CONFIG_FILE;
+		if (is_file($sConfigFile))
+		{
+			copy($sConfigFile, "$sConfigFile.backup");
```
Molkobain: What do you do with that backup file? It stays here? If everything goes fine, we should clean up.

Molkobain (changes requested):

piRGoif (changes requested):
Thanks for the modification ! 
I would just suggest moving the informations in the README.md to a phpdoc block in the unattended-install.php file header ?

piRGoif:
> Could indeed be good to think about #223 as well

@jbostoen are you referring to https://github.com/Combodo/iTop/pull/223#issuecomment-1652440721 ?

I don't recall all the discussions that happened in this PR and the modification asked in this comment on the unattended script, but I would prefer we first integrate the current version : that way the git history will be cristal clear. Any modifications could be done later in subsequent pr / commits.

setup/unattended-install/README.md
odain-cbd: if we keep README.md, it is nicer and i will keep your proposal.

Pierre proposed to move this content in unattented script php doc. it is also a good idea to reduce the amount of files to open.

setup/unattended-install/unattended-install.php
odain-cbd: when $bUseItopConfig enabled, setup is run based on db configuration settings. but there were 2 missing settings that I will take into account in next github push: db_tls_enabled / db_tls_ca

setup/unattended-install/unattended-install.php
odain-cbd: indeed it does not work with current proposal

setup/unattended-install/README.md
Hipska: If you keep this readme file, then GitHub will show this as default context when opening this `setup/unattended-install` directory. No need to do an extra click to see the content of the script in that case 😉

setup/unattended-install/unattended-install.php
odain-cbd: just discussed with Pierre. current PR should remain simple. I remove SaaS option to read db settings from itop configuration. it will come but after merging the PR (same logic applied to Jeffrey PR regarding switch_env).

setup/unattended-install/unattended-install.php
odain-cbd: removed that part also. keep it for another PR.

Hipska (approved):

setup/unattended-install/README.md
piRGoif: Yes you're right about the README being automatically displayed !

But as what we want to add is just a little phrase and mainly the wiki URL,
And because the wiki page is this sole script documentation,
I still think it would be better to add it as a PHPDoc block at the script beginning.

Having documentation as closed to the code is my preferred option...
Plus this is a particular tool, aiming at technical people for automatic installations purposes.

setup/unattended-install/unattended-install.php
Molkobain: Just this, then ok for me

Molkobain (changes requested):

setup/unattended-install/unattended-install.php
Molkobain: Discussed IRL with Olivier. He tested in real conditions with "3" and it works.

Molkobain (approved):

setup/unattended-install/unattended-install.php
Hipska: It probably is "2" when you use `__DIR__`.

Molkobain:
@odain-cbd what's the status on this PR?
Do you have an internal ticket number that we can prepend on the title?

tests/php-unit-tests/unitary-tests/setup/unattended-install/UnattendedInstallTest.php
```diff
@@ -27,8 +39,27 @@ public function testCallUnattendedInstallFromHttp(){
...
+				if (is_file("$sPath")) {
+					rename($sPath, "$sPath.back");
+				}
+			}
+
+			$sJson = $this->callUnattendedFromHttp();
+		}
 
-		$this->assertEquals("Mode CLI only", $sJson);
+		$this->assertEquals("Mode CLI only", $sJson, "even without HTTP protection, script should be called directly by HTTP");
```
Hipska: This comment looks wrong?

odain-cbd: agreed and fixed

PR #626, opened 2024-02-28
N°7302 Fix unit name in \SetupUtils::HumanReadableSize

piRGoif:
Multiple errors in the units sent in the returned value : 
- "b" (bit) instead of "B" (byte)
- Kb instead of kB (IEC 80000-13)
- Hb instead of EB (exabyte)

Also added some phpdoc as reference for the choices made + type hinting

### DIFF
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -899,14 +899,14 @@ class SetupUtils
 
     /**
      * @param float $fBytes size in raw bytes, for example 162594750464.0
-     * @return string formatted string, for example "161.62 Gb"
+     * @return string formatted string, for example "161.62 GB"
      *
      * @link https://en.wiktionary.org/wiki/petabyte petabyte PB
      * @link https://en.wiktionary.org/wiki/exabyte#English exabyte EB
      */
 	public static function HumanReadableSize($fBytes)
 	{
-        $aSizes = array('Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB');
+		$aSizes = array('Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB');
 		$index = 0;
 		while (($fBytes > 1000) && ($index < count($aSizes)))
 		{
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -898,12 +898,15 @@ class SetupUtils
 	}
 
     /**
+     * @param float $fBytes size in raw bytes, for example 162594750464.0
+     * @return string formatted string, for example "161.62 Gb"
+     *
      * @link https://en.wiktionary.org/wiki/petabyte petabyte PB
      * @link https://en.wiktionary.org/wiki/exabyte#English exabyte EB
      */
 	public static function HumanReadableSize($fBytes)
 	{
-        $aSizes = array('bytes', 'Kb', 'Mb', 'Gb', 'Tb', 'Pb', 'Eb');
+        $aSizes = array('Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB');
 		$index = 0;
 		while (($fBytes > 1000) && ($index < count($aSizes)))
 		{
--- a/setup/setuputils.class.inc.php
+++ b/setup/setuputils.class.inc.php
@@ -897,9 +897,13 @@ class SetupUtils
 		return $f;
 	}
 
+    /**
+     * @link https://en.wiktionary.org/wiki/petabyte petabyte PB
+     * @link https://en.wiktionary.org/wiki/exabyte#English exabyte EB
+     */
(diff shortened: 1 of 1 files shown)


Molkobain (approved):

setup/setuputils.class.inc.php
```diff
@@ -897,9 +897,13 @@ public static function CheckDiskSpace($sDir)
 		return $f;
 	}
 
+    /**
+     * @link https://en.wiktionary.org/wiki/petabyte petabyte PB
+     * @link https://en.wiktionary.org/wiki/exabyte#English exabyte EB
+     */
```
Hipska: Are these links really necessary?

setup/setuputils.class.inc.php
```diff
@@ -897,9 +897,13 @@ public static function CheckDiskSpace($sDir)
...
 	public static function HumanReadableSize($fBytes)
 	{
-		$aSizes = array('bytes', 'Kb', 'Mb', 'Gb', 'Tb', 'Pb', 'Hb');
+        $aSizes = array('bytes', 'Kb', 'Mb', 'Gb', 'Tb', 'Pb', 'Eb');
```
Hipska: Actually, the first one says bytes and all the others are about bits?
```suggestion
        $aSizes = array('bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB');
```

See https://en.wiktionary.org/wiki/EB#Symbol vs https://en.wiktionary.org/wiki/Eb#Symbol

Hipska (changes requested):

setup/setuputils.class.inc.php
Molkobain: Good catch!

setup/setuputils.class.inc.php
piRGoif: I wasn't very sure about the units used above terra bytes, they aren't so common :)
So having references for anyone seems like a good idea ?

setup/setuputils.class.inc.php
piRGoif: You're perfectly right ! I was wondering also why we had "b" and not "B" ?
I've just checked, values processed are from `disk_free_space()` of `filesize()`, so they are indeed in bytes.
I'll fixed that, thanks

piRGoif: Fixed in 684fa2d4

setup/setuputils.class.inc.php
```diff
@@ -897,9 +897,16 @@ public static function CheckDiskSpace($sDir)
 		return $f;
 	}
 
+    /**
+     * @param float $fBytes size in raw bytes, for example 162594750464.0
+     * @return string formatted string, for example "161.62 Gb"
```
Hipska:
```suggestion
     * @return string formatted string, for example "161.62 GB"
```

piRGoif: Fixed, thanks !

setup/setuputils.class.inc.php
```diff
@@ -897,9 +897,16 @@ public static function CheckDiskSpace($sDir)
...
+     * @param float $fBytes size in raw bytes, for example 162594750464.0
+     * @return string formatted string, for example "161.62 GB"
+     *
+     * @link https://en.wiktionary.org/wiki/petabyte petabyte PB
+     * @link https://en.wiktionary.org/wiki/exabyte#English exabyte EB
+     */
 	public static function HumanReadableSize($fBytes)
 	{
-		$aSizes = array('bytes', 'Kb', 'Mb', 'Gb', 'Tb', 'Pb', 'Hb');
+		$aSizes = array('Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB');
```
Hipska: Actually it can be lowercase. Also, the k in kilobytes is also lowercase
```suggestion
		$aSizes = array('bytes', 'kB', 'MB', 'GB', 'TB', 'PB', 'EB');
```
See https://en.wikipedia.org/wiki/Kilobyte

Hipska (changes requested):

setup/setuputils.class.inc.php
piRGoif: Yes indeed, thanks !
About bytes VS Bytes : 
https://en.wiktionary.org/wiki/byte#English
https://en.wiktionary.org/wiki/Byte

piRGoif: Fixed in 590c8ce3

Hipska (approved):

Molkobain (approved):

Molkobain:

Molkobain:

setup/setuputils.class.inc.php
piRGoif: Closing conversation as in the reviews there was no objection to keep those url references

PR #628, opened 2024-03-06
In PHPUnit context use php executable used by the running test

piRGoif:
Problem occurred in ImportTest

## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | ❌

When running `ImportTest`, scripts are launched using the `php` command, and not the one configured to run the test. Therefore we could have a Composer platformCheck failure !

The StatusTest class uses the same config parameter, but doesn't use the Composer plateformCheck.

Current workaround : modify the `php_path` config parameter.

Modify `utils::ExecITopScript` so that in a PHPUnit context (ITOP_PHPUNIT_RUNNING_CONSTANT_NAME constant), we will use the already running php executable (PHP_BINARY constant) instead of the config parameter.

Note : base branch is support/3.0 as the ITOP_PHPUNIT_RUNNING_CONSTANT_NAME  constant doesn't exist in support/2.7.

- [✔] I have performed a self-review of my code
- [❌] I have tested all changes I made on an iTop instance
- [✔] I have added a unit test, otherwise I have explained why I couldn't
- [✔] Is the PR clear and detailed enough so anyone can understand digging in the code?

## TODO

- [ ] technical review
- [ ] functional review
- [ ] create bug (don't forget to mention that the datasynchro must be tested !)


Molkobain (approved):

Molkobain:
When creating the corresponding bug in the tracker, please mention that datasynchro must be tested.

piRGoif:
Seen during support PR review : we saw apcEmulationTest was breaking... Because this test class is loading a php file to mock : 
- MetaModel::GetConfig
- utils::GetCachePath
- apc_cache_emulation.max_entries config param value
So as we added a load of the utils class in ItopTestCase, it is now breaking because we have 2 utils class :(

A possibility is to mock differently : using a different env for utils::GetCachePath to return a different path, and setting apc_cache_emulation.max_entries in-memory.

PR #630, opened 2024-03-07
7326-JS error in editing object when a tab with list is deleted

accognet:
With this delta : 
image

we have a js message:
image


Molkobain (changes requested):
Seems like you are mixing 2 different fixes in 1 PR.

accognet:
> Seems like you are mixing 2 different fixes in 1 PR.

No, it's only one fix

Molkobain:
Then I'm not sure it's the right approach, it will be discussed during support PRs review.

PR #632, opened 2024-03-12
N°7336 DeprecatedCallsLog robustness

piRGoif:
When a call to a deprecated method is done outside of a class method, then DeprecatedCallsLog will generate php notices.

Was seen in nightly builds on iTop 3.2 when calling itop-object-copier copy.php : this script contains calls to \Combodo\iTop\Application\WebPage\WebPage::add_linked_script which are deprecated.

What was done : 
* Extract DeprecatedCallsLog message generation to a method
* Add a test with 3 uses cases (call directly in a file, in a method in a file, in a class method)
* Fix the notice

### DIFF
--- a/core/log.class.inc.php
+++ b/core/log.class.inc.php
@@ -3,7 +3,7 @@
 //
 //   This file is part of iTop.
 //
-//   iTop is free software; you can redistribute it and/or modify	
+//   iTop is free software; you can redistribute it and/or modify
 //   it under the terms of the GNU Affero General Public License as published by
 //   the Free Software Foundation, either version 3 of the License, or
 //   (at your option) any later version.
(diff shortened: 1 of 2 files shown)


core/log.class.inc.php
```diff
@@ -1259,6 +1247,33 @@ public static function NotifyDeprecatedPhpMethod(?string $sAdditionalMessage = n
 		static::Warning($sMessage, self::ENUM_CHANNEL_PHP_METHOD);
 	}
 
+	private static function GetMessageFromStack(array $aDebugBacktrace): string
+	{
+		$iStackDeprecatedMethodLevel = 1; // level 0 = current method, level 1 = method containing the `NotifyDeprecatedPhpMethod` call
+		$sDeprecatedObject = $aDebugBacktrace[$iStackDeprecatedMethodLevel]['class'];
```
odain-cbd: why not sharing same defensive code in lines 1253 and 1262? via a refactor inside a dedicated function. it would help lisibility as well.

same remark regarding defensive code for 'class' / 'file' index.

core/log.class.inc.php
```diff
@@ -1259,6 +1247,33 @@ public static function NotifyDeprecatedPhpMethod(?string $sAdditionalMessage = n
...
+		$sDeprecatedMethod = $aDebugBacktrace[$iStackDeprecatedMethodLevel]['function'];
+		$sCallerFile = $aDebugBacktrace[$iStackDeprecatedMethodLevel]['file'];
+		$sCallerLine = $aDebugBacktrace[$iStackDeprecatedMethodLevel]['line'];
+		$sMessage = "Call to {$sDeprecatedObject}::{$sDeprecatedMethod} in {$sCallerFile}#L{$sCallerLine}";
+
+		$iStackCallerMethodLevel = $iStackDeprecatedMethodLevel + 1; // level 2 = caller of the deprecated method
```
odain-cbd: it seems you print level 1 and level 2 from $aStack. it would be simpler to read
$iStackCallerMethodLevel =2;

instead of incrementation

Molkobain (approved):

core/log.class.inc.php
piRGoif: L1253 this is the level containing the call to \DeprecatedCallsLog::NotifyDeprecatedPhpMethod
It should ever be in a class method ? 
It is easy to add defensive tests, but may complexify the code ?

piRGoif:
Hello, just pushed a modification, can you check again, especially @odain-cbd ?

piRGoif:
Discussion done with Olivier.
@Molkobain can you have a final look ?

core/log.class.inc.php
piRGoif: Closing as I discussed directly with Olivier

core/log.class.inc.php
piRGoif: Closing as I discussed directly with Olivier

piRGoif:
Discussed with Guillaume, he's ok, and we can merge.

PR #633, opened 2024-03-13
N°7348 - Enable to param maximum depth of impact analysis in function Ticket::UpdateImpactedItems

accognet:
In this kind of delta
image

we want to limit depth of impact analysis to 1 or 2. It's actually hard coded to 10.

### DIFF
--- a/datamodels/2.x/itop-tickets/main.itop-tickets.php
+++ b/datamodels/2.x/itop-tickets/main.itop-tickets.php
@@ -133,7 +133,15 @@ class ResponseTicketTTR extends ResponseTicketSLT implements iMetricComputer
 class _Ticket extends cmdbAbstractObject
 {
 
-	public function UpdateImpactedItems()
+	/**
+	 * @param $iMaxDepth maximum depth of impact analysis
+	 *
+	 * @return void
+	 * @throws \ArchivedObjectException
+	 * @throws \CoreException
+	 * @throws \CoreUnexpectedValue
+	 */
+	public function UpdateImpactedItems($iMaxDepth=10)
 	{
 		require_once(APPROOT.'core/displayablegraph.class.inc.php');
 		/** @var ormLinkSet $oContactsSet */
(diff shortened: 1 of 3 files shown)


Hipska:
How and where would that be configured then?

accognet:
Directly in the delta on the function call

datamodels/2.x/itop-tickets/main.itop-tickets.php
```diff
@@ -133,7 +133,15 @@ public function ComputeMetric($oObject)
 class _Ticket extends cmdbAbstractObject
 {
 
-	public function UpdateImpactedItems()
+	/**
+	 * @param $iMaxDepth maximum depth of impact analysis
```
Hipska:
```suggestion
	 * @param int $iMaxDepth maximum depth of impact analysis
```

Hipska (approved):

datamodels/2.x/itop-tickets/main.itop-tickets.php
```diff
@@ -133,7 +133,15 @@ public function ComputeMetric($oObject)
...
-	public function UpdateImpactedItems()
+	/**
+	 * @param int $iMaxDepth maximum depth of impact analysis
+	 *
+	 * @return void
+	 * @throws \ArchivedObjectException
+	 * @throws \CoreException
+	 * @throws \CoreUnexpectedValue
+	 */
+	public function UpdateImpactedItems($iMaxDepth=10)
```
Molkobain:
```suggestion
	public function UpdateImpactedItems(int $iMaxDepth = 10)
```

Molkobain (approved):

Molkobain:
Accepted during support PRs review.
A migration check has been done, as this change will break existing overloads of this method.

During functional review, we will discuss whether it's ok or we should rather use a config param.

Molkobain:
Functional review: Accepted for iTop 3.3. Not 3.2 as there are changes requested and we are too close to the code freeze.

Changes to do:
  * Rather than changing the prototype of the existing method, add a new `_Ticket::GetImpactAnalysisMaxDepth()`.
    *  It will avoid breaking existing overloads of the current method
    * It will make customization more clear for support when it is juste about changing the depth and not reworking the whole method
  * Add unit tests based on the test cases reworked by Romain / Eric lately.
    * For the unit tests, we might have to add a config param on the module so it can be changed in the test. We'll have to see if we can find a way to avoid that.

Molkobain:
_Previous message means that once this PR is accepted during "Support PRs review" or "Technical review", it can be merged without going to "Functional review" again_

rquetiez:
A test has been added, showing that "max depth" has an effect, and documenting the default = 10
This can be merged

rquetiez (approved):

PR #634, opened 2024-03-13
Fix incorrect namespace (creates problem in unattended install / toolkit)

jf-cbd:


eespie (approved):

Molkobain:
Looking at the code, it looks like the namesoace was indeed wrong, but did you find why it's causing an issue only on the tailored package? It works fine with the community package.

jf-cbd:
In iTop 3.2 the namespace has been fixed by @bdalsass (https://github.com/Combodo/itop/commit/be99a7c02da23c891649e761f496b8c88017c870).
Closing the PR.

PR #636, opened 2024-03-20
N°7371 -  Remembers the choice done in the popup that appears when saving public and private logs at the same time

accognet:
Internal

### DIFF
--- a/js/layouts/activity-panel/activity-panel.js
+++ b/js/layouts/activity-panel/activity-panel.js
@@ -935,6 +935,7 @@ $(function()
 			_SaveSubmitConfirmationPref: function()
 			{
 				// Note: We have to send the value as a string because of the API limitation
+				this.options.show_multiple_entries_submit_confirmation = false;
 				SetUserPreference('activity_panel.show_multiple_entries_submit_confirmation', 'false', true);
 			},
 			/**
--- a/js/layouts/activity-panel/activity-panel.js
+++ b/js/layouts/activity-panel/activity-panel.js
@@ -32,7 +32,6 @@ $(function()
 					lock_token: null,
 					lock_watcher_period: 30,        // Period (in seconds) between lock status update, uses the "activity_panel.lock_watcher_period" config. param.
 					lock_endpoint: null,
-					show_multiple_entries_submit_confirmation: true,
 					save_state_endpoint: null,
 					last_loaded_entries_ids: {},
 					load_more_entries_endpoint: null,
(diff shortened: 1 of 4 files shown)


js/layouts/activity-panel/activity-panel.js
```diff
@@ -916,6 +916,7 @@ $(function()
 			_SaveSubmitConfirmationPref: function()
 			{
 				// Note: We have to send the value as a string because of the API limitation
+				this.options.show_multiple_entries_submit_confirmation = false;
```
Molkobain: I would rather put this after the SetUserPreference, so the UI is updated only if the call to the server went well.

Molkobain (approved):

Molkobain:
Please set a PR title explaining the symptom; not a generic one.

Hipska:
> Please set a PR title explaining the symptom; not a generic one.

Yes, as it now looks like this PR introduces that bug..

js/utils.js
```diff
@@ -213,10 +213,15 @@ function SetUserPreference(sPreferenceCode, sPrefValue, bPersistent) {
...
-	oUserPreferences[sPreferenceCode] = sPrefValue;
 	if (bPersistent && (sPrefValue != sPreviousValue)) {
-		ajax_request = $.post(GetAbsoluteUrlAppRoot()+'pages/ajax.render.php',
-			{operation: 'set_pref', code: sPreferenceCode, value: sPrefValue}); // Make it persistent
+		return $.post(GetAbsoluteUrlAppRoot()+'pages/ajax.render.php',
+			{operation: 'set_pref', code: sPreferenceCode, value: sPrefValue}, function (data) {
+			//do nothing
+			}).done(function() {
+				oUserPreferences[sPreferenceCode] = sPrefValue;
+			}); // Make it persistent
```
Molkobain: Comment is misleading. Making the POST call actually persist the preference, so it doesn't do "nothing" and it's not L221 that persists it (it updates local state).

js/layouts/activity-panel/activity-panel.js
```diff
@@ -474,7 +473,9 @@ $(function()
 
 				let sStimulusCode = (undefined !== oData.stimulus_code) ? oData.stimulus_code : null
 				// If several entry forms filled, show a confirmation message
-				if ((true === this.options.show_multiple_entries_submit_confirmation) && (Object.keys(this._GetEntriesFromAllForms()).length > 1)) {
+				if ((GetUserPreference('activity_panel.show_multiple_entries_submit_confirmation',true) === true
+					|| GetUserPreference('activity_panel.show_multiple_entries_submit_confirmation', true) === "true")
+					&& (Object.keys(this._GetEntriesFromAllForms()).length > 1)) {
```
Molkobain: This feels messy, if the return type isn't strict maybe you should wrap this in a helper to always return a boolean.

Molkobain (changes requested):

js/layouts/activity-panel/activity-panel.js
accognet: I need to find what the "real" value is. I did a lot of testing and changed the value to 'true' many times that's why this test doesn't look good.

Molkobain:
image

Careful, rebase of the branch wasn't done correctly, you have 237 commits coming to support/3.2 ! :)

PR #637, opened 2024-03-20
N°7331 Fix ListDeniedButStillPresentFoldersAbsPaths

piRGoif:
Was only checking dirs in the denied list that were IsQuestionnable.
In consequence some dirs were in the denied list but never removed !
Eg `ace-builds/src-min-noconflict`

Now all dirs in the denied list and on disk will generate a removal.


Molkobain:
Actually done last week during some libs upgrade.

PR #638, opened 2024-03-22
N°7385 - Trigger on mention executed even if it's not a mention with @

accognet:
Internal

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -3110,7 +3110,7 @@ TXT
 		$sText = html_entity_decode($sText);
 
         $aMentionAllowedClasses = MetaModel::GetConfig()->Get('mentions.allowed_classes');
-        preg_match_all('/<a\s*([^>]*)data-object-class="([^"]*)"\s.*data-object-key="([^"]*)"/Ui', $sText, $aMentionMatches);
+        preg_match_all('/<a\s*([^>]*)data-object-class="([^"]*)"\s.*data-object-key="([^"]*)"\s*([^>]*)>(.*)<\/a>/Ui', $sText, $aMentionMatches);
 		foreach ($aMentionMatches[0] as $iMatchIdx => $sCompleteMatch) {
 			$sMatchedClass = $aMentionMatches[2][$iMatchIdx];
 			$sMatchedId = $aMentionMatches[3][$iMatchIdx];
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -521,7 +521,7 @@ class utils
 
 			// For URL
 			case static::ENUM_SANITIZATION_FILTER_URL:
-                // N°6350 - returns only valid URLs
+				// N°6350 - returns only valid URLs
 				$retValue = filter_var($value, FILTER_VALIDATE_URL);
 				break;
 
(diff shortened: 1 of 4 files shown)


Molkobain (changes requested):
**Review of the current proposition**
Mentions are not only for Contact (`@` marker by default), it can be for any kind of objects. So your current implementation breaks it for all other classes.

If we are to go further with this ticket, then we must check if the mentioned object matches the marker it begins with.

**Use case that led to this PR**
When pasting text from an iTop; if the text contains the hyperlink to an object with the same meta data as the mentions, then the mention will be triggered again even if the hyperlink doesn't start with the marker (`@` for Contact).

@v-dumas had this behaviour recently and thought it was nice. So there is obviously a discussion to be done, so we all decide what behaviour we want.

Molkobain:
I'm moving this PR to "Functional review" even though the technical one didn't occur yet, so the important point can be discussed before going any further.

Molkobain:
Functional review:
  * Expected behavior: Only trigger a mention, if the hyperlink starts with the corresponding marker (e.g. `@` for `Person`)
  * Target version: 3.2.x if code changes are not too risky, 3.3.0 otherwise. Will be determined, once PR is corrected

rquetiez:
Code review : the preceding character must be dehardcoded from '@' to the one given in the configuration, depending on the object class. Also consider a multi-byte safe function.

application/utils.inc.php
```diff
@@ -3110,19 +3110,29 @@ public static function GetMentionedObjectsFromText(string $sText): array
...
-				$aMentionedObjects[$sMatchedClass][] = $sMatchedId;
-			}
+			$sMatchedName = $aMentionMatches[1][$iMatchIdx];
+
+            $sMentionPrefix = array_search($sMatchedClass, $aMentionAllowedClasses);
+            if ($sMentionPrefix === false) {
+                continue;
+            }
+			//tests if the name starts with $sMentionPrefix (@)
+			if (str_starts_with($sMatchedName,$sMentionPrefix) === false) {
```
Molkobain: `str_starts_with()` doesn't handle multibyte caracters (non latin alphabets, emojis, ...), you should rather use `mb_strpos()`

application/utils.inc.php
```diff
@@ -3110,19 +3110,29 @@ public static function GetMentionedObjectsFromText(string $sText): array
...
-			if (!in_array($sMatchedId, $aMentionedObjects[$sMatchedClass])) {
-				$aMentionedObjects[$sMatchedClass][] = $sMatchedId;
-			}
+			$sMatchedName = $aMentionMatches[1][$iMatchIdx];
+
+            $sMentionPrefix = array_search($sMatchedClass, $aMentionAllowedClasses);
+            if ($sMentionPrefix === false) {
+                continue;
+            }
+			//tests if the name starts with $sMentionPrefix (@)
```
```suggestion
			//tests if the name starts with $sMentionPrefix (e.g. '@' for 'Contact' class)
```

Molkobain (changes requested):

application/utils.inc.php
accognet: As the first character always starts as the first byte, mb is not a problem. See specific test case with an emoji.

application/utils.inc.php
```diff
@@ -3110,19 +3110,29 @@ public static function GetMentionedObjectsFromText(string $sText): array
...
-				$aMentionedObjects[$sMatchedClass][] = $sMatchedId;
-			}
+			$sMatchedName = $aMentionMatches[1][$iMatchIdx];
+
+            $sMentionPrefix = array_search($sMatchedClass, $aMentionAllowedClasses);
+            if ($sMentionPrefix === false) {
+                continue;
+            }
+			//tests if the name starts with $sMentionPrefix (e.g. '@' for 'Contact' class)
+			if (str_starts_with($sMatchedName,$sMentionPrefix) === false) {
```
rquetiez: typographic rules: a space is expected after a coma

tests/php-unit-tests/unitary-tests/application/utilsTest.php
```diff
@@ -21,15 +21,19 @@
...
 use ormDocument;
 use utils;
 
 /**
  * @covers utils
  */
-class utilsTest extends ItopTestCase
+class utilsTest extends ItopDataTestCase
 {
+    const USE_TRANSACTION = false;
```
rquetiez: This statement is a trap in the future: adding a test altering the data will leave the DB changed.

Disabling transaction is required here because there is a "data provider" that invokes setUp(), causing a false negative error "two many transactions left open". This issue was revealed when the class inheritance was changed from `ItopTestCase` to `ItopDataTestCase`.

Solution:
- Rework the current proposal to move the new tests into a dedicated test class derived from `ItopDataTestCase`, and leave the existing test class unchanged (derived from `ItopTestCase`)
- Fix the existing test class will be done in a separate flow

tests/php-unit-tests/unitary-tests/application/utilsTest.php
```diff
@@ -653,6 +658,7 @@ public function ToAcronymProvider()
 	 */
 	public function testGetMentionedObjectsFromText($sInput, $aExceptedMentionedObjects)
 	{
+        MetaModel::GetConfig()->Set('mentions.allowed_classes', ['@' => 'Person','😊#' => 'Team']);
```
rquetiez: typographic rules: a space is expected after a coma

rquetiez (commented):
1. Two little typographic rules to fix
2. Having utilsTest derived from iTopDataTestCase seems overkill Is there an alternative?
3. My mistake on my previous review: multi-byte safe function has no sense when it comes to checking the very first character of a byte sequence. Moreover, a test has been added with an emoji as the "marker".

rquetiez (approved):
Ok for merging

PR #640, opened 2024-03-29
N°7306 - Use iTop configuration settings to run unattended installation (instead of XML file settings)

odain-cbd:
### Enhancement
Add an option to use existing configuration file settings (config-itop.php) when installing iTop application via unattended CLI.

```
php unattended-install.php --use-itop-config=1 --response_file=setup_path
```

With this option enabled (=1) you do not have to copy/paste existing settings from config-itop.php to XML setup file (for example DB parameters).  In that case setup XML provided is used to specify how iTop will be installed (fresh installation/ upgrade, which modules/extensions to install, ...).


Molkobain:
Most of this was already done with #624 didn't it? I just merged support/2.7 => develop, you should rebase your branch so only the relevant changes are present. It will be easier to review.

Also, you want this change only in the develop branch?

odain-cbd:
PR 641 is on top of this work. that one could be closed as proposed by Guillaume

PR #641, opened 2024-03-29
N°7407 - Ease iTop installation via unattended CLI by using installation.xml choices

odain-cbd:
### Enhancement
It would be nice to ease iTop installation via unattended CLI. Current PR aims at adding an option to compute extensions/modules activatation via installation.xml choices.

```
php unattended-install.php --use_itop_config=1 --response_file=setup_path --use_installation_xml=./datamodels/2.x/installation.xml
```

When using the option the modules list in _setup_path_ file are not used but computed during installation. 2 possible usecases:

- **no selected_extensions are specified in _setup_path_ file**
In that case unattended CLI will install iTop in non-ITIL mode with all possible extensions based on dependencies (like error and problem management). the list of modules inside _setup_path_ are not used at all.

NB: By the way the term _'selected_extensions'_ is misleading.  in fact this list is in fact installation choices ie further set of modules to install.

Empty _selected_extensions_ section in _setup_path_ file looks like:
```
<selected_extensions type="array"></selected_extensions>
```

- **extensions are specified in _setup_path_ file:**
In that case unattended CLI will install iTop based only on the list of installation choices provided in _selected_extensions_ section.

Example: 
with below _setup_path_  file iTop will be installed with all ITIL modules + error management modules (but without problem management that is not listed)

```
<selected_modules type="array">
    <item>itop-config-mgmt-datacenter</item>
    <item>itop-config-mgmt-end-user</item>
    <item>itop-config-mgmt-storage</item>
    <item>itop-config-mgmt-virtualization</item>
    <item>itop-service-mgmt-enterprise</item>
    <item>itop-ticket-mgmt-itil</item>
    <item>itop-ticket-mgmt-itil-user-request</item>
    <item>itop-ticket-mgmt-itil-incident</item>
    <item>itop-ticket-mgmt-itil-enhanced-portal</item>
    <item>itop-change-mgmt-itil</item>
    <item>itop-kown-error-mgmt</item>
</selected_modules>
```


rquetiez:
As this script is becoming a bit harder to use, it would be nive to have a "usage" explanation when no argument 'or --help) is provided.
The current behavior is:
```
> php .\unattended-install.php
No `--response_file` param specified, using default value !
Param file `default-params.xml` doesn't exist ! Exiting...
```

rquetiez:
A comment on the message given when no response file is provided:
```
No `--response_file` param specified, using default value !
Param file `default-params.xml` doesn't exist ! Exiting...
```
This message is confusing to me because "response file" means "output" of the current script.
On the second line, the "response file" is referred to as "param file".
Furthermore, I do not know where the "default-params.xml" should be placed.

I would rather have this report
```
No `--response_file` param specified, using default value: default-params.xml
Response file `/var/www/itop/setup/unattended-install/default-params.xml` doesn't exist !
Exiting...
```

setup/unattended-install/InstallationFileService.php
```diff
@@ -0,0 +1,252 @@
...
+require_once(APPROOT.'/setup/wizardcontroller.class.inc.php');
+require_once(APPROOT.'/setup/wizardsteps.class.inc.php');
+
+if (version_compare(ITOP_DESIGN_LATEST_VERSION, '2.7', '<=')) {
+	require_once(APPROOT.'/core/config.class.inc.php');
+	require_once(APPROOT.'/core/log.class.inc.php');
+	require_once(APPROOT.'/core/kpi.class.inc.php');
+	require_once(APPROOT.'/core/cmdbsource.class.inc.php');
+	require_once(APPROOT.'/application/clipage.class.inc.php');
+}
```
Molkobain, line 14:
```suggestion
require_once(APPROOT.'/core/config.class.inc.php');
require_once(APPROOT.'/core/log.class.inc.php');
require_once(APPROOT.'/core/kpi.class.inc.php');
require_once(APPROOT.'/core/cmdbsource.class.inc.php');
require_once(APPROOT.'/application/clipage.class.inc.php');
```

Then during the 2.7 => 3.0 merge, remove these lines.
As this file is part of the iTop repo, we rather adapt the code in the branches rather than maintaining "ifs" (which we do in modules)

setup/unattended-install/install-itop.sh
```diff
@@ -0,0 +1,50 @@
...
+	  HELP
+	fi
+else
+  INSTALLATION_XML="$ITOP_DIR/datamodels/2.x/installation.xml"
+fi
+
+echo "$CLI_NAME: Using XML_SETUP ($XML_SETUP) and INSTALLATION_XML ($INSTALLATION_XML) files during unattended itop installation."
+
+rm -rf $ITOP_DIR/data/.maintenance;
+echo php $DIR/unattended-install.php --use-itop-config=1 --use_installation_xml="$INSTALLATION_XML" --response_file="$XML_SETUP"
```
Molkobain: I would use underscores instead of dashes to ease usage for users.
```suggestion
echo php $DIR/unattended-install.php --use_itop_config=1 --use_installation_xml="$INSTALLATION_XML" --response_file="$XML_SETUP"
```

Molkobain (changes requested):
2 comments apart, it's fine by me like that. Installation script seems ok and information messages in the CLI are ok for me.

tests/php-unit-tests/unitary-tests/setup/unattended-install/UnattendedInstallTest.php
```diff
@@ -64,8 +64,11 @@ public function testCallUnattendedInstallFromHttp(){
 
 	public function testCallUnattendedInstallFromCLI() {
 		$cliPath = realpath(APPROOT."/setup/unattended-install/unattended-install.php");
-		$res = exec("php ".$cliPath);
+		exec("php ".$cliPath, $aOutput, $iCode);
```
Molkobain: With this, it will use PHP version from the OS path.
There is a method to use the same PHP binary as PHPUnit, otherwise this can lead to unexpected behavior.

Molkobain (changes requested):

setup/unattended-install/InstallationFileService.php
odain-cbd: ok i will keep it in mind

Hipska:
Can this branch be removed? It's the last one not being in a directory structure..

odain-cbd:
yes you are right. sorry for leaving it as is.

setup/unattended-install/unattended-install.php
```diff
@@ -291,17 +371,24 @@
...
 	// last line: used to check the install
 	// the only way to track issues in case of Fatal error or even parsing error!
 	$sLogMsg = "installed!";
+
+	if ($bUseItopConfig && is_file("$sConfigFile.backup"))
+	{
+		echo "\nuse config file provided by backup in $sConfigFile.";
+		copy("$sConfigFile.backup", $sConfigFile);
+	}
+
```
Hipska, line 385: @odain-cbd I have a question about this section. When no issues are found, the setup is finished and the config file is readonly again. Then why to try overriding the config file with the backup?

PR #643, opened 2024-04-05
N°7416 - Setup: Add warning for optionnal PHP extension "APCu"

Molkobain:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | N°7416

## Objective (enhancement)
Today we use APC to cache things in iTop.

However, this extension is rarely present by default with PHP installations, so we have our own APC emulation layer via the files _apc-compat.php_ and _apc-emulation.php_.

The problem is that this emulation is not perfect and often causes problems.

To try to mitigate this issue, we want to encourage people to activate the official PHP _APCu_ extension on their server.

As _APCu_ is not present on the majority of PHP installations, we put a _warning_ only during setup, no blocking error.

Our compatibility layer remains present just in case.

_Screenshot without the APCu extension_
image

_Screenshot with the APCu extension_
image

  => No unit test on that 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,9 +44,10 @@
     "ext-libsodium": "Required to use the AttributeEncryptedString.",
     "ext-openssl": "Can be used as a polyfill if libsodium is not installed",
     "ext-mcrypt": "Can be used as a polyfill if either libsodium and openssl are not installed (libsodium and openssl are more secure)",
+    "ext-apcu": "For better performance and stability",
+    "ext-imap": "Required by the extension \"Mail to ticket automation\""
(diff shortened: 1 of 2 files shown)


steffunky (approved):

Hipska:
You got to be kidding me: #201

Molkobain:
In the past months we stumbled more and more onto issues related to our poor apc emulation, so our point of view changed on that matter.

Actually we now strongly encourage people to activate APCu to avoid falling in the same issues, hence the (annoying) warning.

Hipska:
Finally! 👍

But it wasn't annoying, it only was said to 'frighten' users 😂

PR #644, opened 2024-04-12
N°7443 - url in read-only attribute are displayed different between console and portal

accognet:
internal


Molkobain:
This was fixed by @bdalsass during iTop 3.2.0

PR #645, opened 2024-04-16
N°2443 - Boolean don't accept yes/no value

bdalsass:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=2443&c[menu]=SearchBugs

Edition of AttributeBoolean is confusing, the component used is a TextField Input, filled with read view representation (yes, no), and cannot be submit until you fill the model representation  (0, 1).

Edit an object with a AttributeBoolean (SynchroAttribute for example)

## Cause (bug)
First implementation incomplete.

Because this kind of developments may be treated by SDK Form, the proposition is to do the minimalist changes.
In CmdbAbstract class, where attribute components are handled, the AttributeBoolean doesn't have a specific case, so it is handle with a simple text input (the default case).
The default case contain a test that can be usefull for our need, it checks allowed_values parameter existence, so if we provide this attribute parameter it could be handle via a select or radio button component (depending on another parameter display_style) 

2024-04-15_16h45_31

- [ ] Verify Behat impact
- [ ] ...
- [ ] ...

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -2613,7 +2613,7 @@ JS
 						// convert AttributeBoolean value due to issue with radio style when value is false
 						// @see N°2443 - Boolean don't accept yes/no value
 						if($oAttDef instanceof AttributeBoolean){
-							$value === false ? $value = 0 : $value = 1;
+							$value = $value === false ? 0 : 1;
 						}
 
 						// Discrete list of values, use a SELECT or RADIO buttons depending on the config
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -2609,6 +2609,13 @@ JS
 					$iFieldSize = $oAttDef->GetMaxSize();
 					if ($aAllowedValues !== null)
 					{
+
+						// convert AttributeBoolean value due to issue with radio style when value is false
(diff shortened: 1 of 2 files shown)


Hipska:
FINALLY a Boolean type 😍 !!! Is this new for 3.2?

THB, I would expect the `radio_horizontal` to be the default.

Hipska (approved):

application/cmdbabstract.class.inc.php
```diff
@@ -2608,6 +2608,13 @@ public static function GetFormElementForField($oPage, $sClass, $sAttCode, $oAttD
 					$iFieldSize = $oAttDef->GetMaxSize();
 					if ($aAllowedValues !== null)
 					{
+
+						// convert AttributeBoolean value due to issue with radio style when value is false
+						// @see N°2443 - Boolean don't accept yes/no value
+						if($oAttDef instanceof AttributeBoolean){
+							$value === false ? $value = 0 : $value = 1;
```
Molkobain:
```suggestion
							$value = $value === false ? 0 : 1;
```

Molkobain (commented):
Apart from the ternary expression, all good!

Molkobain (approved):

PR #650, opened 2024-06-04
N°7459 optimize memory consumption during compilation with php 8.1+

jf-cbd:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? |N°7459
| Type of change?                                               | Enhancement / Translations

Reduce memory peaks during setup

## Cause (bug)
PHP 8.1+ consume more memory when executing operations on XML

. Avoid storing too much XML objects
. Clean dictionaries


tests/php-unit-tests/integration-tests/DictionariesConsistencyTest.php
```diff
@@ -524,4 +553,48 @@ private function FindDictKeyLineNumberInContent(string $sFileContent, string $sD
...
+			'ru'    => ['RU RU', 'Russian', 'Русский'],
+			'sk'    => ['SK SK', 'Slovak', 'Slovenčina'],
+			'tr'    => ['TR TR', 'Turkish', 'Türkçe'],
+			'zh_cn' => ['ZH CN', 'Chinese', '简体中文'],
+		];
+
+		$aPrefixLang = [];
+		foreach ($aLanguages as $key => $value) {
+			$aPrefixLang[$value[1]] = ['lang' => $key];
+		}
```
Hipska: What about moving this to the `if ($bPrefixOnly)` section? And probably better to use `array_map` instead.

tests/php-unit-tests/integration-tests/DictionariesConsistencyTest.php
```diff
@@ -190,7 +175,10 @@ public function DictionaryFileProvider(): array
 
 		$aTestCases = array();
 		foreach ($aDictFiles as $sDictFile) {
-			$aTestCases[$sDictFile] = array('sDictFile' => $sDictFile);
+			preg_match('/^(.*)\\.dict/', basename($sDictFile), $aMatches);
+			$sDictFileLangPrefix = $aMatches[1];
+
+				$aTestCases[$sDictFile] = array('sDictFile' => $sDictFile, 'sLanguagePrefix' => $sDictFileLangPrefix);
```
Hipska, line 181:
```suggestion
			$aTestCases[$sDictFile] = array('sDictFile' => $sDictFile, 'sLanguagePrefix' => $sDictFileLangPrefix);
```

setup/modelfactory.class.inc.php
```diff
@@ -1258,6 +1208,50 @@ public function LoadModule(MFModule $oModule, $aLanguages = array())
...
+			$oXmlEntry = $this->oDOMDocument->CreateElement('entry');
+			$oXmlEntry->setAttribute('id', $sCode);
+			$oXmlValue = $this->oDOMDocument->CreateCDATASection($sLabel);
+			$oXmlEntry->appendChild($oXmlValue);
+			if (array_key_exists($sLanguageCode, $this->aDictKeys) && array_key_exists($sCode, $this->aDictKeys[$sLanguageCode])) {
+				$oXmlEntries->RedefineChildNode($oXmlEntry);
+				$oXmlEntry->RemoveAlteration();
+			} else {
+				// To avoid memory peak during execution of ApplyChanges, just set the node without alteration flag
+				$oXmlEntries->appendChild($oXmlEntry);
```
Hipska, line 1249: Is that having so much difference?

Hipska (changes requested):
I'm wondering if it wouldn't be better to just have the dicts already in xml instead of loading from PHP first?

tests/php-unit-tests/integration-tests/DictionariesConsistencyTest.php
jf-cbd: > I'm wondering if it wouldn't be better to just have the dicts already in xml instead of loading from PHP first?


Thanks for the suggestion. We thought about it, but we choose this solution for performance reason.

> What about moving this to the if ($bPrefixOnly) section? 

 Done, thanks !

setup/modelfactory.class.inc.php
jf-cbd: The memory peak is reduced by around 35% from our tests so yes :)

PR #651, opened 2024-07-09
N°7639 - Switching tab starts reloading the datatable even if data are already loaded

accognet:
On an object display, when you display a tab that contains a data table, it is systematically reloaded even though the data is already all present on the screen. This fix prevents this reload.


Molkobain:
@accognet there is some commits that should be there. This is most likely because the target branch on the PR is not the same as the one you started your branch from. Please fix it.

accognet:
Review PR Support : OK

Molkobain:
Functional review: Accepted for 3.2.1

Molkobain (approved):

PR #652, opened 2024-07-10
Correctly identify as iTop in cURL requests

Hipska:
## Base information

## Objective

Outgoing HTTP requests are identified with "spider" user agent, which can simply be blocked sometimes.

## Proposed solution

Use for example `iTop/3.2.0` as user agent when doing HTTP requests to other services..

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -1929,7 +1929,7 @@ SQL;
 			CURLOPT_HEADER         => false,    // don't return the headers in the output
 			CURLOPT_FOLLOWLOCATION => true,     // follow redirects
 			CURLOPT_ENCODING       => "",       // handle all encodings
-			CURLOPT_USERAGENT      => ITOP_APPLICATION.'/'.ITOP_VERSION, // who am i
+			CURLOPT_USERAGENT      => static::GetConfig()->Get('http.request.user_agent'), // who am i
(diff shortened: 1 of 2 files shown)


Molkobain (approved):

Molkobain:
Good idea. I don't know if it can have any side effects though?

Hipska:
Well, the adverse. So, if an external API refuses to connect anything else than `spider` 😉

jbostoen:
Just wondering if there should be a config setting as well to override this, and set a custom user agent string? (Or we can simply wait for the first real use case where it's necessary and would need to imitate a regular browser string for some reason).

Molkobain:
I'll raise the question during functional review :)

Molkobain (changes requested):
Technical review:
  * Ok for the principle
  * But use a conf. param.
    * Name : `http.request.user_agent` (another name might be proposed)
    * Default value : `ITOP_APPLICATION.'/'.ITOP_VERSION`

The conf. param. will allow to change it to something specific when needed and to revert it to "spider" for cases where people would have used that information.

Molkobain (approved):

jf-cbd:
Accepted during functional review. Will be in next iTop version.

Hipska:
So it can be merged?

jf-cbd:
Yep, thanks @Hipska for the PR !

PR #661, opened 2024-07-30
N°7636 - Too many OQL requests executed in order to display a lnk

accognet:
Improve performance of displaying links on screen

Currently when displaying a user query, displaying a lnk requires 5 sql queries :
Capture d'écran 2024-07-30 124309

After the PR, the same lnk only requires 2 SQL queries:
Capture d'écran 2024-07-30 124629

In total on the screen, the gain is 12 requests:
before:
Capture d'écran 2024-07-30 124540
after:
Capture d'écran 2024-07-30 124551

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -698,8 +698,7 @@ HTML
 			$sLinkedClass = $oAttDef->GetLinkedClass();
 
 			// Filter out links pointing to obsolete objects (if relevant)
-            if (array_key_exists($sAttCode, $this->GetValues()))
-            {
+            if (array_key_exists($sAttCode, $this->GetValues())) {
                 $oOrmLinkSet = $this->GetValues()[$sAttCode];
             } else {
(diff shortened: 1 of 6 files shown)


rquetiez:
Today's review:

- Count set by default to `-1`... that's pretty unusual, why not use `null` instead?
- Add tests in `DBObjectTest`, based on `AssertDBQueryCount`, to show what is being improved at the lowest possible level
- Using `DBObject::GetValues()` is not recommended as compared to `DBObject::Get()`. The former does not give the same results (e.g. lazy loading will lead to missing keys in the array).
- Improving `DBObject::Get()` could be preferable, -depending on fix. The root cause that has to be further clarified, with the help of unit tests

rquetiez:
Code review:
 * Requesting rollback on the optimization of MetaModel::GetHyperlink
 * Rename iCount into object_count in the extra params
 * Tests to simplify a little (no need to test GetValue)

rquetiez:
Code review : ok for merging

PR #662, opened 2024-08-01
New Triggers: TriggerOnAttachmentAdded and TriggerOnAttachmentRemoved

larhip:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | Already discussed with @v-dumas 

Currently there is only a `TriggerOnAttachmentDownload` which is triggered via an event listener, but there are already events for adding and removing attachments (`EVENT_ADD_ATTACHMENT_TO_OBJECT`, `EVENT_REMOVE_ATTACHMENT_FROM_OBJECT`). 
Why not adding triggers here as well, wo make it possible to execute an action each time an attachment is added or removed?

With this PR, two additional triggers (`TriggerOnAttachmentAdded`,`TriggerOnAttachmentRemoved`) are added to be triggered when adding or removing an attachment, with the same logic as currently when downloading an attachment.

So far I have not added any UnitTests as it seems that there is no test case for the existing class `TriggerOnAttachmentDownload`. So I was not sure if and how a unit test would be appropriate here.

- [ ] Clarification of whether unit tests are required
- [ ] Code review by Combodo
- [ ] Translations for all languages beside English and German (at least add the placeholders), might by easier to be done by Combodo?

### DIFF
--- a/datamodels/2.x/itop-attachments/dictionaries/de.dict.itop-attachments.php
+++ b/datamodels/2.x/itop-attachments/dictionaries/de.dict.itop-attachments.php
@@ -59,5 +59,9 @@ oder melden Sie dem '.ITOP_APPLICATION_SHORT.' Administrator diesen Fehler, weil
 	'Class:Attachment/Attribute:user_id+' => '',
 	'Class:TriggerOnAttachmentDownload' => 'Trigger (beim Herunterladen eines Attachment eines Objekts)',
 	'Class:TriggerOnAttachmentDownload+' => 'Trigger für das Herunterladen des Attachments der angegebenen Klasse oder einer Unterklasse',
+	'Class:TriggerOnAttachmentAdded' => 'Trigger (beim Hinzufügen eines Attachments)',
+	'Class:TriggerOnAttachmentAdded+' => 'Trigger für das Hinzufügen des Attachments an einem Objekt der angegebenen Klasse oder einer Unterklasse',
+	'Class:TriggerOnAttachmentRemoved' => 'Trigger (beim Entfernen eines Attachments)',
+	'Class:TriggerOnAttachmentRemoved+' => 'Trigger für das Entfernen des Attachments an einem Objekt der angegebenen Klasse oder einer Unterklasse',
 	'UI:Attachments:DropYourFileHint' => 'Dateien in diesem Bereich ablegen...',
 ]);
--- a/datamodels/2.x/itop-attachments/dictionaries/en.dict.itop-attachments.php
+++ b/datamodels/2.x/itop-attachments/dictionaries/en.dict.itop-attachments.php
@@ -92,3 +92,21 @@ Dict::Add('EN US', 'English', 'English', array(
 	'Class:TriggerOnAttachmentDownload' => 'Trigger (on object\'s attachment download)',
 	'Class:TriggerOnAttachmentDownload+' => 'Trigger on object\'s attachment download of [a child class of] the given class',
 ));
+
+//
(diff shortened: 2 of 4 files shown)


Molkobain:
Nice PR Lars, thanks!

Anne-Catherine worked on this a while back (see #534), and we wondered if these triggers should not display some classes in the "target_class" dropdown. To do so, we added a new `class_exclusion_list` property to the `\AttributeClass`.

Did you discuss that last point with Vincent?

larhip:
Thanks for your Feedback Guillaume!
Wasn't aware of the PR of Ann-Cathrine, sorry. 

May you give me an example of classes which might be excluded?
Since actually for the TriggerOnAttachmentDownload there are no classes excluded, it sounds a little bit confusing to do it for those new triggers but not for the existing one?

Cheers
Lars

v-dumas:
Knowing that the classes which can have attachment aren't yet known at setup time, it's not easy to have a pertinent proposal in terms of classes. Can we capture a change in the configuration which would add a new class and change on the fly the allowed values, I am afraid not.
I agree with Lars, the 3 triggers should behave the same way, whatever way we decide to apply.

Molkobain:
For instance with AC implementation, you could select the `Attachment` class for the trigger, which will be confusing for the user. That's why we proposed to add this new `class_exclusion_list` attribute (the same way that with attribute type exclusion list on other attributes).

What do you think?

larhip:
Hi Guillaume,

that's true with my solution it is possible as well to select `Attachment` as target class (I mean: PHP is in favour of recursion, so why not? :-P)

You are right there might be classes which makes no sense at all to select them as target class (i.e. `Attachment`). That being said I am not 100 percent sure how to avoid that. In my understanding, if we use something like `class_exclusion_list` we have to define that `class_exclusion_list` during the attribute definition? 
Actually the new Triggers are child classes of the class `TriggerOnObject`. The `target_class` attribute is not defined in the child class but in the parent.
So in order to use some new logic like `class_exclusion_list` it is not possible anymore to derive from the class `TriggerOnObject`, is it?

This problem will also affect the already existing Trigger `TriggerOnAttachmentDownload`. In order to keep the same approach for all classes, the class `TriggerOnAttachmentDownload` need to be changed with a changed parent class. This will lead into some migration issues in the database?

So from my point of view the question would be: Is it necessary to refactor the already existing class `TriggerOnAttachmentDownload`, since the problem regarding the target class is already part of the productive code?

- If the answer is yes:  It has to be done in the already existing code, before changing this PR
- If the answer is no: I do not see a reason why it should be different for the new Triggers `TriggerOnAttachmentAdded` and `TriggerOnAttachmentRemoved`.

Anyhow I have got the feeling that this change might be not in the scope of the PR.

What do you think?

Molkobain:
Hello Lars,

I think we should discuss this with the POs first, I'll let you know. :)

Molkobain:
Just discussed this with @v-dumas : there is no need to refactor the code, especially as we didn't do it in our own triggers 😅

Molkobain (approved):

Molkobain:
Technical review: This has been discussed today, Romain & Anne-Catherine (I'll be gone) will compare both PRs and to see what can be kept from each PR, then go forward.

jf-cbd:
Hello @larhip, thanks a lot for your PR ! We are working on merging it into https://github.com/Combodo/iTop/pull/534, you can have a look at it if you want :)

larhip:
Hey @jf-cbd 
Thanks for your feedback!
Since #534 looks promising, I will close this PR :-)

PR #666, opened 2024-09-04
N°8893 - :lipstick: Transition validation markup

jbostoen:
## Base information

When using for example the new event listeners to perform a check before the object is written **during a transition;** the current error modal is ugly. It does not use a nice list; and it shows context data.

The error messages are just examples. 

Before:
image

After:
image

### DIFF
--- a/pages/UI.php
+++ b/pages/UI.php
@@ -1301,6 +1301,8 @@ try
 					if (count($aErrors) == 0)
 					{
 						$sIssues = '';
+						/** @var CoreCannotSaveObjectException|null $oException */
+						$oException = null;
 						$bApplyStimulus = true;
 						list($bRes, $aIssues) = $oObj->CheckToWrite(); // Check before trying to write the object
 						if ($bRes)
@@ -1315,6 +1317,11 @@ try
 								$oObj = MetaModel::GetObject(get_class($oObj), $oObj->GetKey());
 								$oObj->UpdateObjectFromPostedForm('', array_keys($aExpectedAttributes), $aExpectedAttributes);
 								$sIssues = $e->getMessage();
+								
+								if($e instanceof CoreCannotSaveObjectException) {
+									$oException = $e;
+								}
+
 							}
 						}
 						else
@@ -1355,7 +1362,7 @@ try
 								$sMessage = $e->getMessage();
 								$sSeverity = 'info';
 							}
-							$sIssueDesc = Dict::Format('UI:ObjectCouldNotBeWritten',$sIssues);
+							$sIssueDesc = ($oException !== null ? '<div style="display: block;"><style scoped> ul { list-style-type: disc; } </style>'.$oException->getHtmlMessage().'</div>' : Dict::Format('UI:ObjectCouldNotBeWritten', $sIssues));
 							$oP->add_ready_script("CombodoModal.OpenErrorModal('".addslashes($sIssueDesc)."');");
 						}
 						else


pages/UI.php
```diff
@@ -1355,7 +1362,7 @@ function DisplayNavigatorGroupTab($oP)
 								$sMessage = $e->getMessage();
 								$sSeverity = 'info';
 							}
-							$sIssueDesc = Dict::Format('UI:ObjectCouldNotBeWritten',$sIssues);
+							$sIssueDesc = ($oException !== null ? '<div style="display: block;"><style scoped> ul { list-style-type: disc; } </style>'.$oException->getHtmlMessage().'</div>' : Dict::Format('UI:ObjectCouldNotBeWritten', $sIssues));
```
jbostoen, line 1365: While not ideal, this will be easier to maintain than a CSS rule somewhere in a file. 

The main issues:
- Error modal uses flex.
- The "ul" element loses its standard formatting by default in iTop.

pages/UI.php
```diff
@@ -1315,6 +1317,11 @@ function DisplayNavigatorGroupTab($oP)
 								$oObj = MetaModel::GetObject(get_class($oObj), $oObj->GetKey());
 								$oObj->UpdateObjectFromPostedForm('', array_keys($aExpectedAttributes), $aExpectedAttributes);
 								$sIssues = $e->getMessage();
+								
+								if($e instanceof CoreCannotSaveObjectException) {
+									$oException = $e;
+								}
```
Hipska, line 1323: What if you would do this instead? Then there's no need to do the if/else later on.
```suggestion
								$sIssues = $e->getHtmlMessage();
```
(correct styling would still something to be done in CSS, maybe separate PR)

pages/UI.php
Hipska: ErrorModal should be fixed IMHO

pages/UI.php
jbostoen: I'm not sure whether or not we can expect other exceptions here, which don't support the `getHtmlMessage()`, so I was being cautious :)

Hipska: Right, `CoreException` doesn’t seem to have this method 🤔

Then maybe just do getHtmlMessage when it exists and the normal getMessage otherwise?

pages/UI.php
Molkobain: FYI, in the backoffice when you want HTML to retrieve its standard style like for lists, surround it with the `.ibo-is-html-content`.

I would strongly be against inline style like this. A dedicated rule in the right file would be better.

That being said, you should ping @jf-cbd or @rquetiez , they worked on error messages and removing/sanitizing HTML errors last summer.

jbostoen: I'll try the `.ibo-is-html-content` when I find some time. 
Not sure if that would also change the display to block or if there's another CSS class which we can use.

jf-cbd:
Accepted during functional review. Will check with @rquetiez for technical details (3.2 compatibility and way to define the message).

pages/UI.php
jf-cbd: Hey @jbostoen, did you have some time to think about it (moving the inline CSS) ?
On our side, we'll check about the way we want to do the HTML sanitization

jbostoen: Not yet, was waiting to hear more from Combodo on error messages and sanitizing

pages/UI.php
jf-cbd: Hey, we'll work on creating this method in CoreException, so this fix will be more generic (and you'll be able to use Hipska's suggestion). This will probably be in 3.3.

pages/UI.php
jf-cbd: Concerning the error message, I'll check with other member of Combodo to know the best way to do for formatting (probably class + style in a separated file).

jf-cbd:
Hello @jbostoen, thank you for this PR. It has been functionally accepted, and we'll work on it during sprint 3.4

PR #674, opened 2024-10-31
N°7793 Add common SCSS variables between backoffice and end-user portal

steffunky:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | N°7793

## Objective
This feature aims to ease future portal style rework by adding common SCSS variables shared between the portal and the backoffice.
This will ensure that portal and backoffice maintain a similar style and ease overloading both frontends with a single customization. 

## Proposed solution 
This approach adds a set of variables named `$common-` as opposed to itop backoffice `$ibo-` that will distribute default values for front ends.

`$ibo-` variables now consume `$common-` values that are then fed to the stylesheets.

To avoid breaking existing themes in a maintenance release, not all variables can inherit `$common-` values.
Up until now `$ibo-` variables that inherited another `$ibo-` variable value could be indirectly overloaded with a theme:

Flow 1@1x-50fps

But with how we envisioned the addition of `$common` variables, it would break the existing themes by breaking the link between `$ibo-` variables
Flow 2@1x-50fps

In iTop 3.2.1 we'll keep this behavior and add a breaking change in iTop 3.3.0 that will require themes to either overload both `$ibo-` variables, or overload the `$common-` variable
Flow 3@1x-50fps

[x] Ensure previous themes are not broken by this change
     => No differences in custom light/dark themes when comparing dashboards, lists, details
[x] Ensure all variables that can be made common in iTop 3.2.1 are changed in this PR


Molkobain (commented):
Nice job, maybe you could duplicate the `README.md` & `_shame.scss` (empty) into the `common` folder as well.

css/backoffice/utils/functions/_color.scss
Molkobain, line 1: I would migrate usages of these 2 methods and deprecate them as there is no need to have a dedicated implementation for the backoffice.

steffunky: agreed

bdalsass (approved):

PR #676, opened 2024-11-06
N°7968 - ⚡️ Add faster way to rename column on the same table

Hipska:
## Base information

## Objective

The process of moving data from one column to another could be simplified a lot when both are on the same table.

## Proposed solution

Execute a `RENAME COLUMN` SQL statement instead of the following steps:
1. Create new column
2. Copy data from old to new column
3. Delete old column


Molkobain (changes requested):
Careful, this is compatible with MySQL 8.0+, but not MySQL 5.7 which is still [supported by iTop](https://www.itophub.io/wiki/page?id=3_2_0:install:requirements#itop_compatibility_list). You should stick to the `CHANGE` synthax:

| Database       | Version           | `CHANGE` Compatibility | `RENAME COLUMN` Compatibility |
|----------------|-------------------|-------------------------|-------------------------------|
| MySQL          | < 8.0             | Yes                     | [No](https://dev.mysql.com/doc/refman/5.7/en/alter-table.html)                            |
| MySQL          | 8.0+              | Yes                     | [Yes](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html)                           |
| MariaDB        | < 10.5.2            | Yes                     | No                            |
| MariaDB        | 10.5.2+             | Yes                     | [Yes](https://mariadb.com/kb/en/alter-table/#rename-column)                            |

Hipska:
Oh, that's probably the reason I didn't create the PR back then.

Downside with [CHANGE](https://dev.mysql.com/doc/refman/8.4/en/alter-table.html#alter-table-redefine-column) is that it needs the column definition, but there's a method for that, so fixed in 318a7e9.

Molkobain (approved):
`CMDBSource::GetfIeldSpec()` can return false, but I rather have an explicit error than hiding it. Good job.

steffunky:
We already have some tests for this method in `tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php`, adding one testing this use case could be nice as @Molkobain already caught a regression in this method once
If you don't have time I'll try to add one sometime soon

Hipska:
@Molkobain actually, the same method is doing the same without any checks: https://github.com/Combodo/iTop/blob/c6039f4b51fffacedc529df87ae7396c409e0c6e/setup/moduleinstaller.class.inc.php#L277
It will just never occur to return `false` because of the checks before.

steffunky (approved):
Works well on MariaDB 11.3.2 :+1:

Hipska:
Funny, as that version isn't even [supported](https://www.itophub.io/wiki/page?id=latest:install:requirements#itop_compatibility_list) by Combodo.

tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php
```diff
@@ -240,4 +240,48 @@ public function testMoveColumnInDB_MoveMultipleTable(): void
...
+			<<<SQL
+SELECT `{$sDstNonExistingColName}` FROM `{$sOrigTable}` WHERE `id` = 1
+LIMIT 1
+SQL
+		);
+
+		// Put data back in the original table
+		ModuleInstallerAPI::MoveColumnInDB($sOrigTable, $sDstNonExistingColName, $sOrigTable, $sOrigColName);
+
+		$this->assertEquals($sOrigValue, $sDstValue, "Data was not moved as expected");
```
Molkobain, line 285:
```suggestion
		// Try to move data
		ModuleInstallerAPI::MoveColumnInDB($sOrigTable, $sOrigColName, $sOrigTable, $sDstNonExistingColName);

		// Check if data was actually moved
		// - Either way, the column should exist
		$sDstValue = CMDBSource::QueryToScalar(
			<<<SQL
SELECT `{$sDstNonExistingColName}` FROM `{$sOrigTable}` WHERE `id` = 1
LIMIT 1
SQL
		);

		$this->assertEquals($sOrigValue, $sDstValue, "Data was not moved as expected");

		// Put data back in the original table
		ModuleInstallerAPI::MoveColumnInDB($sOrigTable, $sDstNonExistingColName, $sOrigTable, $sOrigColName);
```

I would rather put the assert there with your actual check. With the current proposition, it confused me, as I though you checked the data was put back in it's original column.

Molkobain (changes requested):

tests/php-unit-tests/unitary-tests/setup/ModuleInstallerAPITest.php
Molkobain: Otherwise the test is good, thanks!

Hipska: Doen't it break the state of the table when assert fails? (The way unit tests in php work are weird and confusing anyways)

steffunky: I wanted to keep consistency with the [code written by my ex Lead dev](https://github.com/Combodo/iTop/pull/676/files/007927c852879001fd30552b2ce6322b8371d911#diff-26ac6d70dfb1aa45415774deb778df56af8238e996d5d39f46cdc79cdef36e2dL137-L144)
Joking aside, wouldn't a failed assert stop the test execution and leave an inconsistent table?

Molkobain: 🙈🙈🙈

I think both your statements are valid, I therefore withdraw mine 😁

Molkobain (approved):

rquetiez:
To take all the benefits out of this proposal, the table alteration algorithm must be at least `inplace`.

I've just conducted a quick study to check which DDL algorithm is used, and make sure that the syntax `ALTER TABLE table CHANGE column...` would not imply the infamous `COPY` algorithm.

### Conclusion
The proposed implementation will be quick in any versions supported by iTop, excepted for the rare cases when the `COPY` algorithm is still used (e.g. table created with mariaDB < 10.3 and never rebuilt since the upgrade of the server).

### Methodology
As an alternative to measuring the time spent on executing the `alter table` statement on a "huge" table, one can specify `algorithm=inplace`, so that the `alter table` statement would fail if the engine is not able to execute the `INPLACE` algorithm (or a quicker algorithm).

I've also tested with `lock=none` to document wether the operation involves locking or not (See [Online DDL Performance and Concurrency](https://dev.mysql.com/doc/refman/8.4/en/innodb-online-ddl-performance.html))

```mysql
alter table organization change code new_code varchar(255) default '' null, algorithm=inplace, lock=none;
alter table organization change code new_code varchar(255) default '' null, algorithm=instant;
```

### Results
| Database       | Version           | algorithm=`inplace`, lock=none | algorithm=`instant` |
|----------------|-------------------|-------------------------|-------------------------------|
| MySQL          | 5.7               | Yes                     | n/a |
| MySQL          | 8.3               | Yes                     | Yes |
| MariaDB        | 10.3              | Yes                     | Yes |

rquetiez (approved):

jf-cbd:
Thanks a lot @Hipska for the contribution !

PR #685, opened 2024-11-27
Security hardening

jf-cbd:
Internal - security

### DIFF
--- a/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
@@ -1228,8 +1228,6 @@ class ObjectController extends BrickController
 		$oRequestManipulator = $this->get('request_manipulator');
 		/** @var \Combodo\iTop\Portal\Helper\ScopeValidatorHelper $oScopeValidator */
 		$oScopeValidator = $this->get('scope_validator');
-        /** @var \Combodo\iTop\Portal\Helper\SecurityHelper $oSecurityHelper */
-        $oSecurityHelper = $this->get('security_helper');
 
 		$aData = array();
 
@@ -1266,12 +1264,6 @@ class ObjectController extends BrickController
 		// Retrieving objects
 		while ($oObject = $oSet->Fetch())
 		{
-            $sObjectId = $oObject->Get('id');
-            if (!$oSecurityHelper->IsActionAllowed(UR_ACTION_READ, $sObjectClass, $sObjectId))
-            {
-                IssueLog::Warning(__METHOD__.' at line '.__LINE__.' : User #'.UserRights::GetUserId().' not allowed to read '.$sObjectClass.'::'.$sObjectId.' object.');
-                throw new HttpException(Response::HTTP_NOT_FOUND, Dict::S('UI:ObjectDoesNotExist'));
-            }
 			$aData['items'][] = $this->PrepareObjectInformation($oObject, $aObjectAttCodes);
 		}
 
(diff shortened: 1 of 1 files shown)


datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
```diff
@@ -1263,6 +1266,12 @@ public function GetInformationAsJsonAction(Request $oRequest)
 		// Retrieving objects
 		while ($oObject = $oSet->Fetch())
 		{
+            $sObjectId = $oObject->Get('id');
+            if (!$oSecurityHelper->IsActionAllowed(UR_ACTION_READ, $sObjectClass, $sObjectId))
+            {
+                IssueLog::Warning(__METHOD__.' at line '.__LINE__.' : User #'.UserRights::GetUserId().' not allowed to read '.$sObjectClass.'::'.$sObjectId.' object.');
```
Hipska: Reduce var scope and use specific method.
```suggestion
            if (!$oSecurityHelper->IsActionAllowed(UR_ACTION_READ, $sObjectClass, $sObjectId))
            {
                $iObjectId = $oObject->GetKey();
                IssueLog::Warning(__METHOD__.' at line '.__LINE__.' : User #'.UserRights::GetUserId().' not allowed to read '.$sObjectClass.'::'.$iObjectId.' object.');
```

steffunky: `$iObjectId` is used in the `$oSecurityHelper->IsActionAllowed()` condition, so you can't put it inside the block

Hipska: Oh, I missed that. Well it could still be changed to use the `GetKey()` method 😉

jf-cbd: Yep that would be correct too but GetKey() returns an int, while IsActionAllowed expect (given parameter name) a string

steffunky: Moreover, I prefer storing a value in a variable instead of calling a method twice (even if `GetKey()` isn't very costly, it's always cheaper to access a variable)

steffunky (approved):

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
Hipska: Yes, keep using the variable is fine, never argued that 😉

Molkobain (approved):

datamodels/2.x/itop-portal-base/portal/src/Controller/ObjectController.php
Molkobain: I know it was just the variable naming, but keep in mind that unfortunately there is no consistancy in the app for the expected type of an object key. Someplaces it's an int, others it's a string.

jf-cbd:
$oSecurityHelper->IsActionAllowed seems useless since AddScopeToQuery check if action is allowed

Molkobain:
> $oSecurityHelper->IsActionAllowed seems useless since AddScopeToQuery check if action is allowed

I think it is actually the other way around: `SecurityHelper::IsActionAllowed()` checks if there is a scope for the specified object and check if the object is within that scope.

Calling `ScopeValidatorHelper::AddScopeToQuery()` only tries to add the scope for specified object class, but it doesn't check if you have access to the object or not, whether it would be based on the scopes or the DM rights.

jf-cbd:
Thanks for your message :) If there is no scope, then we should throw an error. Else it should be only filtered data rendered.

That should be enough, right ?

```
		if (!$oScopeValidator->AddScopeToQuery($oSearch, $sObjectClass)
		) {
			IssueLog::Warning(__METHOD__.' at line '.__LINE__.' : User #'.UserRights::GetUserId().' not allowed to read '.$sObjectClass.' object.');
		}     
```

Molkobain:
Exactly yes. You can check other usages of this method in the portal for an example, and see that we check the returned value. :)

PR #690, opened 2024-12-16
N°6282 Fix XSS vulnerability in soap

Timmy38:


Hipska:
The parameter `service_category` is also used in other places like `itop.wsdl.php`.

I think it is best to think about deprecating the SOAP and providing a timeline on when it will be removed from iTop, so that it doesn't depend on `php-soap` anymore for installation.

steffunky (approved):

steffunky:
> I think it is best to think about deprecating the SOAP and providing a timeline on when it will be removed from iTop, so that it doesn't depend on `php-soap` anymore for installation.

We plan on deprecating SOAP in the 3.2 LTS branch and removing it right after, in the 3.3 branch :grin:

Hipska:
Ouch, the vulnerability still isn't fixed in `itop.wsdl.php` as I pointed out before.

PR #695, opened 2024-12-27
N°7383 Show data table even when there is no data after filtering

Timmy38:


Hipska:
Before and after screenshot maybe?

rquetiez (approved):

PR #700, opened 2025-02-25
 N°8724 - Enhance setup feedback in case of module dependency issue

odain-cbd:
This PR aims at enhancing iTop feedback when there are module dependency issues: modules with less missing dependencies are printed first.

Usually modules dependencie issues are due to one missing module. it is hard to find it out when unresolved cascading dependencies are displayed in a "random" order.

Example (BEFORE)
```
Error: The following modules have unmet dependencies:
Module1 (id: id1/123) depends on: ✅ id2/456 + ❌ id4/666 + ❌ id3/789,
Module3 (id: id3/789) depends on: ✅ id2/456 + ❌ id4/666,
Module4 (id: id4/888) depends on: ❌ id5/666,
```

Example (with PR enhancement): Now Module4 will come first.
```
Error: The following modules have unmet dependencies:
Module4 (id: id4/888) depends on: ❌ id5/666,
Module3 (id: id3/789) depends on: ❌ id4/666,
Module1 (id: id1/123) depends on: ❌ id4/666 + ❌ id3/789,
```


eespie (approved):
seems ok for me

tests/php-unit-tests/unitary-tests/setup/ModuleDiscoveryTest.php
```diff
@@ -0,0 +1,110 @@
...
+			for ($j=0; $j<$i; $j++){
+				$aDeps[]=$j;
+			}
+			$aOngoingDependencies[$sKey]=$aDeps;
+		}
+		sort($aExpectedKeys);
+
+		\ModuleDiscovery::SortModulesByCountOfDepencenciesDescending($aOngoingDependencies);
+
+		$this->assertEquals($aExpectedKeys, array_keys($aOngoingDependencies));
```
rquetiez: I'd rather have $aOngoingDependencies and $aExpectedKeys shown explicitely as PHP arrays, thus documenting the input and output of the tested method. We could even inline intermediate variables like this:
```php
$this->assertEquals(
    [3, 2, 1],
    array_keys(\ModuleDiscovery::SortModulesByCountOfDepencenciesDescending(
        [
            1 => [2],
            2 => [3],
            3 => [],
        ]
    ));
);
```
Moreover, I wonder if this is meaningful to test this internal method. This could be the case to explore corner cases. If so, then other cases should be added. Otherwise, the tests implemented after that one should be sufficient to cover this method, and the test of `SortModulesByCountOfDepencenciesDescending` should be removed.

tests/php-unit-tests/unitary-tests/setup/ModuleDiscoveryTest.php
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	public function testOrderModulesByDependencies_CheckMissingDependenciesAreCorrectlyOrderedInTheException()
```
rquetiez: Function naming: "correctly" could be clarified. Alternatively, this could be clarified in the assertion message.

`$this->expectExceptionMessage($sExpectedMessage, 'Modules with fewest count of missing dependencies should reported first');`

setup/modulediscovery.class.inc.php
```diff
@@ -221,55 +228,72 @@ public static function OrderModulesByDependencies($aModules, $bAbortOnMissingDep
...
 		{
+			self::SortModulesByCountOfDepencenciesDescending($aOngoingDependencies);
 			$aModulesInfo = [];
 			$aModuleDeps = [];
-			foreach($aDependencies as $sId => $aDeps)
+			foreach($aOngoingDependencies as $sId => $aCurrentRemainingDeps)
 			{
 				$aModule = $aModules[$sId];
 				$aDepsWithIcons = [];
+				$aDeps=$aDependencies[$sId];
```
rquetiez: Coding convention

tests/php-unit-tests/unitary-tests/setup/ModuleDiscoveryTest.php
```diff
@@ -0,0 +1,110 @@
...
+			],
+			"id2/456" => [
+				'dependencies' => [],
+				'label' => 'label2',
+			],
+		];
+		\ModuleDiscovery::OrderModulesByDependencies($aModules, true);
+	}
+
+	public function testOrderModulesByDependencies_ResolveOk()
```
rquetiez: Coverage for successful cases could be completed by added the following [corner] cases:

1. No dependency at all
2. Several dependencies on a given module

tests/php-unit-tests/unitary-tests/setup/ModuleDiscoveryTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+		$aExpected = [
+			"id4/4",
+			"id3/3",
+			"id2/2",
+			"id1/1",
+		];
+		$this->assertEquals($aExpected, array_keys($aResult));
+	}
+}
```
rquetiez: Coverage for failing cases could be improved by testing:

1. Circular reference
2. bAbortOnMissingDep = false. What is the expected behavior then?

rquetiez (changes requested):
As this method is hard to test manually, changing it means covering it completely. I have added a few suggestions, but it will require an additional review (interactive) to make sure 100% of the rules are covered.

odain-cbd:
close PR temporarly

odain-cbd:
Current branch has been merged in develop apart from PR here

PR #701, opened 2025-02-28
:bug: Use Cancel-Button results in Ajax-Unauthorized Response

Hotade1702:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | no

When calling "cancel button" in edit or creation dialog the ajax request responded with unauthorized 401. This results in login mask when using IIS with Windows Authentication. You can continue to use itop without enter credentials, but it appears everytime canceling a edit/new mask. The reason is the Transaction_id validation in ajax.render.php failed. After changing the ReadPostedParam Parameter "sSanitizationFilter" from default "parameter" to "transaction_id" the validation is true and itop is handleling everything fine.

1. On iTop 3.2.0 
2. With PHP 8.1.30 
3. IIS and Windows Authentication (The Bug also happend everywhere else but you will only see it in the developer console of the browser that the ajax-result answers 401)
1. First go and add a new Object (Server as example) or edit any existing cmdb object or ticket
2. Then do cancel this dialog
3. Finally, see that a login mask appears, which can be canceled and you can continue to use itop.

## Cause (bug)

The reason is when unloading the new object page or edit page the unload function calls a function where the transaction_id verification failed. 

Editing one line in ajax.render.php for header and transaction_id check  from using the default parameters of ReadPostedParam to specific parameter "transaction_id" as sSanitizationFilter

- [x ] I have performed a self-review of my code
- [x ] I have tested all changes I made on an iTop instance
- [x ] no unit test, because only one line changed with existing functions
- [ x] Is the PR clear and detailed enough so anyone can understand digging in the code?

### DIFF
--- a/pages/ajax.render.php
+++ b/pages/ajax.render.php
@@ -62,7 +62,7 @@ try
 	// check if header contains X-Combodo-Ajax for POST request (CSRF protection for ajax calls)
 	// check must be performed after DoLoginEx to be logged in and to be able to check the token (based on the transaction id)
 	if (!isset($_SERVER['HTTP_X_COMBODO_AJAX']) && $_SERVER['REQUEST_METHOD'] !== 'GET') {
-		$sTransactionId = utils::ReadPostedParam("transaction_id");
+		$sTransactionId = utils::ReadPostedParam("transaction_id",'','transaction_id');
 		if (!utils::IsTransactionValid($sTransactionId, false)) { // if a form is submitted without header but contains a token... should be exceptional
 			$sReferer = $_SERVER['HTTP_REFERER'];
 			$sErrorMsg = 'Unauthorized access. Please see https://www.itophub.io/wiki/page?id=3_2_0:release:developer#checking_for_the_presence_of_specific_header_in_the_post_to_enhance_protection_against_csrf_attacks';


jf-cbd:
Hello, thanks for your contribution :)
We'll add the filter you suggest, as it is appropriated and can only be a good thing !
But we couldn't reproduce your problem (with iTop 3.2.0 and PHP 8.1.30, on an Ubuntu server, looking for error in the developer console), and it may be good to know why the transaction_id verification failed.
May you please give use some screenshots and more info to reproduce your problem ?

Hotade1702:
hi, i logged both to the error_log in php, when sending without the filter its doesnt result in a transaction id:
[07-Mar-2025 16:55:39 Europe/Paris] with_transactionidfilter: DANB81.tmp
[07-Mar-2025 16:55:39 Europe/Paris] without_transactionidfilter: 
In the developer log in browser you can only see it when you have a look in the Network tab:
image

When cancel a dialog without transactionfilter i get this in my _**php**_ log with php 8.1.30 (im not shure if version is really relevant) on IIS:
[07-Mar-2025 16:57:04 Europe/Paris] PHP Warning:  Undefined variable $sReferer in D:\itop-dev\pages\ajax.render.php on line 72

which is: 			IssueLog::Error("Unprotected ajax call : $sTransactionId", LogChannels::SECURITY, ['referer' => $sReferer]); (should be 2-3 line higher in your code because i added some lines for debugging (error_log...)

and on IIS with windows-authentication "on" resulting in this popup:
Unbenannt

jf-cbd:
Hello, thanks for your info @Hotade1702 
We think we have found the reason of the bug thanks to it and your fix, could you help us to confirm it ?
When the error occurs, does your $sTransactionId contain a dot, e.g. "xxx.tmp" ?

PR #703, opened 2025-03-18
N°8245 - External key not saved in n-n link in edition

bdalsass:
@see https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8245

Bug introduced with bug N°8245 since N°7264 Update selectize.js from 0.12.4 to 0.15.2
This version of Selectize library reset the input change listeners on initialization.

Needs a better implementation because the code is quite ugly.

### DIFF
--- a/js/links/links_widget.js
+++ b/js/links/links_widget.js
@@ -390,16 +390,17 @@ function LinksWidget(id, sClass, sAttCode, iInputId, sSuffix, bDuplicates, oWizH
 
 	this.RegisterChange = function () {
 		// Listen only used inputs
-		$('#linkedset_'+me.id+' :input[name^="attr_'+me.sAttCode+'["]').off('change').on('change', function () {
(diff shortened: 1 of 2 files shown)


steffunky:
A more elegant way would be to undo your changes and modify how the `change` listener is added in `js/links/links_widget.js`.
Line 393 we could remove the listener from the input element itself (that selectize removes on initialization) and put it on the body with a selector that matches our input.

The code would change from this:
```
$('#linkedset_'+me.id+' :input[name^="attr_'+me.sAttCode+'["]').off('change').on('change', function () {
    ...
});
```
to this: 

```
$('body').off('change', '#linkedset_'+me.id+' :input[name^="attr_'+me.sAttCode+'["]')
    .on('change', '#linkedset_'+me.id+' :input[name^="attr_'+me.sAttCode+'["]', function () {
	.....
});
```

I tested it and it seems to work well

PR #704, opened 2025-03-18
:bug: N°8278 - Avoid autoloaders class name conflict by ensuring all existing have a unique name

Molkobain:
## Base information

## Symptom (bug)
When running a setup, it crashes during compilation saying that an autloader class with the same name is already loaded:
```
( ! ) Fatal error: Cannot declare class ComposerAutoloaderInitd751713988987e9331980363e24189ce, because the name is already in use in C:\wamp64\apps\itop\3.2\extensions\sample-module\vendor\composer\autoload_real.php on line 5
Call Stack
#TimeMemoryFunctionLocation
[...]
110.334411825280require_once( 'C:\wamp64\apps\itop\3.2\extensions\sample-module\vendor\composer\autoload_real.php )...\autoload.php:23

PHP error occured : msg=Cannot declare class ComposerAutoloaderInitd751713988987e9331980363e24189ce, because the name is already in use, no=64, file=C:\wamp64\apps\itop\3.2\extensions\sample-module\vendor\composer\autoload_real.php, line=5
```

This is a very similar symptom than what we had several times (without understanding it at the time) during the development of new portal modules compatible with both iTop 2.6 & 2.7. Which was actually when we introduced autoloaders and composer in portal modules.

1. On iTop 3.2.1+
2. With PHP 8.1.31
2. Create a module with a `composer.json` file that **contains nothing but** the license and the autoload configuration
2. Run `composer install` on the module's `composer.json` file
3. Run the setup and make sure to select the user portal in the options
4. See the error message during compilation

_**Note:** You'll find attach to this PR a minimal module with the corresponding `composer.json` file to ease reproduction._

## Cause (bug)
Both the `itop-portal-base/portal` and my module have the same PHP class name for the autoloader: `d751713988987e9331980363e24189ce`

The regression cause is that in 3.2.1, the `itop-portal-base` module's autoloaders were udpated in [9fadbb5e](https://github.com/Molkobain/iTop/commit/9fadbb5eb1edc337308299e99aa2429df8757962#diff-0353af4e8caf0fb7c33e87e9d702831b35fbd276b8e21a62e96f7d647d1b6c09L25). Changing the autoloader class name to `ComposerAutoloaderInitd751713988987e9331980363e24189ce` which is kind of the default one (see root cause section).

## Root cause
In the latest versions of Composer, when there is a `composer.lock`, the suffix of the autoloader class (e.g. `d751713988987e9331980363e24189ce`) is a hash (md5) based on some of the information in the `composer.json` file.

```php
$relevantKeys = [
    'name',
    'version',
    'require',
    'require-dev',
    'conflict',
    'replace',
    'provide',
    'minimum-stability',
    'prefer-stable',
    'repositories',
    'extra',
];
```

If none of the properties are defined in the `composer.json`, then the hash will always be the same: `d751713988987e9331980363e24189ce`

Composer source references:
  * [AutoloadGenerator](https://github.com/composer/composer/blob/2.8.6/src/Composer/Autoload/AutoloadGenerator.php#L429) calling the `Locker::getContentHash()`
  * [Locker::getContentHash()](https://github.com/composer/composer/blob/2.8.6/src/Composer/Package/Locker.php#L89)

_**Note:** The `composer.lock` file is not generated if you only dump the autoloaders. You have to install/update for it to be generated. Which is not done on all modules._

Add the missing `"name"` information in modules' `composer.json` files and re-dump the autoloaders. It makes sense as module are supposed to be individual packages. Also this is something that is done in many existing Combodo's modules. 

Exception for `itop-portal-base/portal/composer.json` file, as it is not at the root of the module, naming it `combodo/itop-portal-base` would be misleading. Naming it something like `combodo/itop-portal-base_portal` would be weird as well.
So I went with using the `autoloader-suffix` conf. property ([documentation](https://getcomposer.org/doc/06-config.md#autoloader-suffix)) that forces the suffix of the class name.

Mind that we could consider forcing the suffix of autoloaders in all modules, that way we would ensure uniqueness without relying on Composer's black box algorythm. But as it's not the current policy in existing modules, I didn't. Feel free to tell me if you want to adjust the PR accordingly.

## Workaround
For people having the issue with their module, you can juste set a `"name"` property in your `composer.json`:
```json
{
  "name": "molkobain/some-module",
  "autoload": {
    "psr-4": {
      "Molkobain\\iTop\\Extension\\SomeModule\\": "src/"
    }
  }
}
```

- [x] Discuss with Combodo how they would like the unit test to be performed, once we agreed on the solution


Molkobain:
_The sample module [sample-module.zip](https://github.com/user-attachments/files/19329407/sample-module.zip)_

Hipska (approved):

datamodels/2.x/itop-attachments/vendor/autoload.php
```diff
@@ -2,6 +2,24 @@
...
+    trigger_error(
+        $err,
+        E_USER_ERROR
+    );
+}
+
 require_once __DIR__ . '/composer/autoload_real.php';
 
-return ComposerAutoloaderInit478f6fa51842c80229f6a5c4f2ecbd50::getLoader();
+return ComposerAutoloaderInitd751713988987e9331980363e24189ce::getLoader();
```
jf-cbd: Thanks for your PR :D 
On itop-attachments, the new generated string is the one causing trouble (d751713988987e9331980363e24189ce), despite the "name" field added in composer.json in the PR. Is it possible that you performed a `composer install` before adding the "name" field in the composer.json ?

Molkobain: Good catch, thanks! I made so many tests that I made a mistake in the end when sorting everything out. 😅

Molkobain:
Proposal of unit tests:
  * Check every `composer.json` files not in `/lib` nor in `/**/vendor`
    * Assert that its `autoloader_real.php` has not the "default" suffix (will be run by CI for all existing modules included in packages, as well as those with new commits)
    * If on root level of project or module, assert that it has a `"name"` property
    * If not on root level, asset that it has either an `"autoloader-suffix"` or a `"name"` property (for the later it could be a included lib)

Molkobain:
@jf-cbd careful when you change the target branch, sometimes you need to "rebase onto" the original branch as well. Otherwise there is a whole lot of extra commits in the PR like now. You should either do it or poke the author so he does it.

I'll do it :)

Molkobain:
@jf-cbd Done!

Molkobain:
Btw, you changed the target branch to `support/3.2.1` which means that it will be in a micro version (e.g. `3.2.1-1`), is it on purpose?

jf-cbd:
Alright, thanks for doing it @Molkobain !

Concerning the problem of autoloaders class name conflicts, we decided to split the problem in two part : 
The ticket 8278, that fix it in iTop sources, but won't prevent 2 extensions to have conflicts. The fix (yours) will be integrated in 3.2.1-1, so your PR should be merged soon.
The ticket N°8280 (Risk of incompatibility between extensions) is about how to prevent future risks : documentation, tests (we added your proposal in the ticket) and potential other solutions, like warning in the setup.
Its fix will probably arrive in 3.2.2.

Molkobain:
Alright, sounds great!

steffunky:
I met the same issue while working on a client extension with a minimal composer.

image

Adding a `autoloader-suffix` did the trick

jf-cbd:
Perfect, let's merge :)

jf-cbd:
Thank you @Molkobain :wink:

PR #705, opened 2025-03-31
N°8291 - CSV import - column name not automatically recognized in import when name ending by a space

accognet:


pages/ajax.csvimport.php
```diff
@@ -160,7 +160,7 @@ function GetMappingForField($sClassName, $sFieldName, $iFieldIndex, $bAdvancedMo
 			|| ($oAttDef instanceof AttributeFriendlyName)
 		) {
 			$aChoices[$sAttCode] = MetaModel::GetLabel($sClassName, $sAttCode, true);
-			if (($sFieldName == $oAttDef->GetLabel()) || ($sFieldName == $sAttCode)) {
+            if (strcasecmp(trim($sFieldName), trim($oAttDef->GetLabel())) == 0 || (trim($sFieldName) == $sAttCode)) {
```
Hipska, line 163: Should `$oAttDef->GetLabel()` ever return a string with space at the end?

accognet: No, but it can happen.

Hipska: How?

And if so, I think that method should be fixed then..

PR #708, opened 2025-04-01
N°7914 - REST: Allow class based output fields + Allow OQL with multiclass result

accognet:
When performing core/get with multiple return types (parent and child classes) on REST Service, you may either only return fields from the parent object or enforce return all the fields for each object (with *+).

When, for example, requesting information for Tickets of various types and also including some, but not all the custom properties in the output, you can not provide output_fields which may exist in all implementations, if it does not exist in the requested parent implementation.

The following is currently not possible:
```
{
   "operation": "core/get",
   "class": "Ticket",
   "key": "SELECT UserRequest UNION SELECT Change",
   "output_fields": "ref,status,finalclass"
}
```
This is because the "status" is not defined on "Ticket", but rather in each implementation. Querying with "*+" is possible, but then you also request, join, process and transfer a lot of data that was not required.

Proposed solution
Proposed change makes it possible to optionally define a list of output_fields for each class. You can now pass
<Class>:<output_fields>;<Class>:<output_fields>;....
```
{
   "operation": "core/get",
   "class": "Ticket",
   "key": "SELECT UserRequest UNION SELECT Change",
   "output_fields": "Ticket:ref;UserRequest:ref,status,origin;Change:ref,status,outage"
}
```
You must also provide the parents (e.g. "Ticket") output_fields, even though no object of parents type (e.g. "Ticket") could ever be returned directly.
Another solution is to pass all fields and returned classes:
```
{
   "operation": "core/get",
   "class": "UserRequest,Change",
   "key": "SELECT UserRequest UNION SELECT Change",
   "output_fields": "ref,status,origin,outage"
}
```
This 2 call return exactly the same result.

---------------------------------------------------------------
Another case:
The following is currently not possible:
```
{
   "operation": "core/get",
   "class": "Ticket",
   "key": "SELECT P, PR, F
  FROM Person AS P
  JOIN Person AS PR ON P.manager_id = PR.id
  JOIN lnkContactToFunctionalCI AS L ON L.contact_id = P.id
  JOIN FunctionalCI AS F ON L.functionalci_id = F.id",
   "output_fields": "friendlyname, email, organization_name,  move2production"
}
```
Now this kind of call return the same results as in screen

------------------------------------------------------------
This PR merges the code of https://github.com/Combodo/iTop/pull/668 + code send by a client.

### DIFF
--- a/application/applicationextension/rest/RestUtils.php
+++ b/application/applicationextension/rest/RestUtils.php
@@ -107,6 +107,15 @@ class RestUtils
 		};
 	}
 
+	/**
+	 * Check if the requested field list asks for an extended output.
+	 *
+	 * Extended output is requested when using '*+' or class-scoped field definitions.
+	 *
+	 * @param string $sFields Requested field specification.
+	 *
+	 * @return bool
+	 */
 	public static function HasRequestedExtendedOutput(string $sFields): bool
 	{
 		return match($sFields) {
@@ -116,6 +125,13 @@ class RestUtils
 		};
 	}
 
+	/**
+	 * Check if the requested field list asks for all output fields.
+	 *
+	 * @param string $sFields Requested field specification.
+	 *
+	 * @return bool
+	 */
 	public static function HasRequestedAllOutputFields(string $sFields): bool
 	{
 		return match($sFields) {
@@ -129,6 +145,13 @@ class RestUtils
 		return [$sClass => array_keys(MetaModel::ListAttributeDefs($sClass))];
 	}
 
+	/**
+	 * Build a field list for all child classes of the given parent class.
+	 *
+	 * @param string $sClass Parent class name.
+	 *
+	 * @return array Array of class => list of attribute codes.
+	 */
 	protected static function GetFieldListForParentClass(string $sClass): array
 	{
 		$aFieldList = array();
@@ -138,6 +161,17 @@ class RestUtils
 		return $aFieldList;
 	}
 
+	/**
+	 * Build a restricted field list for one class from a comma-separated attribute list.
+	 *
+	 * @param string $sClass Class name.
+	 * @param string $sFields Comma-separated list of requested attribute codes.
+	 * @param string $sParamName Input parameter name used in error messages.
+	 * @param bool $bFailIfNotFound If true, throws when an attribute code is invalid.
+	 *
+	 * @return array Array containing one class => list of attribute codes.
+	 * @throws Exception When an attribute code is invalid and $bFailIfNotFound is true.
+	 */
 	protected static function GetLimitedFieldListForSingleClass(string $sClass, string $sFields, string $sParamName, bool $bFailIfNotFound = true): array
 	{
 		$aFieldList = [$sClass => []];
@@ -154,6 +188,20 @@ class RestUtils
 		return $aFieldList;
 	}
 
+	/**
+	 * Build a restricted field list for one or several classes.
+	 *
+	 * Accepted formats are either "att1,att2" for a single class, or
+	 * "ClassA:att1,att2;ClassB:att3" for class-scoped field definitions.
+	 *
+	 * @param string $sClass Default class name used when no class scope is specified.
+	 * @param string $sFields Requested field specification.
+	 * @param string $sParamName Input parameter name used in error messages.
+	 * @param bool $bFailIfNotFound If true, throws when an attribute code is invalid.
+	 *
+	 * @return array Array of class => list of attribute codes.
+	 * @throws Exception Propagated from GetLimitedFieldListForSingleClass.
+	 */
 	protected static function GetLimitedFieldListForClass(string $sClass, string $sFields, string $sParamName, bool $bFailIfNotFound = true): array
 	{
 		if (!str_contains($sFields, ':')) {
--- a/application/applicationextension/rest/RestUtils.php
+++ b/application/applicationextension/rest/RestUtils.php
@@ -97,33 +97,77 @@ class RestUtils
 	 * @throws Exception
 	 * @api
 	 */
-	public static function GetFieldList($sClass, $oData, $sParamName)
+	public static function GetFieldList($sClass, $oData, $sParamName, $bFailIfNotFound = true)
 	{
 		$sFields = self::GetOptionalParam($oData, $sParamName, '*');
-		$aShowFields = [];
-		if ($sFields == '*') {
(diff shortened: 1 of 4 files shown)


jbostoen:
The idea is nice; but why work with comma separated values?
Perhaps an associative array could be provided for the output fields; with the class name being the key and the value an array of attribute codes?

If the JSON structure is pretty printed, it would look nicer this way :)

Molkobain:
Nice job, I'm very interested by the second use case! 🤩
Like Jeffrey, I would rather have a way to specify output fields for each classes, like it's done in `\DBObjectSet::OptimizeColumnLoad()`.

Hipska:
Yup, I agree..

To keep consistency with original implementation maybe it would look like this?
```json
{
   "operation": "core/get",
   "class": "Ticket",
   "key": "SELECT UserRequest UNION SELECT Change",
   "output_fields": {
      "Ticket": "ref",
      "UserRequest": "ref,status,origin",
      "Change": "ref,status,outage"
   }
}
```

But it looks like @jbostoen and @Molkobain are hinting to this:
```json
{
   "operation": "core/get",
   "class": "Ticket",
   "key": "SELECT UserRequest UNION SELECT Change",
   "output_fields": {
      "Ticket": [
         "ref"
      ],
      "UserRequest": [
         "ref",
         "status",
         "origin"
      ],
      "Change": [
         "ref",
         "status",
         "outage"
      ]
   }
}
```

jbostoen:
An array of strings just makes more sense to me in terms of APIs, rather than concatenating values first. 

The original format was perhaps not optimal to begin with in this aspect.

I'm also missing a point in the example.

From the original proposal (it might just be the example):
```
{
   "operation": "core/get",
   "class": "Ticket",
   "key": "SELECT UserRequest UNION SELECT Change",
   "output_fields": "Ticket:ref;UserRequest:ref,status,origin;Change:ref,status,outage"
}
```

What would the purpose/objective be of 'Ticket' here?
If it's meant to list attributes on parent classes that we want; I understand; but why repeat it for the specific classes?

Hipska:
Hi @accognet, is there any update on this?

accognet:
Hi, this is planned for iTop 3.3

application/applicationextension/rest/RestUtils.php
```diff
@@ -97,33 +97,77 @@ public static function GetClass($oData, $sParamName)
...
-				if (($sAttCode != 'id') && (!MetaModel::IsValidAttCode($sClass, $sAttCode))) {
-					throw new Exception("$sParamName: invalid attribute code '$sAttCode'");
+		return match($sFields) {
+			'*' => self::GetFieldListForClass($sClass),
+			'*+' => self::GetFieldListForParentClass($sClass),
+			default => self::GetLimitedFieldListForClass($sClass, $sFields, $sParamName, $bFailIfNotFound),
+		};
+	}
+
+	public static function HasRequestedExtendedOutput(string $sFields): bool
```
odain-cbd, line 119: please add some phpdoc

application/applicationextension/rest/RestUtils.php
```diff
@@ -97,33 +97,77 @@ public static function GetClass($oData, $sParamName)
...
+	{
+		return match($sFields) {
+			'*' => false,
+			'*+' => true,
+			default => substr_count($sFields, ':') > 1,
+		};
+	}
+
+	public static function HasRequestedAllOutputFields(string $sFields): bool
```
odain-cbd, line 135: please add some phpdoc

odain-cbd: (please add some phpdoc to all freshly created methods)

application/applicationextension/rest/RestUtils.php
```diff
@@ -97,33 +97,77 @@ public static function GetClass($oData, $sParamName)
...
+		}
+		return $aFieldList;
+	}
+
+	protected static function GetLimitedFieldListForSingleClass(string $sClass, string $sFields, string $sParamName, bool $bFailIfNotFound = true): array
+	{
+		$aFieldList = [$sClass => []];
+		foreach (explode(',', $sFields) as $sAttCode) {
+			$sAttCode = trim($sAttCode);
+			if (($sAttCode == 'id') || (MetaModel::IsValidAttCode($sClass, $sAttCode))) {
```
odain-cbd, line 180: ($sAttCode === 'id') is better

tests/php-unit-tests/unitary-tests/webservices/RestUtilsTest.php
```diff
@@ -0,0 +1,110 @@
...
+		$aList = RestUtils::GetFieldList(Ticket::class, (object) ['output_fields' => 'ref,start_date,end_date'], 'output_fields');
+		$this->assertSame([Ticket::class => ['ref', 'start_date', 'end_date']], $aList);
+	}
+
+	public function testGetFieldListForSingleClassWithInvalidFieldNameFails(): void
+	{
+		$this->expectException(\Exception::class);
+		$this->expectExceptionMessage('output_fields: invalid attribute code \'something\' for class \'Ticket\'');
+		$aList = RestUtils::GetFieldList(Ticket::class, (object) ['output_fields' => 'ref,something'], 'output_fields');
+		$this->assertSame([Ticket::class => ['ref', 'start_date', 'end_date']], $aList);
```
odain-cbd, line 25: is this test sucessfull? RestUtils::GetFieldList should raise exception. the no value is returned if i understood correctly

tests/php-unit-tests/unitary-tests/webservices/RestUtilsTest.php
```diff
@@ -0,0 +1,110 @@
...
+
+	/**
+	 * @dataProvider allFieldsOutputDataProvider
+	 */
+	public function testIsAllFieldsOutputRequest(bool $bExpected, string $sFields): void
+	{
+		$this->assertSame($bExpected, RestUtils::HasRequestedAllOutputFields($sFields));
+	}
+
+	public function extendedOutputDataProvider(): array
```
odain-cbd: please move 2 below provider functions close to the testXXX that is relying on it. you could also make them static for PHP 8.4 or 8.5 compatibility.

odain-cbd (approved):
few small feedbacks only. great enhancement. if it is accepted it could be nice to document it by adding a new section in combodo-api-playground extension.

Hipska:
Hi @accognet, could you clarify how the `output_fields` now should look like? What did you think about our proposals?

Hipska:
Hi @accognet, why did you ignore us? Are any of our proposals considered? How should `output_fields` look like in the future?

PR #709, opened 2025-04-02
:bug: N°8345 - Fix double URL encoding in search URLs

Molkobain:
## Base information

## Symptom (bug)
Symptom only appears on some webservers, it seems to be an edge case. But the cause and fix seem legit.

When trying to open a backoffice URL pointing to a search page, the following error occurs everytime:
```
Service Unavailable
The server is temporarily unable to service your request due to maintenance downtime or capacity problems. Please try again later.
```

It happens when the URL is generated by:
  * A dashlet (badge, dynamic, pie chart, bar chart, objects list)
  * An objects list (search page, linkset, ...)
  * An error during CSV import

Unfortunately, I can't find a more explicit error message. There is nothing in iTop's nor Apache's error logs. 🫤

I can't find a reproduction procedure that works everywhere. It only happens on my ionos webserver on which I don't have access to the Apache / PHP configuration. 😭😭😭

1. Go on http://mbc.itop.molkobain.com/pages/UI.php
2. Log with admin / admin
3. Click on the first "objects list" dashlet subtitle "Total: 1 objects."
4. See the error occurring

_P.S: Mind that I patched the badge and dynamic dashlets code in order to verify cause / fix._

## Cause (bug)
As this only happens on my ionos webservers (but on all 3 of them, with different iTop instances / versions), I don't know what the system cause is.

Nonetheless, there is a issue with how search URLs are generated in the backoffice. If you look at the `filter` param, you'll see that it is URL encoded **twice**, hence the 4 hexa characters after the `%`.
```
http://mbc.itop.molkobain.com/pages/UI.php?operation=search&&filter=%255B%2522SELECT%2B%2560Rack%2560%2BFROM%2BRack%2BAS%2B%2560Rack%2560%2BWHERE%2B1%2522%252C%255B%255D%252C%255B%255D%255D&aParams=%7B%22table_id%22%3A%22DashletCUSTOM_WelcomeMenuPage_ID_row1_col0_16%22%7D
```

Why is it encoded twice?
  * First time occurs when serializing the filter in `\DBSearch::serialize()`
  * Second time occurs when building the URL manually, some parts of the code do something like `rawurlencode($oFilter->serialize())`

Changes seem to have been done during the development of the 3.0 and its UI refacto. The complexity of this refacto and its multiple iterations could explain the mixup.

  * Rely on the encoding done by `\DBSearch::serialize()`
  * Remove `rawurlencode()` when `\DBSearch::serialize()` is used (e.g. `rawurlencode($oFilter->serialize())`)

Mind that in some places, something like `rawurlencode($oFilter->ToOQL())` is used. It could be replaced by `$oFilter->serialize()` for better robustness ([Example](https://github.com/Combodo/iTop/blob/support/3.2/application/displayblock.class.inc.php#L1713)). I have **not** made these changes, but I'll be happy to if that what you want.

- [ ] Check with Combodo if they would like a unit test or if E2E is preferrable, and discuss how they would like it to be done

### DIFF
--- a/application/dashlet.class.inc.php
+++ b/application/dashlet.class.inc.php
@@ -2138,7 +2138,7 @@ class DashletHeaderDynamic extends Dashlet
 		$oSet = new DBObjectSet($oFilter);
 		$iCount = $oSet->Count();
 		$oAppContext = new ApplicationContext();
-		$sHyperlink = utils::GetAbsoluteUrlAppRoot().'pages/UI.php?operation=search&'.$oAppContext->GetForLink().'&filter='.rawurlencode($oFilter->serialize());
+		$sHyperlink = utils::GetAbsoluteUrlAppRoot().'pages/UI.php?operation=search&'.$oAppContext->GetForLink().'&filter='.$oFilter->serialize();
 		$oSubTitle->AddHtml('<a class="summary" href="'.$sHyperlink.'">'.Dict::Format(str_replace('_', ':', $sSubtitle), $iCount).'</a>');
 
 		return $oPanel;
--- a/application/displayblock.class.inc.php
+++ b/application/displayblock.class.inc.php
@@ -1125,7 +1125,7 @@ JS
 					}
 					$sHyperlink = utils::GetAbsoluteUrlAppRoot()
 						.'pages/UI.php?operation=search&'.$oAppContext->GetForLink()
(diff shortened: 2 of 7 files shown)


jf-cbd:
Thank you for your PR, unfortunately we can't reproduce your bug (except in the env you provided), and we're not willing to examine it in a future sprint. Feel free to reopen it if you can provide more details about the cause if the bug.

PR #711, opened 2025-04-25
N°6925 - Enable on any search result edition of OQL

accognet:
Adds a new action to the list menu
image
which opens the run_query page with the list query

### DIFF
--- a/application/utils.inc.php
+++ b/application/utils.inc.php
@@ -1555,6 +1555,11 @@ class utils
 				}
 				$aResult[] = new JSPopupMenuItem('UI:Menu:AddToDashboard', Dict::S('UI:Menu:AddToDashboard'), "DashletCreationDlg('$sOQL', '$sContext')");
 				$aResult[] = new JSPopupMenuItem('UI:Menu:ShortcutList', Dict::S('UI:Menu:ShortcutList'), "ShortcutListDlg('$sOQL', '$sDataTableId', '$sContext')");
+                if (ApplicationMenu::IsMenuIdEnabled('RunQueriesMenu')) {
+                    $oMenuItemPlay = new JSPopupMenuItem('UI:Menu:OpenOQL', Dict::S('UI:Edit:TestQuery'), "OpenOql('$sOQL')");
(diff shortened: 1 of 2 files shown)


application/utils.inc.php
```diff
@@ -1555,6 +1555,11 @@ public static function GetPopupMenuItemsBlock(iUIBlock &$oContainerBlock, $iMenu
 				}
 				$aResult[] = new JSPopupMenuItem('UI:Menu:AddToDashboard', Dict::S('UI:Menu:AddToDashboard'), "DashletCreationDlg('$sOQL', '$sContext')");
 				$aResult[] = new JSPopupMenuItem('UI:Menu:ShortcutList', Dict::S('UI:Menu:ShortcutList'), "ShortcutListDlg('$sOQL', '$sDataTableId', '$sContext')");
+                if (ApplicationMenu::IsMenuIdEnabled('RunQueriesMenu')) {
+                    $oMenuItemPlay = new JSPopupMenuItem('UI:Menu:OpenOQL', Dict::S('UI:Edit:TestQuery'), "OpenOql('$sOQL')");
```
Hipska, line 1559: Why is this a JS menu item while it could just be a URL popup menu? (The JS is just generating the URL)

js/utils.js
```diff
@@ -369,6 +369,12 @@ function DashletCreationDlg(sOQL, sContext) {
 	return false;
 }
 
+function OpenOql(sOQL) {
+	sBaseUrl = GetAbsoluteUrlAppRoot() + 'pages/run_query.php?expression=';
+	sQueryEncoded = encodeURI(sOQL);
+	window.open(sBaseUrl + sQueryEncoded, '_blank');
```
rquetiez: Since the OQL can be built programmatically, it may be long enough so that the URL exceeds 2000 characters. To handle any size of OQL, the parameter must be posted instead.

rquetiez (changes requested):
Make it safe with regard of the length of the provided OQL.

application/utils.inc.php
accognet: The code has changed and js is now used for more than just building a url

PR #712, opened 2025-05-06
N°8347 - Bad display of special characters on the portal CheckToWrite message

accognet:
internal


rquetiez (approved):

odain-cbd:
@rquetiez @accognet 
there are no unit test provided with this fix. it does not seem so hard to me.

PR #716, opened 2025-05-22
Fix SynchronizeProfiles()

TdlQ:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | no

We use hybridauth with user provisioning enabled, profiles are defined depending on Okta groups of the external user (it requires some patches to https://github.com/Combodo/combodo-hybridauth to work). The SynchronizeProfiles() function works only when the existing profile list is empty, while creation is OK, updates aren't done correctly and this PR fixes this point.

In current state of iTop, bug can't be triggered because a User object is created once and never updated after that. I've modified HybridAuthLoginExtension.php so it updates the User (and his profiles) everytime.

## Cause (bug)

One issue is related to false and 0 being mixed up (return of array_search). Second issue is that Fetch() does not handle modified list very well and a deletion prevent the next item to be properly evaluated.

See code, fixes are trivial.

### DIFF
--- a/application/loginwebpage.class.inc.php
+++ b/application/loginwebpage.class.inc.php
@@ -128,19 +128,25 @@ class LoginWebPage extends NiceWebPage
 		$oProfilesSet = $oUser->Get('profile_list');
 		//delete old profiles
 		$aExistingProfiles = [];
+		$aProfilestoRemove = [];
 		while ($oProfile = $oProfilesSet->Fetch())
 		{
 			array_push($aExistingProfiles, $oProfile->Get('profileid'));
 			$iArrayKey = array_search($oProfile->Get('profileid'), $aProfiles);
 			if ($iArrayKey === false)
 			{
-				$oProfilesSet->RemoveItem($oProfile->Get('profileid'));
+				$aProfilestoRemove[] = $oProfile;
 			}
 			else
 			{
 				unset($aProfiles[$iArrayKey]);
 			}
 		}
+
+		foreach ($aProfilestoRemove as $oProfile) {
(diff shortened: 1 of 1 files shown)


TdlQ:
To make this PR relevant, the modifications of HybridAuthLoginExtension.php are in this other PR: https://github.com/Combodo/combodo-hybridauth/pull/7

odain-cbd:
Thank you for this contribution. Provisioning enhancement has been moved into combodo-hybridauth extension to release it apart from iTop official releases.
https://github.com/Combodo/combodo-hybridauth/commit/7de4b1aa41e6cc6560dd7bb36d276772102386d0

jf-cbd:
@TdlQ could you please confirm that this PR can be closed, since the edition made in combodo-hybridauth should provide the same functionality ? (If I'm not wrong, you tested it, right ?)
Thank you

TdlQ:
> @TdlQ could you please confirm that this PR can be closed, since the edition made in combodo-hybridauth should provide the same functionality ? (If I'm not wrong, you tested it, right ?) Thank you

Yes, it can be closed. We've been using for a few weeks the full OktaOIDC feature with User provisioning and everything is working properly

jf-cbd:
That's perfect, thanks for your feedback.
Thank you for contribution (on both repos) that helped to fix the issue.
We're grateful for that, and we'd like to send you [stickers](https://github.com/Combodo/iTop/blob/develop/CONTRIBUTING.md#-we-are-thankful). Would you like some ? If yes, please send us your address at community[at]combodo.com

PR #720, opened 2025-05-23
N°8575 - chore(auth-ldap): Prepare for php 8.3 deprecation

Hipska:
## Base information

## Objective

According to [the docs](https://www.php.net/manual/en/function.ldap-connect.php), the signature `ldap_connect(?string $host = null, int $port = 389)` is deprecated as of PHP 8.3.0.

## Proposed solution

Introduce new `uri` module setting and provide migration to be able to use the preferred way of `ldap_connect`.

### DIFF
--- a/datamodels/2.x/authent-ldap/module.authent-ldap.php
+++ b/datamodels/2.x/authent-ldap/module.authent-ldap.php
@@ -43,8 +43,7 @@ SetupWebPage::AddModule(
 		// Default settings
 		//
 		'settings' => array(
-			'host' => 'localhost', // host or IP address of your LDAP server
-			'port' => 389,		  // LDAP port (std: 389)
+			'uri' => 'ldap://localhost', // URI with host or IP address of your LDAP server
 			'default_user' => '', // User and password used for initial "Anonymous" bind to LDAP
 			'default_pwd' => '',  // Leave both blank, if anonymous (read-only) bind is allowed
 			'base_dn' => 'dc=yourcompany,dc=com', // Base DN for User queries, adjust it to your LDAP schema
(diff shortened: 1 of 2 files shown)


datamodels/2.x/authent-ldap/module.authent-ldap.php
```diff
@@ -75,6 +74,27 @@ public static function AfterDataLoad(Config $oConfiguration, $sPreviousVersion,
...
 	}
+
+	public static function BeforeWritingConfig(Config $oConfiguration)
+	{
+		$sURI = $oConfiguration->GetModuleSetting('authent-ldap', 'uri');
+		if (empty($sURI)) {
+			$sLDAPHost = MetaModel::GetModuleSetting('authent-ldap', 'host', 'localhost');
+			$iLDAPPort = MetaModel::GetModuleSetting('authent-ldap', 'port', 389);
+			$sURI = preg_match('#^ldaps?://#i', $sLDAPHost) ? $sLDAPHost : 'ldap://'.$sLDAPHost.':'.$iLDAPPort;
+			$oConfiguration->SetModuleSetting('authent-ldap', 'uri', $sURI);
```
jf-cbd, line 85: I'm getting inspiration from what we did on the LDAP collector : https://github.com/Combodo/itop-data-collector-ldap/blob/master/src/LDAPSearchService.class.inc.php#L193
```php
        if (preg_match('@^(ldap://|ldaps://)@', $sLDAPHost)) {
            if ($iLDAPPort != '') {
                $oConfiguration->SetModuleSetting('authent-ldap', 'uri', "$sLDAPHost:$iLDAPPort");
                return ;
            } else {
                $oConfiguration->SetModuleSetting('authent-ldap', 'uri', $sLDAPHost);
            }
        } else {
            if ($sPort != '389') {
                $oConfiguration->SetModuleSetting('authent-ldap', 'uri', "ldaps://$sLDAPHost:$iLDAPPort");
            } else {
                $oConfiguration->SetModuleSetting('authent-ldap', 'uri', "ldap://$sLDAPHost");
            }
        }
```

datamodels/2.x/authent-ldap/module.authent-ldap.php
```diff
@@ -75,6 +74,27 @@ public static function AfterDataLoad(Config $oConfiguration, $sPreviousVersion,
...
+		}
+
+		$aServers = $oConfiguration->GetModuleSetting('authent-ldap', 'servers', []);
+		foreach ($aServers as &$aServer) {
+			if (!array_key_exists($aServer, 'uri')) {
```
jf-cbd, line 90: Same here

jf-cbd (changes requested):
Hey, thanks for your PR ! Sounds all good, but I'm proposing a change on the way to "compute" the URI, so it's similar to what we did in another extension :)

datamodels/2.x/authent-ldap/module.authent-ldap.php
Hipska: I'm doing the same. Except adding the port, as it does not make sense, since the port is ignored when the host was having the `ldap://` prefix.

When actually doing this like your suggestion, it will break stuff when users already have a port in their URI or when they use for example LDAPS with the URI but didn't update the `port` parameter...

I prefer not to potentially break stuff 😉

jf-cbd: >When actually doing this like your suggestion, it will break stuff when users already have a port in their URI or when they use for example LDAPS with the URI but didn't update the port parameter...

If the URI is defined, then the code below won't be effective (since it's in an "`if (empty($sURI)) {`" statement).

But I guess you are talking about the $sLDAPHost, not the URI, and in this case, your notice makes sense (the 1st one, I didn't get the 2nd one, "when they use for example LDAPS with the URI but didn't update the port parameter").

Hipska: Take this for example:
```
		'host' => 'ldaps://ad1.example.com ldaps://ad2.example.com',
		'port' => 389,
```

You're suggesting to transform this to
```
		'uri' => 'ldaps://ad1.example.com ldaps://ad2.example.com:389',
```
Which will not work as expected. Effectively, the last configured server will not work anymore.

Another example:
```
		'host' => 'ldap://ad1.example.com:3268',
		'port' => 3268,
```
Which will become:
```
		'uri' => 'ldap://ad1.example.com:3268:3268',
```

My suggestion is to keep `host` as a whole when it contains `ldaps://` or `ldap://` and only craft a new URI with the port when it doesn't.

Hipska: @jf-cbd What do you think about that?

jf-cbd: Hey, sounds all good indeed :)

jf-cbd:
Thanks for your contribution @Hipska !

PR #721, opened 2025-05-26
Fix 8198 by setting making ModuleInstallation extends DBObject + bett…

jf-cbd:
Fix 8198 by setting making ModuleInstallation extends DBObject + better check on object creation

### DIFF
--- a/sources/Controller/Base/Layout/ObjectController.php
+++ b/sources/Controller/Base/Layout/ObjectController.php
@@ -82,7 +82,9 @@ class ObjectController extends AbstractController
 		{
 			throw new ApplicationException(Dict::Format('UI:Error:1ParametersMissing', 'class'));
 		}
-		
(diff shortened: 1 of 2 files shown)


sources/Controller/Base/Layout/ObjectController.php
```diff
@@ -82,7 +82,9 @@ public function OperationNew()
 		{
 			throw new ApplicationException(Dict::Format('UI:Error:1ParametersMissing', 'class'));
 		}
-		
+		if (!is_subclass_of($sClass, 'cmdbAbstractObject')) {
```
Hipska:
```suggestion
		if (!is_subclass_of($sClass, cmdbAbstractObject::class)) {
```

jf-cbd: Committed, thanks for the suggestion

Hipska (approved):

bdalsass (approved):

PR #727, opened 2025-07-14
fix: N°8522 - check if org id is null when filtering

hakonharnes:
## Base information

## Symptom 

When querying objects with optional relationships (resulting in LEFT JOINs) where org_id-based access restrictions apply, the query incorrectly filters out rows where the joined object is NULL, effectively converting LEFT JOINs into INNER JOINs.

For example, with the query:
```sql
SELECT VM, OSL
FROM VirtualMachine AS VM
JOIN OSLicence AS OSL ON VM.oslicence_id = OSL.id
```

When `OSLicence` is an optional field for `VirtualMachine` (LEFT JOIN), VMs without an assigned OSLicence should still appear in results with NULL values for the OSL columns. However, for users with organization restrictions, these VMs are completely filtered out. This is because the resulting SQL statement contains this condition: `OSL_Licence.org_id IN ('3', '2', ...)))`, which evaluates to `NULL IN ('3','2',...)`.

This means users with organization restrictions see fewer results than they should; they only see VMs that have an OSLicence assigned, missing all VMs without licenses.

## Reproduction procedure 

1. On iTop 3.2.1-1
2. With PHP 8.1.33
3. Create a user with specific allowed organizations (e.g., org_id 2 and 71)
4. Create several VirtualMachines, some with `oslicence_id` set and some without
5. Execute the OQL query: `SELECT VM, OSL FROM VirtualMachine AS VM JOIN OSLicence AS OSL ON VM.oslicence_id = OSL.id`
6. Only VMs with an assigned OSLicence appear in results (acts like INNER JOIN)
7. Switch to a user with no organization restrictions (empty allowed orgs)
8. Execute the same query - now VMs without OSLicence also appear with NULL values for OSL fields (proper LEFT JOIN behavior)

## Cause
The `MakeSelectFilter` method in `core/userrights.class.inc.php` generates a condition that only checks if `org_id IN (allowed_orgs)`. For LEFT JOINs where the joined object can be NULL, this incorrectly filters out valid rows.

The generated WHERE clause before the fix:
```sql
WHERE (`VM_FunctionalCI`.`org_id` IN ('3', '2', '7')) 
  AND (`OSL_Licence`.`org_id` IN ('3', '2', '7'))
```

When `OSL_Licence` is NULL (no license assigned), `OSL_Licence`.`org_id` is also NULL, and `NULL IN ('3', '2', '7')` returns false, removing the entire row from results.

## Proposed solution 

Modified the organization filter condition in `MakeSelectFilter` (core/userrights.class.inc.php:145-154) to explicitly handle NULL values:

```php
// Create IS NULL condition
$oNullCondition = new BinaryExpression($oExpression, 'IS', new ScalarExpression(null));

// Create IN condition
$oInCondition = new BinaryExpression($oExpression, 'IN', $oListExpr);

// Combine with OR: field IS NULL OR field IN (allowed_orgs)
$oCondition = $oNullCondition->LogOr($oInCondition);

$oFilter->AddConditionExpression($oCondition);
```

This now generates the corrected SQL:
```sql
WHERE ((`VM_FunctionalCI`.`org_id` IS NULL) OR (`VM_FunctionalCI`.`org_id` IN ('3', '2', '7'))) 
  AND ((`OSL_Licence`.`org_id` IS NULL) OR (`OSL_Licence`.`org_id` IN ('3', '2', '7')))
```

This preserves LEFT JOIN semantics by allowing rows where the joined object is NULL, while still enforcing organization restrictions for non-NULL values. 

I would like feedback on this PR before attempting to write unit tests (if required).

### DIFF
--- a/core/userrights.class.inc.php
+++ b/core/userrights.class.inc.php
@@ -142,7 +142,7 @@ abstract class UserRightsAddOnAPI
 		$oFilter  = new DBObjectSearch($sClass);
 		$oListExpr = ListExpression::FromScalars($aAllowedOrgs);
 
-		$oNullCondition = new BinaryExpression($oExpression, 'IS', new ScalarExpression(null));
+		$oNullCondition = new FunctionExpression('ISNULL', [$oExpression]);
 		$oInCondition = new BinaryExpression($oExpression, 'IN', $oListExpr);
 		$oCondition = $oNullCondition->LogOr($oInCondition);
 		$oFilter->AddConditionExpression($oCondition);
--- a/core/userrights.class.inc.php
+++ b/core/userrights.class.inc.php
@@ -142,7 +142,9 @@ abstract class UserRightsAddOnAPI
 		$oFilter  = new DBObjectSearch($sClass);
 		$oListExpr = ListExpression::FromScalars($aAllowedOrgs);
 
-		$oCondition = new BinaryExpression($oExpression, 'IN', $oListExpr);
+		$oNullCondition = new BinaryExpression($oExpression, 'IS', new ScalarExpression(null));
+		$oInCondition = new BinaryExpression($oExpression, 'IN', $oListExpr);
+		$oCondition = $oNullCondition->LogOr($oInCondition);
 		$oFilter->AddConditionExpression($oCondition);
 
 		if ($this->HasSharing())
--- a/tests/php-unit-tests/unitary-tests/core/OQLTest.php
+++ b/tests/php-unit-tests/unitary-tests/core/OQLTest.php
@@ -502,50 +502,46 @@ SELECT
 		];
 	}
 
-	public function testOQLAllowedOrg()
+	private function GivenVMAndLicence(int $iVmOrgId, ?int $iLicenceOrgId): int
 	{
-		$sOrgId = $this->GivenObjectInDB('Organization', [
-			'name' => 'Test organization',
-			'status' => 'active'
-		]);
-
 		$iOsFamilyId = $this->GivenObjectInDB('OSFamily', [
-			'name' => 'Test OS Family'
+			'name' => 'Test OS Family',
 		]);
 
 		$iOsVersionId = $this->GivenObjectInDB('OSVersion', [
-			'name' => 'Test OS Version',
-			'osfamily_id' => $iOsFamilyId
+			'name'        => 'Test OS Version',
+			'osfamily_id' => $iOsFamilyId,
 		]);
 
-		$iOsLicenceId = $this->GivenObjectInDB('OSLicence', [
-			'name' => 'Test OS Licence',
+		$iOSLicenceId = $this->GivenObjectInDB('OSLicence', [
+			'name'         => 'Test OS Licence',
 			'osversion_id' => $iOsVersionId,
-			'org_id' => $sOrgId
+			'org_id'       => $iLicenceOrgId,
 		]);
 
 		$iVClusterId = $this->GivenObjectInDB('Hypervisor', [
-			'name' => 'Test Hypervisor',
-			'org_id' => $sOrgId
+			'name'   => 'Test Hypervisor',
+			'org_id' => $iVmOrgId,
 		]);
 
-		$iVmWithOsLicenceId = $this->GivenObjectInDB('VirtualMachine', [
-			'name' => 'Test VM with OS Licence',
-			'org_id' => $sOrgId,
+		return $this->GivenObjectInDB('VirtualMachine', [
+			'name'           => 'Test VM with OS Licence',
+			'org_id'         => $iVmOrgId,
 			'virtualhost_id' => $iVClusterId,
-			'oslicence_id' => $iOsLicenceId
+			'oslicence_id'   => $iOSLicenceId,
 		]);
+	}
 
-		$iVmWithoutOsLicenceId = $this->GivenObjectInDB('VirtualMachine', [
-			'name' => 'Test VM without OS Licence',
-			'org_id' => $sOrgId,
-			'virtualhost_id' => $iVClusterId
-		]);
+	public function testMultiColumnQueryBehaviorWithOrganizationRestrictions()
+	{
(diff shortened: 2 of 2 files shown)


jf-cbd:
Hello @hakonharnes, thanks for proposing us your fix :)
Since you created a related ticket, your pull request has been reviewed in bug review.
Functionnally, it's very interesting, the fix will probably be included in itop 3.3.
We should perform a technical review on August 1st to check your fix, and we'll come back to you.

core/userrights.class.inc.php
```diff
@@ -142,7 +142,9 @@ public function MakeSelectFilter($sClass, $aAllowedOrgs, $aSettings = array(), $
 		$oFilter  = new DBObjectSearch($sClass);
 		$oListExpr = ListExpression::FromScalars($aAllowedOrgs);
 
-		$oCondition = new BinaryExpression($oExpression, 'IN', $oListExpr);
+		$oNullCondition = new BinaryExpression($oExpression, 'IS', new ScalarExpression(null));
```
jf-cbd: Could you please use : 
		`$oNullCondition = new FunctionExpression('ISNULL', [$oExpression]);`
Which leads to the same result but is compliant with the OQL syntax.

jf-cbd (commented):
Thanks again for this (very) detailed PR !
Just a question : you said you added some unit tests, but I don't see them. Did you forget to add them ? (if not, we can help to create them)

core/userrights.class.inc.php
Hipska: Better to make this as a suggestion: :wink:
```suggestion
		$oNullCondition = new FunctionExpression('ISNULL', [$oExpression]);
```

hakonharnes:
> Thanks again for this (very) detailed PR !

Thanks! It's a bit verbose but it's a subtle bug :)

> you said you added some unit tests, but I don't see them. Did you forget to add them ?

I just wanted to get some feedback on the PR before attempting to write unit tests. Let me know which unit tests you would like me to write and I can give it a try. 

> Could you please use :
`$oNullCondition = new FunctionExpression('ISNULL', [$oExpression]);`
Which leads to the same result but is compliant with the OQL syntax.

I initially had concerns about `ISNULL()` vs `IS NULL` performance (see [this discussion](https://www.sqlservercentral.com/forums/topic/isnull-vs-is-null-performance)). However, I tested both approaches with MariaDB and found that the execution plans are identical. The MariaDB optimizer recognizes that `ISNULL(column)` is functionally equivalent to column `IS NULL` and generates the same index access strategy for both. Performance testing confirmed no measurable difference even in complex queries with multiple JOINs.

Given that there's no performance impact and this aligns with the OQL syntax conventions, I have proceeded with `FunctionExpression('ISNULL')` as suggested.

Note: These were just quick tests on my side. I'd encourage others to verify if they have performance concerns. I'm no SQL expert and didn't test exhaustively.

hakonharnes:
@jf-cbd, @Hipska: Let me know which unit tests you'd like me to write.

jf-cbd:
Hello @hakonharnes,
The verbose really made it easy to understand, so it was really appreciated !
Thanks also for your explanation concerning your researches about ISNULL() vs IS NULL performances.

Concerning unit tests, I'd suggest making a unit test that describe the use case you wrote in the description of the PR.

To do that, you can create a minimum set of data (like Organisations and VirtualMachines); then you can use methods like GivenUserInDB and GivenUserRestrictedToAnOrganizationInDB (available in ItopDataTestCase) to be able to create the 2 different users. Then you can use UserRights::Login to log in with them, and CMDBSource::QueryToArray to get the result and compare them to the expected behavior.

If you want and if you're comfortable with it (we did some documentation to write unit test on iTop [here](https://github.com/Combodo/iTop/tree/develop/tests/php-unit-tests)), you can give it a try, but if you don't feel it, do not hesitate to tell :)

hakonharnes:
Hi again, 

I'm not able to setup the test environment properly. If you could write the unit test that would be very helpful :)

jf-cbd:
Sure, I just did a commit with a unit test failing before the commits of your PR, and succeeding with them :)

Concerning the test environment, could you please tell me the difficulties you're facing ?
We'd like to make it easier for people to run/create test, and we're wondering how we can do that.

hakonharnes:
Thanks for adding the unit test. It works perfectly!

Regarding the test environment difficulties I faced, it was mainly due to `GivenObjectInDB` function.  VirtualMachine and OSLicence classes exist in the main iTop environment but I initially thought they weren't available in the test environment because `GivenObjectInDB('VirtualMachine', ...)` was failing. 

I see you used ` MetaModel::NewObject()` + `DBInsert()` approach in stead. That works just fine in my environment. I think it just boils down to me not understanding the patterns correctly here, rather than the test environment being incorrect. I'm able to run your test just fine.

Let me know if you need something from me before merging the PR :)

jf-cbd:
You're welcome :)

Well you're right, I could have used GivenObjectInDB method - and I just did :D
I think it's better now, thanks to your suggestion. Both ways work, but GivenObjectInDB is more appropriated for tests.
Can you run the test now ?

Concerning your PR, one more person will look at it, and then it should be merged, if there is positive feedback. We'll keep you updated :)

hakonharnes:
I can confirm the refactored test with `GivenObjectInDB` works!

jf-cbd:
Hey @hakonharnes,
Thanks again for your PR and quick responses!
I just merged it, so the change will be effective in the 3.3 version of iTop.
We would like to add you to the contributors list in our README.md file. What would you like us to add? (Of course, there's nothing mandatory, it's up to you)
```Håkon Harnes (a.k.a. @hakonharnes)``` ?
As a thank you, we send stickers to people who contribute to iTop and its ecosystem. Would you like to receive some?
If so, please send your postal address to community[at]combodo.com.

hakonharnes:
Hi, 

Sure! Please add me to the list of contributors. Stickers are welcome too. I'll send my address :)

Thanks.

PR #728, opened 2025-07-16
N°8419 - Improve Setup process by avoiding the update of the attachment table when contact_id is already set

bdalsass:
Setup process is quite long with big database containing lot of attachments.

Launch a setup with lot of Attachments.

## Cause (bug)
When adding the feature:
 https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=6619&c[menu]=SearchBugs
The setup launch attachments process compatibility each time.

Only launch the compatibility process when the setup is done on older databases version than 3.2.0

### DIFF
--- a/datamodels/2.x/itop-attachments/module.itop-attachments.php
+++ b/datamodels/2.x/itop-attachments/module.itop-attachments.php
@@ -201,7 +201,7 @@ SQL;
 			}
 			SetupLog::Info("Initializing attachment/item_org_id - $iUpdated records have been adjusted");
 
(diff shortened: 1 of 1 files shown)


Hipska (approved):

Molkobain (approved):

steffunky (approved):

Hipska:
Be aware that the query itself is still not performant.

This query runs (still slow) much faster:
```sql
UPDATE attachment att 
JOIN priv_user us ON att.user_id = us.id 
SET att.contact_id = us.contactid 
WHERE att.contact_id = 0;
```

PR #729, opened 2025-07-18
N°8570 - :bug: Allow @ as part of url

rudnerbjoern:
## Base information

Some type of URL is not parsed correctly by the url_validation_pattern, like this:

https://calendar.example.com/book/meetingroom42@acme-corp.com/

Paste the given example to a text field in itop and look at the rendered output. The link will stop right before the at-sign.

## Cause (bug)

The DEFAULT_VALIDATION_PATTERN does not allow the @ sign as part of the URL

Add the @ sign to the regex. 
According to RFC 3986 – Uniform Resource Identifier (URI): Generic Syntax, the @ character is a valid character within path segments of a URI. Specifically, it is part of the allowed pchar set and does not need to be percent-encoded (%40) when used in the path component — as long as it is not part of the userinfo or authority sections.

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -8165,7 +8165,7 @@ class AttributeURL extends AttributeString
 	 * @since 3.0.3 moved from Config to AttributeURL constant
 	 */
 	public const DEFAULT_VALIDATION_PATTERN = /** @lang RegExp */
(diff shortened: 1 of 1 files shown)


Molkobain (approved):

jf-cbd:
Hey,
Thanks for your PR. Functionally it's interesting and we'll accept it.
Just notice that since this parameter is flagged as visible, its value is in the default conf file, and the value in the default conf file won't be updated, even on setup.
So it means your correction, when it will be accepted and available in a next iTop version, will only work for new iTop instances.
For the existing one, you'll have to update the "url_validation_pattern" parameter in your conf file.
We'll update the documentation to explain it.

Hipska (approved):

jf-cbd:
Thanks for your contribution @rudnerbjoern !

piRGoif:
Can a corresponding test be added in AttributeURLTest::CheckFormatProvider ?

jf-cbd:
Good idea (done in https://github.com/Combodo/iTop/commit/65c9145b10b867092af137d396b5c0de23992c25), thanks for suggestion !

PR #733, opened 2025-08-01
N°2364 - API : remove old linkedset persistance

accognet:
internal


rquetiez (changes requested):
Caution: this portion of code is also executed when the given "fellow" is an ormLinkSet (could not figure out which use case it matches) which is not deprecated.

rquetiez (approved):

PR #737, opened 2025-08-20
N°8210 - Remove iApplicationObjectExtension

bdalsass:
## Base information
| Related to a Combodo ticket? | [N°8210 - Remove iApplicationObjectExtension](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8210)

Remove old CRUD listening mechanism and update usage to new event system.


Hipska (approved):

eespie (approved):
Ok for me

Hipska:
Wait, this means next version will be iTop 4.0? (since removing something that has been deprecated in 3.1)

application/cmdbabstract.class.inc.php
```diff
@@ -4666,25 +4630,20 @@ public function IsModified()
...
-		}
+		return $this->bIsMarkedAsModified;
+	}
 
-		return false;
+	/**
+	 * Override the default modification state of the object.
+	 *
+	 * The object is considered as modified, whatever the actual state is.
+	 * This is used when an object is modified indirectly (eg. through a linked set)
```
rquetiez, line 4640: To be used within a listener, when a module extends the state of a persistent object (for instance, attachments stored as a link set that cannot be modelized by an attribute)

datamodels/2.x/itop-attachments/main.itop-attachments.php
```diff
@@ -158,45 +159,27 @@ public function EnumAllowedActions(DBObjectSet $oSet)
 		return array();
 	}
 
-	public function OnIsModified($oObject)
+	public function RegisterEventsAndListeners() : void
```
rquetiez, line 162: I wonder why the OnIsModified hook could be abandonned without regressing on the following use case: open an existing ticket, add an attachnement and click on the update button. The attachement should then be recorded. Do you confirm this behavior has been maintained?

rquetiez (changes requested):

jf-cbd:
> Wait, this means next version will be iTop 4.0? (since removing something that has been deprecated in 3.1)

Hey @Hipska, no, next version will be 3.3 (even if, indeed, iApplicationObjectExtension was deprecated in 3.1)
For more details about migration, you can read the following doc : https://www.itophub.io/wiki/page?id=latest:customization:event#migrating_from_iapplicationobjectextension_to_listeners

PR #739, opened 2025-08-25
Returns object if the current user is allowed to do the $sAction on an $sObjectClass object (with $sObjectId id)

accognet:
As MetaModel::GetObject with portal scopes

### DIFF
--- a/datamodels/2.x/itop-portal-base/portal/src/Helper/SecurityHelper.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/Helper/SecurityHelper.php
@@ -20,6 +20,7 @@
 
 namespace Combodo\iTop\Portal\Helper;
 
+use CoreException;
 use LogChannels;
 use UserRights;
 use IssueLog;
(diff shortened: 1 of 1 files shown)


datamodels/2.x/itop-portal-base/portal/src/Helper/SecurityHelper.php
```diff
@@ -174,6 +175,123 @@ public function IsActionAllowed($sAction, $sObjectClass, $sObjectId = null)
...
 	}
 
+
+	/**
+	 * Returns object if the current user is allowed to do the $sAction on an $sObjectClass object (with $sObjectId id)
+	 * Checks are:
+	 * - Has a scope query for the $sObjectClass /$sObjectId/ $sAction
+	 * - Is allowed by datamodel for $sObjectClass / $sAction
+	 *
+	 * @param string $sAction Must be in UR_ACTION_READ|UR_ACTION_MODIFY|UR_ACTION_CREATE
```
jbostoen, line 185: If this is not meant to be extensible, perhaps this could be turned into an enum?

jf-cbd:
@Molkobain will take a first look at this PR

Molkobain:
@accognet please provide examples of use case and code snippets where you would use this. I guess it's to have a shorter way to do something?

PR #741, opened 2025-08-28
Feature/2535 split file

accognet:
Too many commits on this PR, and it's impossible to squash it without losing the Git history. Keep it for split files.

Split files : 
core/action.class.inc.php
core/asynctask.class.inc.php
core/attributedef.class.inc.php
core/event.class.inc.php
core/valuesetdef.class.inc.php
core/event.class.inc.php


Hipska (changes requested):
Great idea!

Please not that most files have no line ending at the end of the file:
image

lib/autoload.php
```diff
@@ -14,7 +14,10 @@
             echo $err;
         }
     }
-    throw new RuntimeException($err);
+    trigger_error(
+        $err,
+        E_USER_ERROR
+    );
```
Hipska: This seems like a backward change..

Molkobain:
Is there a particular reason why you don't move these files in the `/src` folder using the `PSR-4` naming convention so everything is autoloded properly? (It will require to modify the code a bit to discover classes implementing/extending the base classes)

accognet:
I've done the first step of splitting the files. Next week, we'll discuss their final placement.

lib/autoload.php
accognet: Good point, this change was made automatically by the composer

lib/composer/platform_check.php
```diff
@@ -36,7 +36,8 @@
             echo 'Composer detected issues in your platform:' . PHP_EOL.PHP_EOL . str_replace('You are running '.PHP_VERSION.'.', '', implode(PHP_EOL, $issues)) . PHP_EOL.PHP_EOL;
         }
     }
-    throw new \RuntimeException(
-        'Composer detected issues in your platform: ' . implode(' ', $issues)
+    trigger_error(
+        'Composer detected issues in your platform: ' . implode(' ', $issues),
+        E_USER_ERROR
```
Hipska, line 41: Seems to be added again, I would check your composer version/settings to not let this happen again..

Hipska (changes requested):
Also, please mark this PR as Draft while you are still working on it, so people know they don't need to do reviews yet. When finished, you can mark it as ready to review again.

accognet:
Too many commits on this PR, and it's impossible to squash it without losing the Git history. Keep it for split files.

Hipska:
Yeah, I also prefer not to use squash just for that reason; loosing history 😉

PR #742, opened 2025-08-28
 N°7920 - laminas-mail is an abandoned package, replace it with symfony/mailer

steffunky:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | N°7920

## Symptom

Laminas-mail is an [abandoned packaged](https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2023-12-04-TSC-Minutes.md). There will be no enhancement, no security fix and no support for PHP 8.4.

The goal of this pull-request is to replace Laminas-mail for outgoing email with Symfony Mailer.

Symfony Mailer is the recommended replacement, as we already use some of Symfony components we probably avoid bringing in more dependencies into iTop 


rquetiez (approved):
OK
The migration of error reporting seems "too easy".
This should be tested manually with wrong configurations, to make sure that supportability will not be impacted (error messages should lead admins to the root cause)

PR #746, opened 2025-09-09
N°4789 - Parse datamodel module.xxx.php files instead of interpreting them

odain-cbd:
This PR relies on a fork from nikic/PHP-Parser
https://github.com/nikic/PHP-Parser/pull/1107

The idea is to enhance how module configuration are read. It would have been nice to propose another format as well (ie json). but this is another story...


sources/PhpParser/Evaluation/PhpExpressionEvaluator.php
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	public function ParseAndEvaluateExpression(string $sExpr) : mixed
+	{
+		$sPhpContent = <<<PHP
+<?php
+$sExpr;
+PHP;
+		try{
+			$aNodes = ModuleFileParser::GetInstance()->ParsePhpCode($sPhpContent);
```
rquetiez: I thought this class is consumed by ModuleFileParser... not the other way around!

rquetiez (changes requested):
The roles of the three classes MouleFileParser, ModuleFileReader and PhpParser should be furthermore clarified -at least explained.

PR #754, opened 2025-09-22
N°4058 - Setup failed when added an encrypted field due to default value NULL non SODIUM compatible

odain-cbd:
** Symptom**:
Itop setup fails when adding encrypted fields without any default value. in AfterDbCreation step, the application tries to decrypt field value with an invalid value (null/empty). 

`2021-06-02 17:56:20 | Error   | An exception occurred: nonce size should be SODIUM_CRYPTO_SECRETBOX_NONCEBYTES bytes at line 330 in file /var/www/html/core/simplecrypt.class.inc.php | SetupLog`

Fixing it in production could be really painfull and is seen as quite unfriendly...

### DIFF
--- a/core/simplecrypt.class.inc.php
+++ b/core/simplecrypt.class.inc.php
@@ -130,7 +130,15 @@ class SimpleCrypt
      */
     function Decrypt($key, $string)
     {
-        return $this->oEngine->Decrypt($key,$string);
+		try{
+			return $this->oEngine->Decrypt($key,$string);
+		} catch(\Exception $e){
+			if (strlen($string)==0){
+				IssueLog::Error("Cannot decrypt empty/null value", null, ['msg' => $e->getMessage(), 'stack' => $e->getTraceAsString()]);
+				return $string;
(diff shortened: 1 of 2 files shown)


eespie (approved):

core/simplecrypt.class.inc.php
```diff
@@ -422,4 +430,4 @@ public function Decrypt($key, $encrypted_data)
 		return trim($plaintext);
 	}
 
-}
+}
```
Hipska, line 433: Please don't remove EOL's at end of the file!

Molkobain:
@odain-cbd you set the milestone to 3.2.2 but has it was merged on the `develop` branch, it will be only on 3.3.0+ won't it? The `support/3.2` branch was already created.

Molkobain:
@odain-cbd replied in our internal ticket system. This is indeed not in iTop 3.2.2. It will be in iTop 3.2.3 (has to be cherry picked) and iTop 3.3.0

PR #756, opened 2025-10-02
N°7570 - feat(Typology): Add useful tabs to certain typology classes

Hipska:
## Base information

## Objective

When clicking on a Typology item from a linked object, it is sometimes handy to see which other objects are also linking to the same Typology object.

## Proposed solution

Add tabs to the Typology objects to see which other objects are using this Typology.

- [ ] Add dictionary items when the additional fields are accepted


datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
```diff
@@ -4450,14 +4458,23 @@
           </attributes>
         </reconciliation>
       </properties>
-      <fields/>
+      <fields>
+        <field id="osverions_list" xsi:type="AttributeLinkedSet">
```
jf-cbd:
```suggestion
        <field id="osversions_list" xsi:type="AttributeLinkedSet">
```

datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
```diff
@@ -4450,14 +4458,23 @@
...
+        </field>
+      </fields>
       <methods/>
       <presentation>
         <details>
           <items>
             <item id="name">
               <rank>10</rank>
             </item>
+            <item id="osverions_list">
```
jf-cbd:
```suggestion
            <item id="osversions_list">
```

datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
```diff
@@ -4395,6 +4395,11 @@
           <extkey_attcode>osfamily_id</extkey_attcode>
           <target_attcode>name</target_attcode>
         </field>
+        <field id="ospatches_list" xsi:type="AttributeLinkedSet">
+          <linked_class>OSPatch</linked_class>
+          <ext_key_to_me>osversion_id</ext_key_to_me>
+          <edit_mode>add_remove</edit_mode>
```
jf-cbd: Why "add_remove" and not "none" ? (same question for all the added fields)

jf-cbd (commented):
Thank you for your PR, here are a few notices :)

datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
Hipska: To be able to add or remove some of them?

v-dumas:
Hello Hipska,
Have you tested the use case, of adding or removing ospatches (or any other sub-objects)? 
**Add** means moving them from one OS version to another. When do you need to do this ? It is highly subject to errors. And if you really do, the bulk modify is much better to control your changes.
**Remove** means empty the osversion of an ospatch (or multiple), why would you do that? You should rather replace the osversion if it was wrong by the correct one using the bulk modify of ospatches.

As a conclusion, our policy for Typology is to propose "**add_only**", nothing else, which means allow sub-object creation. 
For the "edit_when" option, the policy is more open, if there are no dependencies, it can be omitted, which means
you can create a sub-object in read or edition of the host.
(_This "edit_mode" "add_only" should have been named "create_only"_)

In some case like "model" and "brand", because of dependencies and class restriction, even this "add_only" mode is nearly unusable. So me might prevent even the creation in the standard model.

Translation are missing.

Hipska:
@jf-cbd @v-dumas You were right. It's indeed not possible to remove linked items using the `add_remove` mode. I was confused and actually meant to use the `in_place` mode.

I tried all new lists and was able to create and remove items this way:

- Brand
  - Model
  - IOSVersion
- OSVersion
  - OSPatch
- OSFamily
  - OSVersion

Translations are indeed missing. Are we sure these new lists would be accepted to be added in the datamodel? Then I will make the translations for them..

datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
```diff
@@ -4540,6 +4567,12 @@
             <item id="physicaldevices_list">
               <rank>30</rank>
             </item>
+            <item id="models_list">
+              <rank>40</rank>
+            </item>
```
Hipska, line 4608: IMHO it makes more sense to have the `models_list` before `physicaldevices_list`. WDYT?

jf-cbd:
> @jf-cbd @v-dumas You were right. It's indeed not possible to remove linked items using the `add_remove` mode. I was confused and actually meant to use the `in_place` mode.
> 
> I tried all new lists and was able to create and remove items this way:
> 
>     * Brand
>       
>       * Model
>       * IOSVersion
> 
>     * OSVersion
>       
>       * OSPatch
> 
>     * OSFamily
>       
>       * OSVersion
> 
> 
> Translations are indeed missing. Are we sure these new lists would be accepted to be added in the datamodel? Then I will make the translations for them..

Yes it has been functionally accepted so it will be integrated in iTop :)

datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
jf-cbd: Yes, that's more logical

Hipska: So agree, to change rank of `physicaldevices_list` to 40?

jf-cbd: Yes

v-dumas:
@Hipska 
When providing new dictionary entries, developer is expected to update the 17 translation files with the new entries untranslated but with 2 tildes ~~ at the end, so translators quickly identified untranslated entries.

Hipska:
Hmm, in the past there was an automated tool from Combodo that did that. Why can't it be done for this here as well? It is the first time I see this to be required..

v-dumas:
Ok, will do it ourself.

jf-cbd:
Hey, AFAIK the other translations files are not needed, so I'll merge this PR.
Thanks @Hipska for your (Hacktoberfest) contribution !

Hipska:
Done, no more conflicts..

jf-cbd:
Perfect, thanks :)

jf-cbd:
FI : The test testEveryEnglishEntryShouldHaveItsFrenchCounterpart failed, so it seems 2 languages are required, French and English.
I'll add it.

PR #757, opened 2025-10-06
 N°8796 - Add PHP code style validation in iTop and extensions

odain-cbd:
Nice tool to automatically parse code and find out code style issues.
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/tree/master

```
cd tools; composer install
tools/vendor/bin/php-cs-fixer check 
```

```
tools/vendor/bin/php-cs-fixer fix
```

In a future we could even reformat whole code base. and then be notified by CI when code style is not respected....


Molkobain:
Great initiative! 🙌

Two things we should discuss as a team:
  * Put it in a dedicated folder under `tools/` (e.g. `tools/php-code-standards-fixer`) so if if we add another one (e.g. for JS), its already structured correctly (to avoid the issue we add with tests).
  * Ensure it follows our code standards as described in the wiki or accept that our new coding standard should be PSR-2 or PSR-12.

Extras to check afterwards (did you already take a look?):
  * Add something to the iTop / modules projects so it runs automatically on pre-commit on the dev environment.

Hipska:
A nice to have would also be to have this as an action on GitHub pull requests, so contributors (internal and external) would immediately see if te changes are conform the defined coding standards.

jf-cbd (approved):

tests/php-code-style/.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,27 @@
...
+	->in(__DIR__.'/core')
+	->in(__DIR__.'/datamodels')
+	->in(__DIR__.'/dictionaries')
+	->in(__DIR__.'/pages')
+	->in(__DIR__.'/portal')
+	->in(__DIR__.'/setup')
+	->in(__DIR__.'/sources')
+	->in(__DIR__.'/synchro')
+	->in(__DIR__.'/tests')
+    ->in(__DIR__ . '/webservices')
```
Molkobain:
```suggestion
$finder = PhpCsFixer\Finder::create()
	->in(__DIR__)
    ->exclude(__DIR__.'/core/oql');
```

Is it possible to do something based on opt-out rather than opt-in? It would be easier to maintain and to read.
Also, doesn't this exclude clause risk to exclude also all "oql" folders rather than the one `/core/oql`?

Molkobain:
I've take a look at the rules described by Symfony. From my POV what you set is already a great start to propose to the team.

Hipska (changes requested):
The formatting changes on the files in `/webservices` (why only there?) seem to have changed indentation from tabs to spaces. Please revert.

https://www.itophub.io/wiki/page?id=3_2_0:customization:coding_standards#white_spaces_and_indentation

tests/php-code-style/.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,30 @@
...
+
+$config = new PhpCsFixer\Config();
+return $config->setRiskyAllowed(true)
+    ->setRules([
+	    '@PSR12'       => true,
+        'no_extra_blank_lines' => true,
+	    'array_syntax' => ['syntax' => 'short'],
+    ])
+    ->setFinder($finder)
+;
```
Hipska, line 25: Please add a newline at the end of files. Seems to be part of PSR12.

tests/php-code-style/.php-cs-fixer.dist.php
```diff
@@ -0,0 +1,30 @@
...
+	->in($APPROOT.'/core')
+	->in($APPROOT.'/datamodels')
+	->in($APPROOT.'/dictionaries')
+	->in($APPROOT.'/pages')
+	->in($APPROOT.'/portal')
+	->in($APPROOT.'/setup')
+	->in($APPROOT.'/sources')
+	->in($APPROOT.'/synchro')
+	->in($APPROOT.'/tests')
+    ->in($APPROOT . '/webservices')
```
Hipska: This one looks off..

```suggestion
    ->in($APPROOT.'/webservices')
```

Hipska (changes requested):

tests/php-code-style/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+```
+cd tests/php-code-style/; composer install; cd -
+tests/php-code-style/vendor/bin/php-cs-fixer check --config tests/php-code-style/.php-cs-fixer.dist.php [PATH]
+```
+
+to respect iTop code standards and re-format (no path provided means whole iTop code base):
+
+```
+tests/php-code-style/vendor/bin/php-cs-fixer fix --config tests/php-code-style/.php-cs-fixer.dist.php [PATH]
```
Hipska, line 15: Maybe just move the `.php-cs-fixer.dist.php` to the root directory and add this in the main README?

When also moving composer, the command would then just become `vendor/bin/php-cs-fixer fix [PATH]`, which is much easier/friendly to use.

Hipska (changes requested):

odain-cbd:
> The formatting changes on the files in `/webservices` (why only there?) seem to have changed indentation from tabs to spaces. Please revert.
> 
> https://www.itophub.io/wiki/page?id=3_2_0:customization:coding_standards#white_spaces_and_indentation

we will apply code style to whole code base. for now I did it partially to share future effects with the team/community.

tests/php-code-style/.php-cs-fixer.dist.php
odain-cbd: inception... i should apply code style to code style PHP file... :)

Hipska:
The effect is now apparently to change tabs into spaces? That doesn't seem right?

tests/php-code-style/.php-cs-fixer.dist.php
Hipska: Not fixed yet, please unresolved until it is.

odain-cbd: it should be fine now

odain-cbd:
> The effect is now apparently to change tabs into spaces? That doesn't seem right?

side effect fixed.

tests/php-code-style/README.md
odain-cbd: I agree with you. CLI would be shorter and that's what I started with.

but we prefer to keep such stuff apart. ie phpstan/phpunit/php-cs-fixer are gathered in tests folder to package iTop community easily without them. zip is heavy enough. packaging tools are simpler to maintain.

tests/php-code-style/.php-cs-fixer.dist.php
odain-cbd: just fixed this feedback as well

tests/php-code-style/.php-cs-fixer.dist.php
Hipska: Still not resolved..

tests/php-code-style/README.md
Hipska: Should not impact .zip size when putting in `require-dev` section of composer?

If that's the case, then why not using `tests/composer.json` instead?

PR #759, opened 2025-10-07
N°8921 - feat(AttributeLinkedSet): Allow for export in Spreadsheet HTML format

Hipska:
## Base information

## Objective

Exporting a `LinkedSet` attribute with spreadsheet HTML just returns the list of IDs of the linked objects, which is not very helpful for further processing.

## Proposed solution

* Export the friendly names for the linked objects when localised, IDs otherwise.
* Add option to display the attribute as CSV in templates.


jf-cbd:
Hey @Hipska, thank you for your PR. It seems nice, but why IDs are exported when the object isn't localised ? Shouldn't it be only when there is no friendly name ?
Except this small details, it has been accepted in functional review :)

Hipska:
This was to keep backward compatibility. If somebody relied on having the IDs, then they should still be there when using no localisation.

Hipska:
@jf-cbd when will the technical review occur?

jf-cbd:
Hello, thank you for your PR ! We'll review your code in sprint 3.4, since this bug is related to data and this sprint will be about data management, that should start at the end of summer (from September).

Hipska:
Hi Jimmy, can you link this PR to the correct milestone, so it won't be missed?

PR #761, opened 2025-10-10
N°8528 - ignore silo not applied on 1:n linkedset displayed in form in portal

accognet:
internal

### DIFF
--- a/core/ormlinkset.class.inc.php
+++ b/core/ormlinkset.class.inc.php
@@ -232,6 +232,7 @@ class ormLinkSet implements iDBObjectSetIterator, Iterator, SeekableIterator
 	{
 		$aRet = array();
 		$this->oOriginalSet->Rewind();
+		$this->oOriginalSet->OptimizeColumnLoad([$this->sClass => []]);
 		$iRow = 0;
 		while ($oObject = $this->oOriginalSet->Fetch())
 		{
@@ -321,8 +322,6 @@ class ormLinkSet implements iDBObjectSetIterator, Iterator, SeekableIterator
 	 */
 	public function Fetch()
 	{
-		$this->LoadOriginalIds();
-
 		$ret = $this->current();
 		if ($ret === false)
 		{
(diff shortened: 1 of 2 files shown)


core/ormlinkset.class.inc.php
```diff
@@ -353,7 +352,7 @@ public function current()
 		if ($this->iCursor < $iPreservedCount)
 		{
 			$sId = key($this->aPreserved);
-			$oRet = MetaModel::GetObject($this->sClass, $sId);
+			$oRet = MetaModel::GetObject($this->sClass, $sId, true, true);
```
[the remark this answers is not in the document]
bdalsass: Sounds good 👌

core/ormlinkset.class.inc.php
```diff
@@ -232,6 +232,7 @@ protected function GetArrayOfIndex()
 	{
 		$aRet = array();
 		$this->oOriginalSet->Rewind();
+		$this->oOriginalSet->OptimizeColumnLoad([$this->sClass => []]);
```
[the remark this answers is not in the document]
bdalsass: Sounds good 👌

sources/Core/AttributeDefinition/AttributeLinkedSet.php
bdalsass, line 167: I wonder if there is a way to move this handling inside the portal sources for better isolation 🤔

bdalsass (commented):
Works well but needs another opinion.

Also needs to be rebased on support/3.2 branch.
Be carefull with AttributeDefinition file splitting.

core/ormlinkset.class.inc.php
```diff
@@ -215,34 +230,40 @@ protected function LoadOriginalIds()
 	 */
 	protected function GetArrayOfIndex()
 	{
-		$aRet = [];
+		$aRet = array();
 		$this->oOriginalSet->Rewind();
+		$this->oOriginalSet->OptimizeColumnLoad([$this->sClass => []]);
```
rquetiez, line 235: What is the intention here with regard to the goal of this PR?

sources/Core/AttributeDefinition/AttributeLinkedSet.php
rquetiez, line 160: Core implementation MUST NOT be coupled to the portal (here ScopeValidatorHelper, PORTAL_ID)
Could be acceptable solely for maintenance purposes (branch 2.3)

rquetiez (changes requested):
This proposal needs a rework so that the itop core remains uncoupled from external consumers such as the Portal.

core/ormlinkset.class.inc.php
rquetiez, line 324: For what reason this call would be removed ? Fetch can be called in the first place when browsing a set of links.

PR #771, opened 2025-10-30
N°8891 - feat(DBObject): Add CopyIfNull() helper to copy attribute only if target is null

favincen:
## Base information
| Related to a SourceForge thead / Another PR / Combodo ticket? | [Combodo's partners documentation](https://support.combodo.com/pages/exec.php/documentation/dokuwiki/wikibook/15?exec_module=combodo-support-portal&exec_page=index.php&sPage=partner:phpmethods)

##  Objective (enhancement)

DBObject provides helpers like Set(), SetIfNul(), and Copy(), but misses CopyIfNull().
CopyIfNull() is documented and recommended by Combodo in its partners documentation, and is really useful for its purpose. 
Documentation states a PR to add this helper would be welcomed, so here it is.

Add the missing helper method in iTop, as it is documented.

- [ ] mention this new API method in documentation 
- [ ] update partners documentation to explain how to use instead of how to add, and remove recommandation to make a PR

### DIFF
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -762,12 +762,7 @@ abstract class DBObject implements iDisplay
 	 */
 	public function CopyIfNull($sDestAttCode, $sSourceAttCode)
 	{
-		$oAttDef = MetaModel::GetAttributeDef(get_class($this), $sDestAttCode);
-		$oCurrentValue = $this->Get($sDestAttCode);
-		if ($oAttDef->IsNull($oCurrentValue))
-		{
-			return $this->Copy($sDestAttCode, $sSourceAttCode);
-		}
+		$this->SetIfNull($sDestAttCode, $this->Get($sSourceAttCode));
 		return true;
 	}
 
--- a/core/dbobject.class.php
+++ b/core/dbobject.class.php
@@ -745,7 +745,7 @@ abstract class DBObject implements iDisplay
 	/**
 	 * Helper to copy a value only if the target is currently undefined
 	 *
-	 * Call Copy() only of the internal representation of the attribute is null.
+	 * Call Copy() only of the internal representation of the target attribute is null.
 	 *
 	 * @api
 	 * @see Copy()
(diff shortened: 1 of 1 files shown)


jf-cbd:
Hey @favincen, thank you for your PR. It has been functional accepted - which is normal given the content of the documentation 😁 We'll review it from a technical point of view

core/dbobject.class.php
```diff
@@ -742,6 +742,35 @@ public function SetIfNull($sAttCode, $value)
...
+	 * @throws \Exception
+	 */
+	public function CopyIfNull($sDestAttCode, $sSourceAttCode)
+	{
+		$oAttDef = MetaModel::GetAttributeDef(get_class($this), $sDestAttCode);
+		$oCurrentValue = $this->Get($sDestAttCode);
+		if ($oAttDef->IsNull($oCurrentValue))
+		{
+			return $this->Copy($sDestAttCode, $sSourceAttCode);
+		}
```
Hipska: What about simply using the following code to reduce code duplication?

```suggestion
		$this->SetIfNull($sDestAttCode, $this->Get($sSourceAttCode));
```

favincen: Indeed this seems a smart refactoring. Thanks @Hipska !

core/dbobject.class.php
```diff
@@ -762,12 +762,7 @@ public function SetIfNull($sAttCode, $value)
...
 	public function CopyIfNull($sDestAttCode, $sSourceAttCode)
 	{
-		$oAttDef = MetaModel::GetAttributeDef(get_class($this), $sDestAttCode);
-		$oCurrentValue = $this->Get($sDestAttCode);
-		if ($oAttDef->IsNull($oCurrentValue))
-		{
-			return $this->Copy($sDestAttCode, $sSourceAttCode);
-		}
+		$this->SetIfNull($sDestAttCode, $this->Get($sSourceAttCode));
 		return true;
```
favincen, line 766: Note to Combodo: Set() and Copy() do return true unconditionaly upon completion, but SetIfNul() do not return any value. Seems unconsistant to me.
So how do this new CopyIfNull() method should behave concerning returning value ?

favincen: And should SetIfNull() be updated to return a boolean upon completion ?

Hipska (approved):
The doc needs to be updated a bit, but over-all its ok

PR #783, opened 2025-12-04
N°6327 - Enum, Date, FinalClass in Complementary Name not labelized

accognet:
internal


dictionaries/nl.dictionary.itop.core.php
```diff
@@ -608,7 +608,7 @@
 Dict::Add('NL NL', 'Dutch', 'Nederlands', [
 	'Class:Trigger' => 'Trigger',
 	'Class:Trigger+' => 'Aanleiding tot het uitvoeren van een actie',
-	'Class:Trigger/ComplementaryName' => '%1$s, %2$s',
+	'Class:Trigger/ComplementaryName' => '%1$s, class restriction: %2$s~~',
```
Hipska:
```suggestion
	'Class:Trigger/ComplementaryName' => '%1$s, toegepast op klasse: %2$s',
```

application/ui.extkeywidget.class.inc.php
```diff
@@ -284,7 +284,8 @@ public function DisplaySelect(WebPage $oPage, $iMaxComboLength, $bAllowTargetCre
 				if ($bAddingValue) {
 					$aArguments = [];
 					foreach ($aAdditionalField as $sAdditionalField) {
-						array_push($aArguments, $oObj->Get($sAdditionalField));
+						//getAsCSV to have user friendly value in text format
+						array_push($aArguments, $oObj->GetAsCSV($sAdditionalField, ' ', ''));
```
dflaven: Why use GetAsCSV here ?? It as nothing to do with a CSV. Do we really want double quotes around the value if there are spaces in it ??

accognet: GetAsCSV returns the value in the correct format, which is why I'm using it here even though it's not a CSV.

dictionaries/nl.dictionary.itop.core.php
Hipska:
```suggestion
	'Class:Trigger/ComplementaryName' => '%1$s, toegepast op klasse: %2$s~~',
```

Hipska (changes requested):

PR #786, opened 2025-12-09
N°4032 - On UserRequest, service subcategory no more limited by request_type

v-dumas:
## Base information
| Type of change?                                               | Bug fix / Enhancement / Translations

1. On iTop x.y.z 
2. With PHP x.y.z 
2. First go there
2. Then do that
3. ...
4. Finally, see that...

- [ ] I have performed a self-review of my code
- [ ] Is the PR clear and detailed enough so anyone can understand digging in the code?

- [ ] ...
- [ ] ...
- [ ] ...

### DIFF
--- a/datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
+++ b/datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
@@ -309,7 +309,6 @@
           <filter><![CDATA[SELECT ServiceSubcategory WHERE service_id = :this->service_id AND status != 'obsolete']]></filter>
           <dependencies>
             <attribute id="service_id"/>
-            <attribute id="request_type"/>
           </dependencies>
           <sql>servicesubcategory_id</sql>
           <target_class>ServiceSubcategory</target_class>
--- a/datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
+++ b/datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
@@ -307,9 +307,7 @@
         </field>
         <field id="servicesubcategory_id" xsi:type="AttributeExternalKey">
           <filter><![CDATA[SELECT ServiceSubcategory WHERE service_id = :this->service_id AND status != 'obsolete']]></filter>
-          <dependencies>
-            <attribute id="service_id"/>
-          </dependencies>
+          <dependencies/>
           <sql>servicesubcategory_id</sql>
           <target_class>ServiceSubcategory</target_class>
           <is_null_allowed>true</is_null_allowed>
--- a/datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
+++ b/datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
@@ -306,7 +306,7 @@
           <is_null_allowed>true</is_null_allowed>
         </field>
         <field id="servicesubcategory_id" xsi:type="AttributeExternalKey">
-          <filter><![CDATA[SELECT ServiceSubcategory WHERE service_id = :this->service_id AND (ISNULL(:this->request_type) OR request_type = :this->request_type) AND status != 'obsolete']]></filter>
+          <filter><![CDATA[SELECT ServiceSubcategory WHERE service_id = :this->service_id AND status != 'obsolete']]></filter>
           <dependencies>
             <attribute id="service_id"/>
             <attribute id="request_type"/>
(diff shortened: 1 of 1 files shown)


datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
```diff
@@ -306,7 +306,7 @@
           <is_null_allowed>true</is_null_allowed>
         </field>
         <field id="servicesubcategory_id" xsi:type="AttributeExternalKey">
-          <filter><![CDATA[SELECT ServiceSubcategory WHERE service_id = :this->service_id AND (ISNULL(:this->request_type) OR request_type = :this->request_type) AND status != 'obsolete']]></filter>
+          <filter><![CDATA[SELECT ServiceSubcategory WHERE service_id = :this->service_id AND status != 'obsolete']]></filter>
           <dependencies>
             <attribute id="service_id"/>
             <attribute id="request_type"/>
```
Hipska: I would also remove the dependency on `request_type`.

v-dumas: Thanks Hipska,
I have overlooked that point.
I will fix it

Hipska (approved):
Shouldn't the `request_type` field now be dependant from `subcategory_id` and made read-only?

v-dumas:
> Shouldn't the `request_type` field now be dependant from `subcategory_id` and made read-only?

For the dependency, it would make sense if it was recomputed it in the DB_EVENT_COMPUTE_VALUES, but as it is done on a  DB_EVENT_BEFORE_WRITE, the dependency would have no effect. At the end the result, is the same, so it's good enough, I will tackle other topics.
For the read-only, I would have transformed it in ExternalField. The reason why it is not read-only is that out of the box, in the console, Service subcategory isn't mandatory and user may set the request_type without setting a Service. At least that's what I was told by those who have the most experience in iTop Datamodel Design back in year 2010 ;-)

Hipska:
It would have been nicer indeed if it was implemented as `DB_EVENT_COMPUTE_VALUES`, the user could see the updated value when changing a service/sub.

Anyway, accepted as-is.

datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
```diff
@@ -306,10 +306,9 @@
           <is_null_allowed>true</is_null_allowed>
         </field>
         <field id="servicesubcategory_id" xsi:type="AttributeExternalKey">
-          <filter><![CDATA[SELECT ServiceSubcategory WHERE service_id = :this->service_id AND (ISNULL(:this->request_type) OR request_type = :this->request_type) AND status != 'obsolete']]></filter>
+          <filter><![CDATA[SELECT ServiceSubcategory WHERE service_id = :this->service_id AND status != 'obsolete']]></filter>
           <dependencies>
             <attribute id="service_id"/>
```
rquetiez: Why keep this second explicit dependence (it will be computed anyhow)?

datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
```diff
@@ -1334,19 +1333,21 @@
 		// Compute the priority of the ticket
 		$this->Set('priority', $this->ComputePriority());
```
rquetiez, line 1332: As soon as this changes, I would like it to be fully aligned with the recommendation => event based.
Two handlers could be called : EvtComputePriority and EvtComputeRequestType.

datamodels/2.x/itop-request-mgmt/datamodel.itop-request-mgmt.xml
Hipska: To not be able to select `ServiceSubcategory` from another `Service`. What do you mean with "it will be computed"?

v-dumas: Hi Hipska, since about 2 or 3 years, the field dependencies are useless when due to a filter. iTop automatically add the dependency on field specified in the filter. But dependency remains useful when due to a PHP computation.

Hipska: So you're saying that in the console selecting a Service, and not having this dependency at subcategory, the refresh of that list will happen anyways because `:this->service_id` is used in the filter?

v-dumas: Yes, console and portal

Hipska: Thanks, never heard of that feature/addition.

PR #787, opened 2025-12-15
N°8515 - Add CMDB datamodel for Docker and Kubernetes

v-dumas:
## Base information
| Related to Combodo ticket? |  [N°8515](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8515)

## Objective (enhancement)

Document containerization in iTop CMDB as an installation option 
It's mainly XML datamodel, with very little coded behavior 

- [ ] Set this module as an option below Virtualization
- [ ] Get German/Spanish translation...
- [ ] ...
- [ ] ...


datamodels/2.x/itop-container-mgmt/dictionaries/fr.dict.itop-container-mgmt.php
```diff
@@ -0,0 +1,105 @@
...
+
+    // Class Container Virtual Host
+    'Class:ContainerVirtualHost/Name' => '%1$s',
+    'Class:ContainerVirtualHost/ComplementaryName' => '',
+    'Class:ContainerVirtualHost' => 'Plateforme de conteneurisation',
+    'Class:ContainerVirtualHost+' => 'Plateforme sur laquelle des applications s\'exécutent dans des conteneurs',
+    'Class:ContainerVirtualHost/Attribute:containertype_id' => 'Type de plateforme',
+    'Class:ContainerVirtualHost/Attribute:containertype_id+' => 'Technologie de conteneurisation utilisée',
+    'Class:ContainerVirtualHost/Attribute:status' => 'Etat',
+    'Class:ContainerVirtualHost/Attribute:status+' => 'Etat de la plateforme de conteneurisation',
```
steffunky:
```suggestion
    'Class:ContainerVirtualHost/Attribute:status' => 'État',
    'Class:ContainerVirtualHost/Attribute:status+' => 'État de la plateforme de conteneurisation',
```

datamodels/2.x/itop-container-mgmt/dictionaries/fr.dict.itop-container-mgmt.php
```diff
@@ -0,0 +1,105 @@
...
+    'ContainerImage:baseinfo' => 'Informations générales',
+    'ContainerImage:moreinfo' => 'Spécificités de la conteneurisation',
+
+    // Class Container Application
+    'Class:ContainerApplication/Name' => '%1$s',
+    'Class:ContainerApplication/ComplementaryName' => '%1$s',
+    'Class:ContainerApplication' => 'Application conteneurisée',
+    'Class:ContainerApplication+' => 'Application déployée sur une plateforme de conteneurisation',
+    'Class:ContainerApplication/Attribute:status' => 'Etat',
+    'Class:ContainerApplication/Attribute:status+' => 'Etat de l\'application',
```
steffunky:
```suggestion
    'Class:ContainerApplication/Attribute:status' => 'État',
    'Class:ContainerApplication/Attribute:status+' => 'État de l\'application',
```

Hipska:
To me the link from `ContainerImage` towards `ContainerApplication` looks a bit weird. It feels it should be the other way around? Namely a container has 1 specific image and multiple containers can run the same image.

v-dumas:
> To me the link from `ContainerImage` towards `ContainerApplication` looks a bit weird. It feels it should be the other way around? Namely a container has 1 specific image and multiple containers can run the same image.

The name of 'ContainerApplication' is the most subject to discussion, at a point of time, I was about to name it Deployment.
It contains all the rules on how the various `ContainerImages` must be deployed on the container platform. It is truly a grouping of different Images, to combine them in an Application

Hipska:
Wait, so what is the meaning of `ContainerImage` then? The image or the container that is running?

About the `ContainerApplication`, what kind of rules do you think of? It's not very clear to me what this object actually represents.

And does one of the objects represent that a container is actually running on one of the hosts/cluster? (I thought that would be what `ContainerApplication` meant)

v-dumas:
We don't model in iTop what is actually running on any of the host.
We just document what is suppose to be running, but not how many containers of the same image are currently running on each host.
The ContainerImage is just an Image of a container, it's the reference of a piece of code.
That piece of code is expected to be running in one or multiple containers on the hosts/cluster, but again we don't say more than this piece of code is a piece of that ContainerApplication/Deployment which is running on that ContainerHost/Cluster

Hipska:
Okay, I think I get it. But then again, why does the `ContainerImage` point to a `ContainerApplication`/Deployment and not vice versa?

An image could be used in/for multiple deployments. Where a deployment would indeed also have multiple images. Then, maybe it should be a N:N relation?

PS: `ContainerDeployment` would actually sound good 👍
PS2: Don't forget to add `hash` and (multiple) `tags` to `ContainerImage`. (What about a `ContainerRegistry`?)

datamodels/2.x/itop-container-mgmt/data/en_us.data.itop-container-type.xml
```diff
@@ -0,0 +1,24 @@
...
+    <ContainerType alias="ContainerType" id="4">
+        <name>Docker Swarm</name>
+    </ContainerType>
+    <ContainerType alias="ContainerType" id="5">
+        <name>Google Kubernetes Engine</name>
+    </ContainerType>
+    <ContainerType alias="ContainerType" id="6">
+        <name>Kubernetes</name>
+    </ContainerType>
+    <ContainerType alias="ContainerType" id="7">
```
Hipska, line 21: Maybe also add "OpenShift" as a type?

datamodels/2.x/itop-container-mgmt/module.itop-container-mgmt.php
```diff
@@ -0,0 +1,83 @@
...
+	array(
+		// Identification
+		//
+		'label' => 'Container management',
+		'category' => 'business',
+
+		// Setup
+		//
+		'dependencies' => array(
+			'itop-virtualization-mgmt/3.2.0'
```
Hipska, line 18: A hard dependency on virtualisation?

v-dumas:
> Maybe it should be a N:N relation?
In a sense, it is already an N:N relation between ContainerApplication and Software

> PS: `ContainerDeployment` would actually sound good 👍 
Would it apply / make sense in all type of container platform?

> PS2: Don't forget to add `hash` and (multiple) `tags` to `ContainerImage`. (What about a `ContainerRegistry`?)
You lost me completely, for me this object is just a reference to a software version used inside a containerized application. 
What those 'hash' and 'tags' fields are used for? Is 'hash' related to the way the Image is encoded? And 'tags'

Any information related to the way a particular container image should be deployed on the Hosts/Cluster is not something I want to have in the model, because it is way to dependent on the technology. Maybe a file attribute 'Deployment recipe' on the ContainerApplication?

Hipska:
> You lost me completely, for me this object is just a reference to a software version used inside a containerized application.

So in this regards, a `ContainerImage` does not actually mean an image? 

> What those 'hash' and 'tags' fields are used for? Is 'hash' related to the way the Image is encoded? And 'tags'

See for example the [tags for the iTop image](https://hub.docker.com/r/supervisions/itop/tags) we created at Super-Visions. What I named `hash`, DockerHub names `digest`, but it is actually a hash of the built image+version. Similar as what you have in git for a commit reference. So each build/hash/digest can have multiple tags (`latest`, `3.2.2`, `3.2`, ...). I was under the impression `ContainerImage` was a reference to this kind of images, but apparently I was wrong.

> Any information related to the way a particular container image should be deployed on the Hosts/Cluster is not something I want to have in the model, because it is way to dependent on the technology. Maybe a file attribute 'Deployment recipe' on the ContainerApplication?

I Agree with that, only I don't know where you're referring to or to what you are replying to here?

jbostoen:
Just a general thought: How is the decision taken what should be in iTop core vs. extensions? 
I mean, I think the user base for the Azure datamodel is even greater than for Docker / containerization? 

Besides that, happy to see this available!

v-dumas:
> See for example the [tags for the iTop image](https://hub.docker.com/r/supervisions/itop/tags) we created at Super-Visions. What I named `hash`, DockerHub names `digest`, but it is actually a hash of the built image+version. Similar as what you have in git for a commit reference. So each build/hash/digest can have multiple tags (`latest`, `3.2.2`, `3.2`, ...). I was under the impression `ContainerImage` was a reference to this kind of images, but apparently I was wrong.

Well, for me a ContainerImage **is** a reference of this kind of images, there is no 'hash' nor 'digest' string, and maybe we should, but the url is supposed to be the full path to that image, something like "https://hub.docker.com/layers/supervisions/itop/2.7.13/images/sha256-754eb1f7b671a306432f92a14a80ddc4dac556e8809201e848df5e21305f6beb".

Hipska:
Okay, thanks.

So my question I already asked a few times before is still relevant then:
Why does `ContainerImage` reference a `ContainerApplication` with the `containerapplication_id` field? Why can't a second deployment use the same image?

v-dumas:
I suppose that you refer implicitly to [Database normalization](https://en.wikipedia.org/wiki/Database_normalization).
In this case, if we consider that the only information about a ContainerImage is it URL, the denormalization that I propose, will avoid one lnk table, make it easier for most people to update the URL of a ContainerImage just because they have modified the way to build their Application, with the drawback to have multiple ContainerImages pointing to the same URL, so duplicated information. But this duplication does not induce any risk of one being up-to date while the others are not, because the **only** pertinent information, the URL, is the logical unique identifier of a ContainerImage.
It's a perfectly arbitrary choice.

Hipska:
I would urge to reconsider the choice made now that it is still possible. I foresee this to be more a pain in the ass when using this datamodel as a base for future extensions (Docker, Kubernetes, ... specifics) than what it currently solves (easier documenting deployments).

datamodels/2.x/itop-container-mgmt/module.itop-container-mgmt.php
v-dumas: > A hard dependency on virtualisation?

Because of cloud containerization services, we have a new Cloud class, which inherit from VirtualDevice brought by Virtualization

datamodels/2.x/itop-container-mgmt/module.itop-container-mgmt.php
```diff
@@ -0,0 +1,83 @@
...
+			if ($sPreviousVersion == '') { // First installation
+				$oDataLoader = new XMLDataLoader();
+				CMDBObject::SetTrackInfo("Initialization ContainerTypeInstaller");
+				$oMyChange = CMDBObject::GetCurrentChange();
+				$sFileName = dirname(__FILE__)."/data/en_us.data.itop-container-type.xml";
+				SetupLog::Info("Uploading Typology values for Containers from file: $sFileName");
+				$oDataLoader->StartSession($oMyChange);
+				$oDataLoader->LoadFile($sFileName, false, true);
+				$oDataLoader->EndSession();
+			}
```
Hipska: Isn't this where `data.struct` should be used for?

datamodels/2.x/itop-container-mgmt/datamodel.itop-container-mgmt.xml
```diff
@@ -0,0 +1,1395 @@
...
+        <field id="name" xsi:type="AttributeString">
+          <sql>name</sql>
+          <default_value/>
+          <is_null_allowed>false</is_null_allowed>
+        </field>
+        <field id="publisher" xsi:type="AttributeString">
+          <sql>publisher</sql>
+          <default_value/>
+          <is_null_allowed>false</is_null_allowed>
+        </field>
```
Hipska, line 53: This should probably be an `AttributeExternalKey`?

datamodels/2.x/itop-container-mgmt/data/en_us.data.itop-container-type.xml
```diff
@@ -0,0 +1,36 @@
...
+  </ContainerType>
+  <ContainerImageType alias="ContainerImageType" id="1">
+    <name>Docker Hardened Image</name>
+  </ContainerImageType>
+  <ContainerImageType alias="ContainerImageType" id="2">
+    <name>Docker Official Image</name>
+  </ContainerImageType>
+  <ContainerImageType alias="ContainerImageType" id="3">
+    <name>Verified Publisher</name>
+  </ContainerImageType>
```
Hipska: Will this work? I think this needs its own file, if not just for better overview.

Hipska (commented):
This starts to look good 👍 Some comments..

datamodels/2.x/itop-container-mgmt/datamodel.itop-container-mgmt.xml
v-dumas: I have mirrored the Software object design, because a Container Image is a kind of Software, and in general you don't want to mix up your customer and organizations like PHP or Ngnix

datamodels/2.x/itop-container-mgmt/module.itop-container-mgmt.php
v-dumas: There is a difference between what I have done and using data.struct. With data.struct, if you delete an entry it will come again on each iTop upgrade, with my solution it won't.

datamodels/2.x/itop-container-mgmt/data/en_us.data.itop-container-type.xml
v-dumas: Yes, it does work.

datamodels/2.x/itop-container-mgmt/module.itop-container-mgmt.php
Hipska: Hmm, thanks. Previously I had the impression that `data.struct` only gets processed on initial installation of the module..

datamodels/2.x/itop-container-mgmt/data/en_us.data.itop-container-type.xml
Hipska: Then for better overview 😉

datamodels/2.x/itop-container-mgmt/module.itop-container-mgmt.php
v-dumas: In fact I was wrong, and you are right "data.struct only gets processed on initial installation of the module"
So I could use it, instead of having a dedicated piece of code for this

Hipska: Thanks to confirm, I thought I had missed something in the release notes.

It is indeed better to list it in `data.struct` indeed to have better overview..

datamodels/2.x/itop-virtualization-mgmt/dictionaries/en_gb.dict.itop-virtualization-mgmt.php
```diff
@@ -23,5 +23,22 @@
...
+	'Class:Cloud/Attribute:location_id+' => 'Where is located the cloud~~',
+	'Class:Cloud/Attribute:containerhosts_list' => 'Container Hosts~~',
+	'Class:Cloud/Attribute:containerhosts_list+' => 'List of container hosts hosted in this cloud~~',
 ]);
+
+//
+// Class: LogicalInterface
+//
+
+Dict::Add('EN US', 'English', 'English', [
```
Hipska:
```suggestion
Dict::Add('EN GB', 'British English', 'British English', [
```

datamodels/2.x/itop-virtualization-mgmt/dictionaries/en_gb.dict.itop-virtualization-mgmt.php
```diff
@@ -23,5 +23,22 @@
...
+	'Class:LogicalInterface/Attribute:org_id' => 'Organization',
+	'Class:LogicalInterface/Attribute:org_id+' => '',
+]);
```
Hipska: Please add a newline at the end of each file.

datamodels/2.x/itop-virtualization-mgmt/dictionaries/en.dict.itop-virtualization-mgmt.php
```diff
@@ -22,15 +22,28 @@
  * along with iTop. If not, see <http://www.gnu.org/licenses/>
  */
 
+//
+// Class Cloud
```
Hipska, line 26:
```suggestion
// Class: Cloud
```

datamodels/2.x/itop-virtualization-mgmt/dictionaries/fr.dict.itop-virtualization-mgmt.php
```diff
@@ -21,8 +21,22 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with iTop. If not, see <http://www.gnu.org/licenses/>
  */
+
+//
+// Class Cloud
```
Hipska, line 26:
```suggestion
// Class: Cloud
```

datamodels/2.x/itop-container-mgmt/dictionaries/en.dict.itop-container-mgmt.php
```diff
@@ -0,0 +1,160 @@
...
+//
+
+Dict::Add('EN US', 'English', 'English', [
+	'Container:baseinfo' => 'General',
+	'Container:moreinfo' => 'Container specifics',
+	'Container:otherinfo' => 'Dates and description',
+]);
+
+//
+// Class Container Image
```
Hipska, line 20: Consistent description, same for all other occurrences.
```suggestion
// Class: ContainerImage
```

datamodels/2.x/itop-container-mgmt/dictionaries/en.dict.itop-container-mgmt.php
```diff
@@ -0,0 +1,160 @@
+<?php
+/**
+ * Localized data
+ *
+ * @copyright   Copyright (C) 2013 XXXXX
```
Hipska, line 5: Copyright starts from 2026 when merged, or eventually 2025 could be used.

datamodels/2.x/itop-virtualization-mgmt/dictionaries/nl.dict.itop-virtualization-mgmt.php
```diff
@@ -14,14 +14,22 @@
...
+	'Class:Cloud/Attribute:containerhosts_list+' => 'List of container hosts hosted in this cloud~~',
 ]);
 
 //
 // Class: LogicalInterface
 //
 
 Dict::Add('NL NL', 'Dutch', 'Nederlands', [
-	'Class:LogicalInterface/Attribute:org_id' => 'Org id~~',
+	'Class:LogicalInterface/Attribute:org_id' => 'Organization~~',
```
Hipska:
```suggestion
	'Class:LogicalInterface/Attribute:org_id' => 'Organisatie',
```

datamodels/2.x/itop-container-mgmt/dictionaries/en.dict.itop-container-mgmt.php
v-dumas, line 20:
```suggestion
// Class ContainerImage
```

datamodels/2.x/itop-virtualization-mgmt/dictionaries/en_gb.dict.itop-virtualization-mgmt.php
```diff
@@ -23,5 +23,22 @@
...
 ]);
+
+//
+// Class: LogicalInterface
+//
+
+Dict::Add('EN GB', 'British English', 'British English', [
+	'Class:LogicalInterface/Attribute:org_id' => 'Organization',
+	'Class:LogicalInterface/Attribute:org_id+' => '',
+]);
```
v-dumas:
```suggestion
]);

```

datamodels/2.x/itop-container-mgmt/module.itop-container-mgmt.php
```diff
@@ -0,0 +1,54 @@
...
+
+		// Default settings
+		//
+		'settings' => array(
+			// Module specific settings go here, if any
+		),
+	)
+);
+
+?>
```
v-dumas:
```suggestion

```

v-dumas (commented):
comment

datamodels/2.x/itop-container-mgmt/dictionaries/en.dict.itop-container-mgmt.php
v-dumas: Copyright will be handle globally

datamodels/2.x/itop-container-mgmt/dictionaries/en.dict.itop-container-mgmt.php
v-dumas: Why do you want this ?

datamodels/2.x/itop-virtualization-mgmt/dictionaries/en_gb.dict.itop-virtualization-mgmt.php
v-dumas: I did it but what is the purpose ?

datamodels/2.x/itop-container-mgmt/dictionaries/en.dict.itop-container-mgmt.php
Hipska: Consistency with all other occurrences. I thought this was the guideline.

datamodels/2.x/itop-virtualization-mgmt/dictionaries/en_gb.dict.itop-virtualization-mgmt.php
Hipska: You added 2 extra newlines 😛 that has indeed no use.

A single eol is best practice and doesn’t mess things up when handling files in commandline or programmatically.

PR #788, opened 2025-12-17
Add few APIs to ease tests

odain-cbd:
Add few usefull APIs to ease iTop testing among:
- cleanuping in tearDown
- calling iTop from HTTP easily
- logging via IssueLog when starting a test
- backuping/restoring conf in setup/teardown

### DIFF
--- a/tests/php-unit-tests/src/BaseTestCase/ItopTestCase.php
+++ b/tests/php-unit-tests/src/BaseTestCase/ItopTestCase.php
@@ -14,6 +14,7 @@ use ReflectionMethod;
 use SetupUtils;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
 use Symfony\Component\HttpKernel\KernelInterface;
+
 use const DEBUG_BACKTRACE_IGNORE_ARGS;
 
 /**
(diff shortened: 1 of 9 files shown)


tests/php-unit-tests/src/BaseTestCase/ItopTestCase.php
```diff
@@ -631,4 +641,52 @@ protected static function ReadTail($sFilename, $iLines = 1)
...
 	}
+
+	/**
+	* @param $sUrl
+	* @param array|null $aPostFields
+	* @param array|null $aCurlOptions
+	* @param $bXDebugEnabled
+	* @return string
+	 */
+	protected function CallItopUrl($sUrl, ?array $aPostFields = [], ?array $aCurlOptions = [], $bXDebugEnabled = false): string
```
Molkobain, line 652: @odain-cbd This can call any URL, not just iTop's? In that case we might want to rename it don't you think? Currently the difference between the 2 methods isn't obvious to me.

odain-cbd: right. i will rename it.

PR #800, opened 2026-02-03
N°9010 fix flags when extension missing

Timmy38:


setup/wizardsteps.class.inc.php
```diff
@@ -1937,14 +1937,15 @@ protected function GetStepInfo($idx = null)
 	public function ComputeChoiceFlags(array $aChoice, string $sChoiceId, array $aSelectedComponents, bool $bAllDisabled, bool $bDisableUninstallCheck, bool $bUpgradeMode)
 	{
 		$oITopExtension = $this->oExtensionsMap->GetFromExtensionCode($aChoice['extension_code']);
+		//If the extension is missing from disk, it won't exist in the ExtensionsMap, thus returning null
```
odain-cbd: whole method has become uggly and hard to understand functionally... even if most cases are covered by tests (which is nice).

please refactor it to have sth we can understand. for ex. use [if else] structure to spit things apart. on one side treat the exceptional case (extension missing on disk). then others...

odain-cbd: refactor should not be so long as you have the cover :)

odain-cbd (changes requested):

setup/wizardsteps.class.inc.php
```diff
@@ -1937,14 +1937,15 @@ protected function GetStepInfo($idx = null)
...
 		$bCanBeUninstalled = isset($aChoice['uninstallable']) ? $aChoice['uninstallable'] === true || $aChoice['uninstallable'] === 'yes' : $oITopExtension->CanBeUninstalled();
 		$bSelected = isset($aSelectedComponents[$sChoiceId]) && ($aSelectedComponents[$sChoiceId] == $sChoiceId);
-		$bMandatory = (isset($aChoice['mandatory']) && $aChoice['mandatory']) || $bUpgradeMode && $oITopExtension->bInstalled && !$bCanBeUninstalled && !$bDisableUninstallCheck;
-
 		$bMissingFromDisk = isset($aChoice['missing']) && $aChoice['missing'] === true;
+
+		$bMandatory = (isset($aChoice['mandatory']) && $aChoice['mandatory']) || $bUpgradeMode && !$bMissingFromDisk && $oITopExtension->bInstalled && !$bCanBeUninstalled && !$bDisableUninstallCheck;
```
odain-cbd: variable $bMandatory has become sth much more complex than his name. maybe keep its logic simple (mandatory is mandatory). and the addional stuff can come after to treat each flag.

odain-cbd (changes requested):

odain-cbd (approved):

PR #804, opened 2026-02-16
N°8851 N°8854 - PHP 8.4: Fix explicit nullable in functions parameters + replace trigger_error() usage with E_USER_ERROR

Lenaick:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | [8851](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8851) and [8854](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8854)

## Objective

To support migration to PHP8.4, we must address the new deprecation rules. 
After the initial analysis, the newly deprecated features affecting iTop are :
* [Implicitly nullable parameter](https://www.php.net/manual/en/migration84.deprecated.php#migration84.deprecated.core.implicitly-nullable-parameter)
* [Using trigger_error() with E_USER_ERROR](https://www.php.net/manual/en/migration84.deprecated.php#migration84.deprecated.core.underscore-class-name)

## Proposed solution

* For "Implicitly nullable parameter" : prefix the type name with a question mark, e.g. `?CMDBObject $oObject = null`
* For "trigger_error() with E_USER_ERROR": change to use `E_USER_WARNING`

### DIFF
--- a/application/cmdbabstract.class.inc.php
+++ b/application/cmdbabstract.class.inc.php
@@ -6037,7 +6037,7 @@ JS
 	 * @return void
 	 * @since 3.1.0
 	 */
-	final public function AddAttributeFlags(string $sAttCode, int $iFlags, string $sTargetState = '', string $sReason = null): void
+	final public function AddAttributeFlags(string $sAttCode, int $iFlags, string $sTargetState = '', ?string $sReason = null): void
 	{
 		if (!isset($this->aAttributesFlags[$sTargetState])) {
 			$this->aAttributesFlags[$sTargetState] = [];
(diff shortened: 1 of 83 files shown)


Molkobain (approved):
Adding the nullable type hint on a class that can be extended is ok, even though its child classes don't. So we don't have to worry about APIs implementations 🙌

See [this 3v4l.org test](https://3v4l.org/sBTat#veol) which tests the following snippet:

```php
class A
{
    public function foo(?string $text = null)
    {
        var_dump($text);
    }
}

class B extends A
{
    public function foo(string $text = null)
    {
        var_dump($text);
    }
}

$a = new A();
$a->foo('Hello');

$b = new B();
$b->foo('Hello');
```

The other way around works as well, see [this 3v4l.org test](https://3v4l.org/0g8Ch#veol).

Molkobain:
Ok for merge 🙌

PR #807, opened 2026-02-18
N°8782 - Union queries allow all data

eespie:
## Base information
https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8782
bug fix

AllowAllData bad behaviour on unions

See bug


Molkobain (changes requested):
You pointed the PR to the `develop` branch when it should be `support/3.2` shouldn't it?
Mind to also check your base branch, you might have to do a "rebase onto" if you need to switch it from `develop` to `support/3.2`, ask @steffunky for a hand if necessary, he knows the procedure :)

rquetiez (approved):

PR #818, opened 2026-02-25
N°6977 - Sanitize Excel formulas in exports

steffunky:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | N°6977
| Type of change?                                               | Bug fix / Enhancement / Translations

Sanitize what could be considered as first character of an Excel formula in exports

- [x] Update portal message
- [x] Test if it works query phrasebook


sources/Application/Helper/ExportHelper.php
```diff
@@ -17,12 +20,88 @@
  */
 class ExportHelper
 {
+	public const EXCEL_FORMULA_CHARACTERS = ['=', '＝', '+', '＋', '-', '－', '@', '＠', '|', ' ', "\t", "\n", "\r"];
```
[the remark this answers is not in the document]
steffunky: I'll remove space character as it is far-fetched, but carriage return and line feed are indeed in OWASP recomandations:
https://owasp.org/www-community/attacks/CSV_Injection

core/csvbulkexport.class.inc.php
```diff
@@ -337,6 +349,12 @@ public function GetNextChunk(&$aStatus)
 							$sField = $oObj->GetAsCSV($sAttCode, $this->aStatusInfo['separator'], $this->aStatusInfo['text_qualifier'], $this->bLocalizeOutput, !$this->aStatusInfo['formatted_text']);
 					}
 				}
+
+				// If the option to ignore Excel sanitization is not set or absent, sanitize the field
+				if (!(array_key_exists('ignore_excel_sanitization', $this->aStatusInfo)) || $this->aStatusInfo['ignore_excel_sanitization'] === false) {
+					$sField = ExportHelper::SanitizeField($sField, $this->aStatusInfo['text_qualifier']);
+				}
+
```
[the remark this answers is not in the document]
steffunky: `GetValue` is not called in the `GetNextChunk` context in CSV export (it is though in the Excel export, that's why the code differs).

The sanitization does not add the `text_qualifier`, it strips them from the start of the value to compare its first character and then add it back.

I see no risk of double sanitization

sources/Application/Helper/ExportHelper.php
```diff
@@ -17,12 +20,88 @@
...
+	{
+		if ($sField === '') {
+			return $sField;
+		}
+
+		$sQualifier = $sTextQualifier;
+		if ($sQualifier !== '' && str_starts_with($sField, $sQualifier)) {
+			$sAfterQualifier = substr($sField, strlen($sQualifier));
+			if (self::IsValueFormulaCandidate($sAfterQualifier)) {
+				return $sQualifier."'".$sAfterQualifier;
```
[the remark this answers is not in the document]
steffunky: I see the point if the `IsValueFormulaCandidate` would check later character, but for now it only checks the start of the string so I see no point except for handling an evolution in this kind of attack in a later fix

tests/php-unit-tests/unitary-tests/core/BulkExportTest.php
```diff
@@ -158,4 +158,173 @@ public function testExportWithShowObsoleteParam(
...
+					"'＠org11",
+					"'|org12",
+					"'%3Dorg13",
+					"'%3dorg14",
+					"org14",
+				],
+			],
+		];
+	}
 }
```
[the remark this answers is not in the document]
steffunky: Excel export returns binary data.
In order to test it we would have to intercept its temporary built file between `GetNextChunk` and `GetFooter`.

That would still test incompletely Excel but allow us to test the sanitization implementation.

Though the main part to test was the `ExportHelper`, which is done through the CSV export, it could be done but would add little value to the test as the implementation is very similar to the CSV export

sources/Application/Helper/ExportHelper.php
```diff
@@ -17,12 +20,88 @@
  */
 class ExportHelper
 {
+	public const EXCEL_FORMULA_CHARACTERS = ['=', '＝', '+', '＋', '-', '－', '@', '＠', '|', "\t", "\n", "\r"];
```
Molkobain, line 26:
```suggestion
	/**
	 * @since 3.2.3 3.3.0 N°6977
	 */
	public const EXCEL_FORMULA_CHARACTERS = ['=', '＝', '+', '＋', '-', '－', '@', '＠', '|', "\t", "\n", "\r"];
```

Molkobain (approved):

Molkobain:
Complicated issue, but it seems to get the job done as expected, provide an opt-out option and is tested. So I might be overlooking something, but I looks ok to me.

core/csvbulkexport.class.inc.php
```diff
@@ -55,6 +55,8 @@ public function ReadParameters()
 		$this->aStatusInfo['charset'] = strtoupper(utils::ReadParam('charset', 'UTF-8', true, 'raw_data'));
 		$this->aStatusInfo['formatted_text'] = (bool)utils::ReadParam('formatted_text', 0, true);
 
+		$this->aStatusInfo['ignore_excel_sanitization'] = (bool)utils::ReadParam('ignore_excel_sanitization', 0, true, utils::ENUM_SANITIZATION_FILTER_INTEGER);
```
Hipska, line 58: Why the double negation?
```suggestion
		$this->aStatusInfo['excel_sanitization'] = (bool)utils::ReadParam('excel_sanitization', 0, false, utils::ENUM_SANITIZATION_FILTER_INTEGER);
```

sources/Application/Helper/ExportHelper.php
```diff
@@ -17,12 +20,91 @@
  */
 class ExportHelper
 {
+	/**
+	 * @since 3.2.3 3.3.0 N°6977
+	 */
+	public const EXCEL_FORMULA_CHARACTERS = ['=', '＝', '+', '＋', '-', '－', '@', '＠', '|', "\t", "\n", "\r"];
```
Hipska, line 26: I think only values starting with the `=` sign are considered to be potential formulas?

core/csvbulkexport.class.inc.php
steffunky: I think it's more secure to opt-out the sanitization explicitly, if someone calls the export externally without calling `ReadParameters` (ie portal or query phrasebook), sanitization should be the default.

But if you want the raw data you could add manually a `ignore_excel_sanitization` in `$this->aStatusInfo` before calling the export (like in the unit test)

sources/Application/Helper/ExportHelper.php
steffunky: I'm not an Excel expert, I followed what recommended in:
https://owasp.org/www-community/attacks/CSV_Injection
https://github.com/payload-box/csv-injection-payload-list/

core/csvbulkexport.class.inc.php
Hipska: ~~But You're setting the default parameter (if not set) to `true`, meaning it will be not sanitised?~~

Oh I may have misread.. It is the `0` that is being set. Nevermind then.

PR #821, opened 2026-02-27
N°9141 Log when extension uninstallation is forced

Timmy38:


Molkobain (approved):

setup/wizardsteps/WizStepInstall.php
```diff
@@ -83,14 +83,33 @@ public function Display(WebPage $oPage)
...
+		$aExtensionsRemoved = json_decode($this->oWizard->GetParameter('removed_extensions'), true) ?? [];
+		$aExtensionsNotUninstallable = json_decode($this->oWizard->GetParameter('extensions_not_uninstallable'));
+		$aExtensionsForceUninstalled = [];
+		foreach ($aExtensionsRemoved as $sExtensionCode => $sLabel) {
+			if (in_array($sExtensionCode, $aExtensionsNotUninstallable)) {
+				$aExtensionsForceUninstalled[] = $sExtensionCode;
+			}
+		}
+		if (count($aExtensionsForceUninstalled)) {
+			SetupLog::Warning("Extensions uninstalled forcefully : ".implode(',', $aExtensionsForceUninstalled));
```
odain-cbd, line 106: it seems this section is not dealing with "uninstallation-forced" flag. it logs as if extensions removal is forced. when it is not, i thought it should failf/stop setup, no?

Timmy38: If a non-uninstallable extension is removed, the setup will stop at the extension selection step, the user won't be able to progress further.
If he reaches this step (install), this checks has already been performed.

odain-cbd: why checking option value then? 'if' is not necessary, right?
		if ($this->oWizard->GetParameter('force-uninstall', false)) {
			SetupLog::Warning("User disabled uninstallation checks");
		}

maybe you can add a php comment. to document with below remark. today it is obvious to you that if we reach this piece of code we passed the check. later on some others may not understand.

Timmy38: It could be useful to know when a user disable uninstallation checks, whether it was actually needed or not (to advise them not to unless needed for example)
Good idea for the php comment, I'm adding it.

odain-cbd (approved):

PR #823, opened 2026-02-27
N°9248 - REST API - Updating a link creates an empty caselog on the linked object when there is no change

accognet:
internal


bdalsass (approved):
This may not explain why REST API call this method without modifications but act like a security.

PR #825, opened 2026-02-27
N°7150 - Backup : on MariaDB >= 11.0.1 call mariadb-dump instead of mysqldump

accognet:
internal

### DIFF
--- a/setup/backup.class.inc.php
+++ b/setup/backup.class.inc.php
@@ -606,8 +606,8 @@ EOF;
 
 	public static function GetDumpFunction(): string
 	{
-		$version = CMDBSource::QueryToScalar('SELECT VERSION()');
-		if (is_string($version) && stripos($version, 'MariaDB') !== false) {
+		$sVersion = CMDBSource::GetDBVersion();
+		if (stripos($sVersion, 'MariaDB') !== false) {
 			return 'mariadb-dump';
 		}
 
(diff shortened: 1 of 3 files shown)


setup/backup.class.inc.php
```diff
@@ -603,6 +603,16 @@ public static function MakeSafeMySQLCommand($sMySQLBinDir, string $sCmd)
 
 		return '"'.$sMySQLCommand.'"';
 	}
+
+	public static function GetDumpFunction(): string
+	{
+		$version = CMDBSource::QueryToScalar('SELECT VERSION()');
+		if (is_string($version) && stripos($version, 'MariaDB') !== false) {
```
Hipska:
```suggestion
		$sVersion = CMDBSource::GetDBVersion();
		if (stripos($sVersion, 'MariaDB') !== false) {
```

Hipska (changes requested):

Hipska (approved):

Hipska:
Best way to handle these kind of stuff is using a full path name instead of only the bin dir.

So `/usr/bin/mariadb-dump` vs `/usr/bin` so that it could be any tool the admin wants without doing workarounds. `graphviz_path` is an example where it was implemented the good way.

(I have seen the same problem on different places like collectors as well)

PR #837, opened 2026-03-10
N°9383 - CheckToWrite inside a transition: Incorrect error message display

accognet:
internal


jbostoen:
If working on this - perhaps also a good time to take another look at https://github.com/Combodo/iTop/pull/666 ?

PR #844, opened 2026-03-16
N°9233 - Check user access before acquiring lock on object

Lenaick:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | [N°9233](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=9233)

## Symptom

A user without write permissions for an object is still able to lock it

## Cause

User permissions are not verified during the locking operation

## Proposed solution

Verify that the user has write permissions before allowing the locking operation

### DIFF
--- a/pages/ajax.render.php
+++ b/pages/ajax.render.php
@@ -2054,6 +2054,9 @@ EOF
 				if (false === $oSet->CountExceeds(0)) {
 					throw new SecurityException(Dict::S('UI:ObjectDoesNotExist'));
 				}
+				if (UserRights::IsActionAllowed($sObjClass, UR_ACTION_MODIFY, $oSet) != UR_ALLOWED_YES) {
+					throw new SecurityException(Dict::S('UI:ActionNotAllowed'));
+				}
 
 				$aResult = iTopOwnershipLock::AcquireLock($sObjClass, $iObjKey);
 				if (false === $aResult['success']) {
--- a/pages/ajax.render.php
+++ b/pages/ajax.render.php
@@ -2051,11 +2051,11 @@ EOF
 				$oSearch = new DBObjectSearch($sObjClass);
 				$oSearch->AddCondition(MetaModel::DBGetKey($sObjClass), $iObjKey, '=');
 				$oSet = new CMDBObjectSet($oSearch);
-				if (false === $oSet->CountExceeds(0)) {
+				if (
+					false === $oSet->CountExceeds(0) ||
+					UserRights::IsActionAllowed($sObjClass, UR_ACTION_MODIFY, $oSet) !== UR_ALLOWED_YES
+				) {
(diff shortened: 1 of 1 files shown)


pages/ajax.render.php
```diff
@@ -2047,6 +2047,17 @@ function(data){
...
 
+				// Check user has access to the object before trying to acquire the lock
+				$oSearch = new DBObjectSearch($sObjClass);
+				$oSearch->AddCondition(MetaModel::DBGetKey($sObjClass), $iObjKey, '=');
+				$oSet = new CMDBObjectSet($oSearch);
+				if (false === $oSet->CountExceeds(0)) {
+					throw new SecurityException(Dict::S('UI:ObjectDoesNotExist'));
+				}
+				if (UserRights::IsActionAllowed($sObjClass, UR_ACTION_MODIFY, $oSet) != UR_ALLOWED_YES) {
+					throw new SecurityException(Dict::S('UI:ActionNotAllowed'));
```
steffunky: Maybe we should use the same error code as above ( `UI:ObjectDoesNotExist` ) in this instance too to avoid object enumeration

(This does not seems to be the default behavior of operation `details` yet)

steffunky (approved):

pages/ajax.render.php
```diff
@@ -2047,6 +2047,17 @@ function(data){
...
 				$iObjKey = (int)utils::ReadParam('obj_key', 0, false, 'integer');
 
+				// Check user has access to the object before trying to acquire the lock
+				$oSearch = new DBObjectSearch($sObjClass);
+				$oSearch->AddCondition(MetaModel::DBGetKey($sObjClass), $iObjKey, '=');
+				$oSet = new CMDBObjectSet($oSearch);
+				if (false === $oSet->CountExceeds(0)) {
+					throw new SecurityException(Dict::S('UI:ObjectDoesNotExist'));
+				}
+				if (UserRights::IsActionAllowed($sObjClass, UR_ACTION_MODIFY, $oSet) != UR_ALLOWED_YES) {
```
Molkobain:
```suggestion
				if (UserRights::IsActionAllowed($sObjClass, UR_ACTION_MODIFY, $oSet) !== UR_ALLOWED_YES) {
```

Molkobain: We tend to use strict comparison as much as possible.
If it's not possible because of mixed variable, we add a comment to explain why. :)

Molkobain (changes requested):

pages/ajax.render.php
Lenaick: Fixed in the lastest commit :)

pages/ajax.render.php
Lenaick: Fixed in the lastest commit :)

PR #845, opened 2026-03-17
N°8597 - Fix special characters being escaped for BrowseBrick items in "Tree" mode

Molkobain:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | N°8597

## Symptom (bug)
In the portal, some service catalog items have their titles displaying double encoded characters when displayed in _mosaic mode_.

1. On iTop 3.2.2
2. With PHP 8.1.24
2. Change the name of a service subcategory to `Né&w' mobile phone ordering">`
2. As a portal user, go to the portal and open the service catalog in "mosaic" mode
4. Finally, see that the name of the service has all "special" characters escaped (éàç&'"...)

## Cause (bug)
Regression from [9238527](https://github.com/Combodo/iTop/commit/92385273ffa527783368a1bb7625e4de12fe0585)
and [d72e861](https://github.com/Combodo/iTop/commit/d72e861dfe4d743373fa8e90bfefe3b77e214b78). Escaping as been done systematically in the backend while data usage is not homogenous in the frontend.

  * Avoid to HTML escape data in the backend, the data should always remain raw until used by the consumers (twig, js, ...)
  * Update BrowseBrick templates to ensure that they escape the data correctly to prevent XSS and to make processing homogenous across the different display modes (list, tree, mosaic)

Mind that this must also be done in the _Portal new look for 3.2_ module, which is about to be done in a dedicated PR.
The new module version will only be comaptible with iTop 3.2.3+ in order to not re-open a vulnerability.

_Mosaic mode_
image

_Tree mode_
image

_List mode_
image

### DIFF
--- a/datamodels/2.x/itop-portal-base/portal/src/Helper/BrowseBrickHelper.php
+++ b/datamodels/2.x/itop-portal-base/portal/src/Helper/BrowseBrickHelper.php
@@ -317,7 +317,7 @@ class BrowseBrickHelper
 			$aRow[$key] = [
 				'level_alias' => $key,
 				'id' => $sCurrentObjectId,
-				'name' => utils::EscapeHtml($value->Get($sNameAttCode)),
+				'name' => $value->Get($sNameAttCode),
 				'class' => $sCurrentObjectClass,
 				'action_rules_token' => $this->PrepareActionRulesForItems($aItems, $key, $aLevelsProperties),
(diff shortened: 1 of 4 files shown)


Molkobain:
Good catch, can you re-review now? @greptileai

steffunky (approved):

PR #848, opened 2026-03-18
N°8532 - Apply filters on all DBSearch classes

steffunky:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | N°8532

## Symptom 
When multiple classes are present in a OQL through JOIN statements, it's expected to apply the scope filters to all classes present in the query.

## Reproduction procedure

Run this query with a user that does not have access to certain `VirtualMachine` because of their organization

```SQL
SELECT OSF FROM OSFamily AS OSF JOIN VirtualMachine AS VM ON VM.osfamily_id = OSF.id JOIN Organization AS O ON VM.org_id = O.id
```

See that the `OSFamily` for these are listed

## Cause
The `DBSearch` method of filtering only applies to the selected class. So it does not happen for the joined classes nor for the unioned classes.

## Proposed solution
Make this filter apply to every class in the query.
This also allows to make it work recursively for `UNION` searches.
It's triggered with an opt-in `security.disable_joined_classes_filter` parameter as some iTop functionalities may relay on this behavior.

### DIFF
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -1730,6 +1730,14 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
+		'security.disable_joined_classes_filter' => [
+			'type'                => 'bool',
+			'description'         => 'If true, scope filters aren\'t applied to joined classes or union classes not directly listed in the SELECT clause.',
+			'default'             => true,
+			'value'               => true,
+			'source_of_value'     => '',
+			'show_in_conf_sample' => false,
+		],
 		'security.hide_administrators' => [
 			'type'                => 'bool',
 			'description'         => 'If true, non-administrator users will not be able to see the administrator accounts, the Administrator profile and the links between the administrator accounts and their profiles.',
--- a/core/dbobjectsearch.class.php
+++ b/core/dbobjectsearch.class.php
@@ -1928,8 +1928,9 @@ class DBObjectSearch extends DBSearch
 
 	/**
 	 * @inheritDoc
+	 * @return DBObjectSearch
 	 */
-	protected function ApplyDataFilters()
+	protected function ApplyDataFilters(): DBObjectSearch
 	{
 		if ($this->IsAllDataAllowed() || $this->IsDataFiltered()) {
 			return $this;
--- a/core/dbobjectsearch.class.php
+++ b/core/dbobjectsearch.class.php
@@ -1925,4 +1925,36 @@ class DBObjectSearch extends DBSearch
 	{
 		return $this->GetCriteria()->ListParameters();
 	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected function ApplyDataFilters()
+	{
+		if ($this->IsAllDataAllowed() || $this->IsDataFiltered()) {
+			return $this;
+		}
+
+		$oSearch = $this;
+		$aClassesToFilter = $this->GetSelectedClasses();
+
+		// Opt-in for joined classes filtering, otherwise only filter the selected class(es)
+		if (MetaModel::GetConfig()->Get('security.disable_joined_classes_filter') === false) {
+			$aClassesToFilter = $this->GetJoinedClasses();
+		}
+
+		// Apply filter (this is similar to the one in DBSearch but the factorization could make it less readable)
+		foreach ($aClassesToFilter as $sClassAlias => $sClass) {
+			$oVisibleObjects = UserRights::GetSelectFilter($sClass, $this->GetModifierProperties('UserRightsGetSelectFilter'));
+			if ($oVisibleObjects === false) {
+				$oVisibleObjects = DBObjectSearch::FromEmptySet($sClass);
+			}
+			if (is_object($oVisibleObjects)) {
+				$oVisibleObjects->AllowAllData();
+				$oSearch = $oSearch->Filter($sClassAlias, $oVisibleObjects);
+				$oSearch->SetDataFiltered();
+			}
+		}
+		return $oSearch;
+	}
 }
--- a/core/dbsearch.class.php
+++ b/core/dbsearch.class.php
@@ -1598,7 +1598,7 @@ abstract class DBSearch
 	/**
 	 * Apply data filters to the search, if needed
 	 *
-	 * @return $this|DBSearch
+	 * @return DBSearch
 	 * @throws CoreException
 	 */
 	protected function ApplyDataFilters(): DBSearch
--- a/core/dbsearch.class.php
+++ b/core/dbsearch.class.php
@@ -1601,7 +1601,7 @@ abstract class DBSearch
 	 * @return $this|DBSearch
 	 * @throws CoreException
 	 */
-	protected function ApplyDataFilters()
+	protected function ApplyDataFilters(): DBSearch
 	{
 		if ($this->IsAllDataAllowed() || $this->IsDataFiltered()) {
 			return $this;
--- a/core/dbsearch.class.php
+++ b/core/dbsearch.class.php
@@ -1122,21 +1122,7 @@ abstract class DBSearch
 	 */
 	protected function GetSQLQuery($aOrderBy, $aArgs, $aAttToLoad, $aExtendedDataSpec, $iLimitCount, $iLimitStart, $bGetCount, $aGroupByExpr = null, $aSelectExpr = null)
 	{
-		$oSearch = $this;
-		if (!$this->IsAllDataAllowed() && !$this->IsDataFiltered()) {
-			foreach ($this->GetSelectedClasses() as $sClassAlias => $sClass) {
-				$oVisibleObjects = UserRights::GetSelectFilter($sClass, $this->GetModifierProperties('UserRightsGetSelectFilter'));
-				if ($oVisibleObjects === false) {
-					// Make sure this is a valid search object, saying NO for all
-					$oVisibleObjects = DBObjectSearch::FromEmptySet($sClass);
-				}
-				if (is_object($oVisibleObjects)) {
-					$oVisibleObjects->AllowAllData();
-					$oSearch = $oSearch->Filter($sClassAlias, $oVisibleObjects);
-					$oSearch->SetDataFiltered();
-				}
-			}
-		}
+		$oSearch = $this->ApplyDataFilters();
 
 		if (is_array($aGroupByExpr)) {
 			foreach ($aGroupByExpr as $sAlias => $oGroupByExp) {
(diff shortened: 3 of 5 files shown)


core/config.class.inc.php
```diff
@@ -1730,19 +1730,19 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
-		'security.hide_administrators' => [
+		'security.disable_joined_classes_filter' => [
```
rquetiez, line 1733: I would prefer a positive formulation such as _security.filter_joined_class_

tests/php-unit-tests/unitary-tests/core/DBSearchFilterJoinTest.php
```diff
@@ -0,0 +1,192 @@
...
+
+class DBSearchFilterJoinTest extends ItopDataTestCase
+{
+	private const RESTRICTED_PROFILE = 'Configuration Manager';
+	private $aData = [];
+	protected function setUp(): void
+	{
+		parent::setUp();
+		$this->RequireOnceItopFile('application/startup.inc.php');
+		$this->BackupConfiguration();
```
rquetiez: This is not necessary as the config file is not altered: when changed in memory, the configuration gets restored by ItopDataTestCase::setUp/tearDown

core/dbunionsearch.class.php
```diff
@@ -673,4 +673,29 @@ public function GetExpectedArguments(): array
 
 		return $aVariableCriteria;
 	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected function ApplyDataFilters()
```
rquetiez: If I'm not wrong, this should not be necessary: the overload of Filter does the job.

core/dbsearch.class.php
```diff
@@ -1608,4 +1594,33 @@ public function ListParameters(): array
...
 	 */
 	abstract public function GetExpectedArguments(): array;
+
+	/**
+	 * Apply data filters to the search, if needed
+	 *
+	 * @return $this|DBSearch
+	 * @throws CoreException
+	 */
+	protected function ApplyDataFilters()
```
rquetiez: Could be final as it relies on specifics implementations of **Filter** to the job depending on the type of query

core/dbobjectsearch.class.php
```diff
@@ -1925,4 +1925,36 @@ public function GetExpectedArguments(): array
 	{
 		return $this->GetCriteria()->ListParameters();
 	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected function ApplyDataFilters()
```
rquetiez: Overload not needed. See other comments below.

core/dbobjectsearch.class.php
```diff
@@ -1925,4 +1925,36 @@ public function GetExpectedArguments(): array
...
+		if ($this->IsAllDataAllowed() || $this->IsDataFiltered()) {
+			return $this;
+		}
+
+		$oSearch = $this;
+		$aClassesToFilter = $this->GetSelectedClasses();
+
+		// Opt-in for joined classes filtering, otherwise only filter the selected class(es)
+		if (MetaModel::GetConfig()->Get('security.disable_joined_classes_filter') === false) {
+			$aClassesToFilter = $this->GetJoinedClasses();
```
rquetiez, line 1944: Adding as many conditions as joined classes could cause some queries to hit the 61 tables limit after enabling the feature. Mitigation solutions to be discussed.

rquetiez (changes requested):

core/config.class.inc.php
Molkobain: @rquetiez that was actually his first approach, but as the parameter won't be visible in the configuration file by default I pointed out that if one is setting it, it's because one wants to to disable it. So a negative formulation would be more intuitive from an administrator POV.

steffunky: We already have `security.disable_inline_documents_sandbox` and `security.enable_header_xcontent_type_options` so we could cement our naming standard for this kind of parameter now

We could take into account that we may change all the default values of these kinds of security parameters to `true` one day to follow CRA, so it may be easier to understand that if you want to disable one of these parameters, you need to specifically add the `security.disable_*` configuration parameter

core/dbobjectsearch.class.php
```diff
@@ -1925,4 +1925,36 @@ public function GetExpectedArguments(): array
 	{
 		return $this->GetCriteria()->ListParameters();
 	}
+
+	/**
+	 * @inheritDoc
+	 */
+	protected function ApplyDataFilters()
+	{
+		if ($this->IsAllDataAllowed() || $this->IsDataFiltered()) {
```
rquetiez, line 1935: I was wrong in my first review. Delegation cannot rely solely on the Filter method. Still, the duplicated and unused code (DBSearch::ApplyDataFilter) is questionnable.
For more clarity, I suggest:

1. The condition "_if ($this->IsAllDataAllowed() || $this->IsDataFiltered())_" is implemented in one single point into the calling method in **DBSearch**, or gets encapsulated in a dedicated method.
2. **ApplyDataFilters** be abstract in **DBSearch**

core/dbunionsearch.class.php
```diff
@@ -673,4 +673,29 @@ public function GetExpectedArguments(): array
...
+	{
+		if ($this->IsAllDataAllowed() || $this->IsDataFiltered()) {
+			return $this;
+		}
+
+		// Opt-in for joined classes filtering, otherwise fallback on DBSearch filtering
+		if (MetaModel::GetConfig()->Get('security.disable_joined_classes_filter') === true) {
```
rquetiez, line 688: I am really confused with this: as soon as data filters must be applied, the specialized implementation called at line 695 will handle that correctly.
Hopefully, this will not be needed the hereabove suggested form.

tests/php-unit-tests/unitary-tests/core/DBSearchFilterJoinTest.php
```diff
@@ -0,0 +1,192 @@
...
+	{
+		parent::tearDown();
+	}
+
+	/**
+	 * @dataProvider JoinedAndNestedOqlProvider
+	 */
+	public function testDBSearchFilterAppliedToJoinsWhenEnabled(string $sOqlTemplate, int $iExpectedCount): void
+	{
+		$this->EnableJoinFilterConfig(true);
```
rquetiez, line 40: In that case, the change of the configuration could be inlined, exposing the real name and value of the setting, without having to understand the meaning of the helper method:
`MetaModel::GetConfig()->Set('security.disable_joined_classes_filter', false)`

tests/php-unit-tests/unitary-tests/core/DBSearchFilterJoinTest.php
```diff
@@ -0,0 +1,192 @@
...
+
+		$sOql = sprintf($sOqlTemplate, $this->aData['denied_org_name'], $this->aData['allowed_org_name']);
```
rquetiez: This is rather cryptic. A comment -or another code structure- would help.

Example of another code structure : develop the test case in clear, instead of factorization. This may give more code, but clarity is more important for test code.
```
$oSearch = DBSearch::FromOQL('SELECT .......');
$this->assertEquals(0, $oSet->Count(), 'when... should be filtered');
```

tests/php-unit-tests/unitary-tests/core/DBSearchFilterJoinTest.php
```diff
@@ -0,0 +1,192 @@
...
+				1,
+			],
+			'nested-in-select' => [
+				"SELECT OSF FROM OSFamily AS OSF WHERE OSF.id IN (SELECT OSF1 FROM OSFamily AS OSF1 JOIN VirtualMachine AS VM ON VM.osfamily_id = OSF1.id JOIN Organization AS O ON VM.org_id = O.id WHERE O.name = '%s')",
+				0,
+				1,
+			],
+			'userrequest-join-person-org' => [
+				"SELECT OSF FROM OSFamily AS OSF JOIN VirtualMachine AS VM ON VM.osfamily_id = OSF.id JOIN lnkFunctionalCIToTicket AS L ON L.functionalci_id = VM.id JOIN UserRequest AS UR ON L.ticket_id = UR.id  JOIN Person AS P ON UR.caller_id = P.id JOIN Organization AS O ON P.org_id = O.id WHERE O.name = '%s'",
+				0,
```
rquetiez: For the sake of clarity, is is a good practice to tag the data with the name of the method parameters:
```
'iExpectedCount' => 0,
'iNotExpectedCount' => 1
```

rquetiez (changes requested):

core/dbobjectsearch.class.php
steffunky: Seen with Romain, it's complicated as DBSearch holds the "unsecured" filter way that we still need to apply as it was before this fix if the opt-in parameter is not `true`

core/dbsearch.class.php
Molkobain: Add return type hint

```suggestion
	protected function ApplyDataFilters(): DBSearch
```

core/dbobjectsearch.class.php
Molkobain: Add return type hint

```suggestion
	 * @inheritDoc
	 * @return DBObjectSearch
	 */
	protected function ApplyDataFilters(): DBObjectSearch
```

core/dbunionsearch.class.php
Molkobain: Add return type hint

```suggestion
	 * @inheritDoc
	 * @return DBUnionSearch
	 */
	protected function ApplyDataFilters(): DBUnionSearch
```

Molkobain (changes requested):
Just some suggestions about documentation / type hinting

Molkobain (approved):

core/dbsearch.class.php
Hipska: Don't think `$this` is a supported PHPDoc type, use `static` instead?

core/dbsearch.class.php
```diff
@@ -1608,4 +1594,33 @@ public function ListParameters(): array
 	 * @return array{\VariableExpression}
 	 */
 	abstract public function GetExpectedArguments(): array;
+
+	/**
+	 * Apply data filters to the search, if needed
+	 *
+	 * @return $this|DBSearch
```
Hipska: `$this` is a `DBSearch`.
```suggestion
	 * @return DBSearch
```

steffunky: When I generated the PHPDoc with PHPStorm it gave me these `@return` values so I trusted them, I'll double check if it's relevant

PR #859, opened 2026-03-27
N°8234 - Fix permission checks to conditionally allow display of unauthorized objects

Lenaick:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | [N°8234](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8234)

## Symptom

Following [PR 853](https://github.com/Combodo/iTop/pull/853), notifications for an object are no longer displayed if the user does not have read permissions on the EventNotification{YXZ} class. We would like to display them without being able to access the notification details

### DIFF
--- a/sources/Controller/AjaxRenderController.php
+++ b/sources/Controller/AjaxRenderController.php
@@ -74,7 +74,10 @@ class AjaxRenderController
 		$oSet->SetShowObsoleteData($bShowObsoleteData);
 
 		// N°8606 : Check user permissions on the main class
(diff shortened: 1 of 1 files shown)


steffunky (approved):

Hipska (commented):
Is this `display_unauthorized_objects` already in use somewhere?

Would there be a way to abuse this to gain info otherwise not visible to the user?

Lenaick:
@Hipska This parameter was introduced with [PR 853](https://github.com/Combodo/iTop/pull/853)
We have restricted read access to Events via user permissions, but we want to retain access to the list view in certain cases; this new setting is designed for that purpose

PR #864, opened 2026-03-30
N°9137 logo by network device type

v-dumas:
## Base information

## Objective (enhancement)

Ajouter un champ "logo" sur la typology NetworkDeviceType et ajouter de nouvelles valeurs par défaut

    Router
    Switch
    Firewall
    Load Balancer
    4G Router
    Access Point

- [ ] ...
- [ ] ...
- [ ] ...


datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
```diff
@@ -5313,6 +5345,13 @@
         </reconciliation>
       </properties>
       <fields>
+        <field id="logo" xsi:type="AttributeImage">
+          <display_max_width>96</display_max_width>
+          <display_max_height>96</display_max_height>
+          <storage_max_width>128</storage_max_width>
+          <storage_max_height>128</storage_max_height>
+          <is_null_allowed>true</is_null_allowed>
+        </field>
```
Hipska, line 5354: Maybe also add this attribute as `fields_semantic/image_attribute`? That could also simplify the `NetworkDevice::GetIcon` method. Or maybe just be able to eliminate that method if also adjusting its `field_semantic`?

Hipska (commented):
There are a few invalid svg's:

- ~~icon-acer.svg~~
- icon-sony.svg
- icon-toshiba.svg

I like this PR, however might be improved, see inline comment:

datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
```diff
@@ -5343,7 +5385,7 @@
         </search>
         <list>
           <items>
-            <item id="name">
+            <item id="logo">
               <rank>10</rank>
             </item>
```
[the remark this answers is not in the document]
v-dumas: This assumption is wrong.
By default every displayed class uses in its first column, the magic attribute "friendlyname" which is defined in the XML classes/class/property/naming/attributes. In this case the name is used to build the friendlyname, as a result before this modification both columns where displayed, the first one with an hyperlink to display the deatils of the object and the second identical but without link.

datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
v-dumas: It does not work, because the logo field is not in the class using it as image_attribute and ExternalField on Image does not work (yet) in iTop.

v-dumas:
Renumbering of Brand was made on purpose, it was mixed up in a previous version of itop and the sample data have been incoherent for a while.

v-dumas:
@greptileai review again please

datamodels/2.x/itop-config-mgmt/datamodel.itop-config-mgmt.xml
Hipska: Okay, but could it be added for this class (`NetworkDeviceType`)? It would show the icon when selecting a type on the network device edit page.

Hipska: Not interested in this addition @v-dumas ?

v-dumas: The drawback with your proposal, is that when you open the Typology, the icon does not really show you are on a typology. But I will discuss the topic more widely internally. In general Typology objects have no icon, but your right maybe seeing the icon while selecting the type could help if those icons where widely used and known, but I am not 100% sure they are identified immediately for what they represent...

PR #871, opened 2026-04-07
N°9448 - Fix external auth variable value

steffunky:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | N°9448

`ext_auth_variable` wasn't sanitized and allowed more values than anticipated

Parse the value and interpret it later rather than evaluating the string value

### DIFF
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -873,8 +873,8 @@ class Config
 		'ext_auth_variable' => [
 			'type' => 'string',
 			'description' => 'External authentication expression (allowed: $_SERVER[\'key\'], $_COOKIE[\'key\'], $_REQUEST[\'key\'], getallheaders()[\'Header-Name\'])',
-			'default' => '',
-			'value' => '',
+			'default' => '$_SERVER[\'REMOTE_USER\']',
+			'value' => '$_SERVER[\'REMOTE_USER\']',
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
--- a/core/config.class.inc.php
+++ b/core/config.class.inc.php
@@ -1976,11 +1976,6 @@ class Config
 	protected $m_sAllowedLoginTypes;
 
 	/**
-	 * @var string Name of the PHP variable in which external authentication information is passed by the web server
-	 */
-	protected $m_sExtAuthVariable;
-
-	/**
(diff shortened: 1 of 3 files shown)


core/config.class.inc.php
```diff
@@ -869,6 +870,14 @@ class Config
 			'source_of_value' => '',
 			'show_in_conf_sample' => false,
 		],
+		'ext_auth_variable' => [
+			'type' => 'string',
+			'description' => 'External authentication expression (allowed: $_SERVER[\'key\'], $_COOKIE[\'key\'], $_REQUEST[\'key\'], getallheaders()[\'Header-Name\'])',
+			'default' => '',
+			'value' => '',
```
```suggestion
			'default' => '$_SERVER[\'REMOTE_USER\']',
			'value' => '$_SERVER[\'REMOTE_USER\']',
```

Molkobain (changes requested):
`Config::$m_XXX` are legacy configuration parameters, before the more generic "array" method was introduced.

If you add `ext_auth_variable` to the `Config::$m_aSettings` array (which is the "more generic array method" of defining configuration parameters), then you must remove any references to the legacy class property `Config::$m_sExtAuthVariable` for 2 reasons:
  * For the sake of readability / maintenability, otherwise it will just be confusing
  * In `Config::WriteToFile()` the value will be overwritten by `Config::DEFAULT_EXT_AUTH_VARIABLE`

Other things to do:
  * Remove all references to `Config::$m_sExtAuthVariable`
  * Update `Config::SetExternalAuthenticationVariable()` so it calls `Config::Set()` to update the generic array instead

_Note: We had the exact same discussion with @accognet on another PR or commit, but I can't find it._

accognet:
You might be thinking of my commits regarding the bug N°8786 - configuration: allow conditions on the allowed_login_types field.

Molkobain:
> You might be thinking of my commits regarding the bug N°8786 - configuration: allow conditions on the allowed_login_types field.

Exactly thanks! @steffunky you can check the 4 commits starting with `N°8786` to see how it was handled.

steffunky:
@greptileai review again please
@Molkobain review again please

Molkobain (approved):

jbostoen:
It may be worth mentioning that cookie is definitely not the most secure way to pass info? Contrary to e.g. a server variable, it's quite easy to manipulate?
https://www.itophub.io/wiki/page?id=latest:admin:user_authentication_options

PR #877, opened 2026-04-09
feat(PowerConnection): Add dashboard to list all connected devices

Hipska:
## Base information
| Related to a SourceForge thread / Another PR / Combodo ticket? | #891

## Objective

To be able to see all directly connected devices on the power connection. Currently, only the list of PDUs is presented.

## Proposed solution

Replace the PDU list tab with a dashboard that has a list for both PDUs and connected datacenter devices.

A unit test doesn't seem relevant as it's only a datamodel change.

### DIFF
--- a/datamodels/2.x/itop-datacenter-mgmt/datamodel.itop-datacenter-mgmt.xml
+++ b/datamodels/2.x/itop-datacenter-mgmt/datamodel.itop-datacenter-mgmt.xml
@@ -495,7 +495,49 @@
           </attributes>
         </reconciliation>
       </properties>
-      <fields/>
+      <fields>
+        <field id="powered_devices" xsi:type="AttributeDashboard">
+          <is_user_editable>false</is_user_editable>
+          <definition>
+            <title>Class:PowerConnection/Attribute:powered_devices+</title>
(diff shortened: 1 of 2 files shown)


Molkobain:
@Hipska we discussed this PR during the last review.

Adding the field and tab on the parent class `PowerConnection` doesn't seem relevant as it will also be available on the `PDU` class. We would consider it if you were to add it on the `PowerSource` class though.

Hipska:
Why would it not be relevant on the PDU class? Having a list of devices powered by the PDU seems useful to me..

Molkobain:
Will be discussed with @v-dumas

Hipska:
Would you also be interested in having a 'impact-analysis'-like view where it shows the relation of which device is powered by which up to the power source? "powered by..." and "powers..." as menu items..

Hipska:
@Molkobain @v-dumas ?

Molkobain:
We are currently crunching to release 3.3.0-RC1, I don't think @v-dumas had time to check yet.

Hipska:
Actually, I was hoping this could have been added in iTop 3.3, the PR was created back in April.

PR #880, opened 2026-04-10
N°8766 - Fix wrong author on portal log

jf-cbd:
## Base information

|----------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| Related to a SourceForge thread / Another PR / Combodo ticket? | https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8766

A log entry added by a support agent (e.g., Romy Schneider) to a ticket appears with the wrong author in the portal : when the agent account is disabled, the log entry is incorrectly shown as sent by the ticket creator's initials (e.g., Claude Monet), instead of the one the agent that wrote the log entry.

1. On iTop 3.2.1
2. With PHP 8.3
3. Log in as a support agent (Romy Schneider) and add a log entry to a ticket.
4. Log in as the Portal user (Claude Monet, the ticket requester) and open the ticket. Verify the log entry appears as sent by the agent.
5. Disable the agent’s user account (Romy Schneider).
6. Log in again as the Portal user and open the ticket. Verify the log entry now appears as sent by the Portal user’s initials (Claude Monet) while it should be the one of the agent's user.

## Cause (bug)

When the agent’s account is disabled, the system does not correctly resolve the original author of the log entry and falls back to displaying the Portal user’s initials.

Ensure that log entries display the correct author, even if the agent’s account is disabled.

### DIFF
--- a/core/userrights.class.inc.php
+++ b/core/userrights.class.inc.php
@@ -1345,12 +1345,12 @@ class UserRights
 	 * @return string
 	 * @throws \OQLException
 	 */
-	public static function GetUserFriendlyName($sLogin = '')
+	public static function GetUserFriendlyName(string $sLogin = '', bool $bAllowDisabledUsers = false)
 	{
 		if (empty($sLogin)) {
 			$oUser = self::$m_oUser;
 		} else {
-			$oUser = self::FindUser($sLogin);
+			$oUser = self::FindUser($sLogin, 'any', $bAllowDisabledUsers);
 		}
 		if (is_null($oUser)) {
 			return '';
--- a/core/userrights.class.inc.php
+++ b/core/userrights.class.inc.php
@@ -1345,12 +1345,12 @@ class UserRights
 	 * @return string
 	 * @throws \OQLException
 	 */
-	public static function GetUserFriendlyName(string $sLogin = '', bool $bAllowDisabledUsers = false)
+	public static function GetUserFriendlyName(string $sLogin = '')
 	{
 		if (empty($sLogin)) {
 			$oUser = self::$m_oUser;
 		} else {
-			$oUser = self::FindUser($sLogin, bAllowDisabledUsers: $bAllowDisabledUsers);
+			$oUser = self::FindUser($sLogin);
 		}
 		if (is_null($oUser)) {
 			return '';
@@ -1817,7 +1817,7 @@ class UserRights
 	 * @return User The found user or null
 	 * @throws \OQLException
 	 */
-	protected static function FindUser($sLogin, $sAuthentication = 'any', $bAllowDisabledUsers = false)
+	protected static function FindUser($sLogin, $sAuthentication = 'any', $bAllowDisabledUsers = true)
 	{
 		if ($sAuthentication === 'any') {
 			$oUser = self::FindUser($sLogin, 'internal', $bAllowDisabledUsers);
--- a/core/userrights.class.inc.php
+++ b/core/userrights.class.inc.php
@@ -1350,7 +1350,7 @@ class UserRights
 		if (empty($sLogin)) {
 			$oUser = self::$m_oUser;
 		} else {
-			$oUser = self::FindUser($sLogin, 'any', $bAllowDisabledUsers);
+			$oUser = self::FindUser($sLogin, bAllowDisabledUsers: $bAllowDisabledUsers);
 		}
 		if (is_null($oUser)) {
 			return '';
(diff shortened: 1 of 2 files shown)


core/userrights.class.inc.php
```diff
@@ -1345,12 +1345,12 @@ public static function GetContactObject()
...
 	 * @throws \OQLException
 	 */
-	public static function GetUserFriendlyName($sLogin = '')
+	public static function GetUserFriendlyName(string $sLogin = '', bool $bAllowDisabledUsers = false)
 	{
 		if (empty($sLogin)) {
 			$oUser = self::$m_oUser;
 		} else {
-			$oUser = self::FindUser($sLogin);
+			$oUser = self::FindUser($sLogin, bAllowDisabledUsers: $bAllowDisabledUsers);
```
Molkobain: I think we should always pass `true` for the `$bAllowDisabledUsers` parameter as this method is about getting the friendlyname corresponding to a login. It should always return it no matter whether the user is disabled or not.

It could actually fix the issue in other places.

Leaving other usages of `UserRights::FindUser()` as there are on the other hand sounds good.

_Note: If going forward with my proposition, remove the new parameters from `UserRights::GetUserFriendlyname()`_

core/userrights.class.inc.php
```diff
@@ -1345,7 +1345,7 @@ public static function GetContactObject()
 	 * @return string
 	 * @throws \OQLException
 	 */
-	public static function GetUserFriendlyName($sLogin = '')
+	public static function GetUserFriendlyName(string $sLogin = '')
```
Molkobain:
```suggestion
	public static function GetUserFriendlyName($sLogin = '')
```

We don't add type hinting on existing parameters unless we audited the whole codebase to ensure that there isn't a line where something else is passed (e.g. `null`), otherwise it could crash.

core/userrights.class.inc.php
Molkobain:
```suggestion
			$oUser = self::FindUser($sLogin, bAllowDisabledUsers: true);
```

core/userrights.class.inc.php
```diff
@@ -1817,7 +1817,7 @@ public static function FlushPrivileges($bResetAdminCache = false)
 	 * @return User The found user or null
 	 * @throws \OQLException
 	 */
-	protected static function FindUser($sLogin, $sAuthentication = 'any', $bAllowDisabledUsers = false)
+	protected static function FindUser($sLogin, $sAuthentication = 'any', $bAllowDisabledUsers = true)
```
Molkobain:
```suggestion
	protected static function FindUser($sLogin, $sAuthentication = 'any', $bAllowDisabledUsers = false)
```

sources/Renderer/Bootstrap/FieldRenderer/BsSimpleFieldRenderer.php
```diff
@@ -653,7 +653,7 @@ protected function PreparingCaseLogEntries(RenderingOutput &$oOutput)
 				if (false === empty($iEntryUserId)) {
 					$oEntryUser = MetaModel::GetObject('User', $iEntryUserId, false /* Necessary in case user has been deleted */, true);
 					if (!is_null($oEntryUser)) {
-						$sEntryUserLogin = UserRights::GetUserFriendlyName($oEntryUser->Get('login'));
+						$sEntryUserLogin = UserRights::GetUserFriendlyName($oEntryUser->Get('login'), true);
```
Molkobain:
```suggestion
						$sEntryUserLogin = UserRights::GetUserFriendlyName($oEntryUser->Get('login'));
```

Molkobain (changes requested):
What I had in mind when we discuss it was to change the default behavior of `UserRights::GetUserFriendlyname()` by always passing `true` at line 1353. Not to change the default behavior of `UserRights::FindUser()`.

That way it would fix all use cases of `UserRights::GetUserFriendlyname()` while ensuring `UserRights::FindUser()` still works as expected.

sources/Renderer/Bootstrap/FieldRenderer/BsSimpleFieldRenderer.php
Molkobain, line 656:
```suggestion
						$sEntryUserLogin = UserRights::GetUserFriendlyName($oEntryUser->Get('login'));
```

Molkobain (changes requested):
Mind to remove thisas well, as you removed the second parameter from the method. Then it's ok.

Molkobain (approved):

PR #885, opened 2026-04-17
N°9144 Fix static analysis issues

Timmy38:
Fixing various issues raised by phpstan


setup/wizardsteps/WizStepModulesChoice.php
```diff
@@ -518,6 +515,7 @@ public function GetSelectedModules($aInfo, $aSelectedChoices, &$aModules, $sPare
 						$bSelected = true;
 						if (isset($aModuleInfo[$sModuleId])) {
 							// Test if module has 'auto_select'
```
Hipska: This additional comment can be left out, since it is very clear what the `if` does?

setup/wizardsteps/WizStepSummary.php
```diff
@@ -171,6 +171,7 @@ public function Display(WebPage $oPage)
 		}
 
 		if (isset($aMiscOptions['generate_config'])) {
+			/** @var \DOMDocument $oDoc */
 			$oDoc = new DOMDocument('1.0', 'UTF-8');
```
Hipska: This seems obvious?

setup/ajax.dataloader.php
```diff
@@ -181,6 +181,7 @@ function FatalErrorCatcher($sOutput)
 
 if (function_exists('memory_get_peak_usage')) {
 	if ($sOperation == 'file') {
+		$sFileName = utils::ReadParam('file', '', false, 'raw_data');
 		SetupLog::Info("loading file '$sFileName', peak memory usage. ".memory_get_peak_usage());
 	} else {
```
Hipska: This might be a leftover and should be removed, as operation file would result in the "unsupported operation" exception to be thrown.

setup/compiler.class.inc.php
Hipska: Otherwise, this section could be put together with the next `if (!is_null($oInitialStatePath)) {` at line 1600.

setup/wizardcontroller.class.inc.php
```diff
@@ -67,7 +67,7 @@ protected function PushStep($aStepInfo)
 
 	/**
 	 * Removes information about the previous step from the stack
-	 * @return hash Array('class' => , 'state' => )
+	 * @return array Array('class' => , 'state' => )
```
Hipska: Rather be complete in the type
```suggestion
	 * @return array{'class': string, 'state': string}
```

tests/php-static-analysis/config/base.dist.neon
```diff
@@ -37,4 +37,5 @@ parameters:
             - ../../../env-toolkit (?)              # Irrelevent as it will either already be in `env-production` or might be desynchronized from `env-production` (for local run only, not useful in the CI)
 
             - ../../../tests (?)                    # Exclude tests for now
-            - ../../../toolkit (?)                  # Exlclude toolkit for now
+            - ../../../toolkit (?)                  # Exclude toolkit for now
+            - ../../../setup/compat                 # Exclude fake DOMDocument & DOMElement declarations
```
Hipska: Please add a EOL at EOF.

setup/wizardsteps/WizStepSummary.php
Timmy38: It really is. And phpstan does not raise any issue without.
I wonder why I added it...

Anyway, I removed it

setup/ajax.dataloader.php
Timmy38: You are right, it was a leftover.
Removed the two lines and the if/else statement

setup/compiler.class.inc.php
Timmy38: Good idea, moved the lines in the next if statement

setup/wizardcontroller.class.inc.php
Timmy38: Fixed

Hipska (approved):

odain-cbd (approved):

eespie (approved):

PR #888, opened 2026-04-21
Update contributions guidelines

jf-cbd:
## Base information

Improve contributing guide and add issue in the process

Add more details about the new process

- [x] I have added a unit test, otherwise I have explained why I couldn't => not relevant


.github/issue_template.md
Hipska, line 1: It would be nice to have this as an [Issue form](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) instead of template.

.github/pull_request_template.md
```diff
@@ -1,83 +1,76 @@
...
-|---------------------------------------------------------------|--------
-| Related to a SourceForge thead / Another PR / Combodo ticket? | <!-- Put the URL -->
-| Type of change?                                               | Bug fix / Enhancement / Translations
 
+| Question                                                                        | Answer                               |
+|---------------------------------------------------------------------------------|--------------------------------------|
+| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | <!-- Put the URL -->                 |
+| Type of change?                                                                 | Bug fix / Enhancement / Translations |
 
 ## Symptom (bug) / Objective (enhancement)
```
Hipska, line 17: IMHO, the sections **Symptom/Objective**, **Reproduction procedure** and **Cause** could be removed as SF thread / GH issue / Combodo ticket should already have this, no?

Leaving the sections that are relevant for a PR: **Proposed solution** and **Checklist**.

.github/issue_template.md
jf-cbd: Good idea, I'll take a look at it !

.github/pull_request_template.md
jf-cbd: Maybe we'll have some PR that didn't have any related content (at the author's risk), so it's preferable to keep these sections.

Also, it's good to have these sections because it's almost the same that the one the Bug objects that are created, so it's easier for us to create them from the PRs description :)

Hipska: Then you are doing it the other way round, and very hard to enforce anything..

jf-cbd: If a PR is created without issue, how can we create the bug before ? Because in this case, we need these sections.

Maybe we can indicate to fill these part only if it hasn't been made previously in an issue, but it's not making it easier for anyone reading the PR to understand it (since the persons will have to go to the issue to better understand the context).

Hipska: That's what I'm saying with enforcement. If you enforce an Issue/Bug is created before the PR and otherwise close the PR, then there is no need to duplicate all the info.

If you not enforce this, you can also not enforce users to create issue as well (as why would they, the data is also in the PR), and thus rendering this change useless.

.github/ISSUE_TEMPLATE/issue.yml
```diff
@@ -0,0 +1,80 @@
...
+
+  - type: dropdown
+    id: issue_type
+    attributes:
+      label: Type of issue
+      options:
+        - Bug report
+        - Enhancement
+    validations:
+      required: true
```
Hipska: This could be more logical if you create different template for Bug and Enhancement

.github/ISSUE_TEMPLATE/issue.yml
```diff
@@ -0,0 +1,80 @@
...
+        2. Then do that
+        3. ...
+        4. Finally, see that... (what is expected and what is actually happening)
+    validations:
+      required: false
+
+  - type: textarea
+    id: enhancement_procedure
+    attributes:
+      label: Reproduction procedure (enhancement only)
```
Hipska: I think the label is a bit misleading?

CONTRIBUTING.md
```diff
@@ -4,30 +4,29 @@ You want to contribute to iTop? Many thanks to you! 🎉 👍
 
 Here are some guidelines that will help us integrate your work!
 
-
-## Contributions
-
```
Hipska, line 9: These 2 lines should stay? Otherwise there is no L2 header anymore for **Subjects** and **License and copyright** L3 sections.

.github/pull_request_template.md
jf-cbd: > why would they, the data is also in the PR

I think the main purpose we identified (and that you suggested) about the issues is allowing contributors to ask before starting coding, so they don't spend time coding smt that will be refused.

Tbh we didn't discuss if we should automatically close PR if there is no issue created before (what if a PR makes sens but doesn't have an issue ?).  For now, we simply assume that if a user creates a pull request without reporting an issue, he agrees that it may be refused/closed, even if he has spent time on it.
We may eventually reconsider this, but I'm surprised that it seems to be such a problem.

.github/ISSUE_TEMPLATE/issue.yml
jf-cbd: Good idea, thanks !

.github/pull_request_template.md
Hipska: Aha, then I stand corrected, and maybe these few sentences might be copied and clarified into the CONTRIBUTING.md.

CONTRIBUTING.md
```diff
@@ -4,30 +4,31 @@ You want to contribute to iTop? Many thanks to you! 🎉 👍
...
 
-
 ## Contributions
 
 ### Subjects
+
 You are welcome to create pull requests on any of those subjects:
 
 * 🐛 bug fix
 * 🌐 translation / i18n / l10n
```
Hipska, line 14: Maybe add **Enhancement** in this list as well?

CONTRIBUTING.md
```diff
@@ -4,30 +4,31 @@ You want to contribute to iTop? Many thanks to you! 🎉 👍
...
 
-If you want to implement a **new feature**, please [create a corresponding ticket](https://sourceforge.net/p/itop/tickets/new/) for review.   
+If you want to implement a **new feature**, please [create a corresponding issue](https://github.com/Combodo/iTop/issues) for review.
```
Hipska: Maybe change to reference-style link? Then the second link in this section can also use the same reference.

CONTRIBUTING.md
```diff
@@ -37,47 +38,31 @@ Anyhow, you are encouraged to signal your contribution by the mean of `@author`
 If you want to use another license or keep the code ownership (copyright), you may [create an extension][wiki new ext].
 
 [license.txt]: https://github.com/Combodo/iTop/blob/develop/license.txt
-[wiki new ext]: https://www.itophub.io/wiki/page?id=latest%3Acustomization%3Astart#by_writing_your_own_extension
 
+[wiki new ext]: https://www.itophub.io/wiki/page?id=latest%3Acustomization%3Astart#by_writing_your_own_extension
```
Hipska, line 46: No need to add a newline between link references?

Hipska (commented):
The templates look good!

.github/pull_request_template.md
jf-cbd: Perfect ! I'll check tomorrow if I can make some improvements to make it more clear

jf-cbd: I added a sentence to make it more clear

CONTRIBUTING.md
jf-cbd: Done, thanks for the suggestion

CONTRIBUTING.md
jf-cbd: Done

CONTRIBUTING.md
jf-cbd: Not needed, but it avoids IDE warning haha
No impact on user, just very slightly better when reading the .md file

Hipska: Which linter rule is triggered by this? Couldn't find one..

Hipska (approved):

CONTRIBUTING.md
jf-cbd: Don't know, it's Phpstorm rules saying "incorrect whitespace". Anyway, that's not that important since it's invisible when displaying the file in GitHub.

PR #894, opened 2026-04-24
N°9542 - Portal : add the possibility to choose subclass in search screen

accognet:
internal
image


Hipska:
Shouldn't this new control better be located next to the existing **Filter** control? As this also filters the list. Similar controls close together == better UX.

Molkobain:
@accognet you should not start developing a new feature without designing it with POs first. This is just wasting time.

accognet:
I developed this feature for the Consulting team; I simply integrated the code I wrote in the iTop core.

PR #902, opened 2026-05-04
fix(CMDBSource): getFieldSpec returns NULL as a string

Hipska:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | N/A

## Symptom

Not able to use `ModuleInstallerAPI::MoveColumnInDB` when the destination field is not yet present and it is a nullable integer, date, datetime.

> Module xxxx : error when calling module installer class xxxx for BeforeDatabaseCreation handler: ModulelId = xxxx, ModuleInstallerClass = xxxx, ModuleInstallerHandler = BeforeDatabaseCreation, ExceptionClass = MySQLException, ExceptionMessage = Failed to issue SQL query: query = ALTER TABLE \`ticket\` ADD \`resolution_date\` datetime DEFAULT 'NULL', mysql_errno = 1067, mysql_error = Invalid default value for 'resolution_date'

## Reproduction procedure

1. On iTop 3.2.2
2. With PHP 8.3.20
3. Add `static::MoveColumnInDB('ticket_request', 'resolution_date', 'ticket', 'resolution_date');` to a installer class.
4. Run setup
5. Finally, see error as above.

## Cause

The code in `CMDBSource::getFieldSpec` makes wrong assumptions.

## Proposed solution

First add tests that prove some scenario's fail, then update the mentioned method to better handle `NULL` defaults.

Test results before change: [CMDBSourceTest_testGetFieldSpec.html](https://github.com/user-attachments/files/27340448/CMDBSourceTest_testGetFieldSpec.html)

### DIFF
--- a/core/cmdbsource.class.inc.php
+++ b/core/cmdbsource.class.inc.php
@@ -1279,20 +1279,19 @@ class CMDBSource
 		}
 
 		if (is_numeric($aFieldData["Default"])) {
-			if (strtolower(substr($aFieldData["Type"], 0, 5)) == 'enum(') {
-				// Force quotes to match the column declaration statement
-				$sRet .= ' DEFAULT '.self::Quote($aFieldData["Default"], true);
+			if (self::IsNumericType($aFieldData)) {
+				$sRet .= ' DEFAULT '.$aFieldData["Default"];
 			} else {
-				if (self::IsNumericType($aFieldData)) {
-					$sRet .= ' DEFAULT '.$aFieldData["Default"];
-				} else {
-					$default = $aFieldData["Default"] + 0; // Coerce to a numeric variable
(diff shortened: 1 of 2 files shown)


jf-cbd:
Hello, thank you for your PR. Unfortunately, we're not planning to work on this module in the upcoming 6 months, therefore we prefer to close it, rather than letting it open for years as we did in the past. Sorry :/

jf-cbd:
Hello, thank you for your PR. I'm sorry, but this PR is on a function that's not often used (only when a renaming of a table is needed), and that works in most of the cases. For this reason, we're not willing to assign this PR to a sprint. Please create issue so, in case of refusal, you don't spend time coding it.

PR #906, opened 2026-05-06
N°9205 - Show progress messages in setup sequencer

Lenaick:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | [N°9205](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=9205)                 |

## Objective

Show progress messages in setup sequencer

## Proposed solution

Display a message confirming that the previous step was successful each time we move on to the next step

image

image


Hipska:
Do you have an example how it would look?

eespie (approved):

Molkobain:
> Do you have an example how it would look?

Curious as well :)

odain-cbd (changes requested):
great job. sequencer behaviour has been enhanced. please adapt their tests. i guess they will fail otherwise...

Lenaick:
> Do you have an example how it would look?

I've added some screenshots to the PR description, the messages still need to be approved by our PO

Hipska:
Looks good. Is that "verification" step something new?

jf-cbd (approved):

Lenaick:
> Looks good. Is that "verification" step something new?

@Hipska yes, this is a new step included in the process for uninstalling extensions. It checks whether there is any data that needs to be cleared before running the setup

Hipska:
I have doubts about the "Data consistency check"..

What about custom extensions that do a whole datamodel rework/migration in their 2.0 version? I currently solve consistency problems in `BeforeDatabaseCreation`, and `AfterDatabaseCreation`, but that check seems to be done even before that, so consistency will fail?

BenGrenoble:
@Hipska 

This data consistency is a new step checking if a subclass has been removed. 
If yes it'll suggest to clean the data before uninstalling the extension.
You will still be able to use BeforeDatabaseCreation and AfterDatabaseCreation.
Moreover if you remove a subclass and want to migrate the data AfterDatabaseCreation it's still possible by choosing ignore and continue.

Hipska:
So it will give useless and frightening warnings to the users, even while there are perfect, clean migration steps under the hood?

This will give also warnings for all extensions that somehow use `ModuleInstallerAPI::RenameClassInDB`?

odain-cbd:
> I have doubts about the "Data consistency check"..
> 
> What about custom extensions that do a whole datamodel rework/migration in their 2.0 version? I currently solve consistency problems in `BeforeDatabaseCreation`, and `AfterDatabaseCreation`, but that check seems to be done even before that, so consistency will fail?

data audit step occurs in the middle of iTop setup. after compilation and before real application upgrade (where BeforeDatabaseCreation/AfterDatabaseCreation will be executed among others). 

in case of data possible errors setup hangs and remains as before (unless you force it to go further). admin can switch to data cleanup page. this not a under the hood process...

Hipska:
>in case of data possible errors setup hangs and remains as before (unless you force it to go further). admin can switch to data cleanup page. this not a under the hood process...

So the user will see the consistency check failing, but still able to click to continue setup (where it will run `BeforeDatabaseCreation` and `AfterDatabaseCreation` methods)?

Or how should I see that, do you have a screenshot of such "setup hangs" scenario? And what is that "admin can switch to data cleanup page"?

BenGrenoble:
It's really good to know that you're able to do this "under the hood" but most of our community is only using XML to customize iTop. The uninstallation functionality has a real value for them.
Also it will allow to remove lots of existing modules and extensions without having to code a specific algorithm for each of them.

Hipska:
I don't feel like that's an answer to my questions.

There is no way to do DB migrations with XML.

PR #909, opened 2026-05-12
N°9134 Validate extension codes before installation and refine progress message styling

Lenaick:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? |  [N°9134](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=9134)                |

## Objective

Detect early on during the setup phase whether an extension lacks code

## Proposed solution

Check that all extensions have a code during compilation; raise an error if this is not the case.
Remove unnecessary CSS rules from the progress content to ensure UI consistency.


setup/runtimeenv.class.inc.php
```diff
@@ -1467,6 +1467,14 @@ public function DoCompile(array $aRemovedExtensionCodes, array $aSelectedModules
 		// Removed modules are stored as static for FindModules()
 		$oExtensionsMap->DeclareExtensionAsRemoved($aRemovedExtensionCodes);
 
+		// Check that all the extensions have a code
+		foreach ($oExtensionsMap->GetAllExtensions() as $oExtension) {
+			if (empty($oExtension->sCode)) {
+				$sExtensionLabel = !empty($oExtension->sLabel) ? $oExtension->sLabel : $oExtension->sSourceDir;
+				throw new Exception(sprintf('Extension "%s" cannot be installed: Missing extension code', $sExtensionLabel));
```
odain-cbd, line 1474: what about an earlier failure in iTopExtension constr()?

odain-cbd (dismissed):
maybe my suggestion is not possible? 

extensions with no code is a corner case. does it exist in real life with our customers?

setup/runtimeenv.class.inc.php
Lenaick: I don’t think that’s a good idea from a functional point of view; if we throw an exception in the constructor, won’t the extension no longer be listed among those present?
It seems sensible to me to raise the error only during the compatibility check

Lenaick:
> maybe my suggestion is not possible?
> 
> extensions with no code is a corner case. does it exist in real life with our customers?

No, that shouldn’t exist because it blocks the setup; it’s just an extra safeguard.
As mentioned in the ticket, it blocks iTop, so you then need to manually delete the _env-production_ and _data/cache-production_ folders...

odain-cbd (approved):

PR #915, opened 2026-05-21
N°9622 - Error in the user story life cycle

accognet:
Internal


application/cmdbabstract.class.inc.php
```diff
@@ -3240,7 +3240,13 @@ public function DisplayStimulusForm(WebPage $oPage, $sStimulus, $aPrefillFormPar
 							$aAllowedValues = MetaModel::GetAllowedValues_att($sClass, $sAttCode, $aArgs);
 							if (is_array($aAllowedValues) && count($aAllowedValues) == 1) {
 								$aValues = array_keys($aAllowedValues);
-								$this->Set($sAttCode, $aValues[0]);
+								if ($oAttDef->IsLinkSet()) {
+									$oObjectSet = $this->Get($sAttCode);
+									$oObjectSet->AddItem(MetaModel::GetObject($oAttDef->GetLinkedClass(), $aValues[0]));
```
odain-cbd: is there any way $aValues contains more than one single value? 

can you add a test that cover DisplayStimulusForm with this new fixed usecase (linkset), pls.

odain-cbd: what if GetObject throws an exception when object does not exist? if it could happen, it could be nice for support team to catch the exception, log it and return false (i guess?).

odain-cbd (changes requested):

application/cmdbabstract.class.inc.php
accognet: If the attribute is a LinkSet, GetAllowedValues_att returns an ObjectSet from an OQL query. Therefore, I believe this object must exist.

accognet: And to answer the first question, it is impossible for $Values ​​to contain more than one value, because the test count($allowedValues) == 1 was performed just before.

odain-cbd: ok technically it is not possible. you are right (count==1).

so functionally this PR fixes stimulus form display of current object, by adding first found allowed value to the linkset, right? no all possible allowed values. one is enough... 
if there is one value to add, cant it be default one if it exists? maybe it does not make sense?

Molkobain:
I discussed this fix with @rquetiez, we need to dig deeper into the cause. For now we focus on the 3.3.0-beta, discussion will be done afterwards.

application/cmdbabstract.class.inc.php
```diff
@@ -3240,7 +3240,13 @@ public function DisplayStimulusForm(WebPage $oPage, $sStimulus, $aPrefillFormPar
...
+									$this->Set($sAttCode, $oObjectSet);
```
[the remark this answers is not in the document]
bdalsass: Right, the proposed solution doesn't work with AttributeLinksetIndirect.

PR #920, opened 2026-05-27
:bug: Skip empty result section when accelerator has enable_enlarge=false

### LINKED ISSUE #919: Empty result section is rendered for accelerated classes even when enable_enlarge is false
### Are you willing to create (at a later stage) a PR for that?

Yes

### iTop version

3.2.2-1

### PHP version

8.2

## Context

  When configuring `full_text_accelerators` with `enable_enlarge => false` on a class, iTop still renders an empty "0 result" section for that class in the global search results. The section contains only a header (`Found 0 object(s) of class X.`) with no actionable
  button, since the enlarge action is explicitly disabled by configuration.

  On instances with several accelerated classes configured this way, the user has to scroll past a stack of dead-end "0 result" sections to reach the classes that actually matched the query.

  ## Steps to reproduce

  1. Add the following to `conf/production/config-itop.php` in `$MySettings`:

     ```php
     'full_text_accelerators' => array(
         'Person' => array(
             'query'          => 'SELECT Person WHERE name LIKE :needle OR first_name LIKE :needle OR email LIKE :needle',
             'attributes'     => array('name'),
             'enable_enlarge' => false,
         ),
     ),
     ```

  2. Log in to the iTop backoffice.

  3. In the global search bar, enter a string that does **not** match any `Person` record (e.g. `zzzzzz`) and press Enter.

  4. **Expected:** no `Person` section is rendered in the results — there is no actionable content to show.

     **Actual:** a `<div class="search-class-result search-class-Person">` block is rendered with the header `Found 0 object(s) of class Person.` and an empty action area.

AlxFrst:
## Summary

Fixes #919.

When `full_text_accelerators` is configured with `enable_enlarge => false` for a class, the global search still rendered an empty "0 result" section for that class. The block contained no actionable content (no Enlarge button, by configuration) and only added visual clutter, especially on instances with several accelerators configured this way.

This PR aligns the rendering with the existing `enable_enlarge` semantic: when the action is explicitly disabled, the empty section is no longer rendered.

## Cause

In `pages/ajax.render.php` (~ line 1424), the empty result block was rendered for every accelerated class regardless of the `enable_enlarge` value. The flag only governed whether the button was shown inside the header, not whether the section itself should appear when there was no match.

## Fix

Wrap the empty-section rendering in a check for `enable_enlarge !== false`. The check defaults to ``enabled`` when the key is absent, preserving the current behavior for any configuration that doesn't opt into the new behavior.

## Compatibility

| `enable_enlarge` | 0 result | N > 0 results |
|---|---|---|
| not set (default) | rendered (unchanged) | rendered (unchanged) |
| `true` | rendered (unchanged) | rendered (unchanged) |
| `false` | **not rendered (this PR)** | rendered (unchanged) |

Strictly opt-in via existing configuration. Only impacts configurations that explicitly set `enable_enlarge => false`, where the current behavior was arguably a bug.

## Test plan

- [x] PHP lint passes on the modified file
- [x] Manual: accelerator with `enable_enlarge => false`, search with no match → empty section no longer appears
- [x] Manual: accelerator with `enable_enlarge => true` (or omitted), search with no match → empty section still appears with the Enlarge button
- [x] Manual: results page with N>0 matches is unchanged in both cases

### DIFF
--- a/pages/ajax.render.php
+++ b/pages/ajax.render.php
@@ -1423,12 +1423,19 @@ EOF;
 						}
 					} else {
 						if (array_key_exists($sClassName, $aAccelerators)) {
-							$oPage->add("<div class=\"search-class-result search-class-$sClassName\">\n");
-							$oPage->add("<div class=\"page_header\">\n");
-							$oPage->add('<h2 class="ibo-global-search--result--title">'.MetaModel::GetClassIcon($sClassName).Dict::Format('UI:Search:Count_ObjectsOf_Class_Found', 0, Metamodel::GetName($sClassName)).$sEnlargeButton."</h2>\n");
-							$oPage->add("</div>\n");
-							$oPage->add("</div>\n");
-							$oPage->p('&nbsp;'); // Some space ?
+							// Skip the empty result section when enlarge is disabled:
+							// the block has no actionable content (no Enlarge button) and only
+							// adds visual clutter on instances with many accelerators.
+							$bEnlargeEnabled = !array_key_exists('enable_enlarge', $aAccelerators[$sClassName])
+								|| $aAccelerators[$sClassName]['enable_enlarge'] !== false;
+							if ($bEnlargeEnabled) {
+								$oPage->add("<div class=\"search-class-result search-class-$sClassName\">\n");
+								$oPage->add("<div class=\"page_header\">\n");
+								$oPage->add('<h2 class="ibo-global-search--result--title">'.MetaModel::GetClassIcon($sClassName).Dict::Format('UI:Search:Count_ObjectsOf_Class_Found', 0, Metamodel::GetName($sClassName)).$sEnlargeButton."</h2>\n");
+								$oPage->add("</div>\n");
+								$oPage->add("</div>\n");
+								$oPage->p('&nbsp;'); // Some space ?
+							}
 						}
 					}
 					if ($iTune > 0) {


Hipska (commented):
Wouldn't this have been cleaner to add?
```php
if (array_key_exists($sClassName, $aAccelerators)
    && (!array_key_exists('enable_enlarge', $aAccelerators[$sClassName])
        || $aAccelerators[$sClassName]['enable_enlarge'] !== false)) {
```

jf-cbd:
Hello, thank you, we'll work on this PR in 3.4 sprint. Meanwhile, could you please send us a screenshot of actual/corrected behavior ?

AlxFrst:
Before the fix: 
Capture d’écran 2026-06-05 à 20 12 38
After:
Capture d’écran 2026-06-05 à 20 13 15

AlxFrst:
> Wouldn't this have been cleaner to add?
> 
> ```
> if (array_key_exists($sClassName, $aAccelerators)
>     && (!array_key_exists('enable_enlarge', $aAccelerators[$sClassName])
>         || $aAccelerators[$sClassName]['enable_enlarge'] !== false)) {
> ```

Good point, that does read cleaner. The only thing is it drops the query check the Enlarge button only makes sense when there's a query to widen, so without it the button would show up on accelerators that have nothing to enlarge. I can push a shorter version that keeps that check if you want.

Hipska:
My if does exactly the same as your 2 nested ifs do..

PR #925, opened 2026-06-01
N°9597 - Switch environment to a build environment must not be available

Lenaick:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | [N°9597](https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=9597)                 |

## Symptom

When checking compatibility during setup, a build environment is created. It should not be possible to switch to this environment via the system

## Proposed solution

Add an exception if the target environment of `switch_env` ends with `*-build`

### DIFF
--- a/sources/Service/Startup/StartupService.php
+++ b/sources/Service/Startup/StartupService.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace Combodo\iTop\Service\Startup;
+
+use Combodo\iTop\Application\Helper\Session;
+use CoreException;
+use IssueLog;
+
+class StartupService
+{
+	/**
+	 * @param string|null $sSwitchEnv
+	 * @param bool $bAllowCache
+	 *
+	 * @return string
+	 * @throws CoreException
+	 */
+	public static function SetItopEnvironment(?string $sSwitchEnv, bool &$bAllowCache): string
+	{
+		if (static::IsBuildEnvironment($sSwitchEnv)) {
+			$oException = new CoreException("Switching to environment '$sSwitchEnv' is not allowed since it is a build environment");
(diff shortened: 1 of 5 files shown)


sources/Service/Startup/StartupService.php
```diff
@@ -0,0 +1,58 @@
...
+class StartupService
+{
+	/**
+	 * @param string|null $sSwitchEnv
+	 * @param bool $bAllowCache
+	 *
+	 * @return string
+	 * @throws CoreException
+	 */
+	public static function SetItopEnvironment(?string $sSwitchEnv, bool &$bAllowCache): string
```
odain-cbd: what is the meaning of $bAllowCache? it seems not related to opcache/apccache.

maybe you could document what is its purpose.

odain-cbd: maybe you could change all local method and remove staitc which is not so usefull. and not recommended by phpunit (no mock possible)

odain-cbd (changes requested):

sources/Service/Startup/StartupService.php
Lenaick: This is a boolean that determines whether the APC cache can be used or not, in conjunction with the `apc_cache.enabled` parameter
In the case of the environment switch, the cache must not be used
I've updated the documentation

odain-cbd (approved):

sources/Service/Startup/StartupService.php
Hipska: Maybe a small explanation in the php doc for both parameters would be handy?

PR #926, opened 2026-06-01
N°9553 - Fix on LoadLocalizedData in ModuleInstaller

v-dumas:
## Base information

| Related to a Combodo ticket |                  |
| Type of change?                                                | Bug fix / Enhancement / Translations

Improve code and tests


Molkobain:
@v-dumas I see that Olivier is set as "assignee" to the PR, did you mean "reviewer"?
Only 1 assignee should be assigned to ensure there is an explicit owner :)

v-dumas:
@Molkobain sorry, I fixed my error, you're right he is reviewer.

v-dumas:
@greptile

setup/runtimeenv.class.inc.php
```diff
@@ -1248,7 +1248,7 @@ public function LoadData($aAvailableModules, $bSampleData, $aSelectedModules = n
 		// in the current database
 		foreach ($aPreviouslyLoadedFiles as $sFileRelativePath) {
 			$sFileName = APPROOT.$sFileRelativePath;
-			SetupLog::Info("Loading file: $sFileName (just to get the keys mapping)");
+			SetupLog::Info("Loading file: $sFileName (just to build & cache the keys mapping)");
```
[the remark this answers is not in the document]
v-dumas: No, the translated files uses the same keys, so it's not an issue

setup/moduleinstaller.class.inc.php
```diff
@@ -310,69 +310,101 @@ public static function RenameTableInDB(string $sOrigTable, string $sDstTable)
...
+	 * @param string $sDefaultFileName The default (en_us) file to load, ending with '.en_us.xml' (e.g. 'data.sample.en_us.xml'); the localized variant for the configured language is used when available
 	 *
 	 * @return void
 	 * @throws \ConfigException
 	 * @throws \CoreException
 	 * @throws \CoreUnexpectedValue
 	 */
-	public static function LoadLocalizedData(Config $oConfiguration, string $sPreviousVersion, string $sCurrentVersion, string $sFirstLoadingVersion, string $sFilePattern): void
+	public static function LoadLocalizedDataOnCrossingVersion(Config $oConfiguration, string $sPreviousVersion, string $sCurrentVersion, string $sFirstLoadingVersion, string $sDefaultFileName): void
 	{
```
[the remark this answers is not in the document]
v-dumas: This API has never been published, so no risk to change it...

setup/moduleinstaller.class.inc.php
```diff
@@ -402,26 +362,68 @@ public static function XMLFileLoad(string $sFileName): void
...
+		}
+		if (!self::IsSupportediTopVersion($sCurrentVersion)) {
+			throw new CoreUnexpectedValue("LoadLocalizedData expects sCurrentVersion to match x.y[.z][-name], got '{$sCurrentVersion}'");
+		}
+		if (($sFirstLoadingVersion !== '') && !self::IsSupportediTopVersion($sFirstLoadingVersion)) {
+			throw new CoreUnexpectedValue("LoadLocalizedData expects sFirstLoadingVersion to match x.y[.z][-name], got '{$sFirstLoadingVersion}'");
+		}
+	}
+
+	public static function IsSupportediTopVersion(string $sVersion): bool
```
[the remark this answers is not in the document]
v-dumas: Ok

datamodels/2.x/itop-tickets/module.itop-tickets.php
```diff
@@ -61,6 +61,6 @@ public static function AfterDatabaseCreation(Config $oConfiguration, $sPreviousV
 			}
 		}
 		// Load localized structural data: predefined query phrases for notifications
-		static::LoadLocalizedData($oConfiguration, $sPreviousVersion, $sCurrentVersion, '3.0.0', __DIR__."/data/{{language_code}}.data.itop-tickets.xml");
+		static::LoadLocalizedDataOnCrossingVersion($oConfiguration, $sPreviousVersion, $sCurrentVersion, '3.0.0', utils::GetAbsoluteModulePath('itop-tickets')."/data/data.itop-tickets.en_us.xml");
```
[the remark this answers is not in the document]
v-dumas: ok

setup/moduleinstaller.class.inc.php
```diff
@@ -402,26 +362,68 @@ public static function XMLFileLoad(string $sFileName): void
...
-		if (file_exists($sFileName)) {
-			return $sFileName;
+		if (str_ends_with($sOriginalFileName, '.en_us.xml')) {
+			// If the original file which can be localized, does not exist, then even if the localized file itself exists, we want to raise this design issue.
+			if (!file_exists($sOriginalFileName)) {
+				throw new Exception("File $sOriginalFileName not found, it must exist when required by the module or as iTop fallback language when no localized file exists for the iTop default language");
+			}
+			// Search for a file for the requested language,
+			$sLang = '.'.str_replace(' ', '_', strtolower($sLanguage)).'.xml';
+			$sFileName = str_replace('.en_us.xml', $sLang, $sOriginalFileName);
```
steffunky, line 383: Is it mandatory to replace including the `.xml` ?
That adds an unnecessary requirement in file naming (and is different from file naming standards like dictionaries files for example).

setup/moduleinstaller.class.inc.php
```diff
@@ -310,85 +310,45 @@ public static function RenameTableInDB(string $sOrigTable, string $sDstTable)
...
-	 * @param string $sFilePattern The pattern of the file to load, with {{language_code}} as placeholder for the language code (e.g. 'data.sample.{{language_code}}.xml')
+	 * @param string $sDefaultFileName The file to load (typically the '.en_us.xml' base file); when it ends with '.en_us.xml', the localized variant for the configured language is used when available
 	 *
 	 * @return void
 	 * @throws \ConfigException
 	 * @throws \CoreException
 	 * @throws \CoreUnexpectedValue
 	 */
-	public static function LoadLocalizedData(Config $oConfiguration, string $sPreviousVersion, string $sCurrentVersion, string $sFirstLoadingVersion, string $sFilePattern): void
+	public static function LoadLocalizedDataOnCrossingVersion(Config $oConfiguration, string $sPreviousVersion, string $sCurrentVersion, string $sFirstLoadingVersion, string $sDefaultFileName): void
```
steffunky, line 328: What's the meaning of Crossing and crossed in these functions? I'm not sure I understand the function goal with these namings

setup/moduleinstaller.class.inc.php
```diff
@@ -310,85 +310,45 @@ public static function RenameTableInDB(string $sOrigTable, string $sDstTable)
 	}
 
 	/**
+	 * Helper to load localized data from a file, only if the module is being installed for the first time or if the module is being upgraded and the FirstLoadingVersion is between the PreviousVersion and the CurrentVersion.
+	 *
 	 * @param \Config $oConfiguration
-	 * @param string $sPreviousVersion The previous version of the module (empty string will force the loading)
+	 * @param string $sPreviousVersion The previous version of the module (empty string in case of first install)
 	 * @param string $sCurrentVersion The current version of the module
 	 * @param string $sFirstLoadingVersion The first module version for which the data loading should be performed (e.g. '3.0.0')
```
steffunky: What's the point of `$sFirstLoadingVersion` ?
Will a module bear data only available for a future release of the same module?

setup/moduleinstaller.class.inc.php
```diff
@@ -402,26 +362,68 @@ public static function XMLFileLoad(string $sFileName): void
...
+			throw new CoreUnexpectedValue("LoadLocalizedDataOnCrossingVersion expects sCurrentVersion to match x.y[.z][-name], got '{$sCurrentVersion}'");
+		}
+		if (utils::IsNullOrEmptyString($sFirstLoadingVersion) || !self::IsValidVersionFormat($sFirstLoadingVersion)) {
+			throw new CoreUnexpectedValue("LoadLocalizedDataOnCrossingVersion expects sFirstLoadingVersion to match x.y[.z][-name], got '{$sFirstLoadingVersion}'");
+		}
+	}
+
+	public static function IsValidVersionFormat(string $sVersion): bool
+	{
+		return (preg_match('/^\d+\.\d+(?:\.\d+)?(?:-[A-Za-z0-9]+)?$/', $sVersion) === 1);
```
steffunky, line 423: Could you comment the method/regexp to clarify what's matching it?

setup/moduleinstaller.class.inc.php
```diff
@@ -402,26 +362,68 @@ public static function XMLFileLoad(string $sFileName): void
...
+	/**
+	 * @param string $sPreviousVersion
+	 * @param string $sCurrentVersion
+	 * @param string $sFirstLoadingVersion
+	 *
+	 * @return bool
+	 */
+	public static function IsVersionCrossed(string $sPreviousVersion, string $sCurrentVersion, string $sFirstLoadingVersion): bool
+	{
+		return ($sPreviousVersion === '') ||
```
steffunky, line 435: If the previous version does not exists, don't we check if the current version is greater that the first loading version?

setup/moduleinstaller.class.inc.php
v-dumas: I have used a different convention than the dictionaries, because there can be a lot of sample files of each language and I wanted the various translated versions of each sample or structural data file to be ranked by default in a logical order.
We could ignore the "xml" but I want to keep the ".en_us." to avoid wrong matching if a file is named sample.data.golden_user.xml

setup/moduleinstaller.class.inc.php
v-dumas: The method load the data as soon as you are on a fresh install **or** you pass over the FirstLoadingVersion during your upgrade = PreviousVersion strictly lower that the FirstLoadingVersion (and CurrentVersion is greater that FirstLoadingVersion) - The second part of the test is not really important as the situation will never occur. 
But the first part is important to load the file only once.

setup/moduleinstaller.class.inc.php
v-dumas: I was hoping the unitary tests would explicit the allowed values, should I copy that info as a comment in the code?

setup/moduleinstaller.class.inc.php
v-dumas: I don't really see a case where the **code** of module in version 3.3.0 for eg. would be able to anticipate that in future version 3.4.0 we would load some structural data. But you're right, I do test it in upgrade, so I should be consistent.

setup/moduleinstaller.class.inc.php
v-dumas: Copilot said : So “crossing a particular version during an upgrade” means while upgrading (for example from 2.3 to 3.1), you go past a version like 2.7, and that version may have special migration rules/scripts you must apply.

setup/moduleinstaller.class.inc.php
v-dumas: The method load the data as soon as you are on a fresh install or you pass over the FirstLoadingVersion during your upgrade. So FirstLoadingVersion correspond to the first version which brought those structural data files that you want to load.

setup/moduleinstaller.class.inc.php
steffunky: Right, we could benefit by naming data in modules if they depend on itop version too:
`sample.data.en_us.3_4.xml` wouldn't work with the current rules.
If that's not something we want, you can skip this comment

setup/moduleinstaller.class.inc.php
steffunky: Do we have any other reference to crossing in iTop? Or did we use a different name? Or was it never mentioned before?

setup/moduleinstaller.class.inc.php
steffunky: When we release a package, the tests aren't included. That would make it harder for people debugging their modules on a build to understand what's expected here

setup/moduleinstaller.class.inc.php
steffunky: That was my question in a previous comment, but if the previous version is only used on an upgrade where a certain version threshold is passed, this is not really necessary

setup/moduleinstaller.class.inc.php
v-dumas: Do my knowledge, the loading of QueryPhrases when you cross version 3.0.0 of the itop-ticket module is the only case where we load (structural) data on an upgrade. The code was part of the AfterDatabaseCreation of the module and I moved it as an helper in class ModuleInstallerAPI, so extensions can use it if they need. I will improve the method comment.

setup/moduleinstaller.class.inc.php
v-dumas: It's an excellent idea for modeling easily the CrossingVersion, but as this is currently only used in one single place, I give up, it sounds like over-delivery

steffunky (approved):

PR #928, opened 2026-06-03
N°9169 - adapt itsm designer connector to the new setup

eespie:
Add Return to Designer buttons in the setup wizard and the "Extensions management" module.


Lenaick:
@greptileai

PR #934, opened 2026-06-10
N°9638 - Add setup ContextTag for async operation

Lenaick:
Internal

### DIFF
--- a/sources/Application/WebPage/WebPage.php
+++ b/sources/Application/WebPage/WebPage.php
@@ -1794,11 +1794,9 @@ JS;
 	}
 
 	/**
-	 * @param bool $bReturnOutput
-	 *
 	 * @throws \Exception
 	 */
-	protected function output_dict_entries($bReturnOutput = false)
+	protected function output_dict_entries()
 	{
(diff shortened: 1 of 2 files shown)


sources/Application/WebPage/WebPage.php
```diff
@@ -1794,11 +1794,9 @@ public function GetPopoverMenu(string $sId, array $aActions): ?PopoverMenu
 	}
 
 	/**
-	 * @param bool $bReturnOutput
-	 *
 	 * @throws \Exception
 	 */
-	protected function output_dict_entries($bReturnOutput = false)
+	protected function output_dict_entries()
```
Molkobain, line 1799: @Lenaick why did you remove this? It doesn't seem to be related to the PR, and have you checked it wasn't used in the subclasses (included customizations) ?

Lenaick: @Molkobain that's a mistake. This is a test I ran that I shouldn't have committed. I'll revert it straight away

Molkobain: Thanks :)

PR #935, opened 2026-06-10
N°7060 - create ReadOnly profiles

jf-cbd:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | N°7060                 |

It also introduces new Read Only profiles.

- define new profiles
- write tests


sources/Users/ITopUserCountingRepository.php
```diff
@@ -0,0 +1,260 @@
...
+				$bIsReadOnlyUser = false;
+			}
+			if ($bIsReadOnlyUser) {
+				$aReadOnlyUsers[] = $oUser;
+			}
+		}
+
+		if (MetaModel::IsValidClass('UserToken')) {
+			$aUserToken = $this->GetApplicationUsers();
+			$aReadOnlyUsers = array_diff($aReadOnlyUsers, $aUserToken);
```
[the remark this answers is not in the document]
jf-cbd: Disabled users won't be able to log in, so IsActionAllowed will return true, and IsUserReadOnly will return false, filtering them out.

datamodels/2.x/itop-profiles-itil/datamodel.itop-profiles-itil.xml
```diff
@@ -85,6 +85,13 @@
           <class id="Attachment"/>
         </classes>
       </group>
+      <group id="Ticket" _delta="define">
+        <classes>
+          <class id="Ticket"/>
+          <class id="WorkOrder"/>
+          <class id="Attachment"/>
+        </classes>
```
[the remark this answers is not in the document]
jf-cbd: Not needed.

eespie (approved):
I'm just surprised that read-only profiles don't enforce read-only but only view on several classes

core/userrights.class.inc.php
```diff
@@ -1530,6 +1530,37 @@ public static function GetSelectFilter($sClass, $aSettings = [])
...
+	{
+		$oUserProfiles = $oUser->Get('profile_list');
+		$oUserRights = UserRights::GetModuleInstance();
+		while ($oUserProfile = $oUserProfiles->Fetch()) {
+			$iProfileId = $oUserProfile->Get('profileid');
+			if (in_array($iProfileId, $aExcludedProfilesId)) {
+				return false;
+			}
+			foreach (MetaModel::GetClasses('bizmodel,grant_by_profile') as $sClass) {
+				foreach (['w', 'bw', 'd', 'bd'] as $sWriteActionCode) {
```
Molkobain, line 1552: You should rather use the constants from `\UserRightsProfile::UR_ACTION_XXX` than hardcode them.

datamodels/2.x/itop-profiles-itil/datamodel.itop-profiles-itil.xml
```diff
@@ -205,6 +212,60 @@
       </group>
     </groups>
     <profiles>
+      <profile id="5500" _delta="define">
```
Molkobain, line 215: Are you sure about these profiles IDs? Usually, profiles brought by iTop core are lower than 1024. Check [this page of the documentation](https://www.itophub.io/wiki/page?id=latest:customization:profiles).

Poke @v-dumas

core/userrights.class.inc.php
```diff
@@ -1530,6 +1530,37 @@ public static function GetSelectFilter($sClass, $aSettings = [])
...
+	/**
+	 * @param User $oUser
+	 * @param array $aExcludedProfilesId Administrator by default, but can also be other proofiles depending on needs (e.g. power portal user or REST profile)
+	 * @return bool
+	 * @throws ArchivedObjectException
+	 * @throws CoreException
+	 * @throws CoreUnexpectedValue
+	 * @throws MySQLException
+	 */
+	public static function IsUserReadOnly(User $oUser, array $aExcludedProfilesId = [1]): bool
```
Molkobain, line 1542: Shouldn't we cache this? It's a triple nested loop. I don't know where it will be used (another module?) but if it becomes widely used this could be an issue.

datamodels/2.x/itop-profiles-itil/datamodel.itop-profiles-itil.xml
v-dumas: Agreed, for [Combodo's profiles Ids](https://www.itophub.io/wiki/page?id=latest:admin:managing_user_accounts#itop_core_profiles), I'd recommend to use numbers under 1024, and it's required to add it to this wiki page for others.
Between 22 and 40 there is a risk of collision with Profiles ID assigned by the Designer
5000 range has been used by IPAM

PR #936, opened 2026-06-11
 N°9678 - Datamodel not added after extension management setup

odain-cbd:


setup/extensionsmap.class.inc.php
```diff
@@ -759,7 +759,7 @@ public function GetExtraDirs(): array
 	public function ModuleIsChosenAsPartOfAnExtension($sModuleNameToFind, $sInSourceOnly = iTopExtension::SOURCE_REMOTE)
 	{
 		foreach ($this->GetAllExtensions() as $oExtension) {
-			if (($oExtension->sSource == $sInSourceOnly) &&
+			if (/*($oExtension->sSource == $sInSourceOnly) &&*/
 				($oExtension->bMarkedAsChosen == true) &&
```
Hipska: Commenting out code without an actual comment is useless.
```suggestion
			if (($oExtension->bMarkedAsChosen == true) &&
```

odain-cbd: it will disappear once work is finished

PR #943, opened 2026-06-22
N°9717 - Bring unattended error feedbacks back

odain-cbd:


bdalsass (approved):
E_PARSE, E_CORE_ERROR and E_COMPILE_ERROR may need to be captured ?

PR #947, opened 2026-06-24
 N°9723 - IsActionAllowed should work with non instantiated users objects

jf-cbd:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | N°9723              |
| Type of change?                                                | Bug fix / Enhancement / Translations

The method for calculating permissions (on which the permission's matrix is based) works only on user objects that are instantiated in the database.

This should not be necessary, but it causes a blockage when calculating permissions on non-instantiated user objects.

The method needs to be improved or corrected so that it can calculate the permissions of a non-instantiated user object.

## Reproduction procedure

With iTop 3.2.3, on PHP 8.3, try to compute user permission (with UserRights::IsActionAllowed) on a non instantiated User objects, and see that the method returns wrong values.
See related test case to reproduce the issue (the added test doesn't work with current state of iTop but works with the fix)

## Cause

Methods relies on user key, which can not exist if user is not existing yet.

Replace part of code that relies on user key by methods that do equivalent actions (e.g. get profile list of a user object).

### DIFF
--- a/addons/userrights/userrightsprofile.class.inc.php
+++ b/addons/userrights/userrightsprofile.class.inc.php
@@ -582,16 +582,26 @@ class UserRightsProfile extends UserRightsAddOnAPI
 	 */
 	public function ListProfiles($oUser)
 	{
-		$aRet = [];
-		$oSearch = new DBObjectSearch('URP_UserProfile');
-		$oSearch->AllowAllData();
-		$oSearch->NoContextParameters();
-		$oSearch->Addcondition('userid', $oUser->GetKey(), '=');
-		$oProfiles = new DBObjectSet($oSearch);
-		while ($oUserProfile = $oProfiles->Fetch()) {
-			$aRet[$oUserProfile->Get('profileid')] = $oUserProfile->Get('profileid_friendlyname');
+		if (count($oUser->ListChanges()) === 0) { // backward compatibility
+			$aRet = [];
+			$oSearch = new DBObjectSearch('URP_UserProfile');
+			$oSearch->AllowAllData();
+			$oSearch->NoContextParameters();
+			$oSearch->Addcondition('userid', $oUser->GetKey(), '=');
+			$oProfiles = new DBObjectSet($oSearch);
+			while ($oUserProfile = $oProfiles->Fetch()) {
+				$aRet[$oUserProfile->Get('profileid')] = $oUserProfile->Get('profileid_friendlyname');
+			}
+
+			return $aRet;
+		} else {
+			$aRet = [];
+			$oProfilesSet = $oUser->Get('profile_list');
+			foreach ($oProfilesSet as $oUserProfile) {
+				$aRet[$oUserProfile->Get('profileid')] = $oUserProfile->Get('profileid_friendlyname');
+			}
+			return $aRet;
 		}
-		return $aRet;
 	}
 
 	public function GetSelectFilter($oUser, $sClass, $aSettings = [])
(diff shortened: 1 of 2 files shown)


addons/userrights/userrightsprofile.class.inc.php
```diff
@@ -582,16 +582,26 @@ public function IsPortalUser($oUser)
...
 	{
-		$aRet = [];
-		$oSearch = new DBObjectSearch('URP_UserProfile');
-		$oSearch->AllowAllData();
-		$oSearch->NoContextParameters();
-		$oSearch->Addcondition('userid', $oUser->GetKey(), '=');
-		$oProfiles = new DBObjectSet($oSearch);
-		while ($oUserProfile = $oProfiles->Fetch()) {
-			$aRet[$oUserProfile->Get('profileid')] = $oUserProfile->Get('profileid_friendlyname');
+		if (count($oUser->ListChanges()) === 0) { // backward compatibility
```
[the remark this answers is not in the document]
jf-cbd: This is the expected behavior, listchanges === 0 also includes (in its behavior) that object is new

addons/userrights/userrightsprofile.class.inc.php
```diff
@@ -582,16 +582,26 @@ public function IsPortalUser($oUser)
...
+			$oSearch->AllowAllData();
+			$oSearch->NoContextParameters();
+			$oSearch->Addcondition('userid', $oUser->GetKey(), '=');
+			$oProfiles = new DBObjectSet($oSearch);
+			while ($oUserProfile = $oProfiles->Fetch()) {
+				$aRet[$oUserProfile->Get('profileid')] = $oUserProfile->Get('profileid_friendlyname');
+			}
+
+			return $aRet;
+		} else {
```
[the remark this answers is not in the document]
jf-cbd: Backward compatibility is kept, even if, in theory, we could do everything the same way.

tests/php-unit-tests/unitary-tests/core/UserRightsTest.php
```diff
@@ -81,6 +81,16 @@ protected function CreateUniqueUserAndLogin(string $sLoginPrefix, int $iProfileI
 		return $oUser;
 	}
 
+	public function testReadOnlyUserObject()
```
odain-cbd: please rename the test to know what is covered (method/usecase).

tests/php-unit-tests/unitary-tests/core/UserRightsTest.php
```diff
@@ -81,6 +81,16 @@ protected function CreateUniqueUserAndLogin(string $sLoginPrefix, int $iProfileI
...
+	{
+		$oUser = $this->GivenUserWithProfiles('test1', [3]); // not a readonly profile
```
odain-cbd: please use self::$aURP_Profiles['XXX'] instead of 1/3

tests/php-unit-tests/unitary-tests/core/UserRightsTest.php
```diff
@@ -81,6 +81,16 @@ protected function CreateUniqueUserAndLogin(string $sLoginPrefix, int $iProfileI
...
+		$oAdminUser = $this->GivenUserWithProfiles('test2', [1]);
+		$oAdminUser->DBInsert();
+		$_SESSION = [];
+		UserRights::Login($oAdminUser->Get('login'));
```
odain-cbd, line 90: any reason why you have to log in in this test?

tests/php-unit-tests/unitary-tests/core/UserRightsTest.php
```diff
@@ -81,6 +81,16 @@ protected function CreateUniqueUserAndLogin(string $sLoginPrefix, int $iProfileI
...
+
+		self::assertTrue(UserRights::IsActionAllowed('Server', UR_ACTION_MODIFY, null, $oUser) === UR_ALLOWED_YES);
```
odain-cbd, line 92: I commented the fix in userightsprofile and the test is still OK. am i wrong?

odain-cbd (changes requested):
no so clear what you fixed upon what you tested. pleas adapt the test for documentation and ci coverage

tests/php-unit-tests/unitary-tests/core/UserRightsTest.php
jf-cbd: Because IsActionAllowed method starts with 
```
		if (!self::CheckLogin()) {
			return UR_ALLOWED_YES;
		}
```
So without login, the test isn't relevant

tests/php-unit-tests/unitary-tests/core/UserRightsTest.php
odain-cbd: sorry wrong comment. dont know what i missed but test is failing when commenting the fix. my bad...

eespie (approved):

PR #950, opened 2026-06-30
N°9639 - Uninstallation Analysis result doesn't list all the object

eespie:
N°9639 - Uninstallation Analysis result doesn't list all the object

Refactor the count algorithm to have a result much faster


eespie:
@greptileai

PR #952, opened 2026-07-01
:bug: Fix: Restoring the backup of the iTop config file fails because of permissions

jbostoen:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | no                 |

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | no                 |

When using an install XML file to upgrade an existing instance, and previous_config_file is set to e.g. `/var/www/html/conf/production/config-itop.php` , the restoring of the backup (overwriting the original config) fails due to permissions.

- Have an install.xml file to use for upgrades.
- Have previous_config_file in the install.xml configured and point to the location of config-itop.php
- Execute the unattended install, point to the XML file and use `--use-itop-config` flag.

On my environment, this throws a permissions error.

## Cause (bug)

I think config-itop.php has its permissions already set to something stricter during the installation.

Fix permissions - to be discussed whether this should be 0600 .

- [ ] Is the PR clear and detailed enough so anyone can understand without digging in the code?

### DIFF
--- a/setup/unattended-install/unattended-install.php
+++ b/setup/unattended-install/unattended-install.php
@@ -401,6 +401,7 @@ if (! $bFoundIssues)
 	if ($bUseItopConfig && is_file("$sConfigFile.backup"))
 	{
 		echo "\nuse config file provided by backup in $sConfigFile.";
+		chmod($sConfigFile, 0660);
 		copy("$sConfigFile.backup", $sConfigFile);
 	}
 


Hipska:
Isn't it the normal way to always set correct permissions to config.php before attempting the install?

jbostoen:
> Isn't it the normal way to always set correct permissions to config.php before attempting the install?

Yes, I also do that. But, at the end of the install and with this specific unattended installation approach; it tries to copy back the original config-itop.php file. However, it fails; due to permissions that were already (re)set during iTop's installation process.

PR #956, opened 2026-07-02
N°9679 - Forced uninstallation is not logged when done by extensions management

Lenaick:
N°9679 - Forced uninstallation is not logged when done by extensions management

### DIFF
--- a/datamodels/2.x/combodo-data-feature-removal/src/Controller/DataFeatureRemovalController.php
+++ b/datamodels/2.x/combodo-data-feature-removal/src/Controller/DataFeatureRemovalController.php
@@ -129,6 +129,9 @@ class DataFeatureRemovalController extends Controller
 
 			$aRemovedExtensions = $this->aExtensionsToCheck['to_be_removed'];
 			$aHiddenInputs['removed_extensions'] = $this->ConvertIntoSetupFormat($aRemovedExtensions);
+
+			$aExtensionsNotUninstallable = $this->aExtensionsToCheck['extensions_not_uninstallable'];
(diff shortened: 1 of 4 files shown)


datamodels/2.x/combodo-data-feature-removal/src/Controller/DataFeatureRemovalController.php
```diff
@@ -439,6 +443,9 @@ public function ReadExtensionsDiff(): int
 				if (! $this->bForcedUninstallation && $aExtensionData['extra_flags']['uninstallable']) {
 					$this->bForcedUninstallation = true;
 				}
+				if (false === $aExtensionData['extra_flags']['uninstallable']) {
+					$this->aExtensionsToCheck['extensions_not_uninstallable'][] = $sCode;
+				}
```
odain-cbd, line 448: extensions_not_uninstallable means installable extensions? or uninstallable extensions?

naming is hard to understand with double negative use: "not uninstallable"

(i can't get no satisfaction sorry for that :) )

odain-cbd (changes requested):
please see double negative naming.
also is it possible to cover with tests? too much effort here for display?

datamodels/2.x/combodo-data-feature-removal/src/Controller/DataFeatureRemovalController.php
Lenaick: This flag was taken from the setup, I can change it as it’s not actually clear… It means that these extensions cannot be uninstalled. Do you have a suggestion for renaming this flag?

odain-cbd: seems to me that current double-negative name does not match the logic. what about 'uninstable_extensions' ?

PR #957, opened 2026-07-03
9322 - n:n attributes edition in portal search pop-up hidden behind

accognet:
internal


steffunky (approved):
This will need to be ported ton 3.2 new look (https://github.com/Combodo/itop-portal-new-look-for-3.2-lts/blob/master/public/js/bootstrap-portal-modal.js).
I think this will require a new bug ticket

PR #961, opened 2026-07-07
N°9628 - Fix issue following 5b9e0a1d being applied to non-jQuery elements

steffunky:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | N°9628                 |
| Type of change?                                                                 | Bug fix  |

Some changes made to support jQuery 3.4 were applied to non jQuery elements, leading to js errors or weird behaviour.

## Cause (bug)

5b9e0a1d was commited without checking if every instance modifies were jQuery elements

Revert changes for some elements, or apply 5b9e0a1d to elements that have changed since to accommodate behaviour changes  

### DIFF
--- a/templates/base/components/input/select/select.html.twig
+++ b/templates/base/components/input/select/select.html.twig
@@ -5,7 +5,7 @@
         {% include "base/components/input/inputlabel.html.twig" %}
     {% endif %}
     <select id="{{ oUIBlock.GetId() }}" name="{{ oUIBlock.GetName() }}"
-            {% if oUIBlock.GetSubmitOnChange() %} onChange="$(this).parents('form:first').trigger('submit');" {% endif %}
+            {% if oUIBlock.GetSubmitOnChange() %} onChange="$(this).parents('form').first().trigger('submit');" {% endif %}
             {% if oUIBlock.IsMultiple() %} multiple {% endif %}
             class="{% if oUIBlock.IsHidden() %}ibo-is-hidden{% endif %}{% if oUIBlock.GetAdditionalCSSClassesAsString() %} {{ oUIBlock.GetAdditionalCSSClassesAsString() }}{% endif %}"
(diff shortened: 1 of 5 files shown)


templates/base/components/input/select/select.html.twig
```diff
@@ -5,7 +5,7 @@
         {% include "base/components/input/inputlabel.html.twig" %}
     {% endif %}
     <select id="{{ oUIBlock.GetId() }}" name="{{ oUIBlock.GetName() }}"
-            {% if oUIBlock.GetSubmitOnChange() %} onChange="$(this).parents('form:first').trigger('submit');" {% endif %}
+            {% if oUIBlock.GetSubmitOnChange() %} onChange="$(this).parents('form').first().trigger('submit');" {% endif %}
```
Molkobain, line 8: This one seems legit, why change it?

steffunky: The goal of the original commit was to remove jQuery 3.4 deprecated methods. [first selector](https://api.jquery.com/first-selector/) is one of them:
>As of jQuery 3.4, the :first pseudo-class is deprecated. Remove it from your selectors and filter the results later using [.first()](https://api.jquery.com/first/).

This line was changed in the https://github.com/Combodo/iTop/commit/5b9e0a1d4f4751778386710eb52a1326d3c29423 commit changed the content from `onChange="this.form.submit();"` to `onChange="this.form.trigger('submit');"`, migrating the submit on a non jQuery element.

Since then, commit https://github.com/Combodo/iTop/commit/0802e5abb5fac94b9d03033e99e6cc9df7e694a4 changed the content to a jQuery based element to make the onchange actually work but uses deprecated `first` selector

Molkobain: All good then, thanks.

Molkobain (approved):

PR #964, opened 2026-07-07
N°9759 - Truncate AttributeText don't work as expected in case of multibytes characters

accognet:
internal

### DIFF
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -898,6 +898,26 @@ abstract class AttributeDefinition
 	}
 
 	/**
+	 * Helper to set a value that fits the attribute max size
+	 *
+	 * Default behavior is what DBObject::SetTrim used to do, now delegated to AttributeDefinition
+	 *
+	 * @param string $sValue
+	 *
+	 * @return string
+	 * @since 3.2.0 N°9643
+	 */
+	public function TrimValue(string $sValue)
+	{
+        $iMaxSize = $this->GetMaxSize();
+        if ($iMaxSize && (strlen($sValue) > $iMaxSize))
+        {
+            $sValue = substr($sValue, 0, $iMaxSize);
+        }
+        return $sValue;
+    }
+
+	/**
 	 * @return mixed|null
 	 * @deprecated never used
 	 */
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -910,9 +910,13 @@ abstract class AttributeDefinition
 	public function TrimValue(string $sValue)
 	{
 		$iMaxSize = $this->GetMaxSize();
-		if ($iMaxSize && (strlen($sValue) > $iMaxSize)) {
-			$sValue = substr($sValue, 0, $iMaxSize);
+		$iLength = mb_strlen($sValue);
+		if ($iMaxSize && ($iLength > $iMaxSize)) {
+			$sMessage = " -truncated ($iLength chars)";
+
+			return mb_substr($sValue, 0, $iMaxSize - mb_strlen($sMessage)).$sMessage;
 		}
+
 		return $sValue;
 	}
 
@@ -4266,10 +4270,15 @@ class AttributeText extends AttributeString
 		$iLengthChar = mb_strlen($sValue);
 		if ($iMaxSize && ($iLength > $iMaxSize)) {
 			$sMessage = " -truncated ($iLengthChar chars)";
-			$sVal = substr($sValue, 0, $iMaxSize - strlen($sMessage));
+			$iTruncatedValueMaxSize = $iMaxSize - strlen($sMessage);
+			$sTruncatedValue = substr($sValue, 0, $iTruncatedValueMaxSize);
 
-			//executes the "mb_substr" function to ensure a character is not truncated in the middle.
-			return mb_substr($sValue, 0, mb_strlen($sVal) - 1).$sMessage;
+			// Keep trimming bytes until we have valid UTF-8 to avoid returning a broken multibyte sequence.
+			while (($sTruncatedValue !== '') && !mb_check_encoding($sTruncatedValue, 'UTF-8')) {
+				$sTruncatedValue = substr($sTruncatedValue, 0, -1);
+			}
+
+			return $sTruncatedValue.$sMessage;
 		}
 
 		return $sValue;
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -3510,12 +3510,9 @@ class AttributeString extends AttributeDBField
 	public function TrimValue(string $sValue)
 	{
 		$iMaxSize = $this->GetMaxSize();
-		if ($iMaxSize == null) {
-			return $sValue;
-		}
-		$sLength = mb_strlen($sValue);
-		if ($iMaxSize && ($sLength > $iMaxSize)) {
-			$sMessage = " -truncated ($sLength chars)";
+		$iLength = mb_strlen($sValue);
+		if ($iMaxSize && ($iLength > $iMaxSize)) {
+			$sMessage = " -truncated ($iLength chars)";
 
 			return mb_substr($sValue, 0, $iMaxSize - mb_strlen($sMessage)).$sMessage;
 		}
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -4268,6 +4268,10 @@ class AttributeText extends AttributeString
 	}
 	/**
 	 * @inheritDoc
+	 *
+	 * Truncation is performed on a **byte** budget (MySQL TEXT limit) without ever cutting through a multibyte
+	 * UTF-8 sequence: the returned value is always valid UTF-8 and never exceeds GetMaxSize() bytes,
+	 * truncation suffix included.
 	 */
 	public function TrimValue(?string $sValue)
 	{
--- a/core/attributedef.class.inc.php
+++ b/core/attributedef.class.inc.php
@@ -4261,6 +4261,9 @@ class AttributeText extends AttributeString
 
 	/**
 	 * @inheritDoc
+	 *
+	 * Unlike the default implementation, the size is expressed in **bytes**: MySQL TEXT columns are limited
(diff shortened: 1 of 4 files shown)


webservices/webservices.class.inc.php
```diff
@@ -277,25 +277,13 @@ protected function LogUsage($sVerb, $oRes)
...
 		$oLog->Set('verb', $sVerb);
 		$oLog->Set('result', $oRes->IsOk());
-		$this->TrimAndSetValue($oLog, 'log_info', (string)$oRes->GetInfoAsText());
-		$this->TrimAndSetValue($oLog, 'log_warning', (string)$oRes->GetWarningsAsText());
-		$this->TrimAndSetValue($oLog, 'log_error', (string)$oRes->GetErrorsAsText());
-		$this->TrimAndSetValue($oLog, 'data', (string)$oRes->GetReturnedDataAsText());
+		$oLog->SetTrim('log_info', (string)$oRes->GetInfoAsText());
+		$oLog->SetTrim('log_warning', (string)$oRes->GetWarningsAsText());
+		$oLog->SetTrim('log_error', (string)$oRes->GetErrorsAsText());
+		$oLog->SetTrim('data', (string)$oRes->GetReturnedDataAsText());
```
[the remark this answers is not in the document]
steffunky: Acceptable behavior, we'll need to truncate data in any case. 20 characters out of 65k won't make a difference

core/attributedef.class.inc.php
```diff
@@ -897,6 +897,34 @@ public function GetMaxSize()
 		return null;
 	}
 
+	public function GetSize($value)
```
```suggestion
	/**
	 * @since 3.2.3-2 3.2.4 3.3.0 N°9759
	 */
	public function GetSize($value)
```

core/attributedef.class.inc.php
```diff
@@ -897,6 +897,34 @@ public function GetMaxSize()
...
+
+	/**
+	 * Helper to set a value that fits the attribute max size
+	 *
+	 * Default behavior is what DBObject::SetTrim used to do, now delegated to AttributeDefinition
+	 *
+	 * @param string $sValue
+	 *
+	 * @return string
+	 * @since 3.2.3-2 N°9759
```
```suggestion
	 * @since 3.2.3-2 3.2.4 3.3.0 N°9759
```

core/attributedef.class.inc.php
```diff
@@ -897,6 +897,34 @@ public function GetMaxSize()
...
+	 */
+	public function TrimValue(string $sValue)
```
```suggestion
	public function TrimValue(?string $sValue)
```

Are we sure this will always be a string? It might be `null` IMHO.

core/attributedef.class.inc.php
```diff
@@ -4219,6 +4247,31 @@ protected function GetSQLCol($bFullSpec = false)
 		return "TEXT".CMDBSource::GetSqlStringColumnDefinition();
 	}
 
+	public function GetSize($value)
```
```suggestion
	/**
	 * @inheritDoc
	 */
	public function GetSize($value)
```

core/attributedef.class.inc.php
```diff
@@ -4219,6 +4247,31 @@ protected function GetSQLCol($bFullSpec = false)
...
+	{
+		return strlen($value);
+	}
+	public function TrimValue(string $sValue)
```
```suggestion
	/**
	 * @inheritDoc
	 */
	public function TrimValue(?string $sValue)
```

Same comment

dflaven:
I am afraid that this may not work as expected. The main point is that the method GetMaxSize() is ambiguous: depending on the type of the attribute it returns either a number of characters or a number of bytes. Check the fine-print in the MySQL type definitions:

https://dev.mysql.com/doc/refman/8.4/en/string-type-syntax.html

For the VARCHAR types: the size is specified in chatacters
A variable-length string. M represents the maximum column length in characters. The range of M is 0 to 65,535. The effective maximum length of a VARCHAR is subject to the maximum row size (65,535 bytes, which is shared among all columns) and the character set used. **For example, utf8mb3 characters can require up to three bytes per character**, so a VARCHAR column that uses the utf8mb3 character set can be declared to be a maximum of 21,844 characters.

For the TEXT types: the size is specified in bytes
A TEXT column with a maximum length of 65,535 (216 − 1) bytes. **The effective maximum length is less if the value contains multibyte characters**. Each TEXT value is stored using a 2-byte length prefix that indicates the number of bytes in the value.

core/attributedef.class.inc.php
```diff
@@ -897,6 +897,37 @@ public function GetMaxSize()
 		return null;
 	}
 
+	/**
+	 * @since 3.2.3-2 3.2.4 3.3.0 N°9759
+	 */
+	public function GetSize($value)
```
bdalsass: A little bit confusing because GetSize is defined at AttributeDefinition level, supposing the underlying data is a string

core/attributedef.class.inc.php
```diff
@@ -897,6 +897,37 @@ public function GetMaxSize()
 		return null;
 	}
 
+	/**
+	 * @since 3.2.3-2 3.2.4 3.3.0 N°9759
+	 */
```
Molkobain, line 911:
```suggestion
	/**
	 * Return the size of $value, expressed in the same unit as {@see static::GetMaxSize()} for this attribute class.
	 *
	 * Default unit is a number of **characters**, matching MySQL VARCHAR(M) semantics for VARCHAR-based attributes.
	 * Byte-based attributes (e.g. {@see AttributeText}, stored as MySQL TEXT which is limited to 65535 **bytes**,
	 * not characters) MUST override both this method and {@see TrimValue()} consistently.
	 *
	 * @param string|null $value
	 *
	 * @return int Size of $value in the unit of GetMaxSize() (characters by default)
	 * @since 3.2.3-2 3.2.4 3.3.0 N°9759
	 */
```

core/attributedef.class.inc.php
```diff
@@ -897,6 +897,37 @@ public function GetMaxSize()
...
+	/**
+	 * Helper to set a value that fits the attribute max size
+	 *
+	 * Default behavior is what DBObject::SetTrim used to do, now delegated to AttributeDefinition
+	 *
+	 * @param string $sValue
+	 *
+	 * @return string
+	 * @since 3.2.3-2 3.2.4 3.3.0 N°9759
+	 */
```
Molkobain, line 935:
```suggestion
	/**
	 * Helper to set a value that fits the attribute max size
	 * 
	 * Truncation is performed in the same unit as GetMaxSize() / {@see GetSize()}: a number of **characters**
	 * by default (VARCHAR-based attributes). When truncated, a " -truncated (N chars)" suffix is appended and
	 * the returned value (suffix included) still fits within GetMaxSize().
	 *
	 * Default behavior is what DBObject::SetTrim used to do, now delegated to AttributeDefinition
	 *
	 * @param string $sValue
	 *
	 * @return string $sValue truncated so that it fits within {@see GetMaxSize()}.
	 * @since 3.2.3-2 3.2.4 3.3.0 N°9759
	 */
```

core/attributedef.class.inc.php
```diff
@@ -4219,6 +4250,37 @@ protected function GetSQLCol($bFullSpec = false)
 		return "TEXT".CMDBSource::GetSqlStringColumnDefinition();
 	}
 
+	/**
+	 * @inheritDoc
+	 */
```
Molkobain, line 4281:
```suggestion
	/**
	 * @inheritDoc
	 *
	 * Unlike the default implementation, the size is expressed in **bytes**: MySQL TEXT columns are limited
	 * in bytes (65535), not in characters, and {@see static::GetMaxSize()} for this class returns a number of bytes.
	 */
```

core/attributedef.class.inc.php
```diff
@@ -4219,6 +4250,37 @@ protected function GetSQLCol($bFullSpec = false)
...
+	public function GetSize($value)
+	{
+		return strlen($value);
+	}
+	/**
+	 * @inheritDoc
+	 */
```
Molkobain, line 4297:
```suggestion
	/**
	 * @inheritDoc
	 *
	 * Truncation is performed on a **byte** budget (MySQL TEXT limit) without ever cutting through a multibyte
	 * UTF-8 sequence: the returned value is always valid UTF-8 and never exceeds GetMaxSize() bytes,
	 * truncation suffix included.
	 */
```

core/attributedef.class.inc.php
```diff
@@ -4219,6 +4250,37 @@ protected function GetSQLCol($bFullSpec = false)
...
+			$sMessage = " -truncated ($iLengthChar chars)";
+			$iTruncatedValueMaxSize = $iMaxSize - strlen($sMessage);
+			$sTruncatedValue = substr($sValue, 0, $iTruncatedValueMaxSize);
+
+			// Keep trimming bytes until we have valid UTF-8 to avoid returning a broken multibyte sequence.
+			while (($sTruncatedValue !== '') && !mb_check_encoding($sTruncatedValue, 'UTF-8')) {
+				$sTruncatedValue = substr($sTruncatedValue, 0, -1);
+			}
+
+			return $sTruncatedValue.$sMessage;
```
```suggestion
			// mb_strcut cuts on a byte budget but moves the cut point back to a character boundary,
			// so it never returns a broken multibyte sequence at the end of the value
			$sTruncatedValue = mb_strcut($sValue, 0, $iTruncatedValueMaxSize, 'UTF-8');

			return $sTruncatedValue.$sMessage;
```

Molkobain (changes requested):
Discussed with @steffunky : 
  * The issue is already present in other parts of the code base (e.g. synchrodatasource.class.inc.php, cmdbabstract.inc.php), we rather be cautious as this fix is for a micro version, stability is key
  * Improve PHPDoc of these methods to remove ambiguity about what they do and the format they use internally
  * Improve how the value is truncated by using a proper PHP function instead of looping on it, which could lead to an empty value in the end if the value is made of multi-byte characters only.
  * Improve `TrimValue()` handling of a `null` value to avoid PHP 8.1 depreaction notices

bdalsass (commented):
to be discussed together

bdalsass (approved):
See IRL with @steffunky

Molkobain (approved):

Molkobain:
Let's gooooo!

PR #965, opened 2026-07-07
N°9722 - Setup - extension dependencies must be computed with installation choice change

eespie:
N°9722 - Setup - extension dependencies must be computed with installation choice change


odain-cbd:
@greptileai

PR #967, opened 2026-07-08
N°9692 - Uninstall / Disable install when extension has module with missing dependencies

Lenaick:
N°9692 - Uninstall / Disable install when extension has module with missing dependencies


tests/php-unit-tests/unitary-tests/setup/WizStepModulesChoiceTest.php
```diff
@@ -70,6 +72,8 @@ public function ProviderComputeChoiceFlags()
 					'installed' => false,
 					'disabled' => false,
 					'checked' => true,
+					'cannot-be-installed' => false,
```
odain-cbd: maybe you could rename 'cannot-be-installed' in 'dependency-issue' (or close to it)? maybe...

setup/extensionsmap.class.inc.php
```diff
@@ -562,19 +562,20 @@ public function GetAllExtensionsToDisplayInSetup(bool $bKeepExtensionsHavingMiss
...
-				'missing'        => $oExtension->bRemovedFromDisk,
-				'version'        => $oExtension->sVersion,
+				'extension_code'      => $oExtension->sCode,
+				'title'               => $oExtension->sLabel,
+				'description'         => $oExtension->sDescription,
+				'more_info'           => $oExtension->sMoreInfoUrl,
+				'default'             => true, // by default offer to install all modules
+				'modules'             => $oExtension->aModules,
+				'mandatory'           => $oExtension->bMandatory,
+				'cannot-be-installed' => count($oExtension->aMissingDependencies) > 0,
```
odain-cbd: (only for performances purpose)
count($oExtension->aMissingDependencies) > 0 seems to called further times.

what about providing a HasDependency() method? you could compute it once and cache result in private attribute of Extension class.

odain-cbd (approved):

setup/wizardsteps/WizStepModulesChoice.php
```diff
@@ -751,12 +751,16 @@ public function ComputeChoiceFlags(array $aChoice, string $sChoiceId, array $aSe
 		$bMissingFromDisk = isset($aChoice['missing']) && $aChoice['missing'] === true;
 		$bMandatory = (isset($aChoice['mandatory']) && $aChoice['mandatory']);
 		$bInstalled = $bMissingFromDisk || $oITopExtension->bInstalled;
+		$bCannotBeInstalled = $bMissingFromDisk || count($oITopExtension->aMissingDependencies) > 0;
```
Timmy38: Why would the flag bMissingFromDisk cause bCannotBeInstalled to be set too ?
If the extension is missing from disk it already implies its uninstallation will be forced because it cannot be installed
Setting both flags would cause both badges to be displayed when only one is useful

setup/wizardsteps/WizStepModulesChoice.php
```diff
@@ -882,6 +888,9 @@ protected function DisplayChoice($oPage, $aChoice, $aSelectedComponents, $aDefau
 		if (!$aFlags['uninstallable']) {
 			$sTooltip .= '<div id="badge--'.$sId.'--not-uninstallable" class="ibo-badge ibo-block ibo-is-yellow" title="Once this extension has been installed, it should not be uninstalled." >cannot be uninstalled</div>';
 		}
+		if ($aFlags['cannot-be-installed'] && !$aFlags['installed']) {
```
Timmy38: The flag cannot-be-installed should be visible for installed extensions too.
One of its dependencies could have been removed, forcing it to be uninstalled

eespie (approved):

Timmy38 (approved):

PR #980, opened 2026-07-22
N°9831 - Fix setup error when trying to remove class with a DEL_MANUAL ext. key

Molkobain:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | N°9831                 |

**Please be advised that I tried to make a fix for both bugs but that I don't master these parts of the code, so I may not be aware of the big picture.**

When uninstalling a module through the setup, the "check compatibility" step detects the objects that must be cleaned up to keep iTop consistent and redirects to the "extension management" page, which displays a summary of what will be deleted/updated. When some of these objects **can't** be deleted automatically (e.g. an external key with `DEL_MANUAL` option), the page is supposed to show a blocking message telling the user that it must remove these data manually first, and the "Proceed with deletion" button must not lead to a deletion.

But in this case, for some modules the blocking message isn't shown even though objects cannot be deleted automatically. The "Proceed with deletion" button stays enabled, and clicking it crashes the application with:

```
Combodo\iTop\DataFeatureRemoval\Helper\DataFeatureRemovalException : Deletion Plan cannot be executed due to issues
```

Concrete cases found while troubleshooting (standard datamodel):

| Removed module | External key involved | `is_null_allowed` | `on_target_delete` | Behavior |
|---|---|---|---|---|
| Data center devices | `Enclosure.rack_id` => `Rack` | `false` | `DEL_MANUAL` | ✅ Blocking message shown |
| Virtualization | `Hypervisor.farm_id` => `Farm` | `true` | `DEL_MANUAL` | ❌ No blocking message, crash on "Proceed with deletion" (Bug #1) |
| Virtualization | `VirtualMachine.virtualhost_id` => `VirtualHost` (abstract class) | `false` | `DEL_MANUAL` | ❌ No blocking message, crash on "Proceed with deletion" (Bug #2) |

1. On iTop 3.3.0-dev
3. Install iTop with the sample data (so that virtualization objects exist)
4. Run setup and select all modules
5. In the backoffice, create an `Enclosure` linked to a `Rack`
6. Run the setup again but this time unselect `Virtualization`
7. Continue to the "check compatibility" step
8. Click on "Cleanup my data"
9. See that there is no blocking message below the data table
10. Click on "Proceed with deletion" and see that the application crashes

## Cause (bug)

The "summary" (whether to block) and the "actual execution" rely on two different deletion-plan services that do not do the same processing:

- Summary is done by `StaticDeletionPlan::GetCleanupSummary()` (called by `DataFeatureRemovalController::GetDeletionPlanSummaryTable()`)
- Execution is done by `DataCleanupService::RecursiveDeletion()`, which throws the exception through `ObjectService::SetIssue()`.

Two different bugs made them diverge:

**Cause of bug #1 - Wrong order between `IsNullAllowed()` and `DEL_MANUAL` (`DataCleanupService::RecursiveDeletion()`)**
The "execution" service tested `DEL_MANUAL` before `IsNullAllowed()`, so a nullable external key with `DEL_MANUAL` was treated as an unresolvable issue and threw an exception.

But the reference engine `DBObject::MakeDeletionPlan()` (and `StaticDeletionPlan`) test `IsNullAllowed()` first.

A nullable key is always a simple "reset to null" update, and `DEL_MANUAL` only matters for mandatory ext. keys. That caused for `Hypervisor.farm_id` (nullable + `DEL_MANUAL`) the "summary" reported an update (no issue, no blocking message) while "execution" threw an exception.

**Cause of bug #2 - Deletion-plan entity overwritten (`StaticDeletionPlan::GetStaticDeletionPlan()`)**
While iterating the classes to remove, the method created a new `DeletionPlanEntity` for each classes and assigned it, loosing any entity already populated for that class by a previously processed class (issues/updates are accumulated on the referencing class while its target is processed).

The setup audit feeds the classes **sorted alphabetically** and **skips abstract classes** (`AbstractSetupAudit::RunDataAudit()`). For `VirtualMachine.virtualhost_id` => `VirtualHost`:
- `VirtualHost` is abstract, so only the concrete subclasses `Cloud`/`Farm`/`Hypervisor` are listed;
- `VirtualMachine` (referencing, mandatory `DEL_MANUAL`) is sorted after them, so the issue recorded on the `VirtualMachine` entity while processing `Farm`/`Hypervisor`/`Cloud` was lost when `VirtualMachine` was later processed as a root class => `iIssueCount = 0` => no blocking message, but execution still detected it and threw an exception.

This also explains why the `Data center devices` module works as expected: `Enclosure` (referencing) is sorted before `Rack` (its target), so the issue recorded on `Enclosure` is never overwritten afterwards.

**Fix for bug #1**
In the "summary" service, test `IsNullAllowed()` first like in the "execution" service (`StaticDeletionPlan`) and core `DBObject::MakeDeletionPlan()`.

**Fix for bug #2**
Do not overwrite an existing plan entity, create it only when missing, then merge the initial IDs.
Issues/updates already accumulated are preserved, making issue detection independent of the processing order.

**Tests**
I've tried to add test cases in `StaticDeletionPlanTest` but I really not sure how it's supposed to work, so reviews are welcome 😅

- [X] Is the PR clear and detailed enough so anyone can understand without digging in the code?

### DIFF
--- a/datamodels/2.x/combodo-data-feature-removal/src/Service/DataCleanupService.php
+++ b/datamodels/2.x/combodo-data-feature-removal/src/Service/DataCleanupService.php
@@ -134,10 +134,6 @@ class DataCleanupService
 				/** @var DBObject $oDependentObj */
 				while ($oDependentObj = $oSet->Fetch()) {
 					$iDeletePropagationOption = $oExtKeyAttDef->GetDeletionPropagationOption();
-					if ($iDeletePropagationOption == DEL_MANUAL) {
-						$this->oObjectService->SetIssue(get_class($oDependentObj));
-						continue;
-					}
 
 					if ($oExtKeyAttDef->IsNullAllowed()) {
 						// Optional external key, list to reset
@@ -152,6 +148,12 @@ class DataCleanupService
 							return false;
 						}
 					} else {
+						// Mandatory external key
+						if ($iDeletePropagationOption == DEL_MANUAL) {
+							// Cannot be deleted automatically, must be handled manually
+							$this->oObjectService->SetIssue(get_class($oDependentObj));
+							continue;
+						}
 						// Propagate deletion only if not visited
 						if ($this->IsVisited($oDependentObj)) {
 							continue;
--- a/datamodels/2.x/combodo-data-feature-removal/src/Service/StaticDeletionPlan.php
+++ b/datamodels/2.x/combodo-data-feature-removal/src/Service/StaticDeletionPlan.php
@@ -59,9 +59,11 @@ class StaticDeletionPlan
 	{
 		foreach ($aClasses as $sClass) {
 			$oDeletionPlanItem = $this->GetInitialClassDeletionPlan($sClass);
-			$oDeletionPlanEntity = new DeletionPlanEntity();
-			$oDeletionPlanEntity->oDelete->Merge($oDeletionPlanItem);
-			$this->aDeletionPlan[$sClass] = $oDeletionPlanEntity;
+			// N°9831 Do not overwrite existing entity as it may already exist for this class if a previously processed class references it (issues/updates already accumulated must be kept)
+			if (false === array_key_exists($sClass, $this->aDeletionPlan)) {
+				$this->aDeletionPlan[$sClass] = new DeletionPlanEntity();
+			}
+			$this->aDeletionPlan[$sClass]->oDelete->Merge($oDeletionPlanItem);
 
 			$this->DeletionPlanForReferencingClasses($sClass);
 		}
--- a/tests/php-unit-tests/unitary-tests/datamodels/2.x/combodo-data-feature-removal/DataCleanupServiceTest.php
+++ b/tests/php-unit-tests/unitary-tests/datamodels/2.x/combodo-data-feature-removal/DataCleanupServiceTest.php
@@ -163,6 +163,58 @@ class DataCleanupServiceTest extends \AbstractCleanup
 		$this->AssertSummaryEquals($aExpected, $aRes);
 	}
 
+	/**
+	 * N°9831
+	 *
+	 * A nullable external key configured as DEL_MANUAL must NOT be treated as an issue: it is simply
+	 * reset (set to null), exactly like core DBObject::MakeDeletionPlan does. DEL_MANUAL only blocks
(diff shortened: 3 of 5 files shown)


datamodels/2.x/combodo-data-feature-removal/src/Service/DataCleanupService.php
```diff
@@ -152,6 +148,12 @@ private function RecursiveDeletion(DBObject $oObjectToClean): bool
 							return false;
 						}
 					} else {
+						// Mandatory external key
+						if ($iDeletePropagationOption == DEL_MANUAL) {
```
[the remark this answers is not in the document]
Molkobain: Addressed in the last commit

odain-cbd (commented):
nice work

odain-cbd (approved):

datamodels/2.x/combodo-data-feature-removal/src/Service/StaticDeletionPlan.php
```diff
@@ -59,9 +59,11 @@ public function GetStaticDeletionPlan(array $aClasses): array
 	{
 		foreach ($aClasses as $sClass) {
 			$oDeletionPlanItem = $this->GetInitialClassDeletionPlan($sClass);
-			$oDeletionPlanEntity = new DeletionPlanEntity();
-			$oDeletionPlanEntity->oDelete->Merge($oDeletionPlanItem);
-			$this->aDeletionPlan[$sClass] = $oDeletionPlanEntity;
+			// N°9831 Do not overwrite existing entity as it may already exist for this class if a previously processed class references it (issues/updates already accumulated must be kept)
+			if (false === isset($this->aDeletionPlan[$sClass])) {
```
odain-cbd: array_key_exists is easier to read maybe thant false===isset(...)

odain-cbd (approved):

datamodels/2.x/combodo-data-feature-removal/src/Service/StaticDeletionPlan.php
Hipska: !isset is also easier to read.

Molkobain: Changed to `array_key_exists`.

Regarding the `false` vs `!`, there is no consensus on this in the team 😁 But I personally prefer `false` as I don't always see the `!` symbol.

PR #991, opened 2026-08-04
N°9527 - Avoid multiple notifications on user mentionned on ticket with children

v-dumas:
## Base information

| Related to a Combodo ticket? |                  |
| Type of change?                                                                 | Enhancement |

When a Person is mentioned on a parent Ticket within a log entry, that log entry is copied on the children tickets as is, which result in multiple notification of the same person for the same topic

The idea is to clean the copied log entry, removing the part which trigger the TriggerOnMention

### DIFF
--- a/datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
+++ b/datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
@@ -350,7 +350,7 @@
         <method id="UpdateChildTicketLog">
           <comment><![CDATA[/**
             *
-            * Remove the current user associated Person from the contacts_list of this Ticket
+            * Remove the Person associated with the current user from the contacts_list of this Ticket
             * No error if there is no associated Person or if the Person is not in the list
             * @return bool Return true
              * @param string $sChildClass The class name of the child ticket (e.g. 'Incident', 'UserRequest', etc.)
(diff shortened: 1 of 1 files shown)


datamodels/2.x/itop-tickets/datamodel.itop-tickets.xml
```diff
@@ -383,7 +383,8 @@
 			if (MetaModel::IsValidAttCode($sParentClass, $sParentAttCode) && MetaModel::GetAttributeDef($sParentClass, $sParentAttCode) instanceof AttributeCaseLog
 			&& MetaModel::IsValidAttCode($sChildClass, $sChildAttCode) &&  MetaModel::GetAttributeDef($sChildClass, $sChildAttCode) instanceof AttributeCaseLog
 			&& (!utils::IsNullOrEmptyString($this->Get($sParentAttCode)->GetModifiedEntry('html')))) {
-			  $aChildEntries[$sChildAttCode] = Dict::Format('Class:'.$sParentClass.'/Method:UpdateChildTicketWith:'.$sParentAttCode, $this->GetKey(), $this->Get('ref')).$this->Get($sParentAttCode)->GetModifiedEntry('html');
+			  $sCleanedEntry = preg_replace('/<a\s*([^>]*)data-object-class="([^"]*)"\s.*data-object-key="([^"]*)"\s*([^>]*)>@/Ui','<a\1\4>',$this->Get($sParentAttCode)->GetModifiedEntry('html'));
```
steffunky, line 386: The `@` is hardcoded here. But this value comes from the array configuration parameter `mentions.allowed_classes` which can hold multiple mention triggering character (`@`, `#`, `?` etc..) and may be interpreted not as character but as joker or placeholder

Molkobain: Moreover, this is removing the `data-object-class` and `data-object-key` metadata which may be used by other features, you should not remove them.

Maybe the real fix (but it has to be discussed) would be that a mention adds a dedicated HTML metadata in addition to `data-object-class` / `data-object-key` which should be the one triggering the mention notification. That way you could stip this one without impacting other features.

Molkobain (changes requested):
See comment in Steffunky thread

v-dumas:
There is another PR which aims to replace the regexp with a Xpath, as a result, I will stop this one until we have agreed on a common strategy

PR #992, opened 2026-08-04
N°9876 - A non installed extension should not be selected

Lenaick:
N°9876 - A non installed extension should not be selected if it is not mandatory

### DIFF
--- a/tests/php-unit-tests/unitary-tests/setup/WizStepModulesChoiceTest.php
+++ b/tests/php-unit-tests/unitary-tests/setup/WizStepModulesChoiceTest.php
@@ -232,7 +232,7 @@ class WizStepModulesChoiceTest extends ItopTestCase
 					'mandatory' => true,
 				],
 			],
-			'A mandatory extension should be checked and enabled if the "disable uninstallation check" flag is set' => [
+			'A mandatory but not installed extension should be not checked and enabled if the "disable uninstallation check" flag is set' => [
 				'aExtensionsOnDiskOrDb' => [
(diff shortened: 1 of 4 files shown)


tests/php-unit-tests/unitary-tests/setup/WizStepModulesChoiceTest.php
```diff
@@ -596,44 +692,6 @@ public function testComputeChoiceFlags($aExtensionsOnDiskOrDb, $aWizardStepDefin
 		$this->assertEquals($aExpectedFlags, $aFlags);
 	}
 
-	public function ProviderGetAllExtensionsToDisplayInSetupMandatoryFlag()
```
bdalsass, line 600: this test is not relevant anymore ?

bdalsass (approved):

tests/php-unit-tests/unitary-tests/setup/WizStepModulesChoiceTest.php
Lenaick: Yes, this test checked the condition that I removed from `iTopExtensionsMap::GetAllExtensionsToDisplayInSetup`, which forced the 'mandatory' status for an extension added via the Hub or the Designer, something we no longer want

Timmy38 (approved):

odain-cbd (approved):
great test coverage/documentation on this setup part. it looks good to me. 

quite difficult to keep all desired behaviours and make it simple.... unfortunately

PR #994, opened 2026-08-10
N°9643 - Dataflow: split protocol & type, add url and last change

v-dumas:
## Base information

| Related to a Combodo ticket? | N°9643                 |
| Type of change?                                                                 |  Enhancement  |

## Objective (enhancement)

Implement feedback from partner


datamodels/2.x/itop-flow-map/module.itop-flow-map.php
```diff
@@ -31,7 +31,8 @@
 
 		],
 		'data.struct' => [
-			'data/data.itop-flow-map.en_us.xml',
+			'data/data.itop-dataflowtype.xml',
+			'data/data.itop-dataflowprotocol.xml',
```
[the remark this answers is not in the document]
v-dumas: This module was never released, so this is not pertinent, in addition those data are just example and can be changed by any iTop administrator to the set of values which is pertinent for his organization

datamodels/2.x/itop-flow-map/datamodel.itop-flow-map.xml
```diff
@@ -97,6 +97,31 @@
...
+        <field id="dataflowprotocol_id" xsi:type="AttributeExternalKey">
+          <sql>dataflowprotocol_id</sql>
+          <filter/>
+          <dependencies/>
+          <is_null_allowed>true</is_null_allowed>
+          <target_class>DataFlowProtocol</target_class>
+          <on_target_delete>DEL_MANUAL</on_target_delete>
+          <tracking_level>all</tracking_level>
+        </field>
+        <field id="documentation_url" xsi:type="AttributeURL">
```
Hipska, line 109: Link to documentation should be done by using `Document` tab, adding separate field for this removes unity across objects. 

A URL specifying where te flow goes to would be interesting however.

datamodels/2.x/itop-flow-map/dictionaries/fr.dict.itop-flow-map.php
```diff
@@ -44,10 +44,16 @@
...
-	'Class:DataFlow/Attribute:status+'                           => '',
+	'Class:DataFlow/Attribute:dataflowtype_id'                           => 'Type du flux',
+	'Class:DataFlow/Attribute:dataflowtype_id+'                          => 'Ces valeurs sont gérées dans une typologie',
+	'Class:DataFlow/Attribute:dataflowprotocol_id'                           => 'Protocole',
+	'Class:DataFlow/Attribute:dataflowprotocol_id+'                          => 'Protocole utilisé par le flux. Ces valeurs sont gérées dans une typologie',
+	'Class:DataFlow/Attribute:documentation_url'                           => 'Lien vers la documentation',
+	'Class:DataFlow/Attribute:documentation_url+'                          => 'URL vers la documentation du flux de données',
+	'Class:DataFlow/Attribute:last_change_date' => 'Dernière mise à jour',
+	'Class:DataFlow/Attribute:last_change_date+' => 'Date de la dernière mise à jour du logiciel ou de la configuration du flux de données',
+	'Class:DataFlow/Attribute:status'                => 'Etat',
```
steffunky:
```suggestion
	'Class:DataFlow/Attribute:status'                => 'État',
```

datamodels/2.x/itop-flow-map/datamodel.itop-flow-map.xml
Hipska: @v-dumas what do you think about my comment?

v-dumas: I did not capture what you wanted.
I am only applying the result of the discussion we had a month ago.
Documents tab is available, as it's a FunctionalCI, but it came out that a direct url was quicker to find.
If you want to use it for another purpose than documentation, fare, just rename it.

Hipska: > I did not capture what you wanted.

well, you could have asked.

Also, with this logic, every FunctionalCI should have a documentation_url, I don’t see why DataFlow is special in that regard…

Hipska: @v-dumas?

datamodels/2.x/itop-flow-map/data/data.itop-dataflowtype.xml
```diff
@@ -1,24 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Set>
   <DataFlowType alias="DataFlowType" id="1">
-    <name>HTTP</name>
+    <name>REST API</name>
   </DataFlowType>
   <DataFlowType alias="DataFlowType" id="2">
-    <name>HTTPS</name>
+    <name>KAFKA</name>
   </DataFlowType>
```
Hipska, line 8: Kafka and REST are protocols! Both can have different types of data in their payload. (Like JSON, XML, YAML, ...)

Hipska: Or otherwise MQTT needs to be added in this list and be removed from DataFlowProtocol

v-dumas: Thanks for the feedback. I moved REST and KAFKA as Protocol

PR #1009, opened 2026-08-18
Feature/8420 transaction id lifetime

bdalsass:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | https://support.combodo.com/pages/UI.php?operation=details&class=Bug&id=8420                |
| Type of change?                                                                 | Enhancement |

Transactions ID handler configured with files won't take care of the transation age.
Transition ID are not enaught secured, the complexity may be increase.

## Reproduction procedure (enhancement - if needed)

Open an object in modification, your are able to confirm this form event after a day.

## Cause (bug)

Date creation of the transition as file isn't used.

Add a genral configuration parameters to set the file transition lifetime, all transitions after this limit are considered as invalid.
The garbage collection will also take care of the setting.
The ID will integrate a 48 random hex chars, even for the sessions id.

- [X] Is the PR clear and detailed enough so anyone can understand without digging in the code?


core/config.class.inc.php
```diff
@@ -1289,6 +1289,14 @@ class Config
 			'source_of_value'     => '',
 			'show_in_conf_sample' => false,
 		],
+		'transactions_file_lifetime' => [
+			'type'                => 'integer',
+			'description'         => 'Value in seconds for the lifetime of a transaction file (after this duration, the transaction will be invalid and the garbage collector could delete it).',
```
```suggestion
			'description'         => 'Value in seconds for the lifetime of a transaction file (after this duration, the transaction will be invalid and the garbage collector could delete it). This only works if the "transaction_storage" parameter is set to "File".',
```

Maybe explicit to which other parameter it is bound.

application/transaction.class.inc.php
```diff
@@ -281,6 +308,18 @@ public static function IsTransactionValid($id, $bRemoveTransaction = true)
 			return false;
 		}
 
+		$iTime = filemtime($sFilepath);
+		$iLifetime = (int) MetaModel::GetConfig()->Get('transactions_file_lifetime');
+		$iLimit = time() - $iLifetime;
```
Molkobain:
```suggestion
		$iTime = filemtime($sFilepath);
		$iLifetime = (int) MetaModel::GetConfig()->Get('transactions_file_lifetime');
		$iLimit = $iTime - $iLifetime;
```

Molkobain (approved):

PR #1010, opened 2026-08-18
 N°9875 - Error deprecated on component version object

odain-cbd:


application/utils.inc.php
```diff
@@ -3249,4 +3250,47 @@ public static function GetMemoryLimit(): int
...
+	* @param string $sPrefix
+	* @return array
+	* @throws \Exception
+	 */
+	public static function ZipArchiveOpenWithTempNam(string $sDirectory, string $sPrefix): array
+	{
+		$sTempnam = tempnam($sDirectory, $sPrefix);
+		unlink($sTempnam);
+		$sArchiveName = $sTempnam.'.zip';
+		;
```
bdalsass:
```suggestion
```

application/utils.inc.php
```diff
@@ -3249,4 +3250,47 @@ public static function GetMemoryLimit(): int
...
+		return $oZip;
+	}
+
+	/**
+	* @param string $sDirectory
+	* @param string $sPrefix
+	* @return array
+	* @throws \Exception
+	 */
+	public static function ZipArchiveOpenWithTempNam(string $sDirectory, string $sPrefix): array
```
bdalsass: Maybe ZipArchiveCreate function name will be more clear

bdalsass (changes requested):

application/utils.inc.php
Lenaick: This function creates an archive using a temporary name : I would use `ZipArchiveCreateWithTempNam`

odain-cbd: thanks for the suggestion. i was not inspired at all yesterday to name it correctly

Molkobain (approved):
Apart from the naming proposition, looks good to me.

eespie (approved):

bdalsass (approved):

PR #1011, opened 2026-08-18
N°9934 - OQL with magic field

accognet:
internal


eespie (approved):

Molkobain (approved):
Good job 🙌

tests/php-unit-tests/unitary-tests/core/OQLResolverTest.php
```diff
@@ -0,0 +1,41 @@
...
+{
+	public function GetDatamodelDeltaAbsPath(): string
+	{
+		return __DIR__.'/Delta/delta_oql_resolver.xml';
+	}
+
+	public function testQueryOnMagicalFields()
+	{
+		// Given
+		$sObjectKey = $this->GivenObjectInDB('OQLResolverChild', ['name' => 'toto', 'status' => 'new']);
```
odain-cbd, line 25: maybe a test on a field that will not work would be interesting like Password (for which IsValidFilterAttribute=false)?

maybe it could be interesting to propose some other searchable fields that could be used a lot (attachments, others?)? for documentation purpose

PR #1015, opened 2026-08-20
N°9949 - Cannot uninstall IPAM from setup wizard screen

odain-cbd:


setup/modulediscovery.class.inc.php
```diff
@@ -241,6 +242,19 @@ public static function DeclareRemovedExtensions(array $aRemovedExtension): void
 		self::$m_aRemovedExtensions = $aRemovedExtension;
 	}
 
+	/**
+	* @since 3.3
+	* @return string[]
+	*/
+	public static function GetRemovedExtensionCodes(): array
```
bdalsass, line 249: Strange to get a function exist only for tests purpose but maybe could certainly be used later for core usage.
The function code could also be replaced by an array_map.

setup/modulediscovery.class.inc.php
```diff
@@ -241,6 +242,19 @@ public static function DeclareRemovedExtensions(array $aRemovedExtension): void
...
+	{
+		$aRes = [];
+		foreach (self::$m_aRemovedExtensions as $oExtension) {
+			$aRes [] = $oExtension->sCode;
```
bdalsass, line 253:
```suggestion
			$aRes[] = $oExtension->sCode;
```

bdalsass: strange space between the array name and bracket but it seems more clear for you as you write all you array instructions like this

eespie (approved):

setup/modulediscovery.class.inc.php
odain-cbd: i had other solutions (invoke...). seemed to me useful to have it available for support team.

setup/modulediscovery.class.inc.php
odain-cbd: too bad it is not fixed by php-cs-fixer

odain-cbd: in fact php-cs-fixer is reverting your suggestion!

PR #1021, opened 2026-08-24
N°9919 - Check it "To" exists when sendind mail with PHPMail

jf-cbd:
## Base information

| Related to a SourceForge thread / Another PR / A GitHub Issue / Combodo ticket? | 9919                |

When a notification mail doesn't have a "to" field (e.g. only CC field), mail notification won't be sent with the following error "Call to a member function getBodyAsString() on null in /var/www/html/itop/support/3.2/sources/Core/Email/Transport/SymfonyPHPMailTransport.php:23"

1. On iTop on dev branch (this day)
2. With PHP 8.4
3. Create a notification on an event and only add CC field.
4. Execute an action that will trigger the notification
5. See the error and see that no mail is sent.

## Cause (bug)

In the code, getBodyAsString is called on null, while it could be avoided.

Check is $oHeaders->get('To') exists before calling getBodyAsString

- [x] I have tested all changes I made on an iTop instance => tested with MailPit

### DIFF
--- a/sources/Core/Email/Transport/SymfonyPHPMailTransport.php
+++ b/sources/Core/Email/Transport/SymfonyPHPMailTransport.php
@@ -20,7 +20,7 @@ class SymfonyPHPMailTransport extends AbstractTransport
 	{
 		$oHeaders = $oRawEmail->getHeaders();
 
-		return $oHeaders->get('To')->getBodyAsString();
+		return $oHeaders->get('To') ? $oHeaders->get('To')->getBodyAsString() : '';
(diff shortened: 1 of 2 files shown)


steffunky (approved):
I think this should be merged into `support/3.2` and then merged to `develop`.

Looks good to me, however we lost the return-path compared to 3.2.2, we should look into passing all headers to the `mail` function (maybe not in this ticket)

Image
3.2.2

Image
3.3.0 issue/9919-..

PR #1040, opened 2026-09-04
N°10026 - Problem with apc-emulation when no file is present

eespie:
N°10026 - Problem with apc-emulation when no file is present

### DIFF
--- a/core/apc-emulation.php
+++ b/core/apc-emulation.php
@@ -272,6 +272,7 @@ class apcFile
 			return;
 		}
 		if (!self::$aFilesByTime) {
+			self::$aFilesByTime = [];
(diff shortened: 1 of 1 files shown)


core/apc-emulation.php
```diff
@@ -272,6 +272,7 @@ protected static function AddFile($sNewFilename)
 			return;
 		}
 		if (!self::$aFilesByTime) {
+			self::$aFilesByTime = [];
```
Hipska, line 275: This should probably go into the `ListFilesByTime` method.

odain-cbd (approved):
if you stick yo minimal fix it is ok to me. otherwise init could be done somewhere else.

